PR #29, opened 2020-07-31
Add tests that compare to default.qubit

trbromley:
This PR adds a series of tests that compare the output state of `lightning.qubit` with the output state of `default.qubit` for a variety of different circuits that use all of the supported gates in `lightning.qubit`.

These tests allow us to be confident that `lightning.qubit` is processing states correctly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,5 +76,5 @@ jobs:
       - name: Run PennyLane shared test suite
         run: |
           cd pl
-          pytest pennylane/devices/tests --device lightning.qubit --analytic False --skip-ops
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -76,5 +76,5 @@ jobs:
       - name: Run PennyLane shared test suite
         run: |
           cd pl
-          pytest pennylane/plugins/tests --device lightning.qubit --analytic False --skip-ops
-          pytest pennylane/plugins/tests --device lightning.qubit --analytic True --skip-ops
+          pytest pennylane/devices/tests --device lightning.qubit --analytic False --skip-ops --shots=20000
```
trbromley, line 79: We can change the shots back to the default once the test suite works in non-analytic mode.

PR #32, opened 2020-08-10
Windows compiler and linker options

antalszava:
Modifies the compiler and linker options for Windows. The motivation is to achieve optimized code by using the `/O2` compiler option to achieve faster performance on Windows too.


antalszava:
Tried the latest wheel for Win, the tests pass well and the performance is as expected too. Will go for merging.

PR #33, opened 2020-08-10
Setting up RTD

trbromley:
https://pennylane-lightning.readthedocs.io/en/mocking_out_cpp/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,7 +164,7 @@ ext_modules = [
     ),
 ]
 
-if os.environ.get("USING_RTD", False):
+if os.environ.get("MOCK_DOCS", False):
     ext_modules = []
 
 requirements = [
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -165,6 +164,8 @@ def build_extensions(self):
     ),
 ]
 
+if os.environ.get("USING_RTD", False):
+    ext_modules = []
```
trbromley, line 168: Am thinking to use this: https://docs.readthedocs.io/en/stable/guides/environment-variables.html

josh146: @trbromley how come? Could it be useful to be able to compile the documentation locally even without the C++ requirements?

josh146: Oh I suppose you could always set this locally! Perhaps you could make the env variable more general? `LQ_MOCK_DOCS`?

trbromley: Nice idea!

PR #34, opened 2020-08-12
Modify wheel build

antalszava:
* Adds caching for the Windows wheel build such that Eigen isn't pulled every time


trbromley:
@antalszava what is the status of this PR?

antalszava:
This was an attempt to cache Eigen when building the wheel such that we can decrease the time for building the wheel. As this is not a priority at the moment anymore, I'll close it (could be reopened later if we'd like to revisit).

PR #35, opened 2020-08-12
Move the tests folder one level up (Python)

antalszava:
PennyLane and its plugins have a folder structure where the main folder (e.g. `pennylane`) and the tests of the library (contained in the `tests` folder) are on the same level of the folder tree.

For PennyLane-lightning, the Python tests reside in the `pennylane_lightning/tests` folder, that is `tests` is a subfolder of the main `pennylane_lightning` folder, making the plugin an exception to the convention for plugins.

This PR moves the `tests` directory one level up.

This proves to be useful when [using for example `plugin-matrix-tests`](https://github.com/PennyLaneAI/plugin-test-matrix/runs/977258327?check_suite_focus=true#step:9:11)  for testing PL plugins in a standardized way.


antalszava:
thanks for catching those @trbromley! updated

PR #36, opened 2020-08-12
Tidy up

trbromley:
Tidy up some elements such as the doc installation instructions.


setup.py
```diff
@@ -171,7 +171,7 @@ def build_extensions(self):
 
 
 info = {
-    "name": "PennyLane-lightning",
+    "name": "PennyLane-Lightning",
```
antalszava, line 174: :rocket:

setup.py
```diff
@@ -181,7 +181,7 @@ def build_extensions(self):
     "entry_points": {
         "pennylane.plugins": ["lightning.qubit = pennylane_lightning:LightningQubit",],
     },
-    "description": "PennyLane is a Python quantum machine learning library by Xanadu Inc.",
+    "description": "PennyLane lightning.qubit plugin",
```
antalszava, line 184: (Minor suggestion): Just in case we'd have more devices/would have more functionality not necessarily connected to `lightning.qubit`?
```suggestion
    "description": "PennyLane-Lightning plugin",
```

README.rst
```diff
@@ -23,27 +23,45 @@ Installation
...
 
-.. code-block:: bash
+.. code-block:: console
 
-    git clone https://github.com/XanaduAI/pennylane-lightning.git
-    cd pennylane-lightning
-    pip install -e .
+    $ git clone https://github.com/XanaduAI/pennylane-lightning.git
+    $ cd pennylane-lightning
+    $ pip install -e .
```
antalszava, line 36: Though not tightly connected, thought could be added in this PR. An idea that could be nice to be included in the docs (feel free to rephrase).
```suggestion
    $ pip install -e .

    Subsequent calls to `pip install -e .` will use cached binaries stored in the `build` folder. Use the `python setup.py clean --all` command to clean cached files and to re-build the binaries with `pip install -e .`.
```

trbromley: Thanks, have added!

PR #37, opened 2020-08-12
Add CI checks

trbromley:
Including:
- A check that runs `black`
- A check for codecov
- Tests to increase code coverage
- Fix code quality
- Add shields


README.rst
```diff
@@ -1,6 +1,30 @@
...
+    :target: https://pennylane-lightning.readthedocs.io
+
+.. image:: https://img.shields.io/pypi/v/PennyLane-Lightning.svg?style=flat-square
+    :alt: PyPI
+    :target: https://pypi.org/project/PennyLane-Lightning
+
+.. image:: https://img.shields.io/pypi/pyversions/PennyLane-Lightning.svg?style=flat-square
+    :alt: PyPI - Python Version
+    :target: https://pypi.org/project/PennyLane-Lightning
+
```
trbromley, line 27: image

- Hopefully the build one will work once merged. Although this is the one I'm least sure about why it's working :thinking: 
- Coverage should come up once we merge this into master and upload a report
- Code quality should be improved in this PR, the low quality was due to one test file which is now improved
- The pypi and python shields should update when we release

trbromley: Update: the first one might be fixed with 8951a729a9913f13208517cf60a09389bfd65b2b

pennylane_lightning/src/tests/lightning_unittest.cpp
```diff
@@ -1,21 +1,18 @@
 // Copyright 2020 Xanadu Quantum Technologies Inc.
```
trbromley, line 1: All changes on this file are to fix code quality

.github/workflows/build.yml
```diff
@@ -1,5 +1,9 @@
 name: Testing
-on: [pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
antalszava, line 6: :100:

pennylane_lightning/lightning_qubit.py
```diff
@@ -83,18 +83,21 @@ class LightningQubit(DefaultQubit):
 
     observables = {"PauliX", "PauliY", "PauliZ", "Hadamard", "Identity"}
 
+    _MAX_WIRES = 16
+    """maximum number of supported wires. Beyond this number, lightning.qubit behaves like
```
antalszava, line 87:
```suggestion
    """Maximum number of supported wires. Beyond this number, lightning.qubit behaves like 
```

tests/test_apply.py
```diff
@@ -1022,3 +1022,30 @@ def test_pauliz_hadamard(self, theta, phi, varphi, monkeypatch, qubit_device_3_w
...
+    def test_apply(self, monkeypatch):
+        """Test that the apply method uses the one in default.qubit when the number of wires
+        exceeds the supported number."""
+        dev = qml.device("lightning.qubit", wires=2)
+        dev.num_wires = LightningQubit._MAX_WIRES + 1
+        mock_apply = mock.MagicMock()
+        with monkeypatch.context() as m:
+            m.setattr(DefaultQubit, "apply", mock_apply)
+            dev.apply(0, rotations=1, other=2)
+        call_args = mock_apply.call_args
```
antalszava, line 1046: We've tried avoiding the use of `unittest` and `mock` where possible and keep tests `pytest` only. One way to do that would be to create a `MockApply` class:
```python
class MockApply:
    def __init__(self):
        self.op_call = None
        self.rotations_call = None
        self.kwargs_call = None

    def apply(self, operations, rotations=None, **kwargs):
        self.op_call = operations
        self.rotations_call = rotations
        self.kwargs_call = kwargs
```
and then
```python
mock_apply = MockApply()
with monkeypatch.context() as m:
        m.setattr(DefaultQubit, "apply", mock_apply.apply)
```
(or comparable, this one is not tested code)

trbromley: Nice, thanks @antalszava I will update that.

Out of curiosity, what motivates the drive to avoid `unittest`?

antalszava: I think we'd just want to stick with a single testing framework whenever possible and have one less dependency if possible :thinking: Perhaps @josh146 would have further thoughts here

josh146: @antalszava @trbromley: Note that `unittest.mock` has no pytest equivalent, so it's usually fine to use `unittest.mock` with pytest (better to use an existing package than roll our own mocking framework).

Having said that, why not just use [`pytest-mock`](https://pypi.org/project/pytest-mock/)? This is a simple wrapper around `unittest.mock` to turn it into a fixture (and make it fit better with common pytest patterns):

```python
def test_apply(self, mocker):
    """Test that the apply method uses the one in default.qubit when the number of wires
    exceeds the supported number."""
    mocker.patch("pennylane.devices.DefaultQubit.apply")

    dev = qml.device("lightning.qubit", wires=2)
    dev.num_wires = LightningQubit._MAX_WIRES + 1
    dev.apply(0, rotations=1, other=2)

    DefaultQubit.apply.assert_called_once_with(0, rotations=1, other=2)
```

`pytest.mock` also suports `mocker.spy`, and adheres to the `unittest.mock` api.

codecov.yml
```diff
@@ -0,0 +1,2 @@
+fixes:
+  - "main/::"
```
josh146, line 2: How come this path fix was needed?

trbromley: It's due to the `build` workflow, we clone the repo into a folder called `main` (see line 45). I guess could also have fixed there! But this seemed to work

tests/test_apply.py
trbromley: Thanks a lot to both of you for this! I've updated to @josh146's suggestion - which is also clearer!

PR #39, opened 2020-08-14
Add command to makefile to clean binaries

trbromley:
Add an extra makefile command `make clean-binaries` to remove the c++ compiled code. This allows one to do
```bash
make clean-binaries
pip install -e .
```
such that recompilation will occur. Otherwise a cached version will be used each time.

`make clean-binaries` is kept separate from `make clean`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ clean:
 	rm -rf tmp
 	rm -rf *.dat
 
+clean-binaries:
+	rm -f pennylane_lightning/lightning_qubit_ops*
+
 docs:
 	make -C doc html
 
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,7 @@ Alternatively, to build PennyLane-Lightning from source you can run
     $ pip install -e .
 
 Note that subsequent calls to ``pip install -e .`` will use cached binaries stored in the
-``build`` folder. Run ``make clean`` if you would like to recompile.
+``build`` folder. Run ``make clean`` and ``make clean-binaries`` if you would like to recompile.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -44,6 +44,9 @@ clean:
 	rm -rf tmp
 	rm -rf *.dat
 
+clean-binaries:
+	rm -f pennylane_lightning/lightning_qubit_ops*
```
antalszava, line 48: On linux for me, the wheel gets built in the following directory: `pennylane-lightning/build/lib.linux-x86_64-3.7/pennylane_lightning` and would not be cleaned by this command.

Could we include `python setup.py clean --all` too? Got the following output:
```
/home/antal/anaconda3/lib/python3.7/site-packages/setuptools/dist.py:454: UserWarning: Normalizing '0.11.0-dev' to '0.11.0.dev0'
  warnings.warn(tmpl.format(**locals()))
running clean
removing 'build/temp.linux-x86_64-3.7' (and everything under it)
removing 'build/lib.linux-x86_64-3.7' (and everything under it)
'build/bdist.linux-x86_64' does not exist -- can't clean it
'build/scripts-3.7' does not exist -- can't clean it
removing 'build'
```

trbromley: Actually you're right, now I try again with just `make clean` then it causes me to recompile :thinking: 

Maybe we should just close this PR then?

antalszava: Sounds good

trbromley:
Closing as upon discussion we decided this is not needed.

PR #54, opened 2021-01-21
Further separate function definitions

antalszava:
**Context**

As a preliminary step before adding more test cases and reorganizing the test suite, some function definitions need rearranging.

**Changes**
* To ensure that no compiler errors are raised for multiple function definitions (happens when including the same header in several translation units e.g., separate test files):
    * Function definitions are moved to `lightning_qubit.cpp` from `lightning_qubit.hpp`: the `lightning_qubit.hpp`
    * Smaller functions got `inline`
* The PyBind definitions are separated into a new file (thus PyBind is not a dependency while unit testing with C++)

### DIFF
--- a/pennylane_lightning/src/lightning_qubit.cpp
+++ b/pennylane_lightning/src/lightning_qubit.cpp
@@ -13,10 +13,8 @@
 // limitations under the License.
 /**
  * @file
- * Contains the main `apply()` function for applying a set of operations to a multiqubit
- * statevector.
- *
- * Also includes PyBind boilerplate for interfacing with Python.
+ * Contains the main `apply()` and auxiliary functions for applying a set of
+ * operations to a multiqubit statevector.
  */
 #include "lightning_qubit.hpp"
 
--- a/pennylane_lightning/src/lightning_qubit.cpp
+++ b/pennylane_lightning/src/lightning_qubit.cpp
@@ -18,22 +18,8 @@
  *
  * Also includes PyBind boilerplate for interfacing with Python.
  */
-#include "pybind11/stl.h"
-#include "pybind11/eigen.h"
 #include "lightning_qubit.hpp"
 
-/**
-* Applies specified operations onto an input state of an arbitrary number of qubits.
-*
-* Note that only up to 50 qubits are currently supported. This limitation is due to the Eigen
-* Tensor library not supporting dynamically ranked tensors.
-*
-* @param state the multiqubit statevector
-* @param ops a vector of operation names in the order they should be applied
-* @param wires a vector of wires corresponding to the operations specified in ops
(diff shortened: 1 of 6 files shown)


pennylane_lightning/src/lightning_qubit.cpp
```diff
@@ -99,9 +83,132 @@ VectorXcd apply (
     }
 }
 
+vector<int> calculate_tensor_indices(const vector<int> &wires, const vector<int> &tensor_indices) {
```
antalszava, line 86: These function definitions were moved here from the `.hpp` file. Docstrings were left in `.hpp` with the declaration of each function.

pennylane_lightning/src/lightning_qubit.hpp
```diff
@@ -20,6 +20,7 @@
 #pragma once
 
 #define _USE_MATH_DEFINES
+#include <numeric>
```
antalszava, line 23: Required for `iota` when ran on Windows ([see error in log](https://github.com/PennyLaneAI/pennylane-lightning/runs/1739576949#step:7:835) and [change adding the include guard](https://github.com/PennyLaneAI/pennylane-lightning/pull/54/commits/3b08047d2b1f49b6c6ccf3acb5286689b66ea727)).

josh146 (approved):
@antalszava I did a quick review, and everything looked okay to me. Good idea to separate pybind into its own file for compilation!

However since I'm somewhat unfamiliar with the C++ source code here, better to get @trbromley's thoughts before merging.

pennylane_lightning/src/operations.hpp
```diff
@@ -333,7 +335,7 @@ Gate_Xq<2> CRZ(const double& parameter) {
 * @param omega the third rotation angle
 * @return the controlled rotation tensor
 */
-Gate_Xq<2> CRot(const double& phi, const double& theta, const double& omega) {
+inline Gate_Xq<2> CRot(const double& phi, const double& theta, const double& omega) {
```
trbromley, line 338: Ah so `inline` directly replaces points in the code with the following, rather than calling the function?

trbromley (approved):
:100:

pennylane_lightning/src/operations.hpp
antalszava: Yeah, inlining a function sort of places a request so that the function def is included in every translation unit using it instead of placing a function call and jumping to the function body (mostly recommended for shorter functions).

Also, from [cppref](https://en.cppreference.com/w/cpp/language/inline):
> There may be more than one definition of an inline function or variable (since C++17) in the program as long as each definition appears in a different translation unit and (for non-static inline functions and variables (since C++17)) all definitions are identical. For example, an inline function or an inline variable (since C++17) may be defined in a header file that is #include'd in multiple source files.

PR #61, opened 2021-02-09
No backprop devices

antalszava:
**Context**

As `lighting.qubit` inherits from `default.qubit`, it also inherits the `passthru_devices` entry of the `device.capabilities` dictionary.

This, however, may lead to unexpected behaviour, as the QNode in PennyLane attempts to get a device with the best differentiation method available. Therefore, even without specifying a differentiation method, we may get a `default.qubit.autograd` device.

**Changes**

Removes the `passthru_devices` entry of `device.capabilities`.

### DIFF
--- a/tests/test_apply.py
+++ b/tests/test_apply.py
@@ -454,6 +454,18 @@ class TestLightningQubitIntegration:
         assert dev.analytic
         assert dev.short_name == "lightning.qubit"
 
+    def test_no_backprop(self):
+        """Test that lightning.qubit does not support the backprop
+        differentiation method."""
+
+        dev = qml.device("lightning.qubit", wires=2)
+        def circuit():
+            """Simply quantum function."""
+            return qml.expval(qml.PauliZ(0))
+
+        with pytest.raises(qml.QuantumFunctionError):
+            lightning = qml.QNode(circuit, dev, diff_method="backprop")
+
     def test_args(self):
         """Test that the plugin requires correct arguments"""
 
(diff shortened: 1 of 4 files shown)


josh146:
Didn't seem to fix the windows wheel :(

tests/test_apply.py
```diff
@@ -454,6 +454,18 @@ def test_load_default_qubit_device(self):
...
+        """Test that lightning.qubit does not support the backprop
+        differentiation method."""
+
+        dev = qml.device("lightning.qubit", wires=2)
+        def circuit():
+            """Simple quantum function."""
+            return qml.expval(qml.PauliZ(0))
+
+        with pytest.raises(qml.QuantumFunctionError):
+            lightning = qml.QNode(circuit, dev, diff_method="backprop")
```
josh146, line 467:
```suggestion
            qml.QNode(circuit, dev, diff_method="backprop")
```

tests/test_apply.py
```diff
@@ -454,6 +454,18 @@ def test_load_default_qubit_device(self):
...
+
```
josh146, line 468: Maybe also worth having a test that checks that `diff_method="best"` results in `isinstance(qnode.device, LightningQubit)`? I'm a bit paranoid now 😆

josh146 (approved):
Thanks for getting onto this so quickly @antalszava! Before merging, don't forget to update the changelog.

tests/test_apply.py
antalszava: Good idea, added!

PR #64, opened 2021-02-24
Minor refractor of python loop

albi3ro:
A loop in the python interface was performing more checks than it needed to.  So I moved some things around.

This is just a minor code quality change that should be asymptomatic.

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -116,7 +116,7 @@ class LightningQubit(DefaultQubit):
             super().apply(operations, rotations=rotations, **kwargs)
             return
 
-          # State preparation is currently done in Python
+        # State preparation is currently done in Python
         if isinstance(operations[0], QubitStateVector):
             self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
             del operations[0]
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -117,15 +117,13 @@ class LightningQubit(DefaultQubit):
             return
 
         # State preparation is currently done in Python
-        try:  # if operations[0] doesn't exist, that's ok and suppress the error
+        if operations:  # make sure operations[0] exists
             if isinstance(operations[0], QubitStateVector):
                 self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
                 del operations[0]
             elif isinstance(operations[0], BasisState):
                 self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
                 del operations[0]
-        except IndexError:
-            pass
 
         for operation in operations:
             if isinstance(operation, (QubitStateVector, BasisState)):
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -117,7 +117,7 @@ class LightningQubit(DefaultQubit):
             return
 
         # State preparation is currently done in Python
-        try: # if operations[0] doesn't exist, that's ok and suppress the error
+        try:  # if operations[0] doesn't exist, that's ok and suppress the error
             if isinstance(operations[0], QubitStateVector):
                 self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
                 del operations[0]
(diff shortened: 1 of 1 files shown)


pennylane_lightning/lightning_qubit.py
```diff
@@ -116,21 +116,20 @@ def apply(self, operations, rotations=None, **kwargs):
...
+        for operation in operations[1:]:
             if isinstance(operation, (QubitStateVector, BasisState)):
-                if i == 0:
-
-                    if isinstance(operation, QubitStateVector):
-                        self._apply_state_vector(operation.parameters[0], operation.wires)
-                    else:
-                        self._apply_basis_state(operation.parameters[0], operation.wires)
-
-                    del operations[0]
```
Thenerdstation, line 128: Oof, yeah that's a bug...

pennylane_lightning/lightning_qubit.py
```diff
@@ -116,21 +116,20 @@ def apply(self, operations, rotations=None, **kwargs):
             super().apply(operations, rotations=rotations, **kwargs)
             return
 
-        for i, operation in enumerate(operations):  # State preparation is currently done in Python
+        # State preparation is currently done in Python
```
Thenerdstation, line 119: Maybe also assert len(operations) > 1?

pennylane_lightning/lightning_qubit.py
antalszava: Hmm, it seems to have been okay as it was performed when `if isinstance(operation, (QubitStateVector, BasisState)) and i == 0` is `True`.

pennylane_lightning/lightning_qubit.py
```diff
@@ -116,21 +116,20 @@ def apply(self, operations, rotations=None, **kwargs):
...
+        if isinstance(operations[0], QubitStateVector):
+            self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
+            del operations[0]
+        elif isinstance(operations[0], BasisState):
+            self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
+            del operations[0]
+
+        for operation in operations[1:]:
```
antalszava, line 127: We've already deleted the basis state prep.
```suggestion
        for operation in operations:
```

antalszava, line 127: Care should be taken here, because if there was a basis state prep, that was already deleted.

albi3ro: Thanks for the catch.  Fixed it so that its now just looping over operations.

pennylane_lightning/lightning_qubit.py
albi3ro: using a try-catch now to dismiss index errors.  I'm assuming that's what you were getting at?

pennylane_lightning/lightning_qubit.py
```diff
@@ -116,21 +116,23 @@ def apply(self, operations, rotations=None, **kwargs):
...
+        # State preparation is currently done in Python
+        try:  # if operations[0] doesn't exist, that's ok and suppress the error
+            if isinstance(operations[0], QubitStateVector):
+                self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
+                del operations[0]
+            elif isinstance(operations[0], BasisState):
+                self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
+                del operations[0]
+        except IndexError:
+            pass
```
antalszava, line 128: `operations` should be a list, so we can check `len`
```suggestion
        if len(operations) > 0:
            if isinstance(operations[0], QubitStateVector):
                self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
                del operations[0]
            elif isinstance(operations[0], BasisState):
                self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
                del operations[0]
```

antalszava (approved):
Looks good to me! 💯

pennylane_lightning/lightning_qubit.py
albi3ro: switched to use `if operations:` to check if non-empty.  We don't need to calculate the extensive quantity `len` to check if non-empty.

PR #66, opened 2021-02-26
Add new lightweight backend with performance comparisons to default branch

antalszava:
This PR is identical to #57 except for the `pennylane_lightning/src/rework/GateFactory.cpp` file which was reverted to its state in the  https://github.com/PennyLaneAI/pennylane-lightning/commit/cff579f2c3530b719dd3a42d1bb8ba1790db403e commit. This was required to bypass a segmentation fault error that arose during CI checks.

Further see the [compared diff of this PR and #57](https://github.com/PennyLaneAI/pennylane-lightning/compare/new_backend_with_table...new_backend_original) where the `new_backend_with_table` branch contains the changes applied by #57.

### DIFF
--- a/lightning_benchmark.ipynb
+++ b/lightning_benchmark.ipynb
@@ -1,0 +1,362 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "thrown-argentina",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Requires pip install py-cpuinfo\n",
+    "\n",
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "\n",
+    "import pennylane as qml\n",
+    "import pennylane_lightning\n",
+    "import pennylane_lightning.lightning_benchmark as bench"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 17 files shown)


antalszava:
Hi @ThomasLoke, had to revert the changes to the dispatch table as it turned out that that piece of change caused the segmentation fault. One thing I noticed was that the corresponding `.hpp` file was not changed in the original PR, though wouldn't have expected that to cause such an issue.

Feel free to open another PR that only updates the dispatch table.

ThomasLoke:
Hmm interesting, I didn't run into any segfaults while testing it. Do you have a log of the segfault?

antalszava:
Yes, was also surprised because locally it worked fine, could use the new device on WSL 2. 🤔 Here is the [log on CI](https://github.com/PennyLaneAI/pennylane-lightning/runs/1980294188#step:7:245), the commit right before was the last to [pass the CI checks](https://github.com/PennyLaneAI/pennylane-lightning/runs/1919300494) correctly. Only managed to reproduce this behaviour via trial and error in [a separate PR](https://github.com/PennyLaneAI/pennylane-lightning/pull/65).

ThomasLoke:
I think I managed to narrow it down and fix it--see #68

Issue #60, opened 2021-02-04
Pytest tests hang on Windows


antalszava:
Two test cases were found to cause hanging behaviour when [using `pytest` on a Windows machine](https://github.com/PennyLaneAI/pennylane-lightning/runs/1813396596?check_suite_focus=true):
1. `test_four_qubit_random_circuit` from the PennyLane device test suite;
2. `test_apply_operation_single_wire_with_parameters` from the PennyLane-Lightning test suite when passing the `Rot` gate and the second input parameter.

Running these tests cause the suites to run without termination. When run as a stand-alone case, the code for 1. could be executed with/without using `pytest`.

#59 includes a small fix to split up running 1. and the rest of the device tests when building the wheel.

josh146:
Are the two tests with the issues related? Maybe they are both applying a `Rot` gate with problematic parameters/wires/qubit number?

PR #67, opened 2021-02-26
Replace old backend with new

trbromley:
Fixes https://github.com/PennyLaneAI/pennylane-lightning/issues/60


.github/workflows/wheels.yml
```diff
@@ -9,7 +9,7 @@ env:
   CIBW_SKIP: "cp27-* cp34-* cp35-* *i686 pp* *win32"
 
   # Linux specific build settings
-
+  # TODO - remove eigen
```
antalszava, line 12: Maybe worth merging lines 46 and 47 to check that this PR would indeed fix #60. That separation of tests execution was the hotfix for the issue.

pennylane_lightning/lightning_qubit.py
```diff
@@ -138,7 +116,10 @@ def apply(self, operations, rotations=None, **kwargs):
             self._pre_rotated_state = self._state
 
         if rotations:
-            self._state = self.apply_lightning(self._pre_rotated_state, rotations)
+            if any(isinstance(r, QubitUnitary)for r in rotations):
+                super().apply(operations=[], rotations=rotations)
+            else:
+                self._state = self.apply_lightning(np.copy(self._pre_rotated_state), rotations)
```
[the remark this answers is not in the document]
ThomasLoke: > Being able to pass arbitrary unitaries to the backend would also allow us to directly support QubitUnitary

For gates that need to be defined by a unitary that can't be known ahead of time, passing in the unitary directly is fine.

> also new gates like QFT.

AFAIK, the QFT matrix is just parameterised by its dimension, so I'm not sure I see the need to pass in the matrix directly.

In general, if we can get away with an optimised kernel for gate application, we should. But that's only easily possible for gates that we know ahead of time.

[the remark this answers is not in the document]
trbromley: > For gates that need to be defined by a unitary that can't be known ahead of time, passing in the unitary directly is fine.

Nice! To clarify, this isn't currently supported right? At least, I get an error if I try:
```
TypeError: apply(): incompatible function arguments. The following argument types are supported:
E           1. (arg0: numpy.ndarray[complex128], arg1: List[str], arg2: List[List[int]], arg3: List[List[float]], arg4: int) -> None
```
Since we require `params` to be composed of lists of floats.

> AFAIK, the QFT matrix is just parameterised by its dimension, so I'm not sure I see the need to pass in the matrix directly.
>
> In general, if we can get away with an optimised kernel for gate application, we should. But that's only easily possible for gates that we know ahead of time.

Agreed! I guess what I'm getting at is that we just added `QFT` to PennyLane, and there will inevitably be a gap between gates supported in PL and gates with a defined matrix in the backend. It would be nice for us to have a pipeline to just support _any_ gate that has a matrix method, and we can optionally add in the kernel to the backend if we think it's a popular gate.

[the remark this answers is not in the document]
ThomasLoke: > To clarify, this isn't currently supported right

Yes, that is correct. In theory, it could be made to work with the current pybind bindings, by encoding the matrix in the `vector<double> params` as a flattened array of doubles, but there's little benefit of doing so; modifying the bindings is a trivial thing to do.

> It would be nice for us to have a pipeline to just support any gate that has a matrix method, and we can optionally add in the kernel to the backend if we think it's a popular gate.

Agreed. Conceivably, we could just delegate any unrecognised gate type to the supplied matrix (and error if that's not provided). However, I am wary about always needing to pay the cost of generating the matrix and passing it to the backend, if we don't have to. I don't expect it to be a large penalty (unless we end up supporting controlled unitary operations with arbitrarily many controls), but I'd rather we do without it if possible.

So I think that at the end of the day, what we really want is not to pass the numpy array of the matrix to the backend directly, but rather to pass a functional that when invoked, generates the numpy array for the matrix. This will allow us to lazily generate the matrix as required.

.github/workflows/build.yml
```diff
@@ -16,9 +16,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install dependencies
-        run: sudo apt-get -y -q install cmake gcc libeigen3-dev
```
albi3ro, line 20: Seems like it still uses cmake and gcc? Or are use just assuming the github runner already has those installed?

README.rst
```diff
@@ -65,15 +65,13 @@ Note that subsequent calls to ``pip install -e .`` will use cached binaries stor
...
 * A C++ compiler, such as ``g++``, ``clang``, or ``MSVC``.
-* `Eigen3 <http://eigen.tuxfamily.org/index.php?title=Main_Page>`_ - a C++ header library for linear algebra.
 * `pybind11 <https://pybind11.readthedocs.io/en/stable/>`_ a library for binding C++
   functionality to Python.
 
 On Debian-based systems, these can be installed via ``apt`` and ``pip``:
 
 .. code-block:: console
 
-    $ sudo apt install g++ libeigen3-dev
```
albi3ro, line 76: Should we still have 
`$ sudo apt install g++` ?

README.rst
```diff
@@ -65,15 +65,13 @@ Note that subsequent calls to ``pip install -e .`` will use cached binaries stor
 The following dependencies are required to install PennyLane-Lightning:
 
 * A C++ compiler, such as ``g++``, ``clang``, or ``MSVC``.
```
albi3ro, line 67: Should we mention we want version 2014 and above now?

antalszava: So far the C++14 features that are used by the new backend could be replaced by a custom function. So we could stay with C++11 for now (unless another need arises).

pennylane_lightning/src/Apply.cpp
```diff
@@ -11,43 +11,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#include <set>
-
 #include "Apply.hpp"
 #include "Gates.hpp"
 #include "StateVector.hpp"
 #include "Util.hpp"
+#include "Util.cpp"
```
antalszava, line 18:
```suggestion
```

pennylane_lightning/src/tests/Makefile
```diff
@@ -11,26 +11,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-EIGEN_INCLUDE_DIR ?= /usr/include/eigen3
 CC = g++
 CFLAGS = -std=c++11 -g -Wall -fopenmp -fPIC -I/usr/include \
-	-I../include -I$(EIGEN_INCLUDE_DIR) -I$(GOOGLETEST_DIR)/include \
```
antalszava, line 17: 🥳

tests/test_apply.py
```diff
@@ -1212,42 +1212,3 @@ def test_pauliz_hadamard(self, theta, phi, varphi, monkeypatch, qubit_device_3_w
             - 2 * np.cos(theta) * np.sin(phi) * np.sin(2 * varphi)
         ) / 4
         assert np.allclose(var, expected, atol=tol, rtol=0)
-
```
antalszava, line 1215: 🥳

antalszava (commented):
Looking good so far! 💯 Will do another round after tidying. :)

pennylane_lightning/lightning_qubit.py
```diff
@@ -112,14 +94,10 @@ def capabilities(cls):
...
 
-        if self.num_wires > self._MAX_WIRES:
-            super().apply(operations, rotations=rotations, **kwargs)
-            return
-
         # State preparation is currently done in Python
         if operations:  # make sure operations[0] exists
             if isinstance(operations[0], QubitStateVector):
-                self._apply_state_vector(operations[0].parameters[0], operations[0].wires)
+                self._apply_state_vector(operations[0].parameters[0].copy(), operations[0].wires)
```
trbromley, line 100: This is due to https://github.com/PennyLaneAI/pennylane-lightning/pull/81

pennylane_lightning/lightning_qubit.py
```diff
@@ -138,7 +116,10 @@ def apply(self, operations, rotations=None, **kwargs):
             self._pre_rotated_state = self._state
 
         if rotations:
-            self._state = self.apply_lightning(self._pre_rotated_state, rotations)
+            if any(isinstance(r, QubitUnitary) for r in rotations):
+                super().apply(operations=[], rotations=rotations)
+            else:
+                self._state = self.apply_lightning(np.copy(self._pre_rotated_state), rotations)
```
antalszava, line 122: This is getting me wondering: each time we call `self.apply_lightning`, we are doing conversion between Python and C++. If we combined `operations` and `rotations`, then we could get away with only a single conversion. In exchange, it would not be so simple to obtain the `_pre_rotated_state` as before.

pennylane_lightning/lightning_qubit.py
```diff
@@ -138,7 +116,10 @@ def apply(self, operations, rotations=None, **kwargs):
             self._pre_rotated_state = self._state
 
         if rotations:
-            self._state = self.apply_lightning(self._pre_rotated_state, rotations)
+            if any(isinstance(r, QubitUnitary) for r in rotations):
+                super().apply(operations=[], rotations=rotations)
```
antalszava, line 120: So basically this allowed supporting `Hermitian`, correct?

antalszava (approved):
Looks good to me! 💯 🍾

pennylane_lightning/lightning_qubit.py
trbromley: Agreed! Could we just return the pre-rotated state and the post-rotated state? Also I recall some discussion on the rotations API changing, but can't remember the decision?

antalszava: > Could we just return the pre-rotated state and the post-rotated state?

We should be able to, though it might require some thought to 1. still minimize the number of copies made of the state vector 2. keep the code readable.

A naive way would be to make the bound `apply` C++ function return the pre-rotated state *and* change the state vector in place to yield the post-rotated state. However, wondering if such a function would lead to much confusion because we'd basically return an intermediate result of the computation.

Also unsure at this moment on how much overhead we're actually incurring from calling on bound C++ functions twice.

> Also I recall some discussion on the rotations API changing, but can't remember the decision?

Not quite sure about this one. 🤔

CHANGELOG.md
```diff
@@ -6,21 +6,24 @@
 
 ### Improvements
 
-* Add new lightweight backend with performance improvements
+* Add new lightweight backend with performance improvements.
   [(#57)](https://github.com/PennyLaneAI/pennylane-lightning/pull/57)
 
+* Remove the previous backend using Eigen.
+  [(#57)](https://github.com/PennyLaneAI/pennylane-lightning/pull/67)
```
trbromley, line 13:
```suggestion
* Remove the previous Eigen-based backend.
  [(#67)](https://github.com/PennyLaneAI/pennylane-lightning/pull/67)
```

co9olguy:
:tada:

PR #72, opened 2021-03-02
Add test for Util.hpp

trbromley:
This PR adds some tests for the new backend in `lightning_unittest_new.cpp`.


pennylane_lightning/src/tests/lightning_unittest.cpp
```diff
@@ -11,1066 +11,36 @@
...
-    expected_states[6] = target_state_110;
-    expected_states[7] = target_state_111;
-
-    for (int i=0; i < 8; i++) {
-        auto expected_vector = vectorize(expected_states[i]);
-        auto output_state = vectorize(output_states[i]);
-        EXPECT_TRUE(expected_vector.isApprox(output_state, tol));
+    for (size_t i = 0; i < inputs.size(); i++) {
+        size_t result = Pennylane::exp2(inputs[i]);
+        EXPECT_TRUE(result == outputs[i]);
```
trbromley, line 25: This is what the new file looks like:
https://github.com/PennyLaneAI/pennylane-lightning/blob/72c31a64b62878bde10c646ea33a5983bee95f10/pennylane_lightning/src/tests/lightning_unittest.cpp

albi3ro (commented):
Can you update the makefile?

Also, should we have separate test files for each cpp file we need to test, ex `lightning_util_unittest.cpp`?  Keeping files from getting too long improves readability.

Thenerdstation (approved):
LGTM

trbromley:
Thanks @albi3ro @Thenerdstation! I've updated the makefile.

Let's merge this in and split out into multiple files as we add more.

trbromley:
@albi3ro, I renamed the test to `lightning_util_unittest.cpp` after seeing your PR :+1:

PR #73, opened 2021-03-02
Gates unit tests

albi3ro:
Currently, the `lightning_gates.unittest.cpp` file has been added to the makefile, through the command `make test_new`.

The file also has an example test that should always work.  This allows me to check everything is setup correctly.


pennylane_lightning/src/tests/Makefile
```diff
@@ -24,15 +24,24 @@ all: test
...
 
+cpptests_new: gtest_main.o lightning_gates_unittest.o
+	$(CC) $^ -o $@ $(LFLAGS)
+
 test: cpptests
 	./cpptests
 
+test_new: cpptests_new
+	./cpptests_new
+
```
trbromley, line 44: Do we want to separate this out? Currently, I do `make test-cpp` from the top-level directory, which I believe wouldn't run these tests if we make a `test_new` and `cpptests_new`. In other words, why don't we put everything in `cpptests`, which is accessible by the main Makefile? (I'm also conscious that eventually the new backend will no longer be new, and not wanting to delete lots of `_new` everywhere)

albi3ro: changed the makefile so `lightning_gates_unittest` is now run with `make test`

antalszava:
Hi @albi3ro, the error `undefined reference to `Pennylane::constructGate` arises because the `.cpp` file is not found during compilation.

To aid this and to be able to compile, a couple of things needs to be done:
1. A new object file needs to be created in the `Makefile`, e.g.:
```
Gates.o: ../rework/Gates.cpp
       $(CC) $^ $(CFLAGS) -c

```
2. The def of `cpptests` needs to be updated: `cpptests: gtest_main.o Gates.o lightning_unittest.o lightning_gates_unittest.o`
3. Due to the used `make_unique` C++ 14 feature, the ` -std=c++14` needs to be used as part of `CFLAGS`. For now, this change could help testing, but let's take a bit to think if we'd indeed want to make this change or rather replace the occurrences of `make_unique` with a custom solution.

Overall, as more and more source files will be added, the `Makefile` will become unwieldy, so it will be nice for us to look at another solution to compiling with all the sources and linking correctly.

pennylane_lightning/src/tests/TestingUtils.h
```diff
@@ -0,0 +1,34 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+// Helper macro to check if an exception is thrown with the correct complete error message
+#define EXPECT_THROW_WITH_MESSAGE(stmt, etype, whatstring) EXPECT_THROW( \
```
antalszava, line 17: Added as per https://github.com/google/googletest/issues/952#issuecomment-521361666 and the thread there. Any merged features related to this were before the release version that we are pulling at the moment.

pennylane_lightning/src/rework/Util.hpp
```diff
@@ -44,3 +44,9 @@ namespace Pennylane {
     }
 
 }
+
+// Helper similar to std::make_unique from c++14
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args) {
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
```
Thenerdstation, line 52: Why did we add this? I thought antal removed c++11 support.

pennylane_lightning/src/tests/TestingUtils.h
```diff
@@ -0,0 +1,34 @@
...
+    , etype)
+
+// Helper macro to check if an exception is thrown with the correct error message substring
+#define EXPECT_THROW_WITH_MESSAGE_SUBSTRING(stmt, etype, whatstring) EXPECT_THROW( \
+        try { \
+            stmt; \
+        } catch (const etype& ex) { \
+            EXPECT_NE(std::string(ex.what()).find(whatstring), std::string::npos); \
+            throw; \
+        } \
```
Thenerdstation, line 33: What's the purpose of the backticks?

pennylane_lightning/src/tests/lightning_gates_unittest.cpp
```diff
@@ -0,0 +1,159 @@
...
+        ::testing::Values(
+                std::make_tuple("RX", RX, ONE_PARAM),
+                std::make_tuple("RY", RY, ONE_PARAM),
+                std::make_tuple("RZ", RZ, ONE_PARAM),
+                std::make_tuple("PhaseShift", PhaseShift, ONE_PARAM),
+                std::make_tuple("Rot", Rot, THREE_PARAMS),
+                std::make_tuple("CRX", CRX, ONE_PARAM),
+                std::make_tuple("CRY", CRY, ONE_PARAM),
+                std::make_tuple("CRZ", CRZ, ONE_PARAM),
+                std::make_tuple("CRot", CRot, THREE_PARAMS)
```
Thenerdstation, line 96: Should we consider adding more tests with variable param values?

pennylane_lightning/src/tests/lightning_gates_unittest.cpp
```diff
@@ -0,0 +1,159 @@
...
+// Parameter length validation
+
+class NumParamsThrowsFixture : public ::testing::TestWithParam<std::tuple<string, vector<double> > > {
+};
+
+TEST_P(NumParamsThrowsFixture, CheckParamLength) {
+    const string gate_name = std::get<0>(GetParam());
+    const vector<double> params = std::get<1>(GetParam());
+
+    EXPECT_THROW_WITH_MESSAGE_SUBSTRING(Pennylane::constructGate(gate_name, params), std::invalid_argument, gate_name);
```
Thenerdstation, line 110: Good test to include 👍

pennylane_lightning/src/tests/lightning_gates_unittest.cpp
```diff
@@ -0,0 +1,159 @@
...
+using std::string;
+using std::function;
+
+using Pennylane::CplxType;
+
+namespace test_gates{
+
+const vector<double> ZERO_PARAM = {};
+const vector<double> ONE_PARAM = {0.123};
+const vector<double> THREE_PARAMS = {0.123, 2.345, 1.4321};
```
Thenerdstation, line 32: (See comment below). Should we add additional values here?

pennylane_lightning/src/tests/TestingUtils.h
thisac: Strange that this isn't part of gtest already. 🤔

pennylane_lightning/src/tests/TestingUtils.h
thisac: I'm wondering the same. I would think it's just for line-continuation, but it doesn't make sense having it here, right?

pennylane_lightning/src/tests/lightning_gates_unittest.cpp
```diff
@@ -0,0 +1,159 @@
...
+                std::make_tuple("PauliZ", PauliZ),
+                std::make_tuple("Hadamard", Hadamard),
+                std::make_tuple("S", S),
+                std::make_tuple("T", T),
+                std::make_tuple("CNOT", CNOT),
+                std::make_tuple("SWAP", SWAP),
+                std::make_tuple("CZ", CZ),
+                std::make_tuple("Toffoli", Toffoli),
+                std::make_tuple("CSWAP", CSWAP)
+));
```
thisac, line 66: 🚀

thisac (commented):
Nice use of parameterized testing! 🎖️

pennylane_lightning/src/rework/Util.hpp
antalszava: This was added so that we can keep compiling with c++11. `std::make_unique` was the only feature we would need from c++14 at the moment.

pennylane_lightning/src/tests/lightning_gates_unittest.cpp
antalszava: We could, although personally, I wouldn't worry too much about it. These tests simply check that the matrices are okay, we could test applying the gates more extensively. Having said that I could add more params if others also agreed that it's important.

Thenerdstation: Fair

pennylane_lightning/src/tests/TestingUtils.h
antalszava: The backslashes are indeed added for line continuation because this is a macro so it gets blindly replaced to the code defined before compilation.

pennylane_lightning/src/rework/Util.hpp
trbromley: :thinking: but aren't we already compiling in c++11 in `master` and things are working?

pennylane_lightning/src/tests/GateData.h
```diff
@@ -0,0 +1,214 @@
+// Copyright 2021 Xanadu Quantum Technologies Inc.
```
trbromley, line 1: Should the file be `.hpp`?

pennylane_lightning/src/tests/GateData.h
```diff
@@ -0,0 +1,214 @@
...
+
+static const vector<CplxType> CSWAP = {
+    1, 0, 0, 0, 0, 0, 0, 0,
+    0, 1, 0, 0, 0, 0, 0, 0,
+    0, 0, 1, 0, 0, 0, 0, 0,
+    0, 0, 0, 1, 0, 0, 0, 0,
+    0, 0, 0, 0, 1, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 1, 0,
+    0, 0, 0, 0, 0, 1, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 1 };
```
trbromley, line 213: Part of me wonders if this is recreating what we're doing in `Gates.cpp` :thinking: But then, I'm not sure how we could test better, and maybe doing something twice and seeing if they agree is anyway a nice type of test!

One question, did you recreate these matrices from `Gates.cpp`, or from another source like PennyLane core? (The latter being preferable!).

pennylane_lightning/src/tests/Makefile
```diff
@@ -24,13 +24,16 @@ all: test
 lightning_unittest.o: lightning_unittest.cpp
 	$(CC) $^ $(CFLAGS) -c
 
-lightning_util_unittest.o: lightning_util_unittest.cpp
```
trbromley, line 27: I think we still want to keep `lightning_util_unittest`?

pennylane_lightning/src/tests/Makefile
```diff
@@ -24,13 +24,16 @@ all: test
...
+Gates.o: ../rework/Gates.cpp
+	$(CC) $^ $(CFLAGS) -c
+
+lightning_gates_unittest.o: lightning_gates_unittest.cpp
 	$(CC) $^ $(CFLAGS) -c
 
 gtest_main.o: gtest_main.cpp
 	$(CC) $^ $(CFLAGS) -c
 
-cpptests: gtest_main.o lightning_unittest.o lightning_util_unittest.o
```
trbromley, line 33: I'm confused about `Gates.o` :thinking:

trbromley (commented):
Thanks @albi3ro, looks great - just have a couple of questions. :100:

pennylane_lightning/src/rework/Util.hpp
antalszava: We aren't compiling the `rework/Gates.cpp` source on `master` at the moment when running the tests.

pennylane_lightning/src/tests/GateData.h
antalszava: Yeah, fair point! Thought that we are doing the same anyways in PennyLane core too, so went with this strategy.

> One question, did you recreate these matrices from Gates.cpp, or from another source like PennyLane core? (The latter being preferable!).

Transformed the matrix defs we had with the previous backend (avoided copying from the new backend defs).

trbromley: Great, I'm very happy if we avoided using the new backend ones :+1:

pennylane_lightning/src/tests/Makefile
antalszava: We need to compile (`Gates.o: ../rework/Gates.cpp` option) and link (here adding `Gates.o`) the `.cpp` sources that we depend on in our testing. In this PR it's `Gates.cpp`.

Previously, we could get away with not doing this in the case of `lightning_unittest.cpp` because tests in there only depended on definitions that were available in `.hpp` files. For example,  ​the `lightning_quibt.cpp` file only contained the definition and binding of the `apply` function: a utility function that was a switch. All other function definitions were contained in `lightning_quibt.hpp`, so we didn't need to separately compile & link this file or any other sources.

This is now different with the new backend.

pennylane_lightning/src/tests/TestingUtils.h
antalszava: Something seems to have been added, but the activity was after the latest release of `gtest` available on GitHub. So I haven't really explored too much on what they have added.

pennylane_lightning/src/tests/GateData.h
antalszava: Sure! Changed

pennylane_lightning/src/tests/Makefile
antalszava: Great catch! Changed

pennylane_lightning/src/tests/GateData.hpp
```diff
@@ -0,0 +1,214 @@
...
+    double parameter = pars.at(0);
+
+    const double c = std::cos(parameter / 2);
+    const double s = std::sin(parameter / 2);
+    return {c, -s, s, c};
+}
+
+vector<CplxType> RZ(const vector<double> & pars){
+    double parameter = pars.at(0);
+    const std::complex<double> exponent(0, -parameter/2);
```
albi3ro, line 59: Given a variable is defined above named exponent in the global scope, maybe the local variables should have a different name to improve readability and prevent confusion.

Maybe here they could be phase_1 and phase_2?

pennylane_lightning/src/tests/Makefile
albi3ro: Thanks for figuring that out :) Glad I'm not the only one who got confused by that.

pennylane_lightning/src/tests/GateData.hpp
antalszava: Good idea! Updated.

pennylane_lightning/src/tests/TestingUtils.h
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 Xanadu Quantum Technologies Inc.
```
trbromley, line 1: Also `.hpp`?

trbromley (approved):
:100:

jediMunees:
EXPECT_EQ is expecting "expected_value, actual_value". But the fix is having in reverse order.

Isn't supposed to be like this "EXPECT_EQ(whatstring, std::string(ex.what()));" as below:

`#define EXPECT_THROW_WITH_MESSAGE(stmt, etype, whatstring) EXPECT_THROW( \
        try { \
            stmt; \
        } catch (const etype& ex) { \
            EXPECT_EQ(whatstring, std::string(ex.what())); \
            throw; \
        } \
    , etype)`

antalszava:
@jediMunees good catch! :) Feel free to open a PR that swaps `whatstring` and `std::string(ex.what())`.

jediMunees:
> @jediMunees good catch! :) Feel free to open a PR that swaps `whatstring` and `std::string(ex.what())`.

Please review at https://github.com/PennyLaneAI/pennylane-lightning/pull/101

PR #75, opened 2021-03-02
Get gtest 1.10.0 and parametrize tests

antalszava:
**Changes**

* Changes to CI to get the latest release of Google test available on GitHub (`v1.10.0`), [see release notes](https://github.com/google/googletest/releases/tag/release-1.10.0)
* Creates parametrized tests

### DIFF
--- a/pennylane_lightning/src/tests/lightning_util_unittest.cpp
+++ b/pennylane_lightning/src/tests/lightning_util_unittest.cpp
@@ -19,7 +19,7 @@ namespace test_utils {
 class Exp2TestFixture :public ::testing::TestWithParam<std::tuple<int, int>> {
 };
 
-TEST_P(Exp2TestFixture, CheckPower) {
+TEST_P(Exp2TestFixture, CheckExp2Results) {
     int input = std::get<0>(GetParam());
     int expected = std::get<1>(GetParam());
     ASSERT_EQ(expected, Pennylane::exp2(input));
--- a/pennylane_lightning/src/tests/lightning_util_unittest.cpp
+++ b/pennylane_lightning/src/tests/lightning_util_unittest.cpp
@@ -14,6 +14,8 @@
 #include "gtest/gtest.h"
 #include "../rework/Util.hpp"
 
+#include <tuple>
+
 namespace test_utils {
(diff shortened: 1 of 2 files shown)


pennylane_lightning/src/tests/lightning_util_unittest.cpp
```diff
@@ -14,33 +14,47 @@
...
+class Exp2TestFixture :public ::testing::TestWithParam<std::tuple<int, int>> {
+};
 
-    for (size_t i = 0; i < inputs.size(); i++) {
-        size_t result = Pennylane::exp2(inputs[i]);
-        EXPECT_TRUE(result == outputs[i]);
-    }
+TEST_P(Exp2TestFixture, CheckExp2Results) {
+    int input = std::get<0>(GetParam());
+    int expected = std::get<1>(GetParam());
```
Thenerdstation, line 26: `std::get<0>(GetParam()):` is the python equivalent of `a[0]` lol

Thenerdstation (approved):

pennylane_lightning/src/tests/lightning_util_unittest.cpp
trbromley, line 26: Do you get any warnings when using an `int` type? For me, it was giving a warning of comparing `int` and `size_t`.

trbromley (approved):
Thanks @antalszava, looks great!

I'm still not convinced of the benefits of this approach versus the "good old for loop", since for me there is also a cost to this approach of more complexity and less understandable code.

But maybe where this approach shines is in more advanced test cases. I'm good to go with this approach for now.

pennylane_lightning/src/tests/lightning_util_unittest.cpp
antalszava: Yep, same here, warnings are emitted there.

antalszava:
@trbromley agreed that a `for` loop is actually briefer in these cases! One advantage with parameterized test cases is that they can be easier to parse on test failure as the cases are separated based on parameters. Though in the cases of this PR, I agree that there is not such a big difference.

PR #77, opened 2021-03-03
Cpp14

antalszava:


trbromley:
Will updating to `c++14` have any issues with our CI and deployment? I have some memory of a discussion like this when we first started lightning, and `c++11` was the compatible option.

PR #74, opened 2021-03-02
Add tests for Apply.cpp

trbromley:

PR #79, opened 2021-03-03
Move extra functions out of Apply

trbromley:
This PR moves the following functions:

- `getIndicesAfterExclusion` is moved to `Util.cpp`
- `generateBitPatterns` is moved to `Util.cpp`

The motivation this is that these functions feel closer to utility functions than something that should go within `Apply`.
Undercover motivation - we're also struggling to make PyBind be compatible with GoogleTest!


pennylane_lightning/src/rework/Apply.cpp
```diff
@@ -11,43 +11,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#include <set>
-
 #include "Apply.hpp"
 #include "Gates.hpp"
 #include "StateVector.hpp"
 #include "Util.hpp"
+#include "Util.cpp"
```
trbromley, line 18: Do we have to include this? I couldn't work out how not to :thinking:

pennylane_lightning/src/rework/Apply.cpp
```diff
@@ -63,7 +36,7 @@ void Pennylane::constructAndApplyOperation(
     
     vector<size_t> internalIndices = generateBitPatterns(opWires, qubits);
 
-    vector<unsigned int> externalWires = getIndicesAfterExclusion(opWires, qubits);
+    vector<unsigned int> externalWires = Pennylane::getIndicesAfterExclusion(opWires, qubits);
     vector<size_t> externalIndices = generateBitPatterns(externalWires, qubits);
```
Thenerdstation, line 40:
```suggestion
    vector<size_t> externalIndices = Pennylane::generateBitPatterns(externalWires, qubits);
```

pennylane_lightning/src/rework/Apply.cpp
albi3ro: I don't think we have to include the cpp file here, but we do have to add it to the compilation step.  

I think....

ThomasLoke: If we have to include the cpp file here then something is very wrong indeed. What's wrong with just including the header file?

ThomasLoke (commented):
> The motivation this is that these functions feel closer to utility functions than something that should go within Apply.

Personally, I think that they do belong with Apply, as that's the only point in which we need to generate bit patterns. I'd be wary of extracting everything that looks vaguely generalisable to a monolithic Utils class. Happy to disagree though!

> Undercover motivation - we're also struggling to make PyBind be compatible with GoogleTest!

I'd much rather we just extract the PyBind bindings into a separate file instead!

pennylane_lightning/src/rework/Apply.cpp
antalszava: Yep, agree that this should not be included. If there are undefined references at the linking step, that means that the `.cpp` file was not compiled and the linker cannot find object files that were the result of compilation. So `Util.cpp` would have to be passed with other sources to the compiler e.g., `g++ main.cpp Util.cpp` for a basic command that would use two sources and no external libraries.

pennylane_lightning/src/rework/Apply.cpp
```diff
@@ -63,7 +36,7 @@ void Pennylane::constructAndApplyOperation(
     
     vector<size_t> internalIndices = generateBitPatterns(opWires, qubits);
 
-    vector<unsigned int> externalWires = getIndicesAfterExclusion(opWires, qubits);
+    vector<unsigned int> externalWires = Pennylane::getIndicesAfterExclusion(opWires, qubits);
```
antalszava, line 39: Was is it required to add the `Pennylane::` scope here?

pennylane_lightning/src/rework/Apply.cpp
trbromley: Removed the `Util.cpp`, the solution was that I forgot to update `setup.py`.

pennylane_lightning/src/rework/Apply.cpp
trbromley: Nope, removed.

trbromley:
> > The motivation this is that these functions feel closer to utility functions than something that should go within Apply.
> 
> Personally, I think that they do belong with Apply, as that's the only point in which we need to generate bit patterns. I'd be wary of extracting everything that looks vaguely generalisable to a monolithic Utils class. Happy to disagree though!

Thanks @ThomasLoke! Yes, I guess this is mainly motivated out of "necessity" as I can't find a way to get PyBind and GoogleTest to work together :thinking: 

> I'd much rather we just extract the PyBind bindings into a separate file instead!

Would that involve moving `PYBIND11_MODULE()`, `apply()`, and `constructAndApplyOperation()` into a separate file, since each of these is dependent on PyBind? If so, is that essentially equivalent to moving the complementary two functions to `Util`?

However, I'm really happy if we can find a more natural way than what's done in this PR.

ThomasLoke:
> Would that involve moving PYBIND11_MODULE(), apply(), and constructAndApplyOperation() into a separate file, since each of these is dependent on PyBind? If so, is that essentially equivalent to moving the complementary two functions to Util?

It involves just moving PYBIND_MODULE() to a separate cpp file (say PyBind.cpp). apply() and constructAndApplyOperation() has no dependence on PYBIND11_MODULE, so they can stay where they are in Apply.hpp. In that separate cpp file, From my testing, this was all that was needed in PyBind.cpp:

```
#include "Apply.hpp"

PYBIND11_MODULE(lightning_qubit_new_ops, m)
{
    m.doc() = "lightning.qubit apply() method";
    m.def("apply", Pennylane::apply, "lightning.qubit apply() method");
}
```

and after including this source file in `setup.py`, it seems to build fine.

trbromley:
Thanks @ThomasLoke,

> If its a problem with including the pybind header files themselves

Yes, it's exactly this. For example, we want to add unit tests for `Apply` in this PR: https://github.com/PennyLaneAI/pennylane-lightning/pull/74. However, when running `make test-cpp` we get the error:

```
../Apply.hpp:26:10: fatal error: pybind11/complex.h: No such file or directory
   26 | #include "pybind11/complex.h"
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

This makes sense since we haven't exposed the compiler to PyBind. The natural thing to do is add pybind to the makefile, e.g., as we do in this commit 90615c943bfe56b4898447cc1dc91fbaefdda0ea. However, then we get a very long sequence of errors that look like:
`undefined reference to 'PyErr_Clear'`.

Our ideal scenario would be to just leave `Apply` as-is and work out how to compile. Failing that, I'd be happy with moving the relevant functionality to `PyBind.cpp` or moving some of the functions over to `Util.cpp` (i.e., as in this PR).

Also looping in @antalszava who is going to try fixing this.

ThomasLoke:
> This makes sense since we haven't exposed the compiler to PyBind. The natural thing to do is add pybind to the makefile, e.g., as we do in this commit 90615c9. However, then we get a very long sequence of errors that look like:
undefined reference to 'PyErr_Clear'.

Unfortunately, I'm having jack-all results trying to get the makefile for the tests to build on my windows machine. And the [corresponding checks](https://github.com/PennyLaneAI/pennylane-lightning/pull/74/checks?check_run_id=2058835295) I see for that commit indicate that they failed to resolve the pybind include directories? But anyway. Have you tried compiling a minimal example that references a pybind library without gtest? i.e. something like a `test.cpp`:

```c++
#include "pybind/complex.h"

int main() {
    return 0;
}
```

then adding a make target for that which includes the pybind libraries without gtest, and seeing if that works?

Either way, more detailed build logs would help to clarify the issue.

antalszava:
@ThomasLoke tried with another header, though will explore more, thank you!

> Either way, more detailed build logs would help to clarify the issue.

Another approach that we're considering is to start using `cmake` that could help also with building on Windows seamlessly.

----

@ThomasLoke Slightly unrelated to this PR and more on the performance side: any thoughts on what could be the best next step to fasten [performing the matrix-vector multiplication](https://github.com/PennyLaneAI/pennylane-lightning/blob/ffc69a57dec15951ff8b67fb6356df9d8a6c20f0/pennylane_lightning/src/Apply.cpp#L70) even further? In specific considering to use OpenMP to perform the gate application in a multi-threaded way for example.

Was also wondering how the current approach would compare to using BLAS & Lapack to perform (chunks of) the multiplication, could that be worth benchmarking?

ThomasLoke:
> In specific considering to use OpenMP to perform the gate application in a multi-threaded way for example.

That's probably worthwhile doing, regardless of whatever other optimisations are applied.

> Was also wondering how the current approach would compare to using BLAS & Lapack to perform (chunks of) the multiplication, could that be worth benchmarking?

It might. But I suspect writing specialised kernels for each gate would be the more worthwhile step, rather than continuing to naively write gate applications as matrix-vector multiplications. Though at some point, we'll probably need a generic matrix-vector multiplication kernel to support arbitrary gates for which we don't have an associated kernel anyway.

pennylane_lightning/src/Util.hpp
```diff
@@ -17,10 +17,38 @@
  */
 #pragma once
 
+#include <memory>
```
albi3ro, line 20: What's `<memory>` for?  I see `unique_ptr` is in the memory std, but elsewhere its getting imported with `using std::unique_ptr;`.

antalszava: That's the header itself, required for `unique_ptr`. `using std::unique_ptr;` introduces `unique_ptr` from the `std` namespace.

trbromley:
Hey @ThomasLoke, just to update you that we're still struggling to have compatibility between gtest and pybind - now seeing strange [linking errors](https://github.com/PennyLaneAI/pennylane-lightning/pull/74/checks?check_run_id=2087395068) in https://github.com/PennyLaneAI/pennylane-lightning/pull/74.

Since we want to prioritize developing new features rather than getting stuck on tests, we're planning to merge in this PR and provide some quick tests for `generateBitPatterns` and `getIndicesAfterExclusion` without having to worry about pybind.

ThomasLoke:
> we're still struggling to have compatibility between gtest and pybind - now seeing strange linking errors in #74.

Does the minimal example above compile if you try linking in the pybind supplied paths in the g++ compilation step? I guess I'm trying to be sure that the problem is some sort of conflict between gtest and pybind (which at first glance, I don't see any evidence of in the aforementioned linking errors), rather than something else entirely.

Either way, I think there's value in moving pybind specific things into a single cpp file that's responsible for converting between pybind data structures and custom data structures, and have the rest of the codebase live in ignorance of pybind altogether.

ThomasLoke:
To be clear, this was what I had in mind: [b8371ff6](https://github.com/ThomasLoke/pennylane-lightning/commit/b8371ff6ae3868921e2cca185a9b5a29db8ba6bd)

trbromley:
Thanks @ThomasLoke, awesome! It works, I turned it into a PR: https://github.com/PennyLaneAI/pennylane-lightning/pull/86. We can close this PR now.

PR #83, opened 2021-03-10
[WIP] Add the adjoint method

thisac:
Adds the adjoint method to `lightning.qubit`.

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -16,9 +16,10 @@ This module contains the :class:`~.LightningQubit` class, a PennyLane simulator 
 interfaces with C++ for fast linear algebra calculations.
 """
 from pennylane.devices import DefaultQubit
-from .lightning_qubit_ops import apply
+from .lightning_qubit_ops import apply, adjoint_jacobian
 import numpy as np
-from pennylane import QubitStateVector, BasisState, DeviceError, QubitUnitary
+from pennylane import QubitStateVector, BasisState, DeviceError, QubitUnitary, QuantumFunctionError
+from pennylane.operation import Expectation
 
 from ._version import __version__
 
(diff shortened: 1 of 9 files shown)


pennylane_lightning/src/Bindings.cpp
```diff
@@ -46,8 +46,34 @@ void apply(
     Pennylane::apply(state, ops, wires, params, inverse, qubits);
 }
 
+vector<double> Pennylane::adjointJacobian(
```
antalszava, line 49:
```suggestion
vector<double> adjointJacobian(
```

pennylane_lightning/src/Apply.cpp
```diff
@@ -106,3 +112,101 @@ void Pennylane::apply(
...
+
+        Pennylane::constructAndApplyOperation(
+            phiCopy,
+            observables[i],
+            obsWires[i],
+            obsParams[i],
+            false,
+            obsWires[i].size()
+        );
+        lambdas.push_back(phiCopy);
```
albi3ro, line 146: So we know how large `lambdas` is going in.  Why not preallocate and fill? Is there a reason we need vector functionality?

pennylane_lightning/src/Apply.cpp
```diff
@@ -106,3 +112,101 @@ void Pennylane::apply(
...
+    }
+
+    for (int i = operations.size() - 1; i >= 0; i--) {
+        if (opParams[i].size() > 1) {
+            throw std::invalid_argument("The operation is not supported using the adjoint differentiation method");
+        } else if ((operations[i] != "QubitStateVector") && (operations[i] != "BasisState")) {
+            // copy |phi> to |mu> before applying Uj*
+            CplxType* phiCopyArr = new CplxType[phi.length];
```
albi3ro, line 154: Could you just allocate the memory once and then always copy the `phi` statevector there, overwriting it's value from the last loop?

pennylane_lightning/lightning_qubit.py
```diff
@@ -139,6 +140,57 @@ def apply_lightning(self, state, operations):
...
+        operations, op_params, op_wires = map(list, zip(*op_data))
+
+        obs_data = [(self._remove_inverse_string(obs.name), _unwrap(obs.parameters), obs.wires.tolist()) for obs in tape.observables]
+        observables, obs_params, obs_wires = map(list, zip(*obs_data))
+
+        trainable_params = list(tape.trainable_params)
+
+        # send in flattened array of zeros to be populated by adjoint_jacobian
+        jac = np.zeros(len(tape.observables) * len(tape.trainable_params))
+        adjoint_jacobian(
```
albi3ro, line 180: Maybe have this named something slightly different than the method?

pennylane_lightning/src/Apply.cpp
```diff
@@ -106,3 +112,101 @@ void Pennylane::apply(
...
+            Pennylane::constructAndApplyOperation(
+                phi,
+                operations[i],
+                opWires[i],
+                opParams[i],
+                true,
+                opWires[i].size()
+            );
+
+            if (std::find(trainableParams.begin(), trainableParams.end(), paramNumber) != trainableParams.end()) {
```
albi3ro, line 168: I think here you could take advantage of the fact that `trainableParams` is ordered and `paramNumber` is continuously decreasing. 

You could have a `next_trainableParam_index` and compare `trainableParams[next_trainableParam_index]` to `paramNumber`.  If `paramNumber` is trainable, you decrease `next_trainable_param_index`.

pennylane_lightning/src/Apply.cpp
albi3ro: Ok. I think I see.  `lambdas` is a vector of statevectors? Why not just have an array with additional dimension?

PR #84, opened 2021-03-10
PauliX with swaps

antalszava:
```python
import pennylane as qml

dev = qml.device('lightning.qubit', wires=23)
%timeit dev.apply_lightning(dev.state, [qml.PauliX(0)])
```
`master`:
```
63 ms ± 523 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
PR:
```
28.3 ms ± 1.78 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
`default.qubit`:
```python
import pennylane as qml

dev = qml.device('default.qubit', wires=23)
%timeit dev.apply([qml.PauliX(0)])
```
```
24.6 ms ± 2.23 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,78 @@
+name: Benchmarking
+on:
+  pull_request:
+
+jobs:
+  benchmarks:
+    name: Gate benchmarks
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+
+    steps:
+
+      - name: Checkout PennyLane-Lightning master
+        uses: actions/checkout@v2
+        with:
+          path: main
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Install dependencies
+        run: sudo apt-get -y -q install cmake gcc
+
+      - name: Get required Python packages
+        run: |
+          cd main
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install matplotlib
+
+      - name: Install lightning.qubit device (master)
+        run: |
+          cd main
+          pip install -e .
+
+      - name: Benchmark lightning master device
+        run: |
+          cd main
+          python .github/workflows/benchmarks/run_bench.py lightning.qubit lightning_master.json
+
+      - name: Benchmark default qubit device
+        run: |
+          cd main
+          python .github/workflows/benchmarks/run_bench.py default.qubit default_qubit.json
+
+      - name: Checkout PennyLane-Lightning PR
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          path: main/pr
+
+      - name: Install lightning.qubit device (PR)
+        run: |
+          cd main
+          make clean
+          pip uninstall pennylane-lightning -y
+          cd pr
+          pip install -e .
+
+      - name: Benchmark lightning PR device
+        run: |
+          cd main/pr
+          python .github/workflows/benchmarks/run_bench.py lightning.qubit lightning_pr.json
+          mv lightning_pr.json ..
+
+      - name: Plot results
+        run: |
+          cd main
+          python .github/workflows/benchmarks/plot_results.py
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: gates.png
+          path: ./main/gates.png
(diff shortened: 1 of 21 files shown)


trbromley:
@antalszava can we close this now that https://github.com/PennyLaneAI/pennylane-lightning/pull/85 has been merged?

antalszava:
I'd revisit this as per the [benchmarks here](https://github.com/PennyLaneAI/pennylane-lightning/pull/87#issuecomment-799049743). This approach doesn't use `generateBitPatterns` and have the impression that it will perform a bit better than the `master` version for `PauliX`.

trbromley:
> @trbromley I'd revisit this as per the [benchmarks here](https://github.com/PennyLaneAI/pennylane-lightning/pull/87#issuecomment-799049743). This approach doesn't use `generateBitPatterns` and have the impression that it will perform a bit better than the `master` version for `PauliX`.

Oh awesome!

trbromley:
Although I suspect we are also doing the equivalent of `generateBitPatterns` here.

antalszava:
@trbromley With this approach, we wouldn't generate the bit patterns themselves but rather would index into the vector explicitly.

Ran the following for comparison:
```python
import pennylane as qml

dev = qml.device('lightning.qubit', wires=23)
%timeit dev.apply_lightning(dev.state, [qml.PauliX(0)])
```
#84
```
27.2 ms ± 475 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
`master`
```
38.4 ms ± 2.8 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
---
@ThomasLoke, any thoughts on how the current structure of gates could be modified so that some gates define a different type of `applyKernel`?

As noted, a couple of gates could get away with computing fewer indices, though this would not be general enough to be added to `AbstractGate`. So we would essentially need two types of `applyKernel` functions whose signatures differ. Approaches that came to mind would include casting the pointers such that they point to the derived gate class, though this is slightly more complicated due to using `unique_ptr`s. So maybe we'd need a more general approach where we can access functionality specific to specific gates? 🤔

ThomasLoke:
> With this approach, we wouldn't generate the bit patterns themselves but rather would index into the vector explicitly.

As @trbromley mentioned above, this is doing the equivalent of generating the bit patterns (at the end of the day, indices are bit patterns), so the question isn't quite whether we do or not as much as how we go about doing that. From a quick inspection, if you collected `i + k`, you should end up with the same as the second application of `generateBitPatterns`.

One material difference is that the bit patterns are generated in-place rather than precomputed, so there's slightly less storage requirements. Ah, but I suppose this also does have better locality of reference, given that the same set of memory locations (i, j, k) are being accessed in order to generate the bit patterns. If that's the case, then perhaps a more cache friendly version of `generateBitPatterns` would work. I'll have a think about it.

> As noted, a couple of gates could get away with computing fewer indices

I'm not actually sure this does compute fewer indices? Could you elaborate?

antalszava:
@ThomasLoke thanks so much!

> I'm not actually sure this does compute fewer indices? Could you elaborate?

From a run I had with printing out values, this was my impression, but could indeed be wrong. Mostly thought that this might just be the case due to the performance difference.

> What you're describing is exactly the point of runtime polymorphism (i.e. virtual methods).

Well, what the "So maybe we'd need a more general approach where we can access functionality specific to specific gates?" comment was referring to was a way without having to manage the derived class instances through a pointer to the base class instance (i.e., not through polymorphism). For the base class, it would not make sense to define the virtual method or if it was defined, it would have a dummy definition.

ThomasLoke:
> For the base class, it would not make sense to define the virtual method or if it was defined, it would have a dummy definition.

I'm not sure why this wouldn't make sense. At the end of the day, why do we need a completely different interface to invoke a function that applies a gate to a given state? It should be fairly simple to modify the API of the base class to something like this:

```
virtual void apply(const StateVector& state, const vector<unsigned int>& opWires, const unsigned int qubits);

virtual void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```

Where the base class implementation of the `apply()` method involves doing some of the work that `constructAndApplyOperation()` does in `Apply.cpp`, and then invoking `applyKernel()`. If you have a subclass that you want to bypass the bit pattern generation, then you can just override `apply()` instead of `applyKernel()`.

That said, it's probably worth spending some time to understand why one solution is more performant than another, and seeing if it can be applied more generally :)

PR #85, opened 2021-03-11
Add implementations for optimised gate kernels

ThomasLoke:
I'm still mulling over the most natural API, but this is one attempt at it. This is comprised of two commits:

1. A refactoring commit that simply moves the core of the gate application (matrix multiplication) into the base `AbstractGate` class.
2. Adds optimised kernels for most gates that overrides the default base class implementation. Some gates (e.g. fully populated 2-by-2 matrices for single-qubit gates) have little (if any?) benefit from not using the default kernel, so I left them as-is.

A word of caution: I've yet to test this extensively. In order to eliminate the function call overhead (especially for the single qubit gates), I think what needs to happen is to move the outer loop over `applyKernel` into said function (unfortunately, I don't think runtime polymorphism works with inline functions unless the concrete class can be statically determined). It may be a worthwhile exercise anyway for when we start vectorising, but again, I've not really tested much.

### DIFF
--- a/pennylane_lightning/src/Gates.cpp
+++ b/pennylane_lightning/src/Gates.cpp
@@ -44,24 +44,28 @@ Pennylane::AbstractGate::AbstractGate(int numQubits)
     , length(exp2(numQubits))
 {}
 
-void Pennylane::AbstractGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
+void Pennylane::AbstractGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
     const vector<CplxType>& matrix = asMatrix();
     assert(indices.size() == length && v.size() == length);
 
-    // Gather
-    size_t pos = 0;
-    for (const size_t& index : indices) {
-        v[pos] = state[index];
-        pos++;
-    }
+    vector<CplxType> v(indices.size());
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        // Gather
+        size_t pos = 0;
+        for (const size_t& index : indices) {
+            v[pos] = shiftedState[index];
+            pos++;
+        }
 
-    // Apply + scatter
-    for (size_t i = 0; i < indices.size(); i++) {
-        size_t index = indices[i];
-        state[index] = 0;
-        size_t baseIndex = i * indices.size();
-        for (size_t j = 0; j < indices.size(); j++) {
-            state[index] += matrix[baseIndex + j] * v[j];
+        // Apply + scatter
+        for (size_t i = 0; i < indices.size(); i++) {
+            size_t index = indices[i];
+            shiftedState[index] = 0;
+            size_t baseIndex = i * indices.size();
+            for (size_t j = 0; j < indices.size(); j++) {
+                shiftedState[index] += matrix[baseIndex + j] * v[j];
+            }
         }
     }
 }
@@ -85,8 +89,11 @@ const vector<CplxType> Pennylane::XGate::matrix{
     0, 1,
     1, 0 };
 
-void Pennylane::XGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
-    swap(state[indices[0]], state[indices[1]]);
+void Pennylane::XGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        swap(shiftedState[indices[0]], shiftedState[indices[1]]);
+    }
 }
 
 // -------------------------------------------------------------------------------------------------------------
@@ -102,10 +109,17 @@ const vector<CplxType> Pennylane::YGate::matrix{
     0, -IMAG,
     IMAG, 0 };
 
-void Pennylane::YGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
-    v[0] = state[indices[0]];
-    state[indices[0]] = -IMAG * state[indices[1]];
-    state[indices[1]] = IMAG * v[0];
+void Pennylane::YGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        CplxType v0 = shiftedState[indices[0]];
+        shiftedState[indices[0]] = -IMAG * shiftedState[indices[1]];
+        shiftedState[indices[1]] = IMAG * v0;
+    }
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+
+    }
 }
 
 // -------------------------------------------------------------------------------------------------------------
@@ -121,8 +135,11 @@ const std::vector<CplxType> Pennylane::ZGate::matrix{
     1, 0,
     0, -1 };
 
-void Pennylane::ZGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
-    state[indices[1]] *= -1;
+void Pennylane::ZGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        shiftedState[indices[1]] *= -1;
+    }
 }
 
 // -------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


antalszava:
Awesome, thanks for this PR @ThomasLoke! 🎊 we'll have a look

pennylane_lightning/src/Gates.cpp
```diff
@@ -74,6 +102,12 @@ const vector<CplxType> Pennylane::YGate::matrix{
     0, -IMAG,
     IMAG, 0 };
 
+void Pennylane::YGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
+    v[0] = state[indices[0]];
+    state[indices[0]] = -IMAG * state[indices[1]];
+    state[indices[1]] = IMAG * v[0];
```
antalszava, line 108: How come we need to update `v[0]` here (and not for `PauliX` for example)? Seems like it's used as a temp for the swap, but it would also have effect afterwards, correct?

pennylane_lightning/src/Gates.cpp
```diff
@@ -205,9 +269,13 @@ Pennylane::GeneralRotationGate Pennylane::GeneralRotationGate::create(const vect
...
     , s(std::sin(theta / 2), 0)
+    , r1(c* std::pow(M_E, CplxType(0, (-phi - omega) / 2)))
+    , r2(-s * std::pow(M_E, CplxType(0, (phi - omega) / 2)))
+    , r3(s* std::pow(M_E, CplxType(0, (-phi + omega) / 2)))
+    , r4(c* std::pow(M_E, CplxType(0, (phi + omega) / 2)))
     , matrix{
-      c * std::pow(M_E, CplxType(0, (-phi - omega) / 2)), -s * std::pow(M_E, CplxType(0, (phi - omega) / 2)),
-      s * std::pow(M_E, CplxType(0, (-phi + omega) / 2)), c * std::pow(M_E, CplxType(0, (phi + omega) / 2)) }
+      r1, r2,
+      r3, r4 }
```
antalszava, line 278: Nice!

pennylane_lightning/src/Gates.cpp
```diff
@@ -42,6 +44,28 @@ Pennylane::AbstractGate::AbstractGate(int numQubits)
     , length(exp2(numQubits))
 {}
 
+void Pennylane::AbstractGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
+    const vector<CplxType>& matrix = asMatrix();
```
trbromley, line 48: Possibly a stupid question - now `asMatrix()` is called multiple times in the `for` loop rather than once outside the `for` loop beforehand. However, this shouldn't be a slowdown because we are just re-defining a pointer to an already existing matrix?

antalszava (approved):
Looks good to me! 🥇 💯 Thank you so much for the quick PR, helps a lot! 😊 

A couple of suggestions/questions, but good to go from my side. Ran the PennyLane device test suite and it completed well.

Was wondering: could it make sense to add some special cases too: e.g., when applying PauliX, we skip the swap if the two vector elements are identical. 🤔 Such cases could potentially arise when there's a more complex algorithm and gates are applied on states which are not known in advance. Not in scope here necessarily, just a thought.

Benchmarked `T` for 23 qubits as an example:

`master`
```python

import pennylane as qml

dev = qml.device('lightning.qubit', wires=23)

%timeit dev.apply_lightning(dev._state, [qml.T(wires=[0])])
```
```
49.5 ms ± 1.25 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
#85 
```python
import pennylane as qml

dev = qml.device('lightning.qubit', wires=23)

%timeit dev.apply_lightning(dev._state, [qml.T(wires=[0])])
```
```
19.9 ms ± 421 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
`default.qubit`
```python
import pennylane as qml

dev = qml.device('default.qubit', wires=23)

%timeit dev.apply([qml.T(wires=[0])])
```
```
43.9 ms ± 3.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
*Side note for benchmarking: using `dev.state` instead of `dev._state` with lightning takes ~10 ms.*

trbromley (commented):
Thanks @ThomasLoke! This is really exciting! How can we help with this, perhaps one thing would be to work on adding tests for the kernels.

More generally, we're looking at adding an adjoint method for differentiation in https://github.com/PennyLaneAI/pennylane-lightning/pull/83. For that, we need to:
- Apply the inverse operation
- Apply the generator of an operation, i.e., apply `H` from `e^(i H t)`.

I was wondering how/whether these would impact the API suggested here. In PennyLane, the name of an inverse gate just has `.inv` appended at the end (e.g., `RX.inv` for inverse of `RX`). Perhaps the first line of `constructAndApplyOperation()` could remove the `.inv` and then set a boolean accordingly, then `constructGate()` would accept that boolean to indicate whether to apply the inverse? We may then have to define `applyKernelInverse` for each gate.

For generators of parameterized gates, I imagine we might have a `applyGenerator()` method?

Anyway, just thinking out loud - these ideas could be included in later iterations.

pennylane_lightning/src/Gates.cpp
ThomasLoke: Yep, that's correct; its just returning a reference to an already instantiated vector, so this should be a trivial operation.

pennylane_lightning/src/Gates.cpp
```diff
@@ -61,6 +85,10 @@ const vector<CplxType> Pennylane::XGate::matrix{
     0, 1,
     1, 0 };
 
+void Pennylane::XGate::applyKernel(CplxType* state, vector<size_t>& indices, vector<CplxType>& v) {
```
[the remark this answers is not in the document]
ThomasLoke: I'm okay with adding the `const`; I'm generally not a fan of removing variable names from method signatures, but I guess opinions may differ on this one--if nothing else, its making it clear that its overriding the same `virtual` method.

pennylane_lightning/src/Gates.cpp
ThomasLoke: `v` is really just scratch space. If we move the outer loop in, we can avoid allocating `v` altogether, except in the base implementation. The requirement to pass `v` is one reason why I'm not entirely happy with this interface.

ThomasLoke (commented):
I can't get to these changes tonight, so it'll have to wait until tomorrow. Or just feel free to take this over and edit it as you see fit...

>  could it make sense to add some special cases too: e.g., when applying PauliX, we skip the swap if the two vector elements are identical. 🤔 Such cases could potentially arise when there's a more complex algorithm and gates are applied on states which are not known in advance.

Branching statements may well be detrimental to performance due to branch misprediction, so it's probably not worth it.

> More generally, we're looking at adding an adjoint method for differentiation in #83.

I haven't taken a close look at the PR, but the API you propose does sound sensible to me :)

pennylane_lightning/src/Gates.cpp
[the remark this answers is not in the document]
antalszava: Got it! A reason to make `v` unnamed would be that while reading, having the name `v` was somewhat confusing, as one would expect that to be used within the function (as is in some of them).

pennylane_lightning/src/Gates.cpp
antalszava: Got it! 

> we can avoid allocating v altogether, except in the base implementation

This would be preferable as just creating `v` would result in an overhead, correct? Could we branch of at the beginning of the outer loop based on the type of operation?

pennylane_lightning/src/Gates.cpp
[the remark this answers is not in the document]
ThomasLoke: True; although I think that's indication we should rework the API (i.e. we shouldn't be passing in parameters we don't need if we don't have to). As mentioned in the PR description, its probably beneficial, although it will unfortunately result in some measure of code duplication.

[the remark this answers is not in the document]
antalszava: Or if branching off would result in a performance degradation, how about incorporating the internal loop in `applyKernel` or creating another member function that implements the internal loop and depends on `applyKernel`? That way `inputVector` would only be created if used.

pennylane_lightning/src/Gates.cpp
ThomasLoke: The allocation for `v` is really insignificant in the grand scheme of things, so I don't think branching is necessary. The first rule of optimisation is to not over-optimise, and I don't think this ever becomes a bottleneck, so its more a question of code clarity as to which API we adopt.

pennylane_lightning/src/Gates.cpp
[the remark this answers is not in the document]
ThomasLoke: That would be ideal. However, ultimately I think we want to inline `applyKernel`. And for that to happen I think we need to duplicate the outer loop for every subclass, because of the requirement to be able to statically infer which method is being called. Again, probably beneficial in the long run, but just a bit painful to look at.

ThomasLoke:
I've bitten the bullet and moved the outer loop into `applyKernel()`. I think the API looks cleaner, at the cost of some code duplication, which I can probably live with. I'd speculate that should perform better in terms of reduced function overheads and that the compiler probably has more things it can do now that the outer loop doesn't loop over a runtime polymorphic function, but its hard to say what (if any) measurable difference it'll make without benchmarking it properly, which I'd appreciate some help with.

pennylane_lightning/src/Gates.hpp
```diff
@@ -20,6 +20,7 @@
 #include <memory>
 #include <vector>
 
+#include "StateVector.hpp"
```
antalszava, line 23: @ThomasLoke suspecting that the tests fail on `pybind` because of this include and we'd now need to use it when compiling `Gates.cpp` for the tests

ThomasLoke: Yeah, but I think this becomes a non-issue when #86 is landed.

trbromley (approved):
Approved! Very nice PR!

trbromley:
Thanks @ThomasLoke!

antalszava:
Thanks @ThomasLoke! For benchmarking, created https://github.com/PennyLaneAI/pennylane-lightning/pull/87 which provides benchmarks for certain gates. It's not meant to be extensive but could give good quick feedback when working on PRs. Thoughts on what could be important to be added in a similar fashion are appreciated.

PR #80, opened 2021-03-03
Add test for getIndicesAfterExclusion

trbromley:
Requires https://github.com/PennyLaneAI/pennylane-lightning/pull/79 to be merged first.

PR #86, opened 2021-03-11
Move all pybind-related functionality to a separate file

trbromley:
Credit to @ThomasLoke!

Also includes a test for `getIndicesAfterExclusion`.


pennylane_lightning/src/tests/lightning_apply_unittest.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    vector<unsigned int> input = inputs[index];
+    vector<unsigned int> output = outputs[index];
+    ASSERT_EQ(output, Pennylane::getIndicesAfterExclusion(input, qubits));
+}
+
+INSTANTIATE_TEST_SUITE_P (
+        getIndicesAfterExclusionTests,
+        getIndicesAfterExclusionTestFixture,
+        ::testing::Values(
```
trbromley, line 52: @antalszava thanks for this [comment](https://github.com/PennyLaneAI/pennylane-lightning/pull/80/files#r587621564)! I tried but then realized that `Combine` does the product over inputs, we want more like a `Zip`.

However, I wouldn't say it's worth worrying about much, given that we're just effectively testing `set(a) - set(b)`.

pennylane_lightning/src/Apply.cpp
```diff
@@ -90,17 +90,16 @@ void Pennylane::constructAndApplyOperation(
...
 
 void Pennylane::apply(
-    pybind11::array_t<CplxType>& stateNumpyArray,
-    vector<string> ops,
-    vector<vector<unsigned int>> wires,
-    vector<vector<double>> params,
+    StateVector& state,
+    vector<string>& ops,
+    vector<vector<unsigned int>>& wires,
+    vector<vector<double>>& params,
```
antalszava, line 96: None of these are expected to change within `apply` at the moment, so we should be able to define them as `const` (applies in `.hpp` too if accepted).
```suggestion
    const vector<string>& ops,
    const vector<vector<unsigned int>>& wires,
    const vector<vector<double>>& params,
```

pennylane_lightning/src/PyBind.cpp
```diff
@@ -0,0 +1,52 @@
+// Copyright 2021 Xanadu Quantum Technologies Inc.
```
antalszava, line 1: How about calling this file `bindings.cpp`? The filename `PyBind.cpp` would suggest that this file contains sources from `pybind11`.

pennylane_lightning/src/PyBind.cpp
```diff
@@ -0,0 +1,52 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "pybind11/complex.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/numpy.h"
+#include "pybind11/stl.h"
+
+#include "Apply.hpp"
```
antalszava, line 19: Just wondering about having a separate `.hpp` file to this `.cpp` file instead of using `Apply.hpp` and having two/multiple `.cpp` files. Not super important would just adhere to convention, though this should work as is too.

pennylane_lightning/src/PyBind.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+void apply(
+    pybind11::array_t<CplxType>& stateNumpyArray,
+    vector<string> ops,
+    vector<vector<unsigned int>> wires,
+    vector<vector<double>> params,
+    const unsigned int qubits
+) {
+    StateVector state = create(&stateNumpyArray);
+    Pennylane::apply(state, ops, wires, params, qubits);
```
antalszava, line 45: Could we have this be called `apply_operation` in the `Pennylane` namespace? Would just be perhaps easier to distinguish for readers between the `apply` and `Pennylane::apply`.

antalszava (approved):
Looks good to me! 💯 Thank you @trbromley @ThomasLoke! Had a couple of suggestions which I'd be curious to get thoughts on, but as none are significant blockers, happy to approve.

pennylane_lightning/src/PyBind.cpp
ThomasLoke: This file doesn't provide method bodies for anything in `Apply.hpp`, so its not so much sharing a `.hpp` file as it is using declarations in `Apply.hpp` (just like we'd include say `Util.hpp` if we wanted to use methods from that).

Personally, I don't consider it convention to have a `.hpp` file for source files that provide entry points (e.g. you wouldn't typically have a `main.hpp` if you had a `main.cpp`). I'm somewhat on the fence on whether we should even expose the `create()` and `apply()` methods defined here, or just make them `static` to make it clear they'll only be referenced in this source file.

pennylane_lightning/src/PyBind.cpp
ThomasLoke: As it stands, they're not even in the same namespaces, so it doesn't seem particularly confusing. Personally, I think there's value in sharing the same name and a similar method signature, to make it clear that this is essentially the pybind version of the `Pennylane::apply()` method.

PR #90, opened 2021-03-16
Add generators to gates

thisac:
Adds the generators for the different gates, returning nothing if no generators exist. The generator has a similar type, structure as the generator method in PL and returns a `tuple<vector<CplxType>, double>` with the generator and scaling_factor.

- [x] add generators to gates
- [x] add `applyDerivative` functions
- [ ] add tests


pennylane_lightning/src/Gates.cpp
```diff
@@ -220,6 +230,23 @@ Pennylane::RotationXGate::RotationXGate(double rotationAngle)
       js, c }
 {}
 
+const tuple<vector<CplxType>, double> Pennylane::RotationXGate::generator{
+    { 0, 1,
+      1, 0 }, -0.5
+};
+
+void Pennylane::RotationXGate::applyDerivative(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    CplxType pref = (std::get<1>(Pennylane::RotationXGate::generator), 1.0);
```
trbromley, line 239: Why is it `(-0.5, 1)`?

thisac: Oh, it's just me being confused. Should probably be written as 
```
1j * std::get<1>(Pennylane::RotationXGate::generator)
``` 
if I want to get `-0.5i`, right?

pennylane_lightning/src/Gates.cpp
```diff
@@ -220,6 +230,23 @@ Pennylane::RotationXGate::RotationXGate(double rotationAngle)
       js, c }
 {}
 
+const tuple<vector<CplxType>, double> Pennylane::RotationXGate::generator{
+    { 0, 1,
+      1, 0 }, -0.5
```
trbromley, line 235: I was initially worried about having the shift parameter of the generator hard coded here, and was thinking we could pass this from PennyLane to the backend as this would be more maintainable.

However, I doubt these numbers will change once fixed since it would change the results of any simulation that uses that gate. So maybe there is no maintainability worry! And hardcoding is better for speed!

So TLDR, looks good!

trbromley: Perhaps we should just make the generator and prefactor separate attributes?

thisac:
Closing this in favour of the already merged solution in #92.

PR #91, opened 2021-03-18
Support arbitrary gates

trbromley:


pennylane_lightning/lightning_qubit.py
```diff
@@ -49,9 +49,7 @@ class LightningQubit(DefaultQubit):
     version = __version__
     author = "Xanadu Inc."
 
-    operations = {
-        "BasisState",
-        "QubitStateVector",
+    kernel_operations = {
```
trbromley, line 52: By doing this, `lightning.qubit` will now support any gate that `default.qubit` does. Further, if the gate is in `kernel_operation` then we have an optimized kernel for it. If not, we'll directly use the the gate's matrix to apply the gate through standard matrix multiplication.

pennylane_lightning/src/Gates.cpp
```diff
@@ -556,6 +556,52 @@ void Pennylane::CSWAPGate::applyKernel(const StateVector& state, const std::vect
 
 // -------------------------------------------------------------------------------------------------------------
 
+const string Pennylane::Unitary::label = "QubitUnitary";
+
+Pennylane::Unitary Pennylane::Unitary::create(const vector<double>& parameters) {
+    auto size = parameters.size();
+    unsigned int log2size = 0;
+    while (size >>= 1) ++log2size;
+    unsigned int numQubits = (log2size - 1) / 2;
```
trbromley, line 565: We infer the number of qubits based upon the input parameters.

pennylane_lightning/lightning_qubit.py
```diff
@@ -133,9 +131,24 @@ def apply_lightning(self, state, operations):
...
+        op_wires = []
+        op_param = []
+
+        for o in operations:
+            op_wires.append(self.wires.indices(o.wires))
+            if o.name in self.kernel_operations:
+                op_names.append(o.name)
+                op_param.append(o.parameters)
+            else:
+                op_names.append("QubitUnitary")
```
trbromley, line 144: Here, any gate that does not have an optimized kernel in `lighting.qubit` will be treated as a unitary by the backend. We just need to find the unitary matrix, which we then view as
`[real, imag, real, imag, real, imag...]` using `mat.ravel().view(np.float64)` (no copies required).
These real numbers are then passed as parameters of the gate.

pennylane_lightning/src/Gates.hpp
```diff
@@ -323,6 +323,23 @@ namespace Pennylane {
         void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
     };
 
+    class Unitary : public AbstractGate {
+    private:
+        const std::vector<CplxType> matrix;
```
trbromley, line 328: We actually don't explicitly define `matrix` and so `matrix` and `asMatrix` are not needed. However, they need to be defined in the header because they are virtual in `AbstractGate`.

tests/test_apply.py
```diff
@@ -1212,3 +1213,20 @@ def test_pauliz_hadamard(self, theta, phi, varphi, monkeypatch, qubit_device_3_w
...
+
+    dev = qml.device("lightning.qubit", wires=wires)
+
+    @qml.qnode(dev)
+    def f():
+        qml.QubitUnitary(unitary, wires=range(wires))
+        return qml.state()
+
+    target_state = unitary[:, 0]
+    assert np.allclose(f(), target_state)
```
trbromley, line 1232: Variants of this test are almost certainly provided by the device test suite, but I'm happy to have it explicitly here even if there is some redundancy.

pennylane_lightning/src/Gates.cpp
```diff
@@ -556,6 +556,52 @@ void Pennylane::CSWAPGate::applyKernel(const StateVector& state, const std::vect
 
 // -------------------------------------------------------------------------------------------------------------
 
+const string Pennylane::Unitary::label = "QubitUnitary";
+
+Pennylane::Unitary Pennylane::Unitary::create(const vector<double>& parameters) {
+    auto size = parameters.size();
+    unsigned int log2size = 0;
+    while (size >>= 1) ++log2size;
```
antalszava, line 564: Personally, would not mind having this still be a `while` with syntax of
```cpp
while (size >>= 1){
    ++log2size;
}
```
Also, this functionality could potentially come in handy in other parts of the code base too, could consider defining it in `Utils.hpp`.

pennylane_lightning/src/Gates.hpp
```diff
@@ -323,6 +323,23 @@ namespace Pennylane {
         void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
     };
 
+    class Unitary : public AbstractGate {
```
antalszava, line 326: For the existing gates we knew the matrix representation in advance (and could hardcode them). Down the line, would be nice to check that there's no major performance degradation if we have a circuit with several layers that apply the same `QubitUnitary`. In such a case, parameters will be passed for every `QubitUnitary` and lightning will consider each of them separately. 🤔

pennylane_lightning/src/Gates.hpp
```diff
@@ -323,6 +323,23 @@ namespace Pennylane {
...
+    private:
+        const std::vector<CplxType> matrix;
+        const std::vector<double> real_matrix;
```
antalszava, line 329: Worth leaving a comment on how this vector represents the complex matrix for reference.

pennylane_lightning/src/Gates.cpp
```diff
@@ -556,6 +556,52 @@ void Pennylane::CSWAPGate::applyKernel(const StateVector& state, const std::vect
...
+    unsigned int numQubits = (log2size - 1) / 2;
+
+    return Pennylane::Unitary(numQubits, parameters);
+}
+
+Pennylane::Unitary::Unitary(const int numQubits, const std::vector<double>& real_matrix)
```
antalszava, line 570: Find the name `real_matrix` confusing, `matrix_with_real_and_imag` (or something better)?

pennylane_lightning/src/Gates.cpp
```diff
@@ -556,6 +556,52 @@ void Pennylane::CSWAPGate::applyKernel(const StateVector& state, const std::vect
...
+    : AbstractGate(numQubits)
+    , real_matrix{real_matrix}
+{}
+
+void Pennylane::Unitary::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    const vector<double>& real_matrix = asRealMatrix();
```
antalszava, line 576: Depending on the relative order of the PRs, inverse computations would be required here as per #89?

antalszava (commented):
This is really great, with many clever solutions and workarounds! 💯 😍 Nothing major, maybe just worth commenting on the approach that is taking to allow support. Would just jump in for a quick round before approval.

pennylane_lightning/src/Gates.cpp
```diff
@@ -556,6 +556,52 @@ void Pennylane::CSWAPGate::applyKernel(const StateVector& state, const std::vect
 
 // -------------------------------------------------------------------------------------------------------------
 
+const string Pennylane::Unitary::label = "QubitUnitary";
+
+Pennylane::Unitary Pennylane::Unitary::create(const vector<double>& parameters) {
+    auto size = parameters.size();
```
antalszava, line 562: Probably not use case for now since we are binding PennyLane, though might be worth verifying the size of `parameters`.

ThomasLoke (changes requested):
> with many clever solutions and workarounds!

I find that this is generally indicative of one of three things:
1) There are genuine shortcomings in the language features, in which case  ¯\_(ツ)_/¯; or
2) There are problems with the class design, in which case we should be modify said design to support the case that we're adding; or
3) We need to take a step back, re-evaluate what it is we're trying to do, and see if there's a more natural way of accomplishing things.

This PR feels like we're trying to fit a square into a triangle-shaped hole, so it feels like 3) to me. And, in the first instance, it should already be shaped like a triangle!

pennylane_lightning/src/Gates.cpp
```diff
@@ -583,6 +629,7 @@ static map<string, function<unique_ptr<Pennylane::AbstractGate>(const vector<dou
     addToDispatchTable<Pennylane::CGeneralRotationGate>(dispatchTable);
     addToDispatchTable<Pennylane::ToffoliGate>(dispatchTable);
     addToDispatchTable<Pennylane::CSWAPGate>(dispatchTable);
+    addToDispatchTable<Pennylane::Unitary>(dispatchTable);
```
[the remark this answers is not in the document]
trbromley: I like the idea of (a) `Unitary` being the dispatch table default and (b) `kernel_operations` being removed (that also worried me in terms of maintainability). However, if we do (b), don't we then need to pass the `matrix` from Python to the backend for _every_ operation (including those with a defined kernel)? Is there an overhead in doing that:
- Even if passed by reference, do we expect any overhead from PyBind?
- There would be a (potentially negligible) overhead in creating the matrix in Python when not needed to. E.g., for [Rot](https://github.com/PennyLaneAI/pennylane/blob/34954b560a95125c0852886346ca8c7915142e11/pennylane/ops/qubit.py#L876)?

trbromley (commented):
Thanks @ThomasLoke, really appreciate the feedback and guidance on this! And I agree that this may be a case of taking a step back and potentially re-evaluating, although also balancing that against the cost of slower development.

So the current approach is: if op doesn't have a supported kernel, pass the unitary matrix as a flat, real-valued vector. In the backend, define a tweaked kernel that correctly calculates the matrix multiplication.

What would our ideal approach look like?
- If op doesn't have supported kernel we somehow pass the matrix to the backend
- If op name isn't defined in the dispatch table, we fall back to `Unitary`
- `Unitary` just relies on the `applyKernel` from `AbstractGate`

So for me the question seems to be about how and when to pass the matrix to the backend. I also wonder if we're starting to see limitations in the simple serialization approach we used to communicate between frontend and backend. For example, what if we could pass a list of `Gate` objects to the backend (perhaps this would be more lightweight than PL's current `Operation` class).

pennylane_lightning/src/Gates.hpp
ThomasLoke: > If we have params as a 2*d length flattened and real vector, can we convert that to a d-length complex vector with no overhead/copying?

This solution already involves copying (the `Unitary` constructor is calling the `vector` copy constructor). In fact, because we're relying on the numpy implicit conversion from a python list `op_param` to `vector<vector<double>> params`, the overhead is already awful, scaling with the size of the matrix! So we really should be passing it in separately.

ThomasLoke (commented):
> And I agree that this may be a case of taking a step back and potentially re-evaluating, although also balancing that against the cost of slower development.

That's partially a matter of perspective. Slower development in the present to get things right may save a lot of pain in the future. But I don't dictate development milestones, so ¯\\_(ツ)_/¯

> What would our ideal approach look like?

As mentioned in the preceding comments, ideally Python shouldn't even need to distinguish between optimised and non-optimised gates. Instead, it'd just pass in a functional to generate the matrix, and the backend will invoke that if necessary.

In terms of serialisation, passing in an extra argument to extend the Bindings interface does feel a bit suboptimal. Ideally, we'd just encapsulate all the settings that define a gate within a class of some sort, and just instantiate a single list of those objects to the backend. But reworking this doesn't need to be the domain of this PR.

pennylane_lightning/src/Gates.hpp
ThomasLoke: > because we're relying on the pybind implicit conversion from a python list op_param to vector<vector<double>> params, the overhead is already awful, scaling with the size of the matrix

Actually I take this back; this may not be the case. Its a python list of what's potentially a mixture of python lists and numpy array views. I'm not entirely sure how that gets handled.

pennylane_lightning/lightning_qubit.py
```diff
@@ -133,9 +131,24 @@ def apply_lightning(self, state, operations):
...
+
+                mat = o.matrix
+                if not np.iscomplex(o.matrix).all():
+                    mat = mat.astype(np.complex128)
+
+                unitary_view = mat.ravel().view(np.float64)
+                op_param.append(unitary_view)
```
thisac, line 151: Oh, `view()` splits up the complex number into two floats. Nice! 😮

pennylane_lightning/lightning_qubit.py
thisac: Is the issue here that we cannot pass a complex array with pybind and thus need to represent it as a flattened real array?

thisac (commented):
I have no major thoughts other than some of the comments already expressed here. I agree that it seems like a better solution to somehow use the already existing `applyKernel` here, rather than applying the complex-splitting trick (although I too find it quite clever). Not sure how this could be solved better though. 🤔

trbromley:
Closing this for now while we think about the best way to implement.

PR #92, opened 2021-03-22
`applyGenerator` method for gates

thisac:
Adds an `applyGenerator` method for all supported gates, which applied the corresponding generator to a state, and a `generatorScalingFactor` that returns the shift for the particular gate, for use in the adjoint method.

### DIFF
--- a/pennylane_lightning/src/Apply.cpp
+++ b/pennylane_lightning/src/Apply.cpp
@@ -68,6 +68,20 @@ void Pennylane::constructAndApplyOperation(
     gate->applyKernel(state, internalIndices, externalIndices, inverse);
 }
 
+void Pennylane::applyGateGenerator(
+    StateVector& state,
+    unique_ptr<AbstractGate> gate,
+    const vector<unsigned int>& opWires,
+    const unsigned int qubits
+) {
+    vector<size_t> internalIndices = generateBitPatterns(opWires, qubits);
+
+    vector<unsigned int> externalWires = getIndicesAfterExclusion(opWires, qubits);
+    vector<size_t> externalIndices = generateBitPatterns(externalWires, qubits);
+
+    gate->applyGenerator(state, internalIndices, externalIndices);
+}
+
 void Pennylane::apply(
     StateVector& state,
     const vector<string>& ops,
--- a/pennylane_lightning/src/Apply.hpp
+++ b/pennylane_lightning/src/Apply.hpp
@@ -76,11 +76,11 @@ namespace Pennylane {
     );
     
     /**
-     * Applies the generator to the gate to the state vector.
+     * Applies the generator of the gate to the state vector.
      * 
      * @param state state vector to which to apply the operation
-     * @param gate unique pointer to gate which is to be applied
-     * @param opWires index of qubits on which the gate acts
+     * @param gate unique pointer to the gate whose generator is to be applied
+     * @param opWires index of qubits on which the operation acts
      * @param qubits number of qubits
      */
     void applyGateGenerator(
--- a/pennylane_lightning/src/Apply.hpp
+++ b/pennylane_lightning/src/Apply.hpp
@@ -23,6 +23,7 @@
 #include <string>
 #include <vector>
 
+#include "Gates.hpp"
 #include "StateVector.hpp"
 #include "typedefs.hpp"
 
@@ -55,7 +56,7 @@ namespace Pennylane {
      */
     std::vector<size_t> generateBitPatterns(const std::vector<unsigned int>& qubitIndices, const unsigned int qubits);
 
-    /*
+    /**
      * Constructs the gate defined by the supplied parameters and applies it to the state vector.
      * 
      * @param state state vector to which to apply the operation
@@ -71,6 +72,21 @@ namespace Pennylane {
         const std::vector<unsigned int>& opWires,
         const std::vector<double>& opParams,
         bool inverse,
+        const unsigned int qubits
+    );
+    
+    /**
+     * Applies the generator to the gate to the state vector.
+     * 
+     * @param state state vector to which to apply the operation
+     * @param gate unique pointer to gate which is to be applied
+     * @param opWires index of qubits on which the gate acts
+     * @param qubits number of qubits
+     */
+    void Pennylane::applyGateGenerator(
+        StateVector& state,
+        std::unique_ptr<AbstractGate> gate,
+        const std::vector<unsigned int>& opWires,
         const unsigned int qubits
     );
 
--- a/pennylane_lightning/src/Apply.hpp
+++ b/pennylane_lightning/src/Apply.hpp
@@ -83,7 +83,7 @@ namespace Pennylane {
      * @param opWires index of qubits on which the gate acts
      * @param qubits number of qubits
      */
-    void Pennylane::applyGateGenerator(
+    void applyGateGenerator(
         StateVector& state,
         std::unique_ptr<AbstractGate> gate,
         const std::vector<unsigned int>& opWires,
--- a/pennylane_lightning/src/Gates.cpp
+++ b/pennylane_lightning/src/Gates.cpp
@@ -249,7 +249,8 @@ const double Pennylane::RotationXGate::generatorScalingFactor{ -0.5 };
 
 void Pennylane::RotationXGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
     unique_ptr<Pennylane::XGate> gate;
-    gate->applyKernel(state, indices, externalIndices);
+    gate->applyKernel(state, indices, externalIndices, false);
+}
 
 void Pennylane::RotationXGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool inverse) {
     CplxType js_ = (inverse == true) ? -js : js;
@@ -284,7 +285,8 @@ const double Pennylane::RotationYGate::generatorScalingFactor{ -0.5 };
 
 void Pennylane::RotationYGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
     unique_ptr<Pennylane::YGate> gate;
-    gate->applyKernel(state, indices, externalIndices);
+    gate->applyKernel(state, indices, externalIndices, false);
+}
 
 void Pennylane::RotationYGate::applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool inverse) {
     CplxType s_ = (inverse == true) ? -s : s;
@@ -335,7 +337,7 @@ const double Pennylane::RotationZGate::generatorScalingFactor{ -0.5 };
 
 void Pennylane::RotationZGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
     unique_ptr<Pennylane::ZGate> gate;
-    gate->applyKernel(state, indices, externalIndices);
+    gate->applyKernel(state, indices, externalIndices, false);
 }
 
 
--- a/pennylane_lightning/src/Gates.cpp
+++ b/pennylane_lightning/src/Gates.cpp
@@ -321,7 +321,6 @@ void Pennylane::PhaseShiftGate::applyGenerator(const StateVector& state, const s
     for (const size_t& externalIndex : externalIndices) {
         CplxType* shiftedState = state.arr + externalIndex;
         shiftedState[indices[0]] = 0;
-        shiftedState[indices[1]] *= shift;
     }
 }
 
@@ -452,11 +451,8 @@ const double Pennylane::CRotationXGate::generatorScalingFactor{ -0.5 };
 void Pennylane::CRotationXGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
     for (const size_t& externalIndex : externalIndices) {
         CplxType* shiftedState = state.arr + externalIndex;
-        CplxType v0 = shiftedState[indices[2]];
-        CplxType v1 = shiftedState[indices[3]];
         shiftedState[indices[0]] = shiftedState[indices[1]] = 0;
-        shiftedState[indices[2]] = js * v0 + c * v1;
-        shiftedState[indices[3]] = c * v0 + js * v1;
+        swap(shiftedState[indices[2]], shiftedState[indices[3]]);
     }
 }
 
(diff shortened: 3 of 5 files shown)


pennylane_lightning/src/Gates.cpp
```diff
@@ -70,6 +70,14 @@ void Pennylane::AbstractGate::applyKernel(const StateVector& state, const std::v
     }
 }
 
+const double Pennylane::AbstractGate::generatorScalingFactor{};
+void Pennylane::AbstractGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    (void)state;
+    (void)indices;
+    (void)externalIndices;
+    throw NotImplementedException();
+}
```
antalszava, line 79: Would eventually other gates define an `applyGenerator` function or just the ones where there's a definition provided at the moment?

thisac: Perhaps eventually, if we figure out a way to define the generators for those gates. Although to be honest I'm not sure how that'd be done, or if we'd need to solve it in some other way.

pennylane_lightning/src/Gates.hpp
```diff
@@ -137,6 +147,8 @@ namespace Pennylane {
         inline const std::vector<CplxType>& asMatrix() {
             return matrix;
         }
+        virtual void applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```
trbromley, line 150:
```suggestion
        void applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```
Do we want the virtual here? I think it's for when we are defining base classes.

pennylane_lightning/src/Util.hpp
```diff
@@ -50,3 +50,10 @@ template<typename T, typename... Args>
 std::unique_ptr<T> make_unique(Args&&... args) {
     return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
+
+// Exception for functions that arent't implemented
```
trbromley, line 54:
```suggestion
// Exception for functions that aren't implemented
```

pennylane_lightning/src/Gates.cpp
```diff
@@ -70,6 +70,14 @@ void Pennylane::AbstractGate::applyKernel(const StateVector& state, const std::v
     }
 }
 
+const double Pennylane::AbstractGate::generatorScalingFactor{};
```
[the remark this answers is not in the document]
thisac: It is needed when applying the derivative of a gate for the adjoint method. So, to apply `d/dt gate = d/dt exp(i*s*H*t) = i*s*H * gate`, we'd apply first the `gate`, then the generator `H`, then multiply by the scaling factor `s` times `i` (apologies for rendering maths as code).

pennylane_lightning/src/Gates.cpp
```diff
@@ -285,6 +315,16 @@ void Pennylane::PhaseShiftGate::applyKernel(const StateVector& state, const std:
     }
 }
 
+const double Pennylane::PhaseShiftGate::generatorScalingFactor{ 1.0 };
+
+void Pennylane::PhaseShiftGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        shiftedState[indices[0]] = 0;
+        shiftedState[indices[1]] *= shift;
```
trbromley, line 324: Should this be `shift`? Since the generator is a projector onto the 1 state, maybe we just need `shiftedState[indices[0]] = 0`

pennylane_lightning/src/Gates.cpp
```diff
@@ -407,6 +447,19 @@ void Pennylane::CRotationXGate::applyKernel(const StateVector& state, const std:
...
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        CplxType v0 = shiftedState[indices[2]];
+        CplxType v1 = shiftedState[indices[3]];
+        shiftedState[indices[0]] = shiftedState[indices[1]] = 0;
+        shiftedState[indices[2]] = js * v0 + c * v1;
+        shiftedState[indices[3]] = c * v0 + js * v1;
+    }
+}
+
```
trbromley, line 462: I'm not sure about the generators for the following 3 gates. For me, the generator for this one would be the CNOT gate. The generator for `CRotationYGate` would be a controlled version of the `PauliY` kernel - something like:
```python
void Pennylane::CRotationXGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool) {
    for (const size_t& externalIndex : externalIndices) {
        CplxType* shiftedState = state.arr + externalIndex;
        CplxType v2 = shiftedState[indices[2]];
        shiftedState[indices[2]] = -IMAG * shiftedState[indices[3]];
        shiftedState[indices[3]] = IMAG * v2;
    }
}
```

trbromley (commented):
Thanks @thisac! Looking good! :100:

pennylane_lightning/src/Gates.cpp
thisac: Oh, yep, these are wrong. Thanks for catching it! I managed to confuse applying the derivative with applying the generator here. What I wrote should (I think) apply the derivative and _not_ the generator. I just copied them a bit too quickly. 😆

pennylane_lightning/src/Gates.hpp
thisac: True, just a case of me copying things too quickly without checking.

pennylane_lightning/src/Gates.cpp
thisac: Thanks for noting this. Same here as with the 3-qubit gates. I just managed to keep the `applyDerivative` functions and forgot to update them to just apply the generator. 😞 

I'll fix them right away!

pennylane_lightning/src/Gates.cpp
```diff
@@ -407,6 +446,16 @@ void Pennylane::CRotationXGate::applyKernel(const StateVector& state, const std:
...
 }
 
+const double Pennylane::CRotationXGate::generatorScalingFactor{ -0.5 };
+
+void Pennylane::CRotationXGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    for (const size_t& externalIndex : externalIndices) {
+        CplxType* shiftedState = state.arr + externalIndex;
+        shiftedState[indices[0]] = shiftedState[indices[1]] = 0;
+        swap(shiftedState[indices[2]], shiftedState[indices[3]]);
+    }
```
trbromley, line 456: Optionally, you could also directly apply `CNOT` here

thisac: Is that preferable? I would think a swap would be faster. We still need to set `shiftedState[indices[0]]` and `shiftedState[indices[1]]` to 0. 🤔

trbromley: No you're right! In my head, the generator was exactly the CNOT matrix. But yes it's actually `|1><1| otimes X`! So agree that it's good as-is.

pennylane_lightning/src/Util.hpp
```diff
@@ -50,3 +50,10 @@ template<typename T, typename... Args>
 std::unique_ptr<T> make_unique(Args&&... args) {
     return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
+
+// Exception for functions that aren't implemented
+class NotImplementedException : public std::logic_error
+{
+public:
+    NotImplementedException() : std::logic_error("Function not yet implemented") { };
```
antalszava, line 58: (Minor): perhaps it could be used even when it's not planned that there is going to be an implementation provided?
```suggestion
    NotImplementedException() : std::logic_error("Function is not implemented.") { };
```

pennylane_lightning/src/Gates.cpp
```diff
@@ -237,6 +245,13 @@ Pennylane::RotationXGate::RotationXGate(double rotationAngle)
       js, c }
 {}
 
+const double Pennylane::RotationXGate::generatorScalingFactor{ -0.5 };
```
trbromley, line 248: I still do worry about this number changing in PL and us not noticing, but I guess we'll be able to notice by having good quality integration tests for the gradient.

pennylane_lightning/src/Gates.cpp
```diff
@@ -237,6 +245,13 @@ Pennylane::RotationXGate::RotationXGate(double rotationAngle)
...
+
+void Pennylane::RotationXGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    unique_ptr<Pennylane::XGate> gate;
+    gate->applyKernel(state, indices, externalIndices, false);
```
trbromley, line 252: Ah, sorry for forgetting to add the inverse when I merged!

pennylane_lightning/src/Gates.hpp
```diff
@@ -44,6 +44,16 @@ namespace Pennylane {
          * Generic matrix-multiplication kernel
          */
         virtual void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool inverse);
+        
+        /**
+         * Generic matrix-multiplication kernel for applying the generator of an operation
+         */
+        void applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```
trbromley, line 51: So the reason it's not `virtual void` is because we define `AbstractGate.applyGenerator()`, but just throw a `NotImplementedError` in case it's called?

pennylane_lightning/src/Gates.hpp
```diff
@@ -44,6 +44,16 @@ namespace Pennylane {
          * Generic matrix-multiplication kernel
          */
         virtual void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool inverse);
+        
+        /**
+         * Generic matrix-multiplication kernel for applying the generator of an operation
```
trbromley, line 49:
```suggestion
         * Kernel for applying the generator of an operation
```

trbromley (approved):
Thanks @thisac looks great!

pennylane_lightning/src/Gates.cpp
```diff
@@ -79,6 +79,14 @@ void Pennylane::AbstractGate::applyKernel(const StateVector& state, const std::v
     }
 }
 
+const double Pennylane::AbstractGate::generatorScalingFactor{};
+void Pennylane::AbstractGate::applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices) {
+    (void)state;
+    (void)indices;
+    (void)externalIndices;
```
antalszava, line 86:
```suggestion
void Pennylane::AbstractGate::applyGenerator(const StateVector&, const std::vector<size_t>&, const std::vector<size_t>&) {
```

antalszava (approved):
Looks good to me! 💯 Nice one 😊 Mostly looked at the code itself, haven't verified correctness.

pennylane_lightning/src/Gates.cpp
thisac: Yep, I'm not sure how this would else be done. I think we could leave it like this, rely on testing, and then perhaps update it in the future if we figure out a better way.

trbromley: Agree, and it's possibly faster to have it hard-coded than passing from the Python frontend

pennylane_lightning/src/Gates.hpp
antalszava: Actually a good question, we might need to make this `virtual`, see the [description here](https://stackoverflow.com/a/2392656). We will use `unique_ptr<AbstractGate>->applyGenerator` which can very well always call the base class method if it's not defined as `virtual`.

pennylane_lightning/src/Apply.hpp
```diff
@@ -73,6 +74,21 @@ namespace Pennylane {
         bool inverse,
         const unsigned int qubits
     );
+    
+    /**
+     * Applies the generator to the gate to the state vector.
```
antalszava, line 79:
```suggestion
     * Applies the generator of the gate to the state vector.
```

pennylane_lightning/src/Apply.cpp
```diff
@@ -68,6 +68,20 @@ void Pennylane::constructAndApplyOperation(
     gate->applyKernel(state, internalIndices, externalIndices, inverse);
 }
 
+void Pennylane::applyGateGenerator(
+    StateVector& state,
+    unique_ptr<AbstractGate> gate,
+    const vector<unsigned int>& opWires,
+    const unsigned int qubits
```
trbromley, line 75: Do we need `inverse` here? For the adjoint diff method, I don't think so.

pennylane_lightning/src/Apply.hpp
```diff
@@ -73,6 +74,21 @@ namespace Pennylane {
...
+     * 
+     * @param state state vector to which to apply the operation
+     * @param gate unique pointer to gate which is to be applied
```
antalszava, line 82:
```suggestion
     * @param gate unique pointer to the gate whose generator is to be applied
```

pennylane_lightning/src/Apply.hpp
```diff
@@ -73,6 +74,21 @@ namespace Pennylane {
...
+     * @param opWires index of qubits on which the gate acts
```
antalszava, line 83:
```suggestion
     * @param opWires index of qubits on which the operation acts
```

pennylane_lightning/src/Apply.hpp
```diff
@@ -55,7 +56,7 @@ namespace Pennylane {
      */
     std::vector<size_t> generateBitPatterns(const std::vector<unsigned int>& qubitIndices, const unsigned int qubits);
 
-    /*
+    /**
```
trbromley, line 59:
```suggestion
    /*
```

pennylane_lightning/src/Apply.hpp
trbromley, line 79:
```suggestion
     * Applies the generator of the gate to the state vector.
```

pennylane_lightning/src/Apply.hpp
```diff
@@ -23,6 +23,7 @@
 #include <string>
 #include <vector>
 
+#include "Gates.hpp"
```
trbromley, line 26: Is this for `AbstractGate`?

trbromley (approved):

pennylane_lightning/src/Gates.hpp
```diff
@@ -44,6 +46,16 @@ namespace Pennylane {
          * Generic matrix-multiplication kernel
          */
         virtual void applyKernel(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices, bool inverse);
+
+        /**
+         * Kernel for applying the generator of an operation
+         */
+        void applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```
thisac, line 53:
```suggestion
        virtual void applyGenerator(const StateVector& state, const std::vector<size_t>& indices, const std::vector<size_t>& externalIndices);
```

pennylane_lightning/src/Apply.hpp
thisac: Why remove the extra `*`? I just added them since it seems like they should preferably be there for docstrings (that way the syntax highlighting works in my editor as well).

pennylane_lightning/src/Apply.cpp
thisac: I don't think so either. I think it's fine not having it for now.

pennylane_lightning/src/Apply.hpp
thisac: Yes, it is! 🙂

pennylane_lightning/src/Apply.hpp
trbromley: Oh, if it's standard then feel free to keep :smile:

PR #93, opened 2021-03-27
Update plugin by removing `analytic` argument for device creation

antalszava:
Updates the plugin to cope with removing the `analytic` keyword argument as per https://github.com/PennyLaneAI/pennylane/pull/1079.

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -42,7 +42,7 @@ class LightningQubit(DefaultQubit):
 
     name = "Lightning Qubit PennyLane plugin"
     short_name = "lightning.qubit"
-    pennylane_requires = ">=0.12"
+    pennylane_requires = ">=0.15"
     version = __version__
     author = "Xanadu Inc."
 
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -35,12 +35,9 @@ class LightningQubit(DefaultQubit):
     Args:
         wires (int): the number of wires to initialize the device with
         shots (int): How many times the circuit should be evaluated (or sampled) to estimate
-            the expectation values. Defaults to 1000 if not specified.
-            If ``analytic == True``, then the number of shots is ignored
-            in the calculation of expectation values and variances, and only controls the number
-            of samples returned by ``sample``.
-        analytic (bool): indicates if the device should calculate expectations
-            and variances analytically
+            the expectation values. Defaults to 1000 if not specified. Setting
+            to ``None`` results in computing statistics like expectation values and
+            variances analytically.
     """
 
     name = "Lightning Qubit PennyLane plugin"
@@ -76,8 +73,8 @@ class LightningQubit(DefaultQubit):
 
     observables = {"PauliX", "PauliY", "PauliZ", "Hadamard", "Hermitian", "Identity"}
 
-    def __init__(self, wires, *, shots=1000, analytic=True):
-        super().__init__(wires, shots=shots, analytic=analytic)
+    def __init__(self, wires, *, shots=None):
+        super().__init__(wires, shots=shots)
 
     @classmethod
     def capabilities(cls):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 flaky
 numpy
-pennylane>=0.12
+git+https://github.com/PennyLaneAI/pennylane.git
 pybind11
 pytest
 pytest-cov
--- a/tests/test_apply.py
+++ b/tests/test_apply.py
@@ -1128,13 +1128,19 @@ class TestTensorVar:
         assert np.allclose(res, expected, atol=tol, rtol=0)
 
 
-@pytest.mark.parametrize("theta, phi, varphi", list(zip(THETA, PHI, VARPHI)))
-class TestTensorSample:
-    """Test tensor expectation values"""
+# Tolerance for non-analytic tests
+TOL_STOCHASTIC = 0.05
 
-    def test_paulix_pauliy(self, theta, phi, varphi, monkeypatch, tol):
+
+@pytest.mark.parametrize("theta, phi, varphi", list(zip(THETA, PHI, VARPHI)))
+@pytest.mark.parametrize("shots", [None, 100000])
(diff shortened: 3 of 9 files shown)


josh146:
@thisac @trbromley just a reminder to review this one!

requirements.txt
```diff
@@ -1,6 +1,6 @@
 flaky
 numpy
-pennylane>=0.12
+git+https://github.com/PennyLaneAI/pennylane.git
```
thisac, line 3: Do we want to wait until after release before merging this (and then update the requirements to the latest release instead)?

thisac (approved):
Looks good! Just one question: are the PL device tests updated as well, and should then the line `pl-device-test --device lightning.qubit --analytic False --skip-ops --shots=20000` in `build.yml` be changed here?

josh146:
> Just one question: are the PL device tests updated as well, and should then the line pl-device-test --device lightning.qubit --analytic False --skip-ops --shots=20000 in build.yml be changed here?

Good catch! I'll change it and see

requirements.txt
josh146: In the past we have, but we have so many plugins now, I think it is better to merge it as-is (and simply replace it with `pennylane>=0.15` once it is on PyPI).

By merging it in now, we can:

- Have its entry in the test matrix turn green,
- Keep track of it breaking (again!) between now and release :)

pennylane_lightning/lightning_qubit.py
```diff
@@ -35,12 +35,9 @@ class LightningQubit(DefaultQubit):
     Args:
         wires (int): the number of wires to initialize the device with
         shots (int): How many times the circuit should be evaluated (or sampled) to estimate
-            the expectation values. Defaults to 1000 if not specified.
-            If ``analytic == True``, then the number of shots is ignored
-            in the calculation of expectation values and variances, and only controls the number
-            of samples returned by ``sample``.
-        analytic (bool): indicates if the device should calculate expectations
-            and variances analytically
+            the expectation values. Defaults to 1000 if not specified. Setting
```
trbromley, line 38:
```suggestion
            the expectation values. Defaults to ``None`` if not specified. Setting
```

tests/test_apply.py
```diff
@@ -1135,7 +1134,7 @@ class TestTensorSample:
 
     def test_paulix_pauliy(self, theta, phi, varphi, monkeypatch, tol):
         """Test that a tensor product involving PauliX and PauliY works correctly"""
-        dev = qml.device("lightning.qubit", wires=3, shots=10000)
+        dev = qml.device("lightning.qubit", wires=3, shots=None)
```
trbromley, line 1137: Why does this one need to change?

tests/test_expval.py
```diff
@@ -30,7 +30,7 @@
 
 
 @pytest.mark.parametrize("theta, phi", list(zip(THETA, PHI)))
-@pytest.mark.parametrize("shots", [8192])
+@pytest.mark.parametrize("shots", [None])
```
trbromley, line 33: Why do these ones need to change?

tests/test_var.py
```diff
@@ -30,13 +30,13 @@
 
 
 @pytest.mark.parametrize("theta, phi", list(zip(THETA, PHI)))
-@pytest.mark.parametrize("shots", [8192])
+@pytest.mark.parametrize("shots", [None])
```
trbromley, line 33: Do we need to parametrize over `None` (which is the default)?

trbromley (changes requested):
Thanks @antalszava! I had a few questions/comments, but otherwise looking good!

tests/test_apply.py
antalszava: Because we implicitly had `analytic=True` before

tests/test_expval.py
antalszava: Same as above

tests/test_var.py
antalszava: No, but in this PR rather changed the analytic argument rather than adjust how the testing is done

antalszava:
Thanks for the comments and the catches! Updated the PR

tests/test_apply.py
trbromley: :thinking: This one's still not clear to me - so before we had `analytic=True` and `shots=10000`. In the test, we call `dev.sample()` so it would have generated `10000` samples, yet these samples were not used in `dev.probability()` because we were in analytic mode?

So perhaps the test didn't need to set `shots` or call `dev.sample()`?

Would this just work if we had a parametrize over shots, setting `[None, 10000]`, so we test both analytic and non-analytic cases?

tests/test_expval.py
trbromley: Actually, do you agree that the `shots` fixture is not actively used in any of the tests below? :confused: Unless it's hidden somewhere in a fixture. So we could set to any value.

In which case, ideally we would remove the paramtrize and `shots` argument. But ok for this PR if we want to just change to `None` as you have done :+1:

trbromley (commented):
Thanks @antalszava! Looks close to :heavy_check_mark:, I just had a few final questions.

pennylane_lightning/lightning_qubit.py
```diff
@@ -35,12 +35,9 @@ class LightningQubit(DefaultQubit):
...
-            in the calculation of expectation values and variances, and only controls the number
-            of samples returned by ``sample``.
-        analytic (bool): indicates if the device should calculate expectations
-            and variances analytically
+            the expectation values. Defaults to ``None`` if not specified. Setting
+            to ``None`` results in computing statistics like expectation values and
+            variances analytically.
     """
 
     name = "Lightning Qubit PennyLane plugin"
```
trbromley, line 43: Also, we need to update `pennylane_requires = ">=0.12"` to the latest version

tests/test_expval.py
antalszava: Yep, removed it

tests/test_apply.py
antalszava: Sure, changed this. Also removed the calls to `dev.sample`.

trbromley (approved):
Thanks @antalszava for accommodating those suggestions! :100:

PR #88, opened 2021-03-15
Add the light optimization technique for gate merging

antalszava:
- [X] Add the `QubitUnitary` operation
- [x] Define `merge_gates`
- [x] Add the light optimization
- [x] Create the `create` member of `QubitUnitary` (?)
- [ ] Control qubits and target qubits on gates
- [ ] Polish, docs & checks
- [ ] UI for turning on/off circuit opt

PR #100, opened 2021-05-04
Matrix utility functions for using `std::vector`

antalszava:
PennyLane-Lightning uses a row-major storage approach to represent matrices in a `std::vector`.

This PR adds 4 utility functions for dealing with matrices stored in a `std::vector`:
* `set_block`: sets the block of a matrix with a submatrix,
* `swap_cols`: swaps two columns of a matrix,
* `swap_rows`: swaps two columns of a matrix,
* `create_identity`: creates the identity matrix with the specified dimension.

These functions can be used as a convenience. This PR reduces some of the new functionality added in #88.

### DIFF
--- a/pennylane_lightning/src/tests/Makefile
+++ b/pennylane_lightning/src/tests/Makefile
@@ -13,8 +13,7 @@
 # limitations under the License.
 CC = g++
 CFLAGS = -std=c++11 -g -Wall -fopenmp -fPIC -I/usr/include \
-	-I../include -I$(GOOGLETEST_DIR)/include \
-	-march=native
+	-I../include -I$(GOOGLETEST_DIR)/include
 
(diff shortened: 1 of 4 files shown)


pennylane_lightning/src/tests/Makefile
```diff
@@ -13,8 +13,7 @@
 # limitations under the License.
 CC = g++
 CFLAGS = -std=c++11 -g -Wall -fopenmp -fPIC -I/usr/include \
-	-I../include -I$(GOOGLETEST_DIR)/include \
-	-march=native
```
antalszava, line 17: Had problems locally on a new machine with this command. It should be safe to be removed as it is used for machine-specific optimized code, which is not particularly necessary when running unit tests.

PR #101, opened 2021-05-20
Fix the parameters' order in EXPECT_EQ macro

jediMunees:
EXPECT_EQ is expecting "expected_value, actual_value". But it is having the parameters in reverse order.


jediMunees:
#102 - Issue reference.

antalszava (approved):
Looks good to me! :+1: Thanks for catching this!

Issue #103, opened 2021-06-21
Add CMake build support for tests


mlxd:
Currently, the package has compilation support using a Makefile, and additional build/install support within `setup.py`. Replacing the tests Makefile (Linux/Mac only) with CMake will allow for a much easier compilation experience.

The following CMake features will be required:
- Use features of CMake language from version 3.14 and previous only.
- Enable C++11 as required C++ standard.
- Link against dependent libraries: Pybind11
- Build generates test binary (see [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/src/tests/Makefile)) that runs C++ test suite and mirrors existing Makefile build.

The following features are currently optional:
- Testing and verification of cross-platform compilation (CMake is supported on Win, Mac, Linux).
- Enable compile-time flags for different optimizations (e.g `-march=native`, `-flto`, etc.).
- Tie tests into CMake CTest infrastructure.
- Support for entire package to be compiled using CMake.

chaeyeunpark:
Is pybind11 necessary? It seems like the test code does not use pybind11.

mlxd:
Hi @chaeyeunpark 
The C++-only tests can be built without Pybind11, so feel free to leave this library out if you think it is not needed here.

Though, for the optional feature `Support for entire package to be compiled using CMake.`  Pybind11 will most-definitely be a requirement :slightly_smiling_face:

PR #104, opened 2021-06-23
Adding cmake support

chaeyeunpark:
Hi,

I added simple CMakeLists.txt to manage #103. It now only supports Linux environments (as I don't have other machines). 

Most of CMakeLists.txt is a direct translation of setup.py besides Intel MKL support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,11 +23,13 @@ set(PROJECT_VERSION ${VERSION_STRING})
 set(CMAKE_CXX_STANDARD 11) # At least C++11 is required
 find_package(OpenMP REQUIRED) # find OpenMP
 
+
 if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE Release)
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native -W -DVERSION_INFO=${VERSION_STRING}")
+
+option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)
 
 
 # Add pybind11
@@ -43,17 +45,13 @@ FetchContent_MakeAvailable(pybind11)
 add_library(external_dependency INTERFACE)
 
 
-if (("$ENV{USE_MKL}" OR "${USE_MKL}") AND (DEFINED ENV{MKLROOT}))
-	message(STATUS "Found MKL")
-	target_link_libraries(external_dependency INTERFACE mkl_intel_lp64 mkl_gnu_thread 
-		mkl_core gomp pthread m dl)
-	target_include_directories(external_dependency INTERFACE "$ENV{MKLROOT}/include")
-	target_compile_options(external_dependency INTERFACE "-DMKL=1")
-elseif ("$ENV{USE_LAPACK}" OR "${USE_LAPACK}")
-      message(STATUS "Use LAPACKE")
+if ("$ENV{USE_LAPACK}" OR "${USE_LAPACK}")
+	message(STATUS "Use LAPACKE")
 	target_link_libraries(external_dependency INTERFACE lapacke)
 	target_compile_options(external_dependency INTERFACE "-DLAPACKE=1")
-elseif ("$ENV{USE_OPENBLAS}" OR "${USE_OPENBLAS}")
+endif()
+
+if ("$ENV{USE_OPENBLAS}" OR "${USE_OPENBLAS}")
 	message(STATUS "Use OPENBLAS")
 	target_link_libraries(external_dependency INTERFACE openblas)
 	target_compile_options(external_dependency INTERFACE "-DOPENBLAS=1")
@@ -67,6 +65,15 @@ target_link_libraries(lightning_qubit_ops PRIVATE external_dependency)
 set_target_properties(lightning_qubit_ops PROPERTIES CXX_VISIBILITY_PRESET hidden)
 
 target_link_libraries(lightning_qubit_ops PRIVATE OpenMP::OpenMP_CXX)
+target_compile_options(lightning_qubit_ops PRIVATE "$<$<CONFIG:RELEASE>:-W>")
+target_compile_options(lightning_qubit_ops PRIVATE "$<$<CONFIG:DEBUG>:-Wall>")
+target_compile_definitions(lightning_qubit_ops PRIVATE VERSION_INFO=${VERSION_STRING})
+
+if(ENABLE_NATIVE)
+	message(STATUS "ENABLE_NATIVE is ON. Use -march=native for lightning_qubit_ops.")
+	target_compile_options(lightning_qubit_ops PRIVATE -march=native)
+endif()
+
 
 if (${BUILD_TESTS})
 	enable_testing()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ endif()
 
 
 option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)
+option(BUILD_TESTS "Build cpp tests" OFF)
 
 
 # Add pybind11
@@ -75,7 +76,7 @@ if(ENABLE_NATIVE)
 endif()
 
 
-if (${BUILD_TESTS})
+if (BUILD_TESTS)
 	enable_testing()
 	add_subdirectory("pennylane_lightning/src/tests" "tests")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ add_subdirectory(pybind11)
 add_library(external_dependency INTERFACE)
 
 
-if (("$ENV{USE_MKL}" or "${USE_MKL}") AND (DEFINED ENV{MKLROOT}))
+if (("$ENV{USE_MKL}" OR "${USE_MKL}") AND (DEFINED ENV{MKLROOT}))
 	message(STATUS "Found MKL")
 	target_link_libraries(external_dependency INTERFACE mkl_intel_lp64 mkl_tbb_thread 
 		mkl_core tbb pthread m dl)
@@ -42,6 +42,7 @@ pybind11_add_module(lightning_qubit_ops "pennylane_lightning/src/Apply.cpp"
 	"pennylane_lightning/src/Gates.cpp" "pennylane_lightning/src/Bindings.cpp"
 	"pennylane_lightning/src/StateVector.cpp")
 target_link_libraries(lightning_qubit_ops PRIVATE external_dependency)
+set_target_properties(lightning_qubit_ops PROPERTIES CXX_VISIBILITY_PRESET hidden)
 
 if(OpenMP_CXX_FOUND)
 	target_link_libraries(lightning_qubit_ops PRIVATE OpenMP::OpenMP_CXX)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ message(STATUS "Pennylane_lightning version ${VERSION_STRING}")
 set(PROJECT_VERSION ${VERSION_STRING})
 
 set(CMAKE_CXX_STANDARD 11) # At least C++11 is required
-find_package(OpenMP) # find OpenMP
+find_package(OpenMP REQUIRED) # find OpenMP
 
 if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE Release)
@@ -34,8 +34,9 @@ add_library(external_dependency INTERFACE)
 
 if (("$ENV{USE_MKL}" OR "${USE_MKL}") AND (DEFINED ENV{MKLROOT}))
 	message(STATUS "Found MKL")
-	target_link_libraries(external_dependency INTERFACE mkl_intel_lp64 mkl_tbb_thread 
-		mkl_core tbb pthread m dl)
+	target_link_libraries(external_dependency INTERFACE mkl_intel_lp64 mkl_gnu_thread 
+		mkl_core gomp pthread m dl)
+	target_include_directories(external_dependency INTERFACE "$ENV{MKLROOT}/include")
 	target_compile_options(external_dependency INTERFACE "-DMKL=1")
 elseif ("$ENV{USE_LAPACK}" OR "${USE_LAPACK}")
     message(STATUS "Use LAPACKE")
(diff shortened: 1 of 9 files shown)


mlxd:
Hi @chaeyeunpark 
We will assign some reviewers to have a look at this shortly, and get back to you soon with a code-review as soon as possible :slightly_smiling_face:

co9olguy:
Thanks @chaeyeunpark! Be sure to let us know when the PR is ready for review.

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+set(PROJECT_VERSION ${VERSION_STRING})
+
+set(CMAKE_CXX_STANDARD 11) # At least C++11 is required
+find_package(OpenMP REQUIRED) # find OpenMP
+
+if(NOT CMAKE_BUILD_TYPE)
+	set(CMAKE_BUILD_TYPE Release)
+endif()
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native -W -DVERSION_INFO=${VERSION_STRING}")
```
mlxd, line 19: It may be a good idea to split the native compilation into a user-selectable option, such as:
`option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)` which can then be enabled per target using [target_compile_options](https://cmake.org/cmake/help/v3.14/command/target_compile_options.html) with something like 
```
if(ENABLE_NATIVE)
    target_compile_options(TargetName INTERFACE -march=native)
endif()
```
This way we can more easily see the performance differences with the library depending on architecture-specific features.

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+
+# Add pybind11
+include(FetchContent)
+FetchContent_Declare(
+	pybind11
+	GIT_REPOSITORY https://github.com/pybind/pybind11.git
+	GIT_TAG        stable
```
mlxd, line 27: I was unaware of the `stable` tag. Nice find! 

Though, I think it may be a better idea to pin to a specific version than a rolling version, to ensure we never hit problems with changes on their side (and we can always bump the version numbers later). For now, I think `v2.6.2` would be good here.

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+add_library(external_dependency INTERFACE)
+
+
+if (("$ENV{USE_MKL}" OR "${USE_MKL}") AND (DEFINED ENV{MKLROOT}))
+	message(STATUS "Found MKL")
+	target_link_libraries(external_dependency INTERFACE mkl_intel_lp64 mkl_gnu_thread 
+		mkl_core gomp pthread m dl)
+	target_include_directories(external_dependency INTERFACE "$ENV{MKLROOT}/include")
+	target_compile_options(external_dependency INTERFACE "-DMKL=1")
```
mlxd, line 40: MKL is always a challenge.

One thing that may cause issues is assuming the user will be using GCC, and hence only require the GNU OpenMP library (what happens if they use icpc and hence want `iomp` ?).

Also, are we always guaranteed to use the LP64 interface here?

Rather than explicitly list the MKL libs, following the [link-line advisor](https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/onemkl/link-line-advisor.html), it may be best to opt for linking against the single dynamic library option:
```
 -L${MKLROOT}/lib/intel64 -lmkl_rt -Wl,--no-as-needed -lpthread -lm -ldl
```

and for the headers:

```
 -m64  -I"${MKLROOT}/include"
```

What do you think?

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+project(pennylane_lightning_tests)
+
+set(CMAKE_CXX_STANDARD 11)
+find_package(OpenMP)
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
```
mlxd, line 11: As with Pybind11, it may be best to choose a version tag here, rather than a URL. `release-1.11.0` seems to be the latest. If that allows the tests to build and run, I reckon it is fine to use it. Or, if not, choose the tagged version of the URL given.

mlxd (changes requested):
Hi @chaeyeunpark 
Thanks for the work on this, and good job! I have a few comments and suggestions. Feel free to read below and let me know what you think:

CMakeLists.txt
trevor-vincent: Can we just remove these BLAS/LAPACK lines until BLAS/LAPACK is used? Likely Lapack won't be needed anyway, at least it won't be needed for the fundamental operations in a full state vector simulation.

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+elseif ("$ENV{USE_LAPACK}" OR "${USE_LAPACK}")
+    message(STATUS "Use LAPACKE")
+	target_link_libraries(external_dependency INTERFACE lapacke)
```
Mandrenkov, line 43: The indentation here is a little off:
```suggestion
      message(STATUS "Use LAPACKE")
	target_link_libraries(external_dependency INTERFACE lapacke)
```

trevor-vincent (changes requested):
Thanks chaeyeunpark! Just a few comments.

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+	target_compile_options(external_dependency INTERFACE "-DLAPACKE=1")
+elseif ("$ENV{USE_OPENBLAS}" OR "${USE_OPENBLAS}")
+	message(STATUS "Use OPENBLAS")
+	target_link_libraries(external_dependency INTERFACE openblas)
+	target_compile_options(external_dependency INTERFACE "-DOPENBLAS=1")
+endif()
+
+pybind11_add_module(lightning_qubit_ops "pennylane_lightning/src/Apply.cpp"
+	"pennylane_lightning/src/Gates.cpp" "pennylane_lightning/src/Bindings.cpp"
+	"pennylane_lightning/src/StateVector.cpp")
```
Mandrenkov, line 53: Minor point, but it will be easier to see if we missed a file by listing the sources in a column:
```suggestion
pybind11_add_module(lightning_qubit_ops "pennylane_lightning/src/Apply.cpp"
                                        "pennylane_lightning/src/Bindings.cpp"
                                        "pennylane_lightning/src/Gates.cpp" 
                                        "pennylane_lightning/src/StateVector.cpp")
```

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+)
+# For Windows: Prevent overriding the parent project's compiler/linker settings
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+enable_testing()
```
Mandrenkov, line 16: Is this necessary given that we've already enabled testing in the parent CMakeLists.txt?

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+cmake_minimum_required(VERSION 3.14)
+
+project(pennylane_lightning)
+
+# Read and set pennylane_lightning version
+execute_process(COMMAND "/usr/bin/env" "python"  "-c"
+	"with open('${CMAKE_SOURCE_DIR}/pennylane_lightning/_version.py') as f: print(f.readlines()[-1].split()[-1].strip(\"\\\"'\"), end='')"
```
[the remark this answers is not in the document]
chaeyeunpark: Thank you for your feedback! Of course, the File command is better for consistency, but I called python because handling a file using CMake becomes easily tedious. Still, let me try a bit more to find a compact CMake only way to do this.

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+add_executable(cpptests lightning_util_unittest.cpp lightning_gates_unittest.cpp "../Gates.cpp" lightning_apply_unittest.cpp "../Apply.cpp")
```
Mandrenkov, line 18: Similar to a previous comment, it would be nice to organize the sources as follows:
```suggestion
add_executable(cpptests lightning_apply_unittest.cpp
                        lightning_gates_unittest.cpp
                        lightning_util_unittest.cpp
                        "../Apply.cpp"
                        "../Gates.cpp")
```

CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+	OUTPUT_VARIABLE VERSION_STRING)
+message(STATUS "Pennylane_lightning version ${VERSION_STRING}")
+set(PROJECT_VERSION ${VERSION_STRING})
+
+set(CMAKE_CXX_STANDARD 11) # At least C++11 is required
+find_package(OpenMP REQUIRED) # find OpenMP
```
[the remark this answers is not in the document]
chaeyeunpark: Thanks, trevor-vincent, for your comments!
Me neither, but `setup.py` includes `-fopenmp` always (unless the target is Darwin). I thought it is included for the future, so I also included it. Of course, it can be removed if unnecessary.

pennylane_lightning/src/tests/CMakeLists.txt
chaeyeunpark: I totally agree with you! I just left it as the default version in https://google.github.io/googletest/quickstart-cmake.html as I didn't know which version is the most suitable.

Mandrenkov (commented):
Apart from the comments from Lee and Trevor, this looks good to me; I only have a few optional stylistic suggestions.

pennylane_lightning/src/tests/CMakeLists.txt
chaeyeunpark: Hi Mandrenkov,

thank you for all the comments. I added this to support the build and test of `tests` directory independently, but it can be removed if it is not a common use case.

CMakeLists.txt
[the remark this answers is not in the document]
chaeyeunpark: I added a pure CMake function that extract version from `_version.py` file in 14b2cfd8d48597e7d6b0644d5936cbecea01b349. Does it seem nice? I don't know if there is any better way to do this.

CMakeLists.txt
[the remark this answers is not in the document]
mlxd: I think it is fine right now to mirror the existing Makefile/setup.py directions. Future refactors can remove redundant features across the package.

CMakeLists.txt
mlxd: I think the same directions as earlier --- rather than creeping the scope, mirroring the existing build process is best here, and we can remove these later across the entire package.

pennylane_lightning/src/tests/CMakeLists.txt
mlxd: Sure thing. I think once the tests run locally with the latest release (or a previous iteration or two), we can pin it to that version (release-1.11.0).

pennylane_lightning/src/tests/CMakeLists.txt
mlxd: Keeping this in both is a good call, as they can be run independently. I'm in favour of keeping it.

CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+	file(STRINGS ${VERSION_FILE_PATH} VERSION_FILE_STR)
+	foreach (LINE IN LISTS VERSION_FILE_STR)
+	if("${LINE}" MATCHES "__version__.*")
+		set(VERSION_LINE_STR "${LINE}")
+	endif()
+	endforeach()
+
+	string(REGEX REPLACE "__version__ = \"(.*)\"" "\\1" VERSION_STRING ${VERSION_LINE_STR})
+	set(VERSION_STRING ${VERSION_STRING} PARENT_SCOPE)
+endfunction()
```
mlxd, line 16: Sure, I am fine with this --- as you said earlier, doing file operations with CMake is never fun, but this approach is understandable.

mlxd (changes requested):
Hi @chaeyeunpark just a quick update:

Once you have finished making your edits, it would be great to update the documentation to explain how to use the CMake build for the tests and the package. This will easily allow users to make use of the new feature, and us to test things locally :slightly_smiling_face:

CMakeLists.txt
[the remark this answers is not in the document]
mlxd: Yep, I'm good with this.

pennylane_lightning/src/tests/CMakeLists.txt
Mandrenkov: Sure.  I guess it doesn't hurt to have it in both places.

CMakeLists.txt
chaeyeunpark: Thanks! I incorporate this in f0066d3d417c01c0f0f716851f75fc7447dd5ac8.

pennylane_lightning/src/tests/CMakeLists.txt
chaeyeunpark: Updated in f0066d3d417c01c0f0f716851f75fc7447dd5ac8

CMakeLists.txt
chaeyeunpark: As I am not sure what is the best choice, I just removed the MKL part in f0066d3d417c01c0f0f716851f75fc7447dd5ac8 following `setup.py`, as @mlxd suggested.

mlxd: Sure, I think that is fine for now. We will likely rework the BLAS functionality eventually, and MKL support will be part of this.

chaeyeunpark (commented):
@mlxd I have added a simple CMake instruction in `README.rst`.

CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+target_link_libraries(lightning_qubit_ops PRIVATE OpenMP::OpenMP_CXX)
+target_compile_options(lightning_qubit_ops PRIVATE "$<$<CONFIG:RELEASE>:-W>")
+target_compile_options(lightning_qubit_ops PRIVATE "$<$<CONFIG:DEBUG>:-Wall>")
+target_compile_definitions(lightning_qubit_ops PRIVATE VERSION_INFO=${VERSION_STRING})
+
+if(ENABLE_NATIVE)
+	message(STATUS "ENABLE_NATIVE is ON. Use -march=native for lightning_qubit_ops.")
+	target_compile_options(lightning_qubit_ops PRIVATE -march=native)
+endif()
+
```
mlxd, line 77:
```suggestion
```

mlxd (approved):
Nice work @chaeyeunpark 
I am happy to see this merged in.

mlxd:
It seems the tests are failing due to deprecation of Python 3.6 with PennyLane 0.16.0. I'll create an issue for that to be resolved, and we can merge this in once the issue is fixed.

PR #107, opened 2021-07-01
Adjoint jacobian updates

antalszava:
**Context**

Adjusts some parts of the logic in the `adjoint-jacobian` branch to produce correct results for most cases when using the adjoint differentiation method.

**Changes**

1. Changed the signature of the `adjoint_jacobian` method to correspond to the signature used by `QubitDevice`.
2. Added a new `qubits` argument to the bound `adjoint_jacobian` function to pass the number of qubits required.
3. Calls `self.reset()` to reset the device before passing the device state to `adjoint_jacobian`. Note that this is required for example when using `qml.grad`, where the forward pass already includes a circuit execution and the state vector is computed. Due to the reset, the forward evaluation will be done twice in such cases, however. See 4. point of the Remaining tasks, where adding certain keyword arguments could help.
4. Added a new `current_param_idx` variable and an `if` statement to keep track of the index for the current parametrized gate that is being processed and to only do a check if we have a parameterized gate.
5. Added the circuit execution that applies the unitaries to lambda (step 2. of the algorithm).

**Remaining tasks not addressed in this PR**

1. Revisit the logic of inverting operations when using the adjoint method (`lightning_qubit.py` file).
2. Generalize the logic of using a scaling factor when computing the gradient in C++.
3. Support tensor product observables: currently a QNode with a `return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))` statement is not supported. The logic in C++ would need to apply observables on both wires (currently it is assumed that exactly one observable is inputted).
4. Add support for the `starting_state` and `use_device_state` keyword arguments.

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -140,7 +140,7 @@ class LightningQubit(DefaultQubit):
         apply(state_vector, op_names, op_wires, op_param, op_inverse, self.num_wires)
         return np.reshape(state_vector, state.shape)
 
-    def adjoint_jacobian(self, tape):
+    def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
         for m in tape.measurements:
             if m.return_type is not Expectation:
                 raise QuantumFunctionError(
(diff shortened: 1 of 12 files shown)


pennylane_lightning/lightning_qubit.py
```diff
@@ -140,7 +141,7 @@ def apply_lightning(self, state, operations):
         apply(state_vector, op_names, op_wires, op_param, op_inverse, self.num_wires)
         return np.reshape(state_vector, state.shape)
 
-    def adjoint_jacobian(self, tape):
+    def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
```
antalszava, line 144: This is an update to the new `adjoint_jacobian` signature present in `QubitDevice` too.

pennylane_lightning/src/Apply.hpp
```diff
@@ -109,6 +109,8 @@ namespace Pennylane {
         const unsigned int qubits
     );
 
+    StateVector copy_state_vector(StateVector& state_to_copy);
+
```
antalszava, line 113:
```suggestion
```

pennylane_lightning/src/Apply.cpp
```diff
@@ -156,56 +165,56 @@ void Pennylane::adjointJacobian(
...
-                for (unsigned int j = 0; j < lambdas.size(); j++) {
-                    int lambdaStateSize = lambdas[j].length;
+            // We have a parametrized gate
+            if (!opParams[i].empty()){
+                if (std::find(trainableParams.begin(), trainableParams.end(), current_param_idx) != trainableParams.end()) {
+
+                    // create iH|phi> = d/d dUj/dtheta Uj* |phi> = dUj/dtheta|phi'>
+                    unique_ptr<AbstractGate> gate = constructGate(operations[i], opParams[i]);
+                    // double scalingFactor = gate->generatorScalingFactor;
+                    double scalingFactor = Pennylane::RotationYGate::generatorScalingFactor;
```
antalszava, line 190: The scaling factor will have to be updated here to be specific to the operation (e.g., the phase shift operation will fail due to having another scaling factor than most parameterized gates).

antalszava:
[ch5827]

PR #110, opened 2021-07-09
Build support for multiple architectures

mlxd:
**Context:** This PR enables wheel-building for previously unsupported hardware architectures.

**Description of the Change:** Wheel-build support is added for the following hardware architectures:
   - Linux: PowerPC `(ppc64le)`, ARM64 `(aarch64)`
   - MacOS: Apple M1 ARM `(arm64)`

The wheel-builder actions have also been separated to more easily allow selective building per PR created. This is partly due to the parsing order of the Github Actions YAML, where job `if` statements are evaluated before any matrix statements, introducing skipped tests with undefined potentially overlapping labels.

All x86_64 wheel build operations are run for each PR created. For Linux `ppc64le` and `aarch64`, and MacOS `arm64` these are built only when a PR is merged with master, due to the additional overheads required.

**Benefits:** This will simplify installation and use on the above platforms.

**Possible Drawbacks:** Additional efforts may be required to keep build-system up-to-date with any modifications or changes. Additional cluttering of workflows directory due to lack of support for nested Actions directories.


mlxd:
[ch7059]

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
+name: Wheels
+on:
+  push:
+    branches:
+      - master
+
+env:
+  CIBW_BUILD: 'cp37-* cp38-* cp39-*'
```
trbromley, line 8: Not important for this PR, but I wonder if this will be a pain to update as we update Python versions.

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  # MacOS specific build settings
+  CIBW_BEFORE_ALL_MACOS: |
+    brew uninstall --force oclint
+    brew install gcc libomp
```
trbromley, line 13: Maybe a silly question, but do we only technically need this for the mac builds? But maybe it's easier to keep everywhere?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  # Python build settings
+  CIBW_BEFORE_BUILD: |
+    pip install pybind11
+
+  # Testing of built wheels
+  CIBW_TEST_REQUIRES: numpy scipy pytest pytest-cov pytest-mock flaky
+
+  # Note: breaking up tests so that no test cases hang on Windows, to be reverted
```
trbromley, line 22: I didn't follow what this comment was referring to

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+  CIBW_TEST_COMMAND: |
+    pip install numpy==1.19.5 scipy
```
trbromley, line 24: Do we need this if they're specified in `CIBW_TEST_REQUIRES`?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+    pl-device-test --device=lightning.qubit --skip-ops -x --tb=short --no-flaky-report
```
trbromley, line 25: Are we skipping the `pl-device-test --device lightning.qubit --skip-ops --shots=20000` line? Is this what was meant with the breaking up tests comment?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  # Skip PPC tests due to lack of numpy/scipy wheel support
+  CIBW_TEST_SKIP: "*-manylinux_{aarch64}"
```
trbromley, line 28: Since NumPy and Scipy are dependencies of PL, do people typically build them from source on PPC?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  # Use Centos 7 wheel-builder for ARM, Debian for PPC
+  CIBW_MANYLINUX_AARCH64_IMAGE: manylinux2014
+
+jobs:
+  linux-wheels-aarch64:
+    if: github.ref == 'refs/heads/master'
```
trbromley, line 35: Is this a double check that we're only building on `master`?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+
+      - uses: docker/setup-qemu-action@v1
+        name: Set up QEMU
```
trbromley, line 53: What does this do?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==1.12.0
```
trbromley, line 61: Is it best to version pin?

.github/workflows/wheel_linux_ppc64le.yml
```diff
@@ -0,0 +1,64 @@
...
+  CIBW_BEFORE_ALL_MACOS: |
+    brew uninstall --force oclint
+    brew install gcc libomp
+
+  # Python build settings
+  CIBW_BEFORE_BUILD: |
+    pip install pybind11
+
+  # Skip PPC tests due to lack of numpy/scipy wheel support
+  CIBW_TEST_SKIP: "*-manylinux_{ppc64le}"
```
trbromley, line 20: Sorry for the questions on the previous script, I think I realise that this is supposed to be a common preamble for all of them.

.github/workflows/wheel_linux_ppc64le.yml
```diff
@@ -0,0 +1,64 @@
...
+
+  # Use Debian image for PPC
+  CIBW_MANYLINUX_PPC64LE_IMAGE: manylinux_2_24
```
trbromley, line 23: Does the little-endian vs big-endian subtlety matter? E.g., does it not make sense for PPC64 (i.e. without the LE)?

.github/workflows/wheel_macos_arm64.yml
```diff
@@ -0,0 +1,64 @@
...
+  CIBW_TEST_REQUIRES: numpy scipy pytest pytest-cov pytest-mock flaky
+
+  # Note: breaking up tests so that no test cases hang on Windows, to be reverted
+  CIBW_TEST_COMMAND: |
+    pip install numpy==1.19.5 scipy
+    pl-device-test --device=lightning.qubit --skip-ops -x --tb=short --no-flaky-report
+
+jobs:
+  mac-wheels-arm64:
+    strategy:
```
trbromley, line 29: Do we need an if statement here?

.github/workflows/wheel_linux_aarch64.yml
trbromley: Also, do we skip the local tests (`pytest tests/`) to keep things faster for wheel builds?

trbromley (commented):
Thanks @mlxd! This is really cool!
Have left a few comments, mainly for my understanding/learning.

.github/workflows/wheel_linux_aarch64.yml
mlxd: I agree. Though, this replaces the previous implementation which used an exclude arch for each type we don't intend to include --- I think neither are perfect, but I think this can be something to revisit in future.

.github/workflows/wheel_linux_aarch64.yml
mlxd: Well spotted. I thought I had pulled all of those --- will remove it. Cheers.

.github/workflows/wheel_linux_aarch64.yml
mlxd: Nor I: I think this may be a left-over from a previous refactor. I think it will be safe to remove.

.github/workflows/wheel_linux_aarch64.yml
josh146: I would say inclusion is better than exclusion :+1:

Better to manually decide one day to support Python 3.10, rather than discovering that one day PR CI is failing because 3.10 builds got turned on automatically!

.github/workflows/wheel_linux_aarch64.yml
mlxd: Good point. I have removed these throughout.

.github/workflows/wheel_linux_aarch64.yml
mlxd: It would seem so, though I am not sure. As far as I can tell this may be some legacy options that no longer apply (ie hanging tests). Do you think the shots=20000 test would be a good inclusion here too?

As for skipping the `tests` dir, I think that is safe here, as they are tested as part of the pip install build elsewhere. Though, if you'd prefer, we can run these on a merge with master, rather than on a PR by PR basis?

.github/workflows/wheel_linux_aarch64.yml
mlxd: Yea, that's right. Or, they can be provided on a system by system basis (usually the HPC sysadmins provide default packages to be loaded). Conda also has direct support for them too.   

I toyed with the idea of building our own for this purpose, but it seemed too much a challenge for now to do so in a generalised manner.

.github/workflows/wheel_linux_aarch64.yml
mlxd: Indeed. These can be safely removed, though I'd rather keep the artifact upload checks as an additional safety measure.

.github/workflows/wheel_linux_aarch64.yml
josh146: Oh, I can explain this one. It was a comment added in this PR here: https://github.com/PennyLaneAI/pennylane-lightning/pull/59

In particular, the test `test_four_qubit_random_circuit` was hanging on Windows, likely due to an issue with pytest. We were never able to work out why it hanged, _but_ it must have been something to do with previous tests not cleaning up?

Because if we ran this test by itself, it didn't hang.

.github/workflows/wheel_linux_aarch64.yml
mlxd: QEMU provides an emulation layer for building the ARM and PowerPC libraries. Since we don't have access to the hardware directly, we use the emulator to provide build support for our target machines.

The alternative is to build cross-compilers directly and explicitly handle the compilation ourselves. That would be faster overall, but take a much longer time to set-up, and maintaining the tooling would be a challenge.

.github/workflows/wheel_linux_aarch64.yml
mlxd: We had an older version pinned, so I updated to this, but I agree. I don't see any reason to do so.

mlxd: Actually, on second thought, I think this is fine for now, as it provides all the functionality we currently need.

josh146: Yes, because `cibuildwheel` updates often, changing behaviour and environment variable names. We previously did not pin, and would constantly have to fix the CI.

Having said that, it is important to make sure that we are updating the version of `cibuildwheel` as often as possible, because they are frequently fixing bugs and making sure that all build options and CI services work correctly

.github/workflows/wheel_linux_ppc64le.yml
mlxd: The original single wheel script had a global preamble, which I should have tidied a little on porting to the different arch files. There is a little repetition here, but until GH Actions allow the inclusion of external common files, I think the little repetition is no harm. 

When the actions interpreter is a little more mature, we should be able to refactor these into something more friendly.

josh146: > but until GH Actions allow the inclusion of external common files

This one really breaks my heart 💔 Come on GitHub actions 🤞

.github/workflows/wheel_linux_ppc64le.yml
mlxd: x86 machines are little-endian on whole. Power CPUs tend to support both big and little endian on the same device, but keeping it as `le` makes porting between x86 and PPC much simpler. Most of the builder tooling for this generally assumes the little-endian Power ISA.

.github/workflows/wheel_macos_arm64.yml
mlxd: This should only run on merging with master (top lines of the file).

mlxd: I have also cleaned up the if statements on the other files, as you were right earlier, they were not essential.

.github/workflows/wheel_linux_aarch64.yml
trbromley: > Do you think the shots=20000 test would be a good inclusion here too?

Maybe we can check with @antalszava, who I think fixed the hanging tests last time.

But this looks good to me as-is, we also have `pl-device-test --device lightning.qubit --skip-ops --shots=20000` running in the plugin test matrix.

> As for skipping the tests dir, I think that is safe here, as they are tested as part of the pip install build elsewhere.

Makes sense!

.github/workflows/wheel_linux_aarch64.yml
trbromley: Got it! Thanks!

trbromley (approved):
Thanks @mlxd for the comments! Approved :100:

PR #117, opened 2021-07-23
Update CI image before running tests

mlxd:
**Context:** Github Actions fail to install packages for older Ubuntu 18.04 images, with error messages matching the following pattern:
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/c/cmake/cmake-data_3.10.2-1ubuntu2.18.04.1_all.deb  404  Not Found [IP: 52.252.75.106 80]
```

It is necessary to update the base package list prior to installation of any dependencies.

**Description of the Change:** Add `sudo apt-get update` to Ubuntu 18.04 images.

**Benefits:** Github action fails no more.

**Possible Drawbacks:** None

**Related GitHub Issues:** None

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-20.04]
     steps:
       - uses: actions/checkout@v2
 
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-20.04]
 
     steps:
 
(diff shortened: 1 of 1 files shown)


trbromley (approved):
Thanks @mlxd!

Would an alternative to be just to specify to run on `ubuntu-latest`? We have a couple of other "Install dependencies" actions, but I guess they work because they use `ubuntu-latest` or `ubuntu-20.04`.

mlxd:
> Would an alternative to be just to specify to run on `ubuntu-latest`? We have a couple of other "Install dependencies" actions, but I guess they work because they use `ubuntu-latest` or `ubuntu-20.04`.

Yes, that may be the best idea. I didn't opt for this as it required a temporary deletion of the required tests in the the repo settings (explicitly listed 18.04 test), and then reinstate with the new version. I think updating to the 20.04 LTS release does make sense though. I'll update to this, and merged.

PR #123, opened 2021-08-06
Include test suite in coverage

trbromley:
The tests from PennyLane's device test suite will now be included when generating coverage reports. The makefile has also been updated to make using the test suite easier.

As we further modify some of the `default.qubit`-inherited methods in `lightning.qubit`, it'll be useful to harness the test suite tests for coverage so we don't have to replicated existing tests.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ### Improvements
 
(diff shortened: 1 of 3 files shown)


mlxd (approved):
Great idea and good job on this @trbromley !

PR #125, opened 2021-08-10
Update method signatures for bound class

mlxd:
**Context:** This PR enables direct call of the C++ backed methods with a given wire and parameter argument set. 

**Description of the Change:** Add overloads to Python bidings to allow direct call of the applicable gates without relying on the dispatch functionality.

**Benefits:** Simplifies the integration with PennyLane core.

**Possible Drawbacks:** May be less performant than directly using the C++ dispatcher.

**Related GitHub Issues:** #124

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,12 @@
 
 ### Improvements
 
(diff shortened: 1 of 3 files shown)


trbromley (approved):
Thanks @mlxd, this is great :100:

PR #126, opened 2021-08-10
Fix column major data layout in bindings

mlxd:
**Context:** This PR fixes issues observed when passing numpy data in column-major layout to the row-major backed C++ bindings.

**Description of the Change:** Numpy data is forced to C-style (row-major) layout when passing numpy arrays to the C++ backend.

**Benefits:** This ensures correctness for both row and column major data layouts.

**Possible Drawbacks:** Column major layout data will experience a small overhead (though we should be using row-major everything).

**Related GitHub Issues:** #124

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -47,6 +47,9 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


mlxd:
[ch8073]

trbromley (approved):
:100:

Issue #116, opened 2021-07-22
Update Python version support


trbromley:
We should add support for Python 3.9 and drop support for 3.6, and have this reflected in the wheels uploaded to PyPI.

josh146:
👍 

Also note that Python 3.10 is coming soon 😆

Issue #96, opened 2021-04-20
Support Python 3.9


josh146:
At the upcoming v0.16 release of Lightning, we should deprecate Python 3.6 support and enable building Python 3.9 wheels.

PR #128, opened 2021-08-11
Removes support for Python 3.6 and adds support for Python 3.9

trbromley:
The [deploy](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/.github/workflows/deploy.yml) workflow is already set up for 3.7 - 3.9, so hopefully on the next release we'll have the 3.9 binaries!

Closes https://github.com/PennyLaneAI/pennylane-lightning/issues/96
Closes https://github.com/PennyLaneAI/pennylane-lightning/issues/116

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -208,9 +208,9 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3 :: Only",
     "Topic :: Scientific/Engineering :: Physics",
 ]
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -208,9 +208,9 @@ def build_extensions(self):
     "Operating System :: Microsoft :: Windows",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3 :: Only",
```
antalszava, line 214: @trbromley would it be good to reflect this already in the v0.17.0 release? Will v0.17.0 discontinue support for Python 3.6/add support for Python 3.9?

trbromley: @antalszava yes this would be good to merge into the release branch too - I think we should drop 3.6 support in this release since it's not anyway currently supported in current PL.

antalszava: Sounds great! @albi3ro could you maybe delete the `    "Programming Language :: Python :: 3.6",` line from `setup.py` in #127?

mlxd (approved):
:100:

PR #129, opened 2021-08-11
Allow lightning.qubit to skip C++ compilation and provide a Python-only binary

trbromley:
**Context:**

PennyLane-Lightning can now be installed without compiling its C++ binaries and will fall back
  to using the `default.qubit` implementation. Skipping compilation is achieved by setting the
  `SKIP_COMPILATION` environment variable, e.g., `export SKIP_COMPILATION=True`. This feature
  is intended for building a pure-Python wheel of PennyLane-Lightning as a backup for platforms
  without a dedicated wheel.

**Possible Drawbacks:**

Users installing the non-compiled version will obviously have a device that is as fast as `default.qubit`. A warning has been added for this eventuality.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -12,9 +12,9 @@
 
 * PennyLane-Lightning can now be installed without compiling its C++ binaries and will fall back
   to using the `default.qubit` implementation. Skipping compilation is achieved by setting the
-  `SKIP_COMPILATION` environment variable, e.g., Linux/MacOS: `export SKIP_COMPILATION=True`, Windows: `set SKIP_COMPILATION=True`. This feature
-  is intended for building a pure-Python wheel of PennyLane-Lightning as a backup for platforms
-  without a dedicated wheel.
+  `SKIP_COMPILATION` environment variable, e.g., Linux/MacOS: `export SKIP_COMPILATION=True`,
+  Windows: `set SKIP_COMPILATION=True`. This feature is intended for building a pure-Python wheel of
+  PennyLane-Lightning as a backup for platforms without a dedicated wheel.
   [(129)](https://github.com/PennyLaneAI/pennylane-lightning/pull/129)
 
 * The C++-backed Python bound methods can now be directly called with wires and supplied parameters.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,6 +10,13 @@
 
 ### Improvements
 
+* PennyLane-Lightning can now be installed without compiling its C++ binaries and will fall back
+  to using the `default.qubit` implementation. Skipping compilation is achieved by setting the
+  `SKIP_COMPILATION` environment variable, e.g., `export SKIP_COMPILATION=True`. This feature
+  is intended for building a pure-Python wheel of PennyLane-Lightning as a backup for platforms
+  without a dedicated wheel.
+  [(129)](https://github.com/PennyLaneAI/pennylane-lightning/pull/129)
+
 * The C++-backed Python bound methods can now be directly called with wires and supplied parameters.
   [(125)](https://github.com/PennyLaneAI/pennylane-lightning/pull/125)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 * PennyLane-Lightning can now be installed without compiling its C++ binaries and will fall back
   to using the `default.qubit` implementation. Skipping compilation is achieved by setting the
-  `SKIP_COMPILATION` environment variable, e.g., `export SKIP_COMPILATION=True`. This feature
+  `SKIP_COMPILATION` environment variable, e.g., Linux/MacOS: `export SKIP_COMPILATION=True`, Windows: `set SKIP_COMPILATION=True`. This feature
   is intended for building a pure-Python wheel of PennyLane-Lightning as a backup for platforms
   without a dedicated wheel.
   [(129)](https://github.com/PennyLaneAI/pennylane-lightning/pull/129)
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,7 @@
 name: Deploy
 on:
-  push:
-    branches:
-      - master
-  pull_request:
+  release:
+    types: [published]
 
 env:
   CIBW_BUILD: 'cp37-* cp38-* cp39-*'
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -116,7 +116,7 @@ def build_extensions(self):
         build_ext.build_extensions(self)
 
 
-if not os.environ.get("MOCK_DOCS", False):
+if not os.environ.get("SKIP_COMPILATION", False):
```
trbromley, line 119: Reminder - I believe this will need the environment variable to be updated in RTD.

tests/test_apply.py
```diff
@@ -665,7 +666,7 @@ def circuit(x):
         for _ in range(100):
             runs.append(circuit(p))
 
-        assert np.isclose(np.mean(runs), -np.sin(p), atol=1e-3, rtol=0)
+        assert np.isclose(np.mean(runs), -np.sin(p), atol=1e-2, rtol=0)
```
trbromley, line 669: This was occasionally failing :thinking: So I thought it'd be sufficient to lower the tolerance.

.github/workflows/deploy.yml
```diff
@@ -1,95 +1,120 @@
...
+          python -m pip install --upgrade pip wheel
+          cd main
+          python setup.py bdist_wheel
+        env:
+          SKIP_COMPILATION: True
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: pure-python-wheels.zip
+          path: main/dist/*.whl
```
trbromley, line 84: This is the main change

.github/workflows/deploy.yml
```diff
@@ -1,95 +1,120 @@
...
+
+  upload-wheels:
+    runs-on: ubuntu-latest
+    needs: [build-wheels, build-pure-python-wheel]
```
trbromley, line 88: Also added job as a dependency

.github/workflows/deploy.yml
```diff
@@ -1,95 +1,120 @@
...
-        run: |
-          python -m pip install --upgrade pip wheel
-          python setup.py sdist
-
-      - name: Publish
-        uses: pypa/gh-action-pypi-publish@master
-        with:
-          user: __token__
-          password: ${{ secrets.PYPI }}
+name: Deploy
```
trbromley, line 1: Unfortunately the diff for this file is not great, I believe my editor changed from `\r\n` to `\n` for the new line style. The main changes are highlighted below.

.github/workflows/deploy.yml
```diff
@@ -1,95 +1,120 @@
...
+
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI }}
+
+  upload-source:
+    runs-on: ubuntu-latest
+    needs: [build-wheels, build-pure-python-wheel]
```
trbromley, line 102: Added as dependency, but not sure it's strictly needed.

.github/workflows/tests_without_binary.yml
```diff
@@ -0,0 +1,58 @@
+name: Testing without binary
```
trbromley, line 1: Very similar to the `tests.yml` workflow but with the c++ tests removed and a few tweaks to test only for without binaries.

setup.py
trbromley: Also, maybe we should add docs testing in the CI tests :thinking: I expected the tests to fail because of this.

setup.py
```diff
@@ -168,7 +168,6 @@ def build_extensions(self):
 requirements = [
     "numpy",
     "pennylane>=0.15",
-    "pybind11",
```
trbromley, line 171: Note this removal. This means that all the dependencies of `lightning` are also dependencies of PennyLane.

setup.py
trbromley: I think I just added following [these](https://docs.readthedocs.io/en/stable/pull-requests.html) instructions.

trbromley:
[ch8104]

.github/CHANGELOG.md
```diff
@@ -10,6 +10,13 @@
 
 ### Improvements
 
+* PennyLane-Lightning can now be installed without compiling its C++ binaries and will fall back
+  to using the `default.qubit` implementation. Skipping compilation is achieved by setting the
+  `SKIP_COMPILATION` environment variable, e.g., `export SKIP_COMPILATION=True`. This feature
```
mlxd, line 15: Just wondering, should we also list Windows env var rules?
```suggestion
  `SKIP_COMPILATION` environment variable, e.g., Linux/MacOS: `export SKIP_COMPILATION=True`, Windows: `set SKIP_COMPILATION=True`. This feature
```

or is it fine to just list Linux?

setup.py
mlxd: Just so I'm clear, this change requires the user to ensure they have pybind11 available manually, rather than as an auto-pulled dep when we try to build? As in, explictly running `pip install -r requirements.txt` is needed?

trbromley: Yes, agreed!

.github/CHANGELOG.md
trbromley: Good point, forgot about windows!

.github/workflows/deploy.yml
```diff
@@ -1,95 +1,120 @@
...
+
+
+jobs:
+
+  build-wheels:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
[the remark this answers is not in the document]
trbromley: Can we get the new build scripts (`wheel_linux_aarch64.yml`, etc) to run also on the `published` trigger (e.g., like this deploy workflow). We could then have an additional job in each workflow to upload the wheels _if_ the event is `published` (`if: ${{ github.event_name == 'published' }}`, taken from [here](https://kevsoft.net/2020/06/10/running-github-action-steps-and-jobs-only-on-push-to-master.html))?

We can then:
- Take the pure python build into it's own workflow, as you suggested below
- Reduce this deploy workflow to just uploading the source?

trevor-vincent (approved):
Really good idea Tom, I have nothing to add beyond what Lee has already commented on.

setup.py
josh146: There might be one additional use-case, not sure how common it is though:

> A user wants to `pip` install lightning on an unsupported system and compile from source.

In this case, the user would have to:

1. Make sure all C++ dependencies are installed
2. Run

   ```
   pip install pybind11
   pip install pennylane-lightning --no-binary :all:
   ```

Potentially this could occur in Docker workflows. Say a user is using the Alpine Linux image that uses `musl`; (as far as I am aware?) no wheels are available for this platform, but they

- would like to avoid using git to reduce the docker image size
- would like to use the C++ simulator

.github/workflows/deploy.yml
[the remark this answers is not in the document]
mlxd: Sure, that sounds like a good plan.

setup.py
trbromley: Good idea, have updated the instructions: 8e6ed64364470875634c77c60d23e0fc3190a41f

.github/workflows/deploy.yml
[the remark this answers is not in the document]
trbromley: Nice! Are you ok if we do that in a separate PR?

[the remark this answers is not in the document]
mlxd: Yea, perfectly fine with me.

mlxd (approved):
Great job on this @trbromley !

PR #131, opened 2021-08-13
Update Lightning API documentation

mlxd:
**Context:** This PR ensures all classes, methods and free functions in lightning are documented with doxygen-parseable docstrings. The C++ API is also now built alongside the docs and accessible on the website. This also adds 2 sample C4 architecture diagrams to the documentation, following the Container and Component (levels 2 and 3) processes of the model.

- Updates and adds docstrings for all required code. 
- Builds and hosts C++ API along-side existing documentation
- Adds C4 architecture diagrams to explain current design of Lightning relative to PennyLane ecosystem.
- Moves bindings to anon namespace to selectively scope API generation.

**Benefits:** Enables a more accessible and usable code-base.

**Possible Drawbacks:** More updates with future changes.


doc/_static/arch/system_containers_components.svg
```diff
@@ -0,0 +1,1629 @@
...
+entity python--><rect fill="#1168BD" height="89" style="stroke:#3C7FC0;stroke-width:1.5;" width="100" x="434" y="45.5999"/><text fill="#FFFFFF" font-family="sans-serif" font-size="12" font-style="italic" lengthAdjust="spacing" textLength="52" x="458" y="68.8">«system»</text><text fill="#FFFFFF" font-family="sans-serif" font-size="16" font-weight="bold" lengthAdjust="spacing" textLength="55" x="456.5" y="86.664">Python</text><text fill="#FFFFFF" font-family="sans-serif" font-size="14" lengthAdjust="spacing" textLength="0" x="486" y="104.006"/><text fill="#FFFFFF" font-family="sans-serif" font-size="14" lengthAdjust="spacing" textLength="76" x="448" y="120.8059">Python 3.7+</text><!--MD5=[6b6aa52bae2e86a2b9fe7e5e53b309bd]
```
trbromley, line 12: Minor point, but this seems like one spot we'll easily miss when updating to Python 3.8

doc/_static/arch/system_containers_components.svg
```diff
@@ -0,0 +1,1629 @@
...
+link other to default_qubit--><path d="M386.83,319.5999 C349.44,319.5999 306.39,319.5999 267.37,319.5999 " fill="none" id="other-to-default_qubit" style="stroke:#666666;stroke-width:1.0;"/><polygon fill="#666666" points="259.12,319.5999,267.12,322.5999,267.12,316.5999,259.12,319.5999" style="stroke:#666666;stroke-width:1.0;"/><text fill="#666666" font-family="sans-serif" font-size="12" font-weight="bold" lengthAdjust="spacing" textLength="30" x="308" y="280.7479">Uses</text><text fill="#666666" font-family="sans-serif" font-size="12" font-style="italic" lengthAdjust="spacing" textLength="88" x="277.5" y="297.2001">[Quantum circuit</text><text fill="#666666" font-family="sans-serif" font-size="12" font-style="italic" lengthAdjust="spacing" textLength="56" x="296.5" y="313.4002">backends]</text><!--MD5=[d07d7441133a84b5b8107f0baf4f554f]
```
trbromley, line 19: What about "Quantum device"

doc/_static/arch/system_containers_components.svg
```diff
@@ -0,0 +1,1629 @@
...
+link default_qubit to lightning_qubit--><path d="M151,363.7399 C151,403.9899 151,464.6899 151,511.5599 " fill="none" id="default_qubit-to-lightning_qubit" style="stroke:#666666;stroke-width:1.0;"/><polygon fill="#666666" points="151,519.8499,154,511.8499,148,511.8499,151,519.8499" style="stroke:#666666;stroke-width:1.0;"/><text fill="#666666" font-family="sans-serif" font-size="12" font-weight="bold" lengthAdjust="spacing" textLength="118" x="158.5" y="406.2479">Ops and data pointer</text><text fill="#666666" font-family="sans-serif" font-size="12" font-style="italic" lengthAdjust="spacing" textLength="91" x="170.5" y="422.7001">[Provides default</text><text fill="#666666" font-family="sans-serif" font-size="12" font-style="italic" lengthAdjust="spacing" textLength="128" x="155" y="438.9002">Python plugin methods]</text><!--MD5=[e74efe3d600ba974dbb7e83863c41cd6]
```
trbromley, line 24: What do we mean by "Ops and data pointer"?

doc/arch/arch_overview.rst
```diff
@@ -0,0 +1,34 @@
+Architecture overview
+=====================
+
+.. |br| raw:: html
+
+   <br />
+
+Lightning's purpose is to provide direct C++ kernels to manipulate the statevector data as created with PennyLane. Our design extends that of the "default.qubit" device in PennyLane, allowing direct access to the complex C-backed Numpy array data. This ensures we maintain all existing compatibility with PennyLane plugins, as only the gate-application is offloaded to C++.
```
trbromley, line 8:
```suggestion
Lightning Qubit's purpose is to provide direct C++ kernels to manipulate the statevector data as created with PennyLane. Our design extends that of the ``default.qubit`` device in PennyLane, allowing direct access to the complex C-backed Numpy array data. This ensures we maintain all existing compatibility with the PennyLane device API, as only the gate-application is offloaded to C++.
```

trbromley: Nice! Though going forward will we still be following the philosophy of just offloading gate application to C++? E.g., if we override the whole execute method.

doc/arch/system_context.puml
```diff
@@ -0,0 +1,33 @@
...
+!include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Container.puml
+' uncomment the following line and comment the first to use locally
+' !include C4_Context.puml
+
+LAYOUT_TOP_DOWN()
+
+title System-level component diagram for PennyLane using Lightning Qubit device
+
+Person(developer, "Quantum software developer")
+System(python, "Python", "Python 3.7+")
```
trbromley, line 12: Now I know that these are autogenerated from an easily-searcheable script, maybe putting Python 3.7 isn't so bad here!

doc/devices.rst
```diff
@@ -26,10 +26,16 @@ Supported operations and observables
 
     ~pennylane.BasisState
     ~pennylane.CNOT
+    ~pennylane.CRot
+    ~pennylane.CRX
+    ~pennylane.CRY
+    ~pennylane.CRZ
```
trbromley, line 32: Good spot!

trbromley: In the long run, maybe we shouldn't even have this list of supported ops! Every op should be supported :zap: (and saves us having to maintain this)

pennylane_lightning/src/Bindings.cpp
```diff
@@ -18,45 +18,65 @@
 #include "pybind11/pybind11.h"
 #include "pybind11/stl.h"
 
+/// @cond DEV
```
trbromley, line 21: What does this bit do? :thinking:

pennylane_lightning/src/Bindings.cpp
```diff
@@ -18,45 +18,65 @@
...
-template <class T = double>
-static StateVector<T> create(const py::array_t<complex<T>> *numpyArray) {
+/**
+ * @brief Create a `%StateVector` object from a 1D numpy complex data array.
+ *
+ * @tparam fp_t Precision data type
+ * @param numpyArray Numpy data array.
+ * @return StateVector<fp_t> `%StateVector` object.
+ */
+template <class fp_t = double>
```
trbromley, line 36: What is the relevance of `fp_t`?

trbromley (approved):
Thanks @mlxd, looks really cool in the [built docs](https://pennylane-lightning--131.org.readthedocs.build/en/131/api/library_root.html)!

Wonder why the lines are so close together:
image

doc/_static/arch/system_containers_components.svg
mlxd: This will be auto-generated from the `.puml` files.

doc/_static/arch/system_containers_components.svg
mlxd: Sure, that should work.

doc/_static/arch/system_containers_components.svg
mlxd: Good point --- "Operations and pointer to data". I should be more explicit on this

doc/devices.rst
mlxd: Sure --- do you think this entire list can be pulled? Or should we opt for that later?

pennylane_lightning/src/Bindings.cpp
mlxd: This is understood as a conditional option for Doxygen --- anything within the `@cond LABELNAME` regions will be excluded for generating API docs. We should be publishing the binding API with the C++ API, so this just ensures that entire region is ignored.

pennylane_lightning/src/Bindings.cpp
mlxd: I had thought of using `fp_t` as floating point type template type, with a default value of double. Happy to rename if you think something more relevant makes sense.

PR #132, opened 2021-08-16
Update tests due to the deprecated `qml.sample` usage when `shots=None`

antalszava:
Some test cases needed adjustment due to: https://github.com/PennyLaneAI/pennylane/pull/1522.

Note: since only the test cases are updated, there is no need to incorporate the changes in the release candidate branch. Having them in `master` will help with passing the checks in the plugin test matrix.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 flaky
 numpy
-git+https://github.com/PennyLaneAI/pennylane.git@master
+pennylane>=0.15
 pybind11
 pytest
 pytest-cov
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 flaky
 numpy
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git@master
 pybind11
 pytest
 pytest-cov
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 flaky
 numpy
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git@master
```
mlxd, line 3: Will this be changed to `pennylane>=0.16` after the release?

albi3ro (approved):
seems all good.

requirements.txt
antalszava: The idea would be to merge the release candidate into `master` after the release. It seems, that there we're still pinning to `pennylane>=0.15`:
https://github.com/PennyLaneAI/pennylane-lightning/blob/937592796828ad131469a9258e715c7c7994c724/requirements.txt#L3

Having said that, let me undo this line, the change was made just to make sure that tests pass well with the `master` branch of PennyLane.

antalszava:
```suggestion
pennylane>=0.15
```

PR #135, opened 2021-08-18
Fix CI wheel deploy

mlxd:
**Context:** This PR ensures all required and supported architecture wheels are built for deployment. Additionally, all wheel artifacts are stored as ZIP files with OS and arch type names, and contain the given Python versions for the supported build. The zip files require unpacking before deploying.

**Description of the Change:** Updates the `deploy.yml` CI action file to build the full architecture suite, and unpacks the zip files of the given artifacts for deployment of the wheels to PyPI.

**Benefits:** Automated wheel delivery.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,7 @@ on:
     types: [published]
 
 env:
+  CIBW_BUILD: 'cp37-* cp38-* cp39-*'
   CIBW_SKIP: "*-win32 *-manylinux_i686"
 
   # MacOS specific build settings
@@ -37,8 +38,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         cibw_archs: [auto64]
-        py_version: ["cp37-*", "cp38-*", "cp39-*"]
-
         include:
           - os: ubuntu-latest
             cibw_archs: aarch64
@@ -73,11 +72,10 @@ jobs:
           python -m cibuildwheel --output-dir dist
         env:
           CIBW_ARCHS: ${{ matrix.cibw_archs }}
-          CIBW_BUILD: ${{ matrix.py_version }}
 
       - uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.os }}-${{ matrix.cibw_archs }}-${{ matrix.py_version }}-wheel
+          name: ${{ matrix.os }}-${{ matrix.cibw_archs }}-wheel
           path: ./dist/*.whl
 
   build-pure-python-wheel:
@@ -114,7 +112,9 @@ jobs:
           path: dist/
       
       - name: Extract wheels from zip
-
+        run: |
+          cd dist
+          7z x *.zip
 
       - name: Publish
         uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,7 +4,7 @@ on:
     types: [published]
 
 env:
-  CIBW_BUILD: 'cp37-* cp38-* cp39-*'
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -5,6 +5,7 @@ on:
 
 env:
   CIBW_BUILD: 'cp37-* cp38-* cp39-*'
+  CIBW_SKIP: "*-win32 *-manylinux_i686"
```
trbromley, line 8: Why do we skip these? :thinking:

.github/workflows/deploy.yml
```diff
@@ -40,23 +54,29 @@ jobs:
...
       - name: Install cibuildwheel
         run: |
           python -m pip install --upgrade pip
-          python -m pip install cibuildwheel==1.5.5
+          python -m pip install cibuildwheel==2.1.1
 
       - name: Build wheels
         run: |
-          python -m cibuildwheel --output-dir wheelhouse
+          python -m cibuildwheel --output-dir dist
```
trbromley, line 72: The fact we were using `wheelhouse` before as the directory but later on looking for `dist`, is this why things weren't working?

.github/workflows/deploy.yml
```diff
@@ -40,23 +54,29 @@ jobs:
...
+        env:
+          CIBW_ARCHS: ${{ matrix.cibw_archs }}
 
       - uses: actions/upload-artifact@v2
         with:
-          name: ${{ runner.os }}-wheels.zip
-          path: ./wheelhouse/*.whl
+          name: ${{ matrix.os }}-${{ matrix.cibw_archs }}-wheel
```
trbromley, line 78: Do we still need `.zip` at the end?

trbromley: Also out of curiosity, is the name of this zipped file something that needs to follow a strict convention for PyPI? Or does the naming just need to be internally consistent?

.github/workflows/deploy.yml
```diff
@@ -80,17 +100,22 @@ jobs:
 
       - uses: actions/upload-artifact@v2
         with:
-          name: pure-python-wheels.zip
+          name: pure-python-wheels
```
trbromley, line 103: Sorry if I'm getting confused, but also here isn't the `.zip` needed since we unzip the `.zip` files in line 117?

trbromley (approved):
Thanks @mlxd, looks great!

.github/workflows/deploy.yml
mlxd: Do we need 32-bit wheels? If so I can pull this out (they were necessary on my local tests, but may not be required on the CI itself).

.github/workflows/deploy.yml
mlxd: I think one of the main reasons, as the pypa CI step expects a `dist` directory.
The artifacts are also stored as `zip.zip`, which may or not not be supported by the upload to PyPI, so I added a separate steps to explicitly extract them before any attempt made.

.github/workflows/deploy.yml
mlxd: The zip is implied in this instance. The artifacts are currently getting built as `.zip.zip` files. We can remove it from the name and the zip is still applied.

.github/workflows/deploy.yml
mlxd: As above :slightly_smiling_face:

mlxd: Regarding the naming, since we arent uploading the artifacts directly from the zip (the names clobber one another in the previous format, hence the added architecture builds). We will unzip all the wheels into a common directory and upload from there, so we should be fine.

PR #136, opened 2021-08-19
Add C++ adjoint diff methods

mlxd:
**Context:** This PR will add C++ support for adjoint diff, based on arXiV: arXiV:2009.02823. 

- `StateVectorManaged` class added as a subclass of `StateVector`, where memory management is handled internally by the class.
- `AdjointDiff.hpp` file added, with `AdjointJacobian` and helper classes to prepare and calculate the Jacobian for a given set of parameters and observables. For multiple observables, the calculations are handled independently by OpenMP threads and populate the Jacobian. 
- `std::variant` based functionality used to distinguish between parametric gates, non parametric gates, and arbitrary gates defined by a list of complex parameters.
- MacOS wheelbuilder compiler updated to use brew-install clang for x86, as default compiler uses an older version of the Xcode SDK not supporting `std::variant`.
- Repo refactored into subdirectories for multiple separate components:
  `algorithms` for algorithms using and to be used by the simulator;  `bindings` for source files directly related to binding the C++ code with Python; `simulator` for source-files directly related to the simulator; `util` for utility functions to be used across the package;
- Each subdirectory has now been compartmentalized with CMake, allowing separable compilation of each translation unit, enabling C++ system binaries as well as Python bindings to be built.
- Extensive testing for all of the above.
- Bug-fix for the `CRY` gate, which had an incorrect index.

**Benefits:** C++ backed adjoint method is much faster (4x-6x) than Python.

**Possible Drawbacks:** Requires modern C++17 knowledge to maintain. Design may require some later refactor.

**Related GitHub Issues:**


mlxd:
Just a note, CodeFactor seems to be caching old runs and using those --- please ignore its Action failure.

trbromley (commented):
Thanks @mlxd! Thanks for putting in so much awesome work to get this done! Everything looks great to me, my only question is the failing tests in https://github.com/PennyLaneAI/pennylane-lightning/pull/142.

pennylane_lightning/lightning_qubit.py
```diff
@@ -131,6 +145,62 @@ def apply_lightning(self, state, operations):
...
+
+        if len(tape.trainable_params) == 0:
+            return np.array(0)
+
+        for m in tape.measurements:
+            if m.return_type is not Expectation:
+                raise QuantumFunctionError(
+                    "Adjoint differentiation method does not support"
+                    f" measurement {m.return_type.value}"
+                )
```
josh146, line 167: A quick flyby comment to ask - since probabilities are essentially just a list of expectation values of projectors, would it also be fairly trivial to support differentiating probabilities? This is a very common use-case

pennylane_lightning/lightning_qubit.py
```diff
@@ -131,6 +145,62 @@ def apply_lightning(self, state, operations):
...
+
+        for op in tape.operations:
+            if op.num_params > 1 and not isinstance(op, Rot):
+                raise QuantumFunctionError(
+                    f"The {op.name} operation is not supported using "
+                    'the "adjoint" differentiation method'
+                )
```
josh146, line 174: Another flyby comment just to ask why the other parametrized ops (RX, RY, RZ, PhaseShift, etc.) are not supported? Is that coming in a later PR?

pennylane_lightning/lightning_qubit.py
```diff
@@ -131,6 +145,62 @@ def apply_lightning(self, state, operations):
...
+    def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
+        if not CPP_BINARY_AVAILABLE:
+            return super().adjoint_jacobian(tape, starting_state, use_device_state)
+
+        if self.shots is not None:
+            warn(
+                "Requested adjoint differentiation to be computed with finite shots."
+                " The derivative is always exact when using the adjoint differentiation method.",
+                UserWarning,
+            )
```
josh146, line 157: It would be great it we could add support for sampling from the inner-products computed for the adjoint method 💪

tests/test_adjoint_jacobian.py
```diff
@@ -0,0 +1,505 @@
...
+        with qml.tape.JacobianTape() as tape:
+            qml.RX(0.4, wires=[0])
+            qml.Rot(x, y, z, wires=[0])
+            qml.RY(-0.2, wires=[0])
+            qml.expval(qml.PauliZ(0))
+
+        tape.trainable_params = {1, 2, 3}
+
+        grad_D = dev.adjoint_jacobian(tape)
+        grad_F = tape.jacobian(dev, method="numeric")
```
josh146, line 238: Note that, while it is not officially deprecated yet, there is a plan to remove `tape.jacobian` in a later release of PL!

The new canonical way of computing Jacobians will be via the [`qml.gradients`](https://pennylane.readthedocs.io/en/latest/code/qml_gradients.html#transforming-tapes) package:

```pycon
>>> gradient_tapes, fn = qml.gradients.finite_diff(tape)
>>> fn(qml.execute(gradient_tapes, device, None))
```

This API is now available in PL master, and due for the upcoming v0.18 release.

josh146: The reason for this new API:

- It separates the tape-manipulation and classical post-processing logic from evaluation of quantum devices
- The gradient transforms _themselves_ are fully differentiable, allowing for nth order derivatives.

pennylane_lightning/lightning_qubit.py
josh146: Nevermind, I missed the `op.num_params > 1` 🤦

.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -53,6 +53,8 @@ jobs:
         run: python -m cibuildwheel --output-dir wheelhouse
         env:
           CIBW_ARCHS_MACOS: ${{matrix.arch}}
+          CC: /usr/local/opt/llvm/bin/clang
+          CXX: /usr/local/opt/llvm/bin/clang++
```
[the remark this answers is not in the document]
mlxd: Good point, and you are right. Removing CC should be fine (though I'll test it out). There may be a way to avoid use of this too, though I must check this out first.

pennylane_lightning/_serialize.py
```diff
@@ -0,0 +1,146 @@
...
+
+        ob = ObsStructC128(name, params, wires)
+        obs.append(ob)
+
+    return obs
+
+
+def _serialize_ops(
+    tape: QuantumTape, wires_map: dict
+) -> Tuple[List[List[str]], List[np.ndarray], List[List[int]], List[bool], List[np.ndarray]]:
```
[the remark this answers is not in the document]
mlxd: While this should be fine, I wonder will it mess with the conversion to C++ (though I'd imagine pybind is clever enough that we do need to be explicit).

.github/workflows/tests.yml
```diff
@@ -31,7 +31,7 @@ jobs:
             cmake . -BBuild  -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=1
             cmake --build ./Build
             mkdir -p ./Build/tests/results
-            ./Build/tests/runner --order lex --reporter junit --out ./Build/tests/results/report.xml
+            ./Build/pennylane_lightning/src/tests/runner --order lex --reporter junit --out ./Build/tests/results/report.xml
```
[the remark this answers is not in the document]
mlxd: Oops, good point

pennylane_lightning/lightning_qubit.py
```diff
@@ -131,6 +145,62 @@ def apply_lightning(self, state, operations):
...
+            ket = np.ravel(self._pre_rotated_state)
+
+        adj = AdjointJacobianC128()
+
+        obs_serialized = _serialize_obs(tape, self.wire_map)
+        ops_serialized, use_sp = _serialize_ops(tape, self.wire_map)
+
+        ops_serialized = adj.create_ops_list(*ops_serialized)
+
+        tp_shift = tape.trainable_params if not use_sp else {i - 1 for i in tape.trainable_params}
```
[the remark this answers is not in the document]
mlxd: Since the state-prep ops aren't passed to the cpp layer, this accounts for a shift in the parametric gates numbers selected by the trainable params list. Once state prep is handled cpp side, we can pass through the appropriate prep ops, and this can vanish.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -0,0 +1,655 @@
+#pragma once
```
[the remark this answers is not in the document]
mlxd: Do we need to add a copyright notice in every file? Usually adding it once to the repo is enough in a LICENSE file

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -0,0 +1,655 @@
...
+     * @param adj Take the adjoint of the given operations.
+     */
+    inline void applyOperation(StateVectorManaged<T> &state,
+                               const OpsData<T> &operations, size_t op_idx,
+                               bool adj = false) {
+        state.applyOperation(operations.getOpsName()[op_idx],
+                             operations.getOpsWires()[op_idx],
+                             operations.getOpsInverses()[op_idx] ^ adj,
+                             operations.getOpsParams()[op_idx]);
+    }
```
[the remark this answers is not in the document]
mlxd: This was an inlining method that gets inserted where it is used. It encapulates the call and offset indices to `state.applyOperation` with this method call, making it easier to understand and more readable.

I find 
```
applyOperation(state, operations, op_idx)
```
 to be more easily understood than 

```
state.applyOperation(operations.getOpsName()[op_idx],
                             operations.getOpsWires()[op_idx],
                             operations.getOpsInverses()[op_idx] ^ adj,
                             operations.getOpsParams()[op_idx])
```

pennylane_lightning/src/simulator/StateVectorManaged.hpp
```diff
@@ -0,0 +1,102 @@
...
+    out << "num_qubits=" << qubits << std::endl;
+    out << "data=[";
+    out << data[0];
+    for (size_t i = 1; i < length - 1; i++) {
+        out << "," << data[i];
+    }
+    out << "," << data[length - 1] << "]";
+
+    return out;
+}*/
```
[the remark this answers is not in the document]
mlxd: Oops, good point.

pennylane_lightning/lightning_qubit.py
mlxd: That should be a straight-forward extension. Once we have projectors (we can easily support this with some python-layer magic right now), the probabilities should fall out. Though , I think we can add this in a follow-up PR.

mlxd: Added as a future story

pennylane_lightning/lightning_qubit.py
mlxd: Added as a story.

tests/test_adjoint_jacobian.py
mlxd: Thanks Josh. @trbromley Is this something you have time to change? If not, I can do it as a separate PR.

pennylane_lightning/lightning_qubit.py
trbromley: @josh146, what do you mean by this? Is this following the idea of supporting adjoint in a finite-shots setting?

pennylane_lightning/src/algorithms/AdjointDiff.hpp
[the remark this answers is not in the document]
josh146: This was a convention I accidentally started a couple of years ago, as we had a few MIT licensed files in the repo

pennylane_lightning/lightning_qubit.py
josh146: > Once we have projectors (we can easily support this with some python-layer magic right now), the probabilities should fall out. 

Yep, I think it should be relatively straightforward! Or potentially you might not even need to support projectors, maybe a tweak to the adjoint logic in a future story/PR should allow probability gradients to fall out?

pennylane_lightning/lightning_qubit.py
josh146: Yep - the final inner product that is performed as part of the adjoint method, rather than computing that inner product exactly, could we sample from it using the number of shots the user has requested?

This is partly a research question though, since I don't think the answer here is straightforward 🤔 It would give an approximation of 'parameter shift + finite shots', but I don't know if it's a reasonable approximation or not

tests/test_adjoint_jacobian.py
josh146: Oh was just leaving this here more as a comment! This is currently in PL master, so changing it now won't work if the lightning tests are run against PL stable.

trbromley: @josh146, using
```python
gradient_tapes, fn = qml.gradients.finite_diff(tape)
numeric_val = fn(qml.interfaces.batch.execute(gradient_tapes, dev, None))
```
Does this assume that the tape is already suitably expanded into device-supported ops? E.g., doing this naively and I'm getting `Gate MottonenStatePreparation not supported on device lightning.qubit`. I guess the second piece is to ensure devices are responsible for their own expansion into supported gates?

josh146: > I guess the second piece is to ensure devices are responsible for their own expansion into supported gates?

Yep, exactly right - that is my goal for the next iteration :)

pennylane_lightning/lightning_qubit.py
trbromley: Thanks! Yes for me this is also quite an open question. Since the adjoint method itself depends on evolving the _state_ and finding overlaps, I'm not sure how this would be hardware compatible.

josh146: > I'm not sure how this would be hardware compatible.

Oh it definitely wouldn't be, but I was thinking more maybe we could _approximate_ the stochasticity of finite-shots with the adjoint method via some approach?

josh146: Basically: even if we can't sample directly from this (unphysical) expectation value, can we treat the exact inner product as the mean of some probability distribution, with variance dependent on shot value?

pennylane_lightning/src/tests/Test_AdjDiff.cpp
```diff
@@ -0,0 +1,325 @@
...
+        REQUIRE(std::is_constructible<AdjointJacobian<>>::value);
+    }
+    SECTION("AdjointJacobian<TestType> {}") {
+        REQUIRE(std::is_constructible<AdjointJacobian<TestType>>::value);
+    }
+}
+
+TEST_CASE("AdjointJacobian::adjointJacobian", "[AdjointJacobian]") {
+    AdjointJacobian<double> adj;
+    // std::vector<double> param{1, -2, 1.623, -0.051, 0};
```
trevor-vincent, line 40:
```suggestion
```

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -0,0 +1,617 @@
...
+        sv.generateBitPatterns(externalWires);
+    for (const size_t &externalIndex : externalIndices) {
+        std::complex<T> *shiftedState = sv.getData() + externalIndex;
+        shiftedState[internalIndices[0]] = shiftedState[internalIndices[1]] = 0;
+        shiftedState[internalIndices[3]] *= -1;
+    }
+}
+
+template <class T = double, class SVType = Pennylane::StateVector<T>>
+void applyGeneratorControlledPhaseShift(SVType &sv,
```
trevor-vincent, line 105: Are these functions too simple to need doxygen?

trevor-vincent (changes requested):
Awesome work Lee, just two minor comments.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
mlxd: These were internal only to the translation unit, and will not be exposed to the user, so no need for Doxygen comments

pennylane_lightning/lightning_qubit.py
mlxd: I think this is something we can explore. I've created a story for this, so we can revisit it and figure out the best path.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -0,0 +1,655 @@
...
+     * @param scaling_coeff Generator coefficient for given gate derivative.
+     * @param index Position of Jacobian to update.
+     */
+    inline void updateJacobian(const std::complex<T> *sv1,
+                               const std::complex<T> *sv2,
+                               std::vector<std::vector<T>> &jac,
+                               size_t num_elements, T scaling_coeff,
+                               size_t obs_index, size_t param_index) {
+        jac[obs_index][param_index] =
+            -2 * scaling_coeff * std::real(innerProdC(sv1, sv2, num_elements));
```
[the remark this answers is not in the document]
mlxd: That real was a mistake --- replaced real with `imag`, and carrying the imag into the inner product causes a -1 to arise.

pennylane_lightning/lightning_qubit.py
josh146: Nice! I had an idea; if we can work out the adjoint rule for differentiating probabilities, and then write all expval derivatives in terms of this (via `expval = \sum_i \lambda_i probs_i`), finite-shots would make sense and probably 'just work', since we have a probability distribution we can sample from!

Although, there is a partial downside --- I think this might scale (linearly) with the number of qubits. This is because reverse-mode differentiation scales as per the size of the output, unlike forward-mode differentiating (e.g., parameter-shift), which scales with the size of the input.

mlxd:
Thanks for the reviews all. As per our chat on Friday I'm ignoring the codefactor report and merging this, as it still fails to remove the cached version.

.github/CHANGELOG.md
```diff
@@ -9,6 +9,17 @@
...
+* A new state-vector class `StateVectorManaged` was added, enabling memory use to be bound to 
+  statevector lifetime.
+  [(136)](https://github.com/PennyLaneAI/pennylane-lightning/pull/136)
+
+* The repository now has a well-defined component hierarchy, allowing each indepedent unit to be 
+  compiled and linked separately.
+  [(136)](https://github.com/PennyLaneAI/pennylane-lightning/pull/136)
+
+* PennyLane-Lightning now has support for the adjoint Jacobian method of http://arxiv.org/abs/2009.02823.
+  [(136)](https://github.com/PennyLaneAI/pennylane-lightning/pull/136)
```
trbromley, line 21: For me this is a major new feature :rocket: :tada: We can probably move it up during release :grin:

PR #137, opened 2021-08-19
Add adjoint method to Python frontend

trbromley:


pennylane_lightning/lightning_qubit.py
```diff
@@ -131,6 +134,51 @@ def apply_lightning(self, state, operations):
...
+            ket = self._pre_rotated_state
+
+        # TODO: How to accommodate for tensor product observables?
+        adj = AdjointJacobian(ket)
+
+        obs_serialized = _serialize_obs(tape, self.wire_map)
+        ops_serialized = _serialize_ops(tape, self.wire_map)
+
+        jac = adj.adjoint_jacobian(
+            *obs_serialized, *ops_serialized, tape.trainable_params, tape.num_params
```
mlxd, line 169: This is still up for debate, but I think we can pass in the tuples directly to CPP, without incurring the expansion overhead.
```suggestion
            obs_serialized, ops_serialized, tape.trainable_params, tape.num_params
```
and I can unpack them in CPP. I have already made some helper struct to simplify data passing around here, so this might just be tidier. I'll let you know once I have it bound up.

PR #144, opened 2021-09-13
Ensure C++17 compiler builds wheel for MacOS

mlxd:
**Context:** This fixes issues with the MacOS wheelbuilder, requiring C++17 support that is unavailable within the default CI setup. We opt for an externally installed (brew) clang compiler, allowing both X86_64 and ARM64 wheels to be compiled.

**Description of the Change:** Uses brew clang as default compiler for MacOS.

**Benefits:** Enables Cpp17 features for MacOS under multiple architectures.

**Possible Drawbacks:** May have possible issues with older MacOS versions.

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,7 +11,7 @@ env:
 
   CIBW_BEFORE_ALL_MACOS: |
     brew cask uninstall --force oclint
-    brew install gcc libomp
+    brew install clang libomp
 
   # Python build settings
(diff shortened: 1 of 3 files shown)


mlxd:
As a note, both `CC` and `CXX` are supplied, since many places in the compilation process the C-compiler is favored over the C++ compiler.

mlxd:
I can confirm the above built wheels do not require an installation of brew Clang on a Mac to run successfully.

trbromley (approved):

PR #147, opened 2021-09-16
Fix C++ compiler warnings

maliasadi:
**Context:**
Fix C++ compiler warnings.  

Fix warnings related to the unused variables. Fix inconsistencies within `std::conj` and `conj`, and `std::swap` and `swap`. Fix a couple of minor issues with functions' descriptions. Update Util calls relying on `USE_CBLAS`. 

**Benefits:**
A cleaner `make install` for the PL-lightning

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -631,7 +631,7 @@ template <class T = double> class AdjointJacobian {
                                 !operations.getOpsInverses()[op_idx]) *
                             (2 * (0b1 ^ operations.getOpsInverses()[op_idx]) -
                              1);
-                        size_t index;
```
mlxd, line 634: :+1: well spotted

pennylane_lightning/src/util/Util.hpp
```diff
@@ -46,8 +44,8 @@ namespace Util {
 /**
  * @brief Compile-time scalar real times complex number.
  *
- * @tparam T Precision of complex value and result.
- * @tparam U Precision of real value.
+ * @tparam U Precision of real value `a`.
+ * @tparam T Precision of complex value `b` and result.
```
mlxd, line 48: Oops, my bad!

pennylane_lightning/src/util/Util.hpp
```diff
@@ -214,17 +212,17 @@ std::complex<T> innerProd(const std::complex<T> *data_1,
...
-            result = cblas_cdotu_sub(data_size, data_1, 1, data_2, 1, &result);
-        else if constexpr (std::is_same_v<T, double>)
-            result = cblas_zdotu_sub(data_size, data_1, 1, data_2, 1, &result);
-    } else {
-        result = std::inner_product(
-            data_1, data_1 + data_size, data_2, std::complex<T>(), ConstSum<T>,
-            static_cast<std::complex<T> (*)(std::complex<T>, std::complex<T>)>(
-                &ConstMult<T>));
-    }
+#ifdef USE_CBLAS
```
mlxd, line 215: Is there a good reason to favour the preprocessor rather than the `if constexpr` support for this?

setup.py
```diff
@@ -75,12 +75,12 @@ class BuildExt(build_ext):
 
     c_opts = {
         "msvc": ["-EHsc", "-O2", "-W1", "-std:c++17", "-D_USE_MATH_DEFINES"],
-        "unix": ["-O3", "-W", "-fPIC", "-shared", "-fopenmp"],
+        "unix": ["-O3", "-W", "-Wno-reorder", "-fPIC", "-shared", "-fopenmp"],
```
mlxd, line 78: It may be better to reorder the constructor initializer lists rather than ignore this. 

Also, we should opt to enable all warnings as errors both here and in the CMake (see the [Jet repo](https://github.com/XanaduAI/jet/blob/cb4ab418cfae45a74c226ab354abb6a86e77825d/CMakeLists.txt#L205) for how we did it there).

Lastly, we need these in the Windows and MacOS flags too to keep it consistent (eventually I expect to move away from the build with `setup.py` to instead have it call CMake, but that will need to wait for now). The CI checks here should be enough to know if it works on there.

mlxd (changes requested):
Hey @maliasadi nice work. I have a few comments to address:

pennylane_lightning/src/util/Util.hpp
maliasadi: The generated assemblies from using `if constexpr` and the preprocessor are identical! I updated this part mainly to emphasise the use of USE_CBLAS within the code, and I thought that making use of preprocessors are more common for such situations. I can simply revert it back to what it was if you think differently?

setup.py
maliasadi: well caught! my bad here. This was a temporary fix of variable reordering warnings.

maliasadi (commented):
The only warning message that is requiring an update is  
`pennylane_lightning/src/bindings/Bindings.cpp:780:46: warning: unused parameter ‘adj’ [-Wunused-parameter]
  780 |              [](AdjointJacobian<PrecisionT> &adj
`
Any suggestions?

pennylane_lightning/src/util/Util.hpp
mlxd: Sounds good. I'd prefer to avoid use of the preprocessor unless we need to use it (or unless it makes things much simpler).  In this case given that everything happens at build time anyway, I'd rather we keep to C++ constructs where we can.

setup.py
mlxd: No worries :slightly_smiling_face:

mlxd:
> The only warning message that is requiring an update is
> `pennylane_lightning/src/bindings/Bindings.cpp:780:46: warning: unused parameter ‘adj’ [-Wunused-parameter] 780 | [](AdjointJacobian<PrecisionT> &adj `
> Any suggestions?

That is quite strange --- the `adj` argument here is (at least I thought) necessary for the Python bindings, as the first argument is generally the object created. I'll have a look and see if there is something to be done.

mlxd:
> > The only warning message that is requiring an update is
> > `pennylane_lightning/src/bindings/Bindings.cpp:780:46: warning: unused parameter ‘adj’ [-Wunused-parameter] 780 | [](AdjointJacobian<PrecisionT> &adj `
> > Any suggestions?
> 
> That is quite strange --- the `adj` argument here is (at least I thought) necessary for the Python bindings, as the first argument is generally the object created. I'll have a look and see if there is something to be done.

So I can confirm that this line is needed by Python to ensure the signature is correct (each bound member function will always pass in the object as `self` whether we use it or not). We can move the above function to be static, but then we have a function with and without the first argument being the class. For consistency, I think we can just patch this with the following workaround: adding `static_cast<void>(adj);` to the function body should suppress that warning, will be optimised away at compile time, and ensures both signatures are the same.

setup.py
```diff
@@ -83,6 +83,10 @@ class BuildExt(build_ext):
         "unix": ["-O3", "-W", "-fPIC", "-shared", "-fopenmp"],
     }
 
+    if os.environ.get("ENABLE_WARNINGS", False):
+        c_opts["unix"].extend(["-Wall", "-Wextra", "-Werror"])
+        l_opts["unix"].extend(["-Wall", "-Wextra", "-Werror"])
+
```
mlxd, line 89:
```suggestion
        c_opts["msvc"].extend(["/Wall", "/WX"])
        l_opts["msvc"].extend(["/WX"])
```
See [here](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-160) for MSVC flags. I assume the `/` is correctly parsed and does not need to be escaped, but if failures occur it may be worth a try.

mlxd (changes requested):

mlxd (approved):

mlxd:
Nice work on this @maliasadi . I'll merge it in now.

PR #153, opened 2021-10-08
Enable clang-tidy in CI checks

mlxd:
**Context:** To ensure modern C++ standards, we require testing of newly added code to ensure it is standards-conforming with all newly supported features. This PR enables clang-tidy with warnings-as-errors support for a variety of tidy options as defined from [here](https://clang.llvm.org/extra/clang-tidy/)

**Description of the Change:** Add CI-enabled checking of code-base builds with clang-tidy for the following options:
```
clang-tidy \
-extra-arg=-std=c++17 \
-warnings-as-errors=* \
-header-filter=.* \
-checks=-*,-llvmlibc-*,clang-analyzer-*,modernize-*,clang-analyzer-cplusplus*,openmp-*,performance-*,portability-*,readability-*
```

**Benefits:** This will ensure our codebase uses non-legacy and/or deprecated language features, as well as ensuring strict conformance to the modern C++ standard (currently C++17). In addition, this can prevent compiler-specific bugs being uncaught, which can be difficult to find between compiler versions.

**Possible Drawbacks:** May add additional development time to new features as developer become comfortable with the requirements of the standard.

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -44,13 +44,14 @@ jobs:
 
       - name: Install dependencies
         run: sudo apt update && sudo apt -y install clang-tidy-11 python3 cmake g++
+        env: DEBIAN_FRONTEND=noninteractive
 
       - name: Checkout code
         uses: actions/checkout@v2
 
       - name: Run clang-tidy compilation
         run: |
-          cmake -BBuild -DENABLE_CLANG_TIDY=on . 
+          cmake -BBuild -DENABLE_CLANG_TIDY=on -DCLANT_TIDY_BINARY=clang-tidy-11 .
           cmake --build ./Build
 
   format-cpp:
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -32,6 +32,27 @@ jobs:
       - name: Run Black
         run: black -l 100 pennylane_lightning/ --check
 
+  tidy-cpp:
+    name: Tidy (C++)
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Cancel previous runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Install dependencies
+        run: sudo apt update && sudo apt -y install clang-tidy-11 python3 cmake g++
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run clang-tidy compilation
+        run: |
+          cmake -BBuild -DENABLE_CLANG_TIDY=on . 
+          cmake --build ./Build
+
   format-cpp:
     name: Format (C++)
     runs-on: ubuntu-20.04
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,15 +1,9 @@
 name: Formatting check
 on:
   pull_request:
-    paths:
-      - "pennylane_lightning/**"
-      - "tests/**"
   push:
     branches:
       - master
-    paths:
-      - "pennylane_lightning/**"
-      - "tests/**"
 
 jobs:
   black:
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,12 +1,15 @@
 name: Formatting check
 on:
   pull_request:
-    push:
-      branches:
-        - master
-      paths:
-        - "pennylane_lightning/**"
-        - "tests/**"
+    paths:
+      - "pennylane_lightning/**"
+      - "tests/**"
+  push:
+    branches:
+      - master
+    paths:
+      - "pennylane_lightning/**"
+      - "tests/**"
 
 jobs:
   black:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,10 @@ option(ENABLE_WARNINGS "Enable warnings" ON)
 option(ENABLE_CLANG_TIDY "Enable clang-tidy build checks" OFF)
 
 if(ENABLE_CLANG_TIDY)
-    set(CMAKE_CXX_CLANG_TIDY clang-tidy;
+    if (NOT DEFINED CLANG_TIDY_BINARY)
+        set(CLANG_TIDY_BINARY clang-tidy)
+    endif()
+    set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_BINARY};
                             -extra-arg=-std=c++17;
                             -warnings-as-errors=*;
                             -header-filter=.*;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,16 @@ endif()
 option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)
 option(BUILD_TESTS "Build cpp tests" OFF)
 option(ENABLE_WARNINGS "Enable warnings" ON)
+option(ENABLE_CLANG_TIDY "Enable clang-tidy build checks" OFF)
+
+if(ENABLE_CLANG_TIDY)
+    set(CMAKE_CXX_CLANG_TIDY clang-tidy;
+                            -extra-arg=-std=c++17;
+                            -warnings-as-errors=*;
+                            -header-filter=.*;
+                            -checks=-*,-llvmlibc-*,modernize-*,clang-analyzer-cplusplus*,openmp-*,performance-*,portability-*,readability-*;
+    )
+endif()
 
 # Add pybind11
 include(FetchContent)


pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -291,44 +305,51 @@ template <class T> class OpsData {
      *
      * @return size_t Number of operations.
      */
-    size_t getSize() const { return ops_name_.size(); }
+    [[nodiscard]] auto getSize() const -> size_t { return ops_name_.size(); }
```
maliasadi, line 308: Not sure how `[[nodiscard]]` would be helpful except making some dev challenges to ensure not-discarding the return value. I think, the `[[nodiscard]]` attribute is more practical where discarding the result leads to a computational error.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -339,7 +360,7 @@ template <class T> class OpsData {
      * @return true Gate is parametric (has parameters).
      * @return false Gate in non-parametric.
      */
-    inline bool hasParams(size_t index) const {
+    [[nodiscard]] inline auto hasParams(size_t index) const -> bool {
```
maliasadi, line 363: The same concern here.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -490,6 +513,100 @@ template <class T = double> class AdjointJacobian {
...
+        // Globally scoped exception value to be captured within OpenMP block.
+        // See the following for OpenMP design decisions:
+        // https://www.openmp.org/wp-content/uploads/openmp-examples-4.5.0.pdf
+        std::exception_ptr ex = nullptr;
+        size_t num_states = states.size();
+#if defined _OPENMP
+#pragma omp parallel default(none)                                             \
+    shared(states, operations, op_idx, ex, num_states)
+        {
+#pragma omp for
```
maliasadi, line 584: clang-tidy `openmp-use-default-none` check would fix this, I think.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -525,15 +642,16 @@ template <class T = double> class AdjointJacobian {
      * @param adj Indicate whether to take the adjoint of the operation.
      * @return T Generator scaling coefficient.
      */
-    inline T applyGenerator(StateVectorManaged<T> &sv,
-                            const std::string &op_name,
-                            const std::vector<size_t> &wires, const bool adj) {
+    inline auto applyGenerator(StateVectorManaged<T> &sv,
```
maliasadi, line 645: Wouldn't pass-by-value and using `std::move` be a better practice here?

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -546,12 +664,13 @@ template <class T = double> class AdjointJacobian {
      * @param ops_matrices Matrix definition of an operation if unsupported.
      * @return const OpsData<T>
      */
-    const OpsData<T> createOpsData(
+    auto createOpsData(
         const std::vector<std::string> &ops_name,
```
maliasadi, line 668: Wouldn't pass-by-value and using `std::move` be a better practice here?

pennylane_lightning/src/.clang-tidy
```diff
@@ -0,0 +1,233 @@
+---
+Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,-llvmlibc-*,modernize-*,-modernize-use-trailing-return-type,clang-analyzer-cplusplus*,openmp-*,performance-*,portability-*,readability-*'
+WarningsAsErrors: '*'
+HeaderFilterRegex: '.*'
+AnalyzeTemporaryDtors: false
+FormatStyle:     none
+InheritParentConfig: true
+User:            mlxd
+CheckOptions:
+  - key:             modernize-replace-auto-ptr.IncludeStyle
```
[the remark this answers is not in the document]
mlxd: This was an autogenerated option from the clang-tidy export. I agree, probably not needed, but no harm to keep in.

pennylane_lightning/src/.clang-tidy
```diff
@@ -0,0 +1,233 @@
...
+    value:           '10'
+  - key:             llvm-else-after-return.WarnOnConditionVariables
+    value:           '0'
+  - key:             llvm-qualified-auto.AddConstToQualified
+    value:           '0'
+  - key:             llvm-else-after-return.WarnOnUnfixable
+    value:           '0'
+  - key:             openmp-exception-escape.IgnoredExceptions
+    value:           ''
+...
```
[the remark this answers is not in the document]
mlxd: These are already caught by the top-level checks enabled by:
`Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,-llvmlibc-*,modernize-*,-modernize-use-trailing-return-type,clang-analyzer-cplusplus*,openmp-*,performance-*,portability-*,readability-*'`

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -658,17 +661,19 @@ void lightning_class_bindings(py::module &m) {
                             auto buffer = param.request();
                             auto ptr = static_cast<std::complex<Param_t> *>(
                                 buffer.ptr);
-                            if (buffer.size > 0)
+                            if (buffer.size > 0) {
```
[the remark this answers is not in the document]
mlxd: That is now updated to check for ! empty instead.

pennylane_lightning/src/simulator/Gates.hpp
```diff
@@ -182,7 +193,7 @@ template <class T> static constexpr std::vector<std::complex<T>> getToffoli() {
  * data.
  */
 template <class T, class U = T>
-static const std::vector<std::complex<T>> getPhaseShift(U angle) {
+static auto getPhaseShift(U angle) -> std::vector<std::complex<T>> {
```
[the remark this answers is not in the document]
mlxd: It was recommended as not essential. Returning a const newly created object is not recommended apparently (and I can understand why).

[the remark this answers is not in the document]
mlxd: It leaves the const-ness up to the user of that function, rather than the function itself.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
mlxd: As above.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
mlxd: Yea, Trevor is correct. We are not making any copies inside the method here, so copy-and-move would be extra work than reference only. The copy-and-move semantic assumes only works in those instances, and only when moving is cheap also.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
mlxd: Potentially, though `OpsData` has 2 constructors (one for ref, one for copy and move). In this instance it may be better, though clang did not recognize the need for modification here.

It could be how the method is constructed. Though, this method is only used in C++ tests, as the bound method for Python is exposed differently.

pennylane_lightning/src/.clang-tidy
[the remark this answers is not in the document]
maliasadi: Nice! I'm learning clang-tidy while reviewing. thanks!

pennylane_lightning/src/simulator/Gates.hpp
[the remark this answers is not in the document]
maliasadi: Nice recommendation!

.github/workflows/tests.yml
```diff
@@ -21,6 +21,11 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
```
trbromley, line 27: Does the version matter? It's 3.8 in the `format.yml` workflow.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -178,36 +182,42 @@ template <class T = double> class ObsDatum {
      * @param obs_wires Wires upon which to apply operation. Each observable
      * operation will be a separate nested list.
      */
-    ObsDatum(std::vector<std::string> &&obs_name,
+    /*ObsDatum(std::vector<std::string> &&obs_name,
              std::vector<param_var_t> &&obs_params,
              std::vector<std::vector<size_t>> &&obs_wires)
         : obs_name_{std::move(obs_name)}, obs_params_{std::move(obs_params)},
-          obs_wires_{std::move(obs_wires)} {};
+          obs_wires_{std::move(obs_wires)} {};*/
```
trbromley, line 189: Should we delete if no longer needed?

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -490,6 +513,100 @@ template <class T = double> class AdjointJacobian {
...
+    inline void applyObservables(std::vector<StateVectorManaged<T>> &states,
+                                 const StateVectorManaged<T> &reference_state,
+                                 const std::vector<ObsDatum<T>> &observables) {
+        // Globally scoped exception value to be captured within OpenMP block.
+        // See the following for OpenMP design decisions:
+        // https://www.openmp.org/wp-content/uploads/openmp-examples-4.5.0.pdf
+        std::exception_ptr ex = nullptr;
+        size_t num_observables = observables.size();
+#if defined _OPENMP
+#pragma omp parallel default(none)                                             \
```
trbromley, line 533: Is it normal for these compiler directives to be non-indented?

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -490,6 +513,100 @@ template <class T = double> class AdjointJacobian {
...
+    shared(states, reference_state, observables, ex, num_observables)
+        {
```
trbromley, line 535: Does this need an indent?
(I know indentation doesn't matter in C++, but more so for readability.)

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -490,6 +513,100 @@ template <class T = double> class AdjointJacobian {
...
+            if (ex) {
+#pragma omp cancel parallel
+            }
+        }
+#endif
+        if (ex) {
+            std::rethrow_exception(ex);
+        }
+    }
+
```
trbromley, line 609: Are the above additions purely resulting from `clang-tidy` recommendations? :thinking:

.github/workflows/format.yml
```diff
@@ -32,6 +29,34 @@ jobs:
...
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: sudo apt update && sudo apt -y install clang-tidy-11 python3 cmake g++
```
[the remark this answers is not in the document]
mlxd: That's a good point. I'd imagine not in this case.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
mlxd: As above, I disabled the formatter here and added appropriate indentation.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
trbromley: Nice!

mlxd:
Thanks for the reviews and suggestions all!

Issue #133, opened 2021-08-16
Provide LRU caching support for index operations


mlxd:
#### Issue description

Often calls to the gate methods of the `StateVector` class can be repeated, and hence have some amount of repeated work. By enabling the use of [LRU](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) caching for the outputs from the `StateVector::generateBitPatterns` and `StateVector::getIndicesAfterExclusion` methods, we can avoid some amount of repeated work, and reuse the previously called method results, at the expense of some additional RAM.

Tasks:
- Implement a C++ LRU cache class in [pennylane_lightning/src/util](https://github.com/PennyLaneAI/pennylane-lightning/tree/master/pennylane_lightning/src/util).
- Adapt calls to `StateVector::generateBitPatterns` and `StateVector::getIndicesAfterExclusion` to enable use of caching (see [StateVector.hpp](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/src/simulator/StateVector.hpp)).
- Add suitable [tests](https://github.com/PennyLaneAI/pennylane-lightning/tree/master/pennylane_lightning/src/tests) to verify correct operation of the LRU cache.

mlxd:
Hi @AmintorDusko, great! Feel free to open a WIP PR. Also, if there are any questions or clarifications needed feel free to ask here or in the PR.

PR #154, opened 2021-10-12
LRU caching support for index operations

AmintorDusko:
**Context:**
Implement a C++ LRU cache class.
Implement a C++ LRU cache class and adapt calls to `StateVector::generateBitPatterns` and `StateVector::getIndicesAfterExclusion` to enable use of caching. Also, add suitable tests to verify the correct operation of the LRU cache.
**Benefits:**
Support to caching may avoid some amount of repeated work, improving efficiency.
**Possible Drawbacks:**

**Related GitHub Issues:**
#133 created by @mlxd


AmintorDusko:
Hi @mlxd, thanks for your input.
I already push the `LRU_cache class` implementation. 
My next steps will be:
- Integrate the LRU cache into the `class StateVector`.
- Implement tests for the `LRU_cache class`.

The `LRU_cache class` constructor has the cache size as an input parameter (if this parameter is set as 0, the cache is considered to be large). The default cache size is 1. 
How do you think we should set the cache size? Should we keep it a fixed pre-determined value or take its value from the `class StateVector` constructor?


By now, the `LRU_cache class` is not thread-safe. Depending on how the parallelization is implemented in PennyLane this would never be a problem. If more than one thread could be accessing the same cache at the same time, we would need to implement some lock guards (mutex). Please let me know your thoughts.

pennylane_lightning/src/util/Cache.hpp
```diff
@@ -0,0 +1,178 @@
...
+typedef std::pair<std::vector<size_t>, size_t> key_type;
+typedef std::vector<size_t> stored_type;
+typedef std::unordered_map<key_type, stored_type>::const_iterator
+    const_iterator_type;
+
+namespace Pennylane {
+
+// Boost implementation of a hash combine:
+// https://www.boost.org/doc/libs/1_35_0/doc/html/boost/hash_combine_id241013.html
+struct pair_hash {
```
mlxd, line 41: Can this struct be made visible inside this translation unit only?

pennylane_lightning/src/util/Cache.hpp
```diff
@@ -0,0 +1,178 @@
...
+#include <limits>
+#include <utility>
+
+typedef std::pair<std::vector<size_t>, size_t> key_type;
+typedef std::vector<size_t> stored_type;
+typedef std::unordered_map<key_type, stored_type>::const_iterator
+    const_iterator_type;
+
+namespace Pennylane {
+
```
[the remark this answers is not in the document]
AmintorDusko: OK. will do! Thanks!

mlxd:
> Hi @mlxd, thanks a lot for your nice comments and inputs. I changed the `LRU_cache class`, to accommodate your suggestions.
> 
> First I intended to introduce the cache containers for `StateVector::generateBitPatterns` and `StateVector::getIndicesAfterExclusion` as class members constructed when constructing the StateVector object. Since these functions are static members I needed to do some refactoring to the `LRU_cache class`, so it could operate nicely as a static object. This static cache container has a default size of 10, that can be changed with a function member of the class.
> 
> It seems like the integration is working nicely.
> 
> My next step will be: Implement tests for the `LRU_cache class`.
> 
> Please let me know if you have any thoughts or suggestions.

Thanks @AmintorDusko . Do you think it better to modify the static member methods, or to modify the equivalent non-static methods instead? I'd imagine the non-static methods would be easier, and avoid cache creation for members which may be called outside the statevector.

Also, creating the cache as a static objects in the namespace will also likely invalidate the previous mention of concurrency, as a single cache class will persist across all StateVector objects. In this case, the caches need to be class member variables to avoid this issue.

AmintorDusko:
> > Hi @mlxd, thanks a lot for your nice comments and inputs. I changed the `LRU_cache class`, to accommodate your suggestions.
> > First I intended to introduce the cache containers for `StateVector::generateBitPatterns` and `StateVector::getIndicesAfterExclusion` as class members constructed when constructing the StateVector object. Since these functions are static members I needed to do some refactoring to the `LRU_cache class`, so it could operate nicely as a static object. This static cache container has a default size of 10, that can be changed with a function member of the class.
> > It seems like the integration is working nicely.
> > My next step will be: Implement tests for the `LRU_cache class`.
> > Please let me know if you have any thoughts or suggestions.
> 
> Thanks @AmintorDusko . Do you think it better to modify the static member methods, or to modify the equivalent non-static methods instead? I'd imagine the non-static methods would be easier, and avoid cache creation for members which may be called outside the statevector.
> 
> Also, creating the cache as a static objects in the namespace will also likely invalidate the previous mention of concurrency, as a single cache class will persist across all StateVector objects. In this case, the caches need to be class member variables to avoid this issue.

Thanks, @mlxd, you are right. I change the code files to make use of the non-static methods.

AmintorDusko:
Hi @mlxd, with this last push I believe I covered most of what we discuss. 
Could you please give a final review?
Please let me know your thoughts.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
+#include <catch2/catch.hpp>
+
+#include "Cache.hpp"
+
+using namespace Pennylane::Util;
+
+namespace {
```
mlxd, line 7: The test-case here does not need to be within an anonymous namespace block.

pennylane_lightning/src/util/Cache.hpp
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * @brief Least Recently Updated (LRU) cache policy class.
+ *
+ * This class define a container to store previously calculated indices.
+ * Data is stored up to a maximum cache size, defined by the class constructor.
+ * The key for accessing the cache storage is a vector and a scalar
+ * representing respectively, the wires and number of qubits involved in the
+ * calculation.
+ *
```
mlxd, line 63: To help with the documentation generation, the comment here should have the template parameters as `@tparam` entries (e.g see [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/238f4d05ae6b2aa3d06fd3e77113dc847650d2a6/pennylane_lightning/src/simulator/StateVector.hpp#L66) for how this is done with the StateVector class).

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    size_t number_of_keys = keys.size();
+
+    std::vector<std::vector<size_t>> indices = {
+        {1, 1, 1, 1},     {3, 3, 3, 3},     {5, 5, 5, 5},     {7, 7, 7, 7},
+        {9, 9, 9, 9},     {11, 11, 11, 11}, {13, 13, 13, 13}, {15, 15, 15, 15},
+        {17, 17, 17, 17}, {19, 19, 19, 19}, {21, 21, 21, 21}, {23, 23, 23, 23},
+        {25, 25, 25, 25}, {27, 27, 27, 27}};
+
+    size_t nn = 0;
```
mlxd, line 31: What does `nn` refer to here? Is there a more descriptive name for this variable we could use? Maybe it is obvious, but I am not sure yet.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    SECTION("Getting a single value.") {
+        auto element_it = cache_container.check_cache(keys.back());
+        REQUIRE(cache_container.get_key(element_it) == keys.back());
+        REQUIRE(cache_container.get(element_it) == indices.back());
+    }
+
+    SECTION("Checking if elements in the cache are the ones most recently "
+            "introduced.") {
+        for (int nn = 0; nn < (number_of_keys - cache_container.size()); nn++) {
```
mlxd, line 57: If using sized values that never go negative, we should opt for `size_t` instead of `int` here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+            auto element_it = cache_container.check_cache(keys[nn]);
+            REQUIRE(element_it == cache_container.end());
+        }
+
+        for (int nn = (number_of_keys - cache_container.size());
```
mlxd, line 62: As above for `int` -> `size_t`

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    cache_container.set_cache_size(5);
+    SECTION("Checking new container capacity. ") {
+        REQUIRE(cache_container.size() == 5);
+        REQUIRE(cache_container.capacity() == 5);
+    }
+
+    SECTION("After re-scaling, check if elements in the cache are the ones "
+            "most recently introduced.") {
+        for (int nn = 0; nn < (number_of_keys - cache_container.size()); nn++) {
```
mlxd, line 77: Same here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+            auto element_it = cache_container.check_cache(keys[nn]);
+            REQUIRE(element_it == cache_container.end());
+        }
+
+        for (int nn = (number_of_keys - cache_container.size());
```
mlxd, line 82: Same here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    cache_container.set_cache_size(7);
+    SECTION("Checking new container capacity. ") {
+        REQUIRE(cache_container.size() == 5);
+        REQUIRE(cache_container.capacity() == 7);
+    }
+
+    for (int nn = 0; nn < number_of_keys; nn++) {
```
mlxd, line 95: Same here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    cache_container.set_cache_size(7);
+    SECTION("Checking new container capacity. ") {
+        REQUIRE(cache_container.size() == 7);
+        REQUIRE(cache_container.capacity() == 7);
+    }
+
+    SECTION("Checking if we only kept the most recently updated elements.") {
+        {
+            int nn = 0;
```
mlxd, line 116: Same here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+            auto element_it = cache_container.check_cache(keys[nn]);
+            REQUIRE(cache_container.get_key(element_it) == keys[nn]);
+        }
+
+        for (int nn = 1; nn < (number_of_keys - cache_container.size() + 1);
```
mlxd, line 121: Same here.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,135 @@
...
+             nn++) {
+            auto element_it = cache_container.check_cache(keys[nn]);
+            REQUIRE(element_it == cache_container.end());
+        }
+
+        for (int nn = (number_of_keys - cache_container.size() + 1);
```
mlxd, line 127: Same here.

pennylane_lightning/src/util/Cache.hpp
```diff
@@ -0,0 +1,196 @@
...
+            combined_hash_value ^= std::hash<size_t>()(term) + 0x9e3779b9 +
+                                   (combined_hash_value << 6) +
+                                   (combined_hash_value >> 2);
+        };
+        combined_hash_value ^= std::hash<size_t>()(key.second) + 0x9e3779b9 +
+                               (combined_hash_value << 6) +
+                               (combined_hash_value >> 2);
+        return combined_hash_value;
+    }
+};
```
mlxd, line 53: I wonder since the hash will be dependent upon the datatype being cached, if this should be housed elsewhere, such as within the source file using the cache. Ideally, we want to be as generalisable as possible with this, and having a hash function for a single key type alongside the cache utility may be somewhat restrictive.
Having the user of the cache define the necessary hash where it is used may be a better idea. Otherwise, if this file can be appropriately templated to work with any (reasonable) type, this could also work.

What do you think?

mlxd (commented):
Great job @AmintorDusko . I have a few more comments and queries on the current implementation.

mlxd:
Just to note, the recent changes to master may require some updates to integrate, but I think we can concentrate on having all tests pass here before worrying about integration.

pennylane_lightning/src/tests/Test_Cache.cpp
AmintorDusko: I will remove it. Thanks.

pennylane_lightning/src/util/Cache.hpp
AmintorDusko: I think you are right. I will move it to the StateVector source code and adapt the class accordingly.

AmintorDusko:
Hi @mlxd, Thank you very much for your comments and suggestions. I believe that they really improved the quality of this work. Now that we can optionally provide the `struct hash` as a template parameter, the `LRU_cache class` is much more general. To test this, I expanded the test files to include the use of its default `struct std::hash`. We can also store the specific `struct hash` used by the StateVector class in an unnamed subspace in its header, covering one of our previous discussions.

Once again, please let me know your thoughts.

mlxd (commented):
Great, thanks for this I'll have a deeper look over the coming days, but I think this is exactly the solution we were looking for with this. As an FYI, here is the runtime changes with your solution using a single thread and representative circuit:
image

For small qubit numbers, as expected the overhead is a little higher, but the changes are on the order of 1-10 milliseconds in runtime differences a 20% loss in this region doesn't concern me much. Beyond the 11 qubit region the cache really helps, as prepping the indexing arrays becomes more of a challenge, and as you can see we get up to 20% runtime improvement at 18 qubits. 

I think we can figure out the integration changes early next week, as we may scope out some enable rules for beyond 10 qubits or similar. I'll also have another colleague have a look over this as an additional check. Once again, good job on this.

pennylane_lightning/src/simulator/StateVector.hpp
```diff
@@ -108,7 +138,7 @@ template <class fp_t = double> class StateVector {
      * @param length Number of elements in complex data array. Must be
      * power-of-2 (qubits only).
      */
-    StateVector(CFP_t *arr, size_t length)
+    StateVector(CFP_t *arr, size_t length, size_t cache_size = 10)
```
trevor-vincent, line 141: cache_size needs a doxygen comment here

trevor-vincent (changes requested):
Awesome work Amintor. The pr is really clean and I can't see any issues. I made one small comment about a missing doxygen @param for cache_size. Once that comment is added, it's ready to merge.

trevor-vincent (approved):
Awesome, looks ready to merge.

pennylane_lightning/src/tests/Test_Cache.cpp
```diff
@@ -0,0 +1,289 @@
...
+        combined_hash_value ^= std::hash<size_t>()(key.second) + offset +
+                               (combined_hash_value << 6) +
+                               (combined_hash_value >> 2);
+        return combined_hash_value;
+    }
+};
+
+// NOLINTNEXTLINE
+TEST_CASE("LRU_Cache for doubles, with size_t keys", "[Cache]") {
+    LRU_cache<size_t, double> cache_container{10};
```
mlxd, line 29: Can we add a test also for a 0-sized container, to test that the cache does not operate? In this case, it should default to the current behaviour with no caching.

pennylane_lightning/src/simulator/StateVector.hpp
```diff
@@ -106,8 +138,10 @@ template <class fp_t = double> class StateVector {
      * @param arr Pointer to the complex data array.
      * @param length Number of elements in complex data array. Must be
      * power-of-2 (qubits only).
+     * @param cache_size Cache size to store most calculated indices throughout
+     * operations.
      */
-    StateVector(CFP_t *arr, size_t length)
+    StateVector(CFP_t *arr, size_t length, size_t cache_size = 10)
```
mlxd, line 144:
```suggestion
    StateVector(CFP_t *arr, size_t length, size_t cache_size = 0)
```

We should default to the cache being disabled. This also means we require to be no cache, rather than infinite.

pennylane_lightning/src/tests/Test_Cache.cpp
AmintorDusko: I commit your suggestion to default the cache size to 0. I implemented the cache_size=0 to be the no-cache scenario when you suggested it sometime during the beginning of this PR.  I will implement tests for the 0-cache case as soon as I can.

PR #155, opened 2021-10-13
Add Multi-threaded DOTC, GEMV, GEMM along with BLAS Support 

maliasadi:
**Context:**
Add Cache-Friendly and Multi-threaded DOTC, GEMV, GEMM along with BLAS Support to PL-lightning

Update `innerProd` and `innerProdC`.
Add `Transpose`, `matrixVecProd`, and `matrixMatProd`. 

**Benefits:**
Utilizing BLAS operations: inner, matrix-vector, and matrix-matrix products are orders of magnitude faster than plain implementation of such arithmetic. This PR adds BLAS support as well as a cache-friendly and multi-threaded implementation of these operations in PL-lightning. 

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+    if (r1 >= s1 && r1 > BLOCK_THRESHOLD) {
+        r2 = (m1 + m2) / 2;
+        CFTranspose(mat, mat_t, m, n, m1, r2, n1, n2);
+        m1 = r2;
+        goto tail;
+    } else if (s1 > BLOCK_THRESHOLD) {
+        s2 = (n1 + n2) / 2;
+        CFTranspose(mat, mat_t, m, n, m1, m2, n1, s2);
+        n1 = s2;
+        goto tail;
```
mlxd, line 548: Same here

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+                              std::complex<T> *m_out, size_t m, size_t n,
+                              size_t k, bool transpose = false) {
+    if (!m_out) {
+        return;
+    }
+#if defined _OPENMP
+#pragma omp parallel default(none)
+#endif
+    {
+        std::size_t r, c, b;
```
mlxd, line 606: Is there maybe a better naming scheme for these (and similarly used) variables? 
I think naming a variable explicitly with its intention allows the code to be better understood by others.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+                for (c = 0; c < n; ++c) {
+                    for (b = 0; b < k; ++b) {
+                        m_out[r * n + c] +=
+                            m_left[r * k + b] * m_right[c * n + b];
+                    }
+                }
+            }
+        } else {
+            std::size_t i, j, l;
+            std::size_t s = 2; // stride
```
mlxd, line 621:
```suggestion
            std::size_t i, j, l;
            std::size_t stride = 2; 
```
As mentioned above, this is the way to go

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+                            for (j = c; j < std::min(c + s, n); ++j) {
+                                t = 0;
+                                for (l = b; l < std::min(b + s, k); ++l) {
+                                    t += m_left[i * k + l] * m_right[l * n + j];
+                                }
+                                m_out[i * n + j] += t;
+                            }
+                        }
+                    }
+                }
```
mlxd, line 640: This may levels of looping will annihilate the code-complexity tests. Is there another way to achieve the same logic?

pennylane_lightning/src/util/Util.hpp
maliasadi: A good call! I thought these names are explicit enough representing row, column, and block :) I will use a better naming.

pennylane_lightning/src/util/Util.hpp
maliasadi: This is fixed!

pennylane_lightning/src/util/Util.hpp
maliasadi: The other way around is loop-unrolling for small strides, however, this makes the function too long with many if-else conditions to consider all the corner cases.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+/**
+ * @brief Partition [0, data_size] into n subset of size data_size/n.
+ *
+ * @param n Number of partitions.
+ * @param data_size Size of data array.
+ * @return std::vector of partitions indices.
+ *
+ * @note the last index sets to data_size.
+ */
+inline static auto partition(size_t n, size_t data_size)
```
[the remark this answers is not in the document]
maliasadi: Ops! This is part of the old code... deleted!

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+    constexpr size_t BLOCK_THRESHOLD = 16;
+    size_t r, s, r1, s1, r2, s2;
+tail:
+    r1 = m2 - m1;
+    s1 = n2 - n1;
+    if (r1 >= s1 && r1 > BLOCK_THRESHOLD) {
+        r2 = (m1 + m2) / 2;
+        CFTranspose(mat, mat_t, m, n, m1, r2, n1, n2);
+        m1 = r2;
+        goto tail;
```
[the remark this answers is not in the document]
maliasadi: I'm not a big fan of `goto` either. But, I found it easy enough to do in this way :)

pennylane_lightning/src/util/Util.hpp
mlxd: Ah yea, makes sense. I guess we can disable the complexity tests on these loops if/when they start to complain

pennylane_lightning/src/util/Util.hpp
```diff
@@ -199,87 +205,509 @@ template <class T> inline size_t dimSize(const std::vector<T> &data) {
...
+ * @param v_out Pre-allocated complex data array to store the result.
+ * @param m Number of rows of `mat`.
+ * @param n Number of columns of `mat`.
+ * @param transpose If `true`, considers transposed version of `mat`.
+ * @param omp_critical If `true`, uses column-wise parallel-for, otherwise
+ * row-wise.
+ */
+template <class T>
+inline static void
+omp_matrixVecProd(const std::complex<T> *mat, const std::complex<T> *v_in,
```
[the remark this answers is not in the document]
mlxd: If it becomes too much work to refactor to appease the tests, we can find another route.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -204,89 +207,465 @@ template <class T> inline auto dimSize(const std::vector<T> &data) -> size_t {
...
     std::complex<T> result(0, 0);
 
-    if constexpr (USE_CBLAS) {
-        if constexpr (std::is_same_v<T, float>) {
-            cblas_cdotu_sub(data_size, data_1, 1, data_2, 1, &result);
-        } else if constexpr (std::is_same_v<T, double>) {
-            cblas_zdotu_sub(data_size, data_1, 1, data_2, 1, &result);
-        }
-    } else {
+#if defined(USE_CBLAS) && USE_CBLAS
```
mlxd, line 253: Does `if constexpr` not work here? I'd prefer it if we could use it instead of the preprocessor where possible. Any non-enabled `#ifdef` code will never hit the compiler after the preprocessing step, whereas anything with `if constexpr` will, and hence can be checked for correctness, even if not used.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -204,89 +207,465 @@ template <class T> inline auto dimSize(const std::vector<T> &data) -> size_t {
...
+    if constexpr (std::is_same_v<T, float>) {
+        cblas_cdotu_sub(data_size, v1, 1, v2, 1, &result);
+    } else if constexpr (std::is_same_v<T, double>) {
+        cblas_zdotu_sub(data_size, v1, 1, v2, 1, &result);
+    }
+#else
+    if (data_size > DOTU_STD_CROSSOVER) {
```
mlxd, line 260: Should this be 
```suggestion
    if (data_size < DOTU_STD_CROSSOVER) {
```
Or does OMP get enabled for small sizes only?

pennylane_lightning/src/util/Util.hpp
```diff
@@ -204,89 +207,465 @@ template <class T> inline auto dimSize(const std::vector<T> &data) -> size_t {
...
+                       const size_t data_size) -> std::complex<T> {
     std::complex<T> result(0, 0);
 
-    if constexpr (USE_CBLAS) {
-        if constexpr (std::is_same_v<T, float>) {
-            cblas_cdotc_sub(data_size, data_1, 1, data_2, 1, &result);
-        } else if constexpr (std::is_same_v<T, double>) {
-            cblas_zdotc_sub(data_size, data_1, 1, data_2, 1, &result);
-        }
+#if defined(USE_CBLAS) && USE_CBLAS
```
mlxd, line 317: As previously mentioned.

pennylane_lightning/src/util/Util.hpp
maliasadi: `if constexpr` works only for innerProd and innerProdC, but not for matrixVecProd and matrixMatProd. So, to be consistent, I switched using `#ifdef` for all places using BLAS operations. Let me know if you still think that using `if constexpr` is the better idea for innerProd and innerProdC?

mlxd: I think we should figure out how to keep `if constexpr` in here as it will be much safer and maintainable.

maliasadi: In this case, using define directive to define undefined BLAS variables in matrixVecProd and matrixMatProd is a safe and practical approach to tackle this problem. What do you think?

mlxd: I still think enabling this with `if constexpr` is the way to go. Since half the code gets disabled before hitting the compiler, going with `if constexpr` allows for better static code analysis, as well as a feature of the modern uses of the language. Avoiding the preprocessor unless required has become a common theme in many C++ articles (https://cor3ntin.github.io/posts/undef_preprocessor/ is a good example). If using `if constexpr` is not possible, maybe in this instance many of the helpers in the `type_traits` library could help out here (https://en.cppreference.com/w/cpp/header/type_traits), such as `enable_if` for example.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -36,6 +36,13 @@
 #define USE_CBLAS 1
 #else
 #define USE_CBLAS 0
+
+#define CblasTrans 0
+#define CblasNoTrans 0
+#define CblasRowMajor 0
```
mlxd, line 42: Since these flags are used to select logical options, defining them with an `enum` is a safer move. This can be placed in an anonymous namespace to enable visibility within the file only.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -36,6 +36,13 @@
...
+
+#define DOTU_STD_CROSSOVER (1 << 10)
+#define DOTC_STD_CROSSOVER (1 << 10)
```
mlxd, line 45: These may also be better used as [template non-type arguments](https://www.ibm.com/docs/en/zos/2.1.0?topic=arguments-template-non-type) in the functions they are used in. This way we can conditional compile it using templating, and not be required to specify it as a defined magic value.

pennylane_lightning/src/util/Util.hpp
mlxd: [This article by Sy Brand](https://blog.tartanllama.xyz/if-constexpr/) who works on Microsoft's MSVC compiler has some good examples of doing conditional compilation for different operating systems.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -204,89 +212,466 @@ template <class T> inline auto dimSize(const std::vector<T> &data) -> size_t {
...
+#pragma omp declare \
+            reduction (sm:std::complex<T>:omp_out=ConstSum(omp_out, omp_in)) \
+            initializer(omp_priv=std::complex<T> {0, 0})
+#endif
+
+#if defined _OPENMP
+#pragma omp parallel for default(none) shared(v1, v2, data_size)               \
+    reduction(sm                                                               \
+              : result)
+#endif
```
mlxd, line 238: All good using `ifdef` here, since OpenMP doesn't really allow any other way.

mlxd (changes requested):

mlxd (approved):

PR #157, opened 2021-10-20
Optimise x86_64 builds with AVX friendly opts

mlxd:
**Context:** This enables AVX operations by default for Intel/AMD machines which can improve performance of vectorisable operations.

**Description of the Change:** AVX support is added to Linux x86_64 wheels. In addition, the width of various registers is reduced for many parametric ops by explicitly favouring 64-bit wide operations instead of 128-bit (double vs std::complex<double>) where applicable.

**Benefits:** Improves overall runtime performance.

**Possible Drawbacks:** Machines not suport AVX (pre-2011) will require manual builds to run lightning.

**Related GitHub Issues:**


trevor-vincent (approved):
Awesome work Lee. I see no issues!

PR #159, opened 2021-11-01
Added auto version bump functionality to pl lightning

Jaybsoni:
**Context:**
- Automated version bumping of the lightning plugin

- added a two new GitHub workflows to bump the version # on release

### DIFF
--- a/.github/workflows/post_release_version_bump.yml
+++ b/.github/workflows/post_release_version_bump.yml
@@ -35,5 +35,5 @@ jobs:
           title: Version Bump
           body: updated changelog and _version.py
           branch: post-release-version-bump
-          reviewers: Jaybsoni
+          reviewers: antalszava
           base: main
--- a/.github/workflows/post_release_version_bump.yml
+++ b/.github/workflows/post_release_version_bump.yml
@@ -26,7 +26,7 @@ jobs:
         run: >
           python .github/workflows/vb_script.py
           --version_path "./pennylane_lightning/_version.py"
-          --changelog_path "./.github/CHANGELOG.md" --post_release --post_pl_release
+          --changelog_path "./.github/CHANGELOG.md" --post_release
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
--- a/.github/workflows/post_release_version_bump.yml
+++ b/.github/workflows/post_release_version_bump.yml
@@ -1,0 +1,39 @@
+name: Automated Release Version Bumps
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  post_release_version_bump:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pennylane
+
+      - name: Run version bump
+        run: >
+          python .github/workflows/vb_script.py
+          --version_path "./pennylane_lightning/_version.py"
+          --changelog_path "./.github/CHANGELOG.md" --post_release --post_pl_release
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: post release version bump
+          title: Version Bump
+          body: updated changelog and _version.py
+          branch: post-release-version-bump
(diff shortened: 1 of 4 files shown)


.github/workflows/post_release_version_bump.yml
```diff
@@ -0,0 +1,39 @@
...
+          --changelog_path "./.github/CHANGELOG.md" --post_release --post_pl_release
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: post release version bump
+          title: Version Bump
+          body: updated changelog and _version.py
+          branch: post-release-version-bump
+          reviewers: Jaybsoni
```
Jaybsoni, line 38: Will change this to whom ever this plugin is assigned to as a reviewer

.github/workflows/pre_release_version_bump.yml
```diff
@@ -0,0 +1,38 @@
...
+          --changelog_path "./.github/CHANGELOG.md" --pre_release --post_pl_release
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: pre release version bump
+          title: Version Bump
+          body: updated changelog and _version.py
+          branch: pre-release-version-bump
+          reviewers: Jaybsoni
```
Jaybsoni, line 37: same as above

.github/workflows/post_release_version_bump.yml
antalszava: Feel free to add me here, ccing @mlxd too: Jay is adding an automated version bump for the pre and post-release of Lightning. The new actions would tidy the changelog and increment the version number. For now, the action opens a new PR and this field determines who'll be put as the reviewer(s) to the PR. Would you also like to be tagged? :slightly_smiling_face:

.github/workflows/vb_script.py
```diff
@@ -0,0 +1,179 @@
...
+    replaces it with the bumped version depending on the pre/post
+    release flag.
+
+    Args:
+         version_line (string): The string containing the current
+            version of the plugin.
+         pre_release (bool): A flag which determines if this is a
+            pre-release or post-release version bump.
+        pl_release (bool): A flag which determines if the new
+            version of PennyLane has already been released to Pypi
```
antalszava, line 30: Likely, we could do without this option. For the pre-release version bump, there will likely not be a scenario where we have

a) not done the pre-release version bump;
b) have released PennyLane.

Where there's a minor difference in the case of Lightning, is the post-release version bump: we'll likely be releasing PennyLane-Lightning first and then PennyLane. If the opposite order was assumed and used somewhere, then we'd need a change in logic there.

antalszava (commented):
Looking great overall :100: Just one comment regarding `pl_release_status`.

.github/workflows/vb_script.py
Jaybsoni: Ahh I see. Currently it can support performing both pre or post release version bumps both before or after we have released Pennylane. I can remove this flag as for post-release version bumps I do not use the current version of PL, but instead just increment the version found in the change log and `_version.py` file. My personal recommendation would be to keep it in as this kind of functionality might become useful if in the future we decide to change up when we are going to be releasing some of the plugins.

.github/workflows/post_release_version_bump.yml
Jaybsoni:
```suggestion
          reviewers: antalszava
```

.github/workflows/pre_release_version_bump.yml
Jaybsoni:
```suggestion
          reviewers: antalszava
```

.github/workflows/pre_release_version_bump.yml
```diff
@@ -0,0 +1,38 @@
...
+  workflow_dispatch:
+
+jobs:
+  pre_release_version_bump:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
```
Jaybsoni, line 13:
```suggestion

```

.github/workflows/vb_script.py
antalszava: If it's not used at the moment, then maybe it's best to remove it. It can lead to confusion to have an option that is not needed.

antalszava (approved):
Looks good to me :100:

antalszava:
[sc-10712]

Issue #158, opened 2021-10-28
Build stand-alone lightning binary for gate aggregate performance tests


mlxd:
#### Issue description
Benchmarking lightning directly without the Python interface can help in isolating performance regressions between the bindings and the underlying kernels. The purpose of this task is to create a PennyLane Lightning binary without the Python interface, and test the aggregate gate performance.

Your task will be the following:
- Build a quantum circuit simulation binary directly in C++ using PennyLane Lightning in the new `examples` directory. You can make use of the class `StateVectorManaged` to create and allocate the required memory for your simulation. Examples of use are located within the `src/tests/Test_StateVectorManaged_*.cpp` files. You can use either the dispatcher (gate-string call) or direct gate call to make use of gates. You can also use CMake to assist with the build process (see the [JET](https://quantum-jet.readthedocs.io/en/latest/dev/guide.html#example-cmake-project) guide for an example of how this may be achieved).

Your simulation binary should make a user-defined number of calls to a user-defined number of qubits for the following built-in parametric and non-parametric gates:
    - Non-parametric: (PauliX, PauliY, PauliZ, Hadamard, CNOT, CZ)
    - Parametric: (RX, RY, RZ, CRX, CRY, CRZ)
    - User-defined call number can be read from a command-line argument (eg `./binary 10 6` will run each gate ten times in the circuit per qubit, with 6 qubits overall).

Run your simulation for qubit counts ranging from 6 - 22 in steps of 2, and plot the timing results, noting your compiler type, and optimization level. Wall-clock time will be sufficient here.

PR #165, opened 2021-11-05
Gate Aggregate Performance Tests; Issue #158

isschoch:
**Context:**
Technical part of coding challenge for Xanadu residency program.
Added a new directory called "examples". In this directory one can benchmark the performance of the gates in Pennylane-Lightning with qubits ranging from 6 to 22 in steps of 2 by running the run_gate_benchmark.sh script. 
**Benefits:**
Can be used to assess performance of gates in Pennylane-Lightning.
**Possible Drawbacks:**
Using gprof will give you a more nuanced overview.
**Related GitHub Issues:**
Issue #158


mlxd:
Thank for the submission @isschoch 
We'll have a look and get back to you early next week!

mlxd:
Hi @isschoch thanks again for the contribution. To finish off this work, I will ask for some small changes in your code review. There is no hurry here, so feel free to get back to it when suits you.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    GIT_TAG         master
+)
+FetchContent_MakeAvailable(Pennylane-lightning)
+
+#############################
+## III. Create project target
+#############################
+
+set(CMAKE_CXX_STANDARD 17) # At least C++17 is required
+# set(CMAKE_CXX_FLAGS "-O3") # Not including flags here as I pass them via commandline args in bash script (I would normally set it here)
```
mlxd, line 31: I think it is fine to set `O3` here. For real-world use cases, we don't want to pass compile time options through bash scripts, instead relying on the library we use to already be compiled with the optimal flags.

examples/README.md
```diff
@@ -0,0 +1,20 @@
+Run "bash run_gate_benchmark.sh" in the terminal. The script will automatically build the gate_benchmark project.
```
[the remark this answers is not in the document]
mlxd: Using backticks here renders the output as code (single for inline, triple for blocks).

examples/README.md
```diff
@@ -0,0 +1,20 @@
...
+It will set the CXX and CMAKE_CXX_FLAGS environment variables to "which clang++" and "-O3".
```
mlxd, line 2: It would be great if this were user-selectable, rather than set in the script (even as a command-line argument). Other potential options could be `g++`, `icpc`, `icpx`, `nvc++` and others.

examples/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+gate_benchmark.cpp:
+* I decided to generate a single random angle per parameterised gate as I feel like this makes use of parametric property of the gates while keeping the performance penalty due to cache access to a bare minimum
+* I decided to apply the gates in the order that they were listed in the task; furthermore I perform the gate repitition after every single gate has been applied to avoid performing identity operations for gates which are involutory
+* I have included many different libraries without much regard as to which libraries are already used in pennylane-lightning as I don't believe this is the focus of this task
+
+gate_benchmark_plotter.py:
+* The first plot is supposed to be close to the task description
+* Due to the exponential nature of quantum computing one cannot read out the time for few qubits; I therefore also included a loglog plot depicting the scaling of the program
+* Only the average is plotted as described in the task description. Normally I would average over many more runs and plot the 95% confidence interval or standard deviation
```
mlxd, line 20: Great job on the description here. Though, one thing to note is this readme could server as a way to document the example benchmark of the library using your implemented work. In this case, I think the phrasing needs to focus more on what is happening as is, with less focus on your decisions.


eg
*I decided to only run the lightning code in the cpp file and to control the compilation, file-creation etc via a bash script*
could be replaced with

*The compile-time options are controlled by the bash script `scriptname.sh`

examples/CMakeLists.txt
mlxd: Also, one thing we may/should do to ensure adherence with the rest of the library is make sure the clang-tidy option is run (and no errors given) at build-time. We strictly adhere to clang-tidy recommendations for code structure and formatting, treating all warnings as errors to avoid any future problems. The following is an excerpt from the Lightning CMake

```
option(ENABLE_WARNINGS "Enable warnings" ON)
option(ENABLE_CLANG_TIDY "Enable clang-tidy build checks" OFF)

if(ENABLE_CLANG_TIDY)
    if (NOT DEFINED CLANG_TIDY_BINARY)
        set(CLANG_TIDY_BINARY clang-tidy)
    endif()
    set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_BINARY};
                            -extra-arg=-std=c++17;
    )
endif()
```

which will build with support when enabled (`-DENABLE_CLANG_TIDY=ON`), and treat all warnings as errors.

examples/run_gate_benchmark.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+crt_dir=$(pwd)
+
+# Export env variables in case cmake reverts to default values
+export CXX=$(which clang++)
```
mlxd, line 6: Maybe defining a bash array with the expected compilers above here, and iterating over them, producing a build/output per compiler queried.

mlxd (commented):
Great work on this. I have a few small comments to make below:

.github/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Added examples folder containing aggregate gate performance test
```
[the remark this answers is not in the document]
isschoch: Included your changes and also added "### New features since last release" heading.

examples/README.md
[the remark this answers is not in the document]
isschoch: Incorporated your changes. Thanks for the hint!

examples/README.md
isschoch: The compiler name now has to be specified as a command-line argument after `run_gate_benchmark.sh` script.

examples/README.md
isschoch: Changed the `README.md` to be more concise and to actually describe what was implemented. The first version was meant to clear up any questions regarding my thought-process in the context of the coding challenge.

examples/run_gate_benchmark.sh
isschoch: Don't believe I fully understand the suggestion. Is the idea to test and compare multiple compilers by listing them as command-line arguments? E.g. `bash run_gate_benchmark.sh clang++ g++` would benchmark with both clang++ and g++?
If the idea is to iterate through possible installed compilers to find one that works then I believe the solution of passing the compiler name via a command-line argument circumvents this problem.

mlxd: Sure, passing the compiler explicitly as an argument would work. In that case, ignore my comment about having a list of available compilers in the file.

examples/CMakeLists.txt
isschoch: Dear @mlxd,
I have been working on adhering to the clang-tidy configuration that has been set up in PL lightning and added the suggested lines to the `CMakeLists.txt` file in the examples folder. I've experienced some difficulties while trying to CMake with `-DENABLE_CLANG_TIDY=ON` on macOS which I will outline:
  The macOS clang compiler does not come preinstalled with clang-tidy so I installed llvm@11 and llvm@13 via brew and  specify the clang-tidy binary by adding `DCLANG_TIDY_BINARY=/opt/homebrew/opt/llvm@11/bin/clang-tidy` when executing CMake. There is a problem with this setup which I cannot resolve: When executing `cmake -DENABLE_CLANG_TIDY=ON -DCLANG_TIDY_BINARY=/opt/homebrew/opt/llvm@11/bin/clang-tidy .. && make` in the master branch (same error in the examples folder) of PL lightning I get the error message 
  `{PATH_TO_PL_LIGHTNING}/pennylane-lightning/pennylane_lightning/src/simulator/StateVector.hpp:278:5: error: function 'getData' should be marked [[nodiscard]] [modernize-use-nodiscard,-warnings-as-errors]
      auto getData() const -> CFP_t * { return arr_; }
      ^
      [[nodiscard]] 
  8344 warnings generated.
  Suppressed 8351 warnings (8342 in non-user code, 9 NOLINT).
  Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
  1 warning treated as error
  make[2]: *** [pennylane_lightning/src/simulator/CMakeFiles/lightning_simulator.dir/StateVector.cpp.o] Error 1
  make[1]: *** [pennylane_lightning/src/simulator/CMakeFiles/lightning_simulator.dir/all] Error 2
  make: *** [all] Error 2`.
  
  I have tried explicitly including the `pennylane-lightning/src/.clang-tidy` config file using llvm@13 and the flag `--config-file` to no avail.

The only way I have been able to run clang-tidy on the `gate_benchmark.cpp` is to execute `cmake -DENABLE_CLANG_TIDY=on -DCLANG_TIDY_BINARY=/opt/homebrew/opt/llvm@13/bin/clang-tidy -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..` in the `/examples/build` directory as described here  https://stackoverflow.com/questions/39455090/clang-tidy-cant-find-my-header-files.
I then move the generated `compile_commands.json` file to the `examples` directory and execute clang-tidy on `gate_benchmark.cpp` which does not result in any errors. As this solution is sub-optimal I was wondering if you have any suggestions of what I could try next to make sure my code fulfils the clang-tidy configuration of PL lightning.

mlxd: Hey @isschoch thanks for your diligence on this. The clang-tidy warning (being reported as an error) is complaining about this line in lightning's `StateVector.hpp` file: https://github.com/PennyLaneAI/pennylane-lightning/blob/4d8df75da8b8c9f5576f464fb390a9b895988480/pennylane_lightning/src/simulator/StateVector.hpp#L278

Namely, the `[[nodiscard]]` qualifier (https://en.cppreference.com/w/cpp/language/attributes/nodiscard) means that the return type of the method must be captured. I think adding `[[nodiscard]]` similar to:
https://github.com/PennyLaneAI/pennylane-lightning/blob/4d8df75da8b8c9f5576f464fb390a9b895988480/pennylane_lightning/src/simulator/StateVector.hpp#L319
may fix the complaint. It may be possible to auto-fix it using the `-fix` and `-fix-error` arguments to clang-tidy on the Lightning source (see [here](https://stackoverflow.com/questions/53195335/clang-tidy-fix-does-not-apply-any-changes)).

Using CMake this way pulls the lightning repo into a `_deps` directory, which can be modified to test this out during compilation. If this fixes the problem, I think it would be good to add the missing statements as a separate Lightning PR, since this will likely happen again in future.

mlxd: Actually, on second thought, fixing in this PR should be fine, rather than opening a second one.

isschoch: Thank you for the quick and detailed reply. I got the same error at multiple locations so I also modified all of those. In addition clang-tidy requested that the namespaces be concatenated instead of nested in some parts of the code (i.e. `namespace Pennylane::Util{...}` instead of `namespace Pennylane{namespace Util{...}}`). Here's a quick overview of the code that caused the warnings:

https://github.com/isschoch/pennylane-lightning/blob/075671cf9a9c929fa062534ce21cce1fa99486c5/pennylane_lightning/src/algorithms/AdjointDiff.hpp#L139-L140

https://github.com/isschoch/pennylane-lightning/blob/075671cf9a9c929fa062534ce21cce1fa99486c5/pennylane_lightning/src/simulator/Gates.hpp#L15-L16

https://github.com/isschoch/pennylane-lightning/blob/075671cf9a9c929fa062534ce21cce1fa99486c5/pennylane_lightning/src/simulator/StateVector.hpp#L278

https://github.com/isschoch/pennylane-lightning/blob/075671cf9a9c929fa062534ce21cce1fa99486c5/pennylane_lightning/src/simulator/StateVectorManaged.hpp#L74

https://github.com/isschoch/pennylane-lightning/blob/075671cf9a9c929fa062534ce21cce1fa99486c5/pennylane_lightning/src/util/Util.hpp#L56-L57

I'll push the fixes to this PR if you give me the go ahead to keep it in the same PR (double checking since the fix isn't a one-liner anymore). With this clang-tidy doesn't give me any errors in the examples folder as long as I replace the `examples/build/_deps/pennylane-lightning-src/pennylane_lightning/src` with the fixed version.

mlxd: Nice work! I think pushing to this PR makes sense. If there are any issues, it will be caught by the automated CI test-suite. I'm happy to have a look at this PR once the changes are made too, and we can always adapt if needed.

examples/README.md
```diff
@@ -0,0 +1,19 @@
+Run `bash run_gate_benchmark.sh $CXX_COMPILER`, where `$CXX_COMPILER` is the compiler you wish to use, in the terminal (e.g. `bash run_gate_benchmark.sh clang++`). The script will automatically build the gate_benchmark project.
+It will set the CXX environment variable to "$CXX_COMPILER".
+
+Some comments on the implementation:
+General comments: 
+* The compile-time options are controlled by the bash script `run_gate_benchmark.sh`
+* The lightning code is run in the cpp file
+* Plotting is done with the python script `gate_benchmark_plotter.py`
```
mlxd, line 8:
```suggestion
* Plotting is accomplished with the Python script `gate_benchmark_plotter.py`. 
* Plotting requires the packages listed in `requirements.txt`
```

Be sure to add a `requirements.txt` file for the dependent Python packages to enable plotting. Looking at the Python file, this will need to include `numpy`, `matplotlib` and `pandas`.

examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+* The number of gate repitions is set to 3 and can be changed in the bash script `run_gate_benchmark.sh` by changing the `num_gate_reps` variable
```
mlxd, line 9:
```suggestion
* The number of gate repetitions is set to 3 and can be changed in the bash script `run_gate_benchmark.sh` by modifying the `num_gate_reps` variable
```

examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+gate_benchmark.cpp:
```
mlxd, line 11:
```suggestion
### `gate_benchmark.cpp`:
```

examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+* A single random angle is generated per gate repetition and qubit; the same random angle is used once for all of the parameterised gates
+* The gates are applied in the order X, Y, Z, H, CNOT, CZ, RX, RY, RZ, CRX, CRY, CRZ
+* The above order is repeated `num_gate_reps`-times
+
+gate_benchmark_plotter.py:
```
mlxd, line 16:
```suggestion
### `gate_benchmark_plotter.py`:
```

If the backticks renders the page strangely, feel free to ignore them.

examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+* The first plot shows the absolute runtime
+* The second plot is on a loglog scale which better depicts the exponential scaling of the relative runtime with respect to the number of simulated qubits
+* We plot the time needed to execute the gate sequence averaged over the repititions
```
mlxd, line 19:
```suggestion
* We plot the time needed to execute the gate sequence averaged over the repetitions
```

examples/gate_benchmark_plotter.py
```diff
@@ -0,0 +1,39 @@
...
+    plt.plot(data[:, num_qubits_idx], avg_time_arr, "rX")
+    plt.figtext(0.05,0.0, ("Compiler:\t" + compiler_info[0] + "Optimization:\t" + optimization).expandtabs(), fontsize=7, va="bottom", ha="left")
+    plt.subplots_adjust(bottom=0.2)
+    plt.savefig("avg_time.png", dpi=200)
+    plt.close()
+
+    # Plot relative values in log-log plot
+    plt.title("Scaling Behaviour: Relative Time vs Number of Qubits")
+    plt.xlabel("Number of Qubits in $[1]$")
+    plt.ylabel("Relative Time (compared to 1 qubit) in $[1]$")
```
mlxd, line 33: Is this time relative to `1` or time relative to the first simulation taken?

mlxd (changes requested):
Nice work again! I have a few more small suggestions.

examples/gate_benchmark_plotter.py
isschoch: I divide all times by the average time it took to run the circuit with 1 qubit, so yes if by first simulation you also mean the (averaged) simulation with 1 qubit. I do this in order to get a dimensionless quantity.

.github/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+* Added examples folder containing aggregate gate performance test.
+[(#165)](https://github.com/PennyLaneAI/pennylane-lightning/pull/165)
+
+### Contributors
+
+This release contains contributions from (in alphabetical order):
+
+Isidor Schoch
+
+---
```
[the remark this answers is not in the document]
isschoch: Moved bullet point under new features of 0.20.0-dev and added my name to the contributors list.

examples/gate_benchmark_plotter.py
mlxd: Got it, thanks.

mlxd:
Hey @isschoch great job on this. I am happy to see it added to the repo.

The testing seems to be failing with the recent merge from master --- I think it should be fine to try adding an empty commit to trigger the CI with: ` git commit -m "Trigger CI" --alow-empty` and I'll reapprove the runs.

isschoch:
@mlxd thank you for notifying me about the runs. Sorry for pushing the change so late, is it still early enough for the 0.20-dev release?

mlxd:
> @mlxd thank you for notifying me about the runs. Sorry for pushing the change so late, is it still early enough for the 0.20-dev release?

All good --- I'll let the runs go, and if everything passes I'll merge immediately for `0.20`. Good job again!

mlxd (approved):

PR #170, opened 2021-11-09
Merge v0.19.0 release fixes onto master

mlxd:
**Context:** Bugfixes from v0.19.0 release

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -55,7 +57,7 @@ jobs:
           CIBW_ARCHS_LINUX: ${{matrix.arch}}
 
       - uses: actions/upload-artifact@v2
-        if: github.ref == 'refs/heads/master'
+        if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master' }}
```
trbromley, line 60: Why do we need the `or` here?

.github/workflows/wheel_linux_x86_64.yml
```diff
@@ -1,10 +1,12 @@
-name: Wheels
+name: Wheel::Linux::x86_64
 on:
+  release:
```
trbromley, line 3: Do we need the "types" line, as we do e.g. for `wheel_linux_ppc64le.yml`? (same question applies to some of the other workflows)

trbromley (approved):
Thanks @mlxd!

.github/CHANGELOG.md
```diff
@@ -16,30 +16,34 @@ This release contains contributions from (in alphabetical order):
 
 ---
 
-# Release 0.19.0-dev
+# Release 0.19.0
 
 ### New features since last release
```
antalszava, line 21:
```suggestion
```

.github/workflows/wheel_linux_aarch64.yml
mlxd: Good question. I think `||` is supported (link https://docs.github.com/en/actions/learn-github-actions/expressions#operators) though happy to modify if you think `or` is better.

mlxd: Ah, I misunderstood. Yea, we want this to run in both cases. We can issue a release from a non master branch (the tagged v0.xy branch), and ensure the wheels are built.  Master may have since move on, so we need to be able to support both ( I think).

.github/workflows/wheel_linux_x86_64.yml
mlxd: Good catch. I think these need to be added to all.

mlxd: Good catch -- I think these need to be added to all workflows.

mlxd: Rather than changing the RC, maybe we merge this as is and modify in a subsequent PR. This won't change the wheel, and help with consistency across the wheel history. What do you think?

trbromley: Sounds good! Thanks @mlxd!

PR #176, opened 2021-11-12
[WIP] setup.py uses CMake

chaeyeunpark:
* `setup.py` is rewritten to use CMake.
* Compiling guide for MSVC is added (README.rst).


chaeyeunpark:
Windows tests fail as python does not find a correct DLL path. https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew can be a reason?

README.rst
```diff
@@ -97,7 +97,6 @@ while the C++ code can be tested with
 
     $ make test-cpp
 
-Testing the C++ code requires the `GoogleTest <https://github.com/google/googletest>`__ framework.
```
mlxd, line 100: Ooops -- my bad on forgetting to remove this! Thanks!

README.rst
```diff
@@ -121,19 +120,47 @@ To test the C++ code:
...
+    $ python --version
+
+Then a common command will work.
+
+.. code-block:: console
+
+    $ pip install -r requirements.txt
+    $ pip install -e .
+
+Note that OpenMP and BLAS are disabled in this setting.
```
mlxd, line 153: :100:

pennylane_lightning/lightning_qubit.py
```diff
@@ -32,12 +33,25 @@
...
-    from .lightning_qubit_ops import (
-        apply,
-        StateVectorC64,
-        StateVectorC128,
-        AdjointJacobianC128,
-    )
+    print(os.path.dirname(os.path.abspath(__file__)))
+    if platform.system() == "Windows" and sys.version_info[:2] >= (3, 8):
+        # Add the current directory to DLL path.
+        # See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
```
mlxd, line 39: Will this have any problem with Python 3.7 support on Windows?

pennylane_lightning/src/algorithms/CMakeLists.txt
```diff
@@ -4,16 +4,9 @@ set(CMAKE_CXX_STANDARD 17)
...
-
-find_package(OpenMP)
-if(OpenMP_FOUND)
-    target_link_libraries(lightning_algorithms PRIVATE OpenMP::OpenMP_CXX)
-endif()
+target_link_libraries(lightning_algorithms PRIVATE pennylane_lightning_compile_options
+                                                   pennylane_lightning_external_libs)
 
+target_include_directories(lightning_algorithms PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} )
+target_link_libraries(lightning_algorithms PRIVATE lightning_simulator lightning_utils)
```
mlxd, line 11: Since the `lightning_algorithms` target can be compiled as a separate module, do you think we should keep the OpenMP target link step in here?

pennylane_lightning/src/simulator/CMakeLists.txt
```diff
@@ -4,17 +4,10 @@ set(CMAKE_CXX_STANDARD 17)
 set(SIMULATOR_FILES StateVector.cpp StateVector.hpp StateVectorManaged.hpp Gates.hpp CACHE INTERNAL "" FORCE)
 add_library(lightning_simulator STATIC ${SIMULATOR_FILES})
 
-find_package(OpenMP)
-
+target_link_libraries(lightning_simulator PRIVATE pennylane_lightning_compile_options
+                                                   pennylane_lightning_external_libs)
+
```
mlxd, line 9: Same again here.

setup.py
```diff
@@ -14,167 +14,90 @@
...
-            library_dirs=library_dirs,
-            extra_compile_args=extra_compile_args,
-            extra_link_args=extra_link_args,
-        ),
-    ]
-else:
-    ext_modules = []
 
 requirements = [
+    "ninja",
```
mlxd, line 100: I assume ninja is a requirement for building on Windows?

setup.py
```diff
@@ -14,167 +14,90 @@
...
-        "unix": ["-O3", "-fPIC", "-shared", "-fopenmp", "-Wall", "-Wextra", "-Werror"],
-    }
+        super().finalize_options()
 
-    l_opts = {
-        "msvc": ["/WX"],
-        "unix": ["-O3", "-fPIC", "-shared", "-fopenmp", "-Wall", "-Wextra", "-Werror"],
-    }
-    if platform.system() == "Linux" and platform.machine() == "x86_64":
-        c_opts["unix"].append("-mavx")
```
mlxd, line 86: We should make sure the AVX build options get added somewhere here for x86_64 platforms.

mlxd (commented):
Nice work on this @chaeyeunpark 
I have a few small queries and comments.

pennylane_lightning/lightning_qubit.py
chaeyeunpark: Python version below 3.8 follows the legacy (which means the `else` part will be executed without a problem). `cibuildwheel` indeed tests Windows+Python3.7 which do not report any problem.

setup.py
chaeyeunpark: Ninja works on all platforms (and usually faster than Makefile). I used Ninja as it gives the same build script for all environments. I also could compile the library without Ninja in Linux and MSVC, but it was harder to find out what is the problem (when there is as the output of `CMake` was tons of XML files for msbuild).

setup.py
chaeyeunpark: I agree. I will add some lines in `setup.py` and `CMakeFiles.txt` to deal with this.

.github/workflows/deploy.yml
```diff
@@ -62,6 +62,9 @@ jobs:
         with:
           python-version: '3.7'
 
+      - name: Install dependencies
+        run: sudo apt-get -y -q install cmake gcc
```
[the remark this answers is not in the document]
chaeyeunpark: I found clang also works well here. By the way, `deploy.yml` is removed in `master`. I am slightly confused about what I should do.

CMakeLists.txt
```diff
@@ -26,13 +28,20 @@ if(NOT CMAKE_BUILD_TYPE)
...
 endif()
 
+# Clang-tidy
+option(ENABLE_CLANG_TIDY "Enable clang-tidy build checks" OFF)
+
+# Compile options
+option(ENABLE_WARNINGS "Enable warnings" ON)
 option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)
+option(ENABLE_OPENMP "Enable OpenMP" ON)
+option(ENABLE_BLAS "Enable BLAS" ON)
```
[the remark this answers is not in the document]
chaeyeunpark: I will turn it off.

pennylane_lightning/src/algorithms/CMakeLists.txt
chaeyeunpark: I found that `lightning_algorithms` anyway cannot be compiled independently, as it depends on other sub-targets. Let's just keep this and fix it when we really use sub-targets independently.

setup.py
chaeyeunpark: Fixed in 9322482ecd60f334a5f58aa8a386e291b058e9ba

.github/workflows/wheel_linux_ppc64le.yml
```diff
@@ -9,7 +9,7 @@ env:
 
   # Python build settings
   CIBW_BEFORE_BUILD: |
-    pip install pybind11
+    pip install pybind11 ninja && yum install -y openblas openblas-devel cmake gcc gcc-c++
```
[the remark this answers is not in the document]
chaeyeunpark: Fixed in 9322482

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -9,7 +9,7 @@ env:
 
   # Python build settings
   CIBW_BEFORE_BUILD: |
-    pip install pybind11
+    pip install pybind11 ninja && yum install -y openblas openblas-devel cmake gcc gcc-c++
```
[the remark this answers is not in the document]
chaeyeunpark: Fixed in 9322482

CMakeLists.txt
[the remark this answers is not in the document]
chaeyeunpark: Fixed in 9322482

setup.py
```diff
@@ -14,172 +14,91 @@
...
-            ext.extra_link_args = link_opts
-
-        build_ext.build_extensions(self)
-
-
-if not os.environ.get("SKIP_COMPILATION", False):
+        build_args = []
+
+        if platform.system() == "Darwin":
+            configure_args += []
```
mlxd, line 72: One more change I think we need here is the OpenMP support being disabled on the M1 Macs. I had a workaround until Github Actions add M1 machines to their CI at https://github.com/PennyLaneAI/pennylane-lightning/blob/63929edacc85840b88208bfc0ef9dd75901e2cf6/setup.py#L98

Basically, if we enable the env variable the Mac build with or without OpenMP support. This comes from us using libomp in brew, which only pulls in the x86_64 version, even when building for M1 Macs (as they are still running on x86_64 in Github actions).

mlxd:
It may also be worth enabling the CI wheel builders to export the built wheels for this PR (they are normally disabled outside master). As a quick workaround, we can disable this line: https://github.com/PennyLaneAI/pennylane-lightning/blob/63929edacc85840b88208bfc0ef9dd75901e2cf6/.github/workflows/wheel_linux_x86_64.yml#L57

and re-add it after we pull down the wheel. Same as for Mac x86_64 and M1 (to verify OpenMP no longer fails).

setup.py
chaeyeunpark: Thank you for noting this. I just thought it was fine as Wheel::MacOS::ARM didn't work for PRs. Now a wheel builder for M1 also works without an error.

chaeyeunpark:
> It may also be worth enabling the CI wheel builders to export the built wheels for this PR (they are normally disabled outside master). As a quick workaround, we can remove this line:
> 
> https://github.com/PennyLaneAI/pennylane-lightning/blob/63929edacc85840b88208bfc0ef9dd75901e2cf6/.github/workflows/wheel_linux_x86_64.yml#L57
> 
> and re-add it after we pull down the wheel. Same as for Mac x86_64 and M1 (to verify OpenMP no longer fails).

See https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/1467603819 and https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/1467603816

mlxd:
Hi @chaeyeunpark I had time to compare the wheels on Python3.9 for Linux. The performance is pretty much the same for both (nice work on that). I did notice that the size of the wheels, as well as the exported symbols, differ between the current version and this version. I think for now to minimize the differences we can add some additional compile-time flags that are turned on by-default for the Linux/Mac setuptools versions: 

```
## Allows overflows of integer ops; should never be a problem and can potentially add some improvements
-fwrapv

## Allows more optimal external library calls
-fno-plt

## Avoid intermediate files hitting the disk; speeds up compilation in some cases
-pipe
```

I think this should be enough to get an almost apples-to-apples binary comparison. With the above changes, and disabling the wheels being stored, I am happy to see this being merged in.

I suspect the CodeCov issue will resolve itself too with another commit. If not, we can see about whether it makes sense to add another test, or just override its complaint.

chaeyeunpark:
Hi @mlxd, I made those changes https://github.com/PennyLaneAI/pennylane-lightning/pull/176/files#diff-f8588b38421f6ebc8fdb5eaaca2d2a160b4e38a5092a2ffa045f7cd5075823eaR10-R17. However, tests still fail (probably because of the new Pennylane features). We may merge this after https://github.com/PennyLaneAI/pennylane-lightning/pull/179 is merged.

mlxd:
> Hi @mlxd, I made those changes https://github.com/PennyLaneAI/pennylane-lightning/pull/176/files#diff-f8588b38421f6ebc8fdb5eaaca2d2a160b4e38a5092a2ffa045f7cd5075823eaR10-R17. However, tests still fail (probably because of the new Pennylane features). We may merge this after #179 is merged.

Awesome work! Thanks @chaeyeunpark

maliasadi:
> We may merge this after #179 is merged.

Nice work @chaeyeunpark! PR #179 is also merged.

.github/workflows/wheel_noarch.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         with:
           python-version: '3.7'
 
+      - name: Get latest CMake and ninja
+        uses: lukka/get-cmake@latest
```
mlxd, line 23: Is there any reason to prefer this over a pip installation of both cmake and ninja?

pennylane_lightning/lightning_qubit.py
```diff
@@ -32,12 +33,24 @@
...
-    from .lightning_qubit_ops import (
-        apply,
-        StateVectorC64,
-        StateVectorC128,
-        AdjointJacobianC128,
-    )
+    if platform.system() == "Windows" and sys.version_info[:2] >= (3, 8):
+        # Add the current directory to DLL path.
+        # See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        os.add_dll_directory(os.path.dirname(os.path.abspath(__file__)))
```
mlxd, line 39: Maybe we can add a Pylint ignore comment here to ignore the warnings. Or, should we add tests as recommended?

mlxd (commented):
Hey Chae-Yeun, I added tow more comments. Happy to merge after these.

mlxd (approved):

PR #179, opened 2021-11-19
Update PL-Lightning to support new features in PL

maliasadi:
**Context:**
Current PL-Lightning is not compatible with new features in PL and as the result almost all `gates` and `adjoint_jacobian` tests are FAILED. This PR updates the PL-lightning along with all tests to support new features in PL. 

These files are mainly updated: 
- `lightning_qubit.py`
- `tests/test_gates.py`
- `tests/test_adjoint_jacobian.py`

**Benefits:**
PL-Lightning is compatible with new features in PL and there shouldn't be any failed tests 

**Related GitHub Issues:**
n/a

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -18,6 +18,7 @@ interfaces with C++ for fast linear algebra calculations.
 from warnings import warn
 
 import numpy as np
+from numpy.core.fromnumeric import transpose
 from pennylane import (
     BasisState,
     DeviceError,
@@ -220,11 +221,12 @@ class LightningQubit(DefaultQubit):
 
         ops_serialized = adj.create_ops_list(*ops_serialized)
 
-        tape_trainable_params_set = set(tape.trainable_params)
+        # tape_trainable_params_set = set(tape.trainable_params)
+        trainable_params = sorted(tape.trainable_params)
+        first_elem = 1 if trainable_params[0] == 0 else 0
+
         tp_shift = (
-            tape_trainable_params_set
-            if not use_sp
-            else {i - 1 for i in tape_trainable_params_set.difference({0})}
+            trainable_params if not use_sp else [i - 1 for i in trainable_params[first_elem:]]
         )  # exclude first index if explicitly setting sv
 
         jac = adj.adjoint_jacobian(
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -18,7 +18,6 @@ interfaces with C++ for fast linear algebra calculations.
 from warnings import warn
 
 import numpy as np
-from numpy.core.fromnumeric import transpose
 from pennylane import (
     BasisState,
     DeviceError,
@@ -221,7 +220,6 @@ class LightningQubit(DefaultQubit):
 
         ops_serialized = adj.create_ops_list(*ops_serialized)
 
-        # tape_trainable_params_set = set(tape.trainable_params)
         trainable_params = sorted(tape.trainable_params)
         first_elem = 1 if trainable_params[0] == 0 else 0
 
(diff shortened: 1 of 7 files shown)


pennylane_lightning/lightning_qubit.py
```diff
@@ -220,10 +220,11 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
 
         ops_serialized = adj.create_ops_list(*ops_serialized)
 
+        tape_trainable_params_set = set(tape.trainable_params)
```
josh146, line 223: @maliasadi just something to keep in mind is that the reason we changed `trainable_params` from being a set is that we were affected by a rare python bug where set order of integers is not guaranteed to be stable!

So something to be careful with if you plan to iterate over a set, and expect the integers to be ordered.

mlxd: I think it should be fine --- we don't really care about order as much as the existence of indices. We iterate over them in the order given to CPP, and assume no guarantees on being sorted (I think).

maliasadi: Thank you @josh146 @mlxd! I update `AdjointDiff` method to work with `trainable_params` as a list.

tests/test_adjoint_jacobian.py
```diff
@@ -250,18 +252,29 @@ def test_multiple_rx_gradient(self, tol):
...
-        args = np.linspace(0.2, 0.5, op.num_params)
 
+        # op.num_wires and op.num_params must be initialized a priori
         with qml.tape.JacobianTape() as tape:
             qml.Hadamard(wires=0)
             qml.RX(0.543, wires=0)
             qml.CNOT(wires=[0, 1])
 
-            op(*args, wires=range(op.num_wires))
+            op
```
mlxd, line 277: Just curious if the operation is being applied here with this? Or is this a GH formatting issue that I cannot see the rest of the changes?

tests/test_adjoint_jacobian.py
```diff
@@ -462,15 +475,13 @@ def circuit(params):
             qml.Rot(params[1], params[0], 2 * params[0], wires=[0])
             return qml.expval(qml.PauliX(0))
 
-        spy_numeric = mocker.spy(qml.tape.JacobianTape, "numeric_pd")
```
mlxd, line 465: Should we be checking this?

tests/test_adjoint_jacobian.py
```diff
@@ -462,15 +475,13 @@ def circuit(params):
...
         spy_analytic = mocker.spy(dev, "adjoint_jacobian")
 
         cost = QNode(circuit, dev, diff_method="finite-diff")
 
         grad_fn = qml.grad(cost)
         grad_F = grad_fn(params)
 
-        spy_numeric.assert_called()
```
mlxd, line 473: And here?

mlxd (commented):
Nice one Ali. Just a few small questions

tests/test_adjoint_jacobian.py
maliasadi: Previously, `op` was defined as a list of operations' inits and this test function worked properly as values of `op.num_params` and `op.num_wires` were known for each operation. Now, with the recent updates, these values are not known and one must pass arguments a priori. Hence, each operation is being applied here with each `op` call without any needs on passing args and wires.

tests/test_adjoint_jacobian.py
maliasadi: It seems this is not required anymore as they are also deprecated in PL test-suite.

josh146: Yep, `JacobianTape` (and all the other tape subclasses!) are being deprecated in favour of the functions in the `qml.gradients` subpackage.

josh146: Note that these checks _could_ be updated here, _but_ I don't think it makes sense to unittest PL core functionality in the plugin test.

mlxd (approved):
:100: Great job @maliasadi

PR #181, opened 2021-11-25
Add VJP support to PL-Lightning

maliasadi:
**Context:**
Implement and Integrate Vector Jacobian Product (VJP) methods in PL-Lightning along with adding tests following PL implementation [here](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/gradients/vjp.py) and using `pybind11`.

Add `vecMatrixProd` to `Util.hpp` and  class `VectorJacobianProduct` to `AdjointDiff.cpp` with two public methods: 
- `tensorDot`: Calculates tensor-dot of a given vector of gradient outputs and a Jacobian.
- `vectorJacobianProduct`: Calculates the vjp for the selected set of parametric gates utilizing the `AdjointJacobian` method. 

These methods are fully tested at `Test_VectorJacobianProduct.cpp`. 

Python bindings are added to `Bindings.cpp` and `lightning.qubit` is now offering two methods to compute vjps:
- `vector_jacobian_product`: Generate the the vector-Jacobian products of a tape.
- `batch_vjp`: Generate the the vector-Jacobian products of a batch of tapes.

These methods are fully tested at `test_vector_jacobian_product`.


pennylane_lightning/src/algorithms/AdjointDiff.cpp
```diff
@@ -18,6 +18,9 @@
 template class Pennylane::Algorithms::AdjointJacobian<float>;
 template class Pennylane::Algorithms::AdjointJacobian<double>;
 
+template class Pennylane::Algorithms::VectorJacobianProduct<float>;
+template class Pennylane::Algorithms::VectorJacobianProduct<double>;
```
mlxd, line 22: Do you think this should be included in the Adjoint file, or separated into its own file in the algorithms directory?

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -32,13 +32,14 @@ include(Catch)
...
-                                Test_Bindings.cpp
+target_sources(runner PRIVATE   Test_Util.cpp
+                                Test_AdjDiff.cpp
+                                Test_VectorJacobianProduct.cpp
                                 Test_StateVector_Nonparam.cpp 
                                 Test_StateVector_Param.cpp 
                                 Test_StateVectorManaged_Nonparam.cpp 
                                 Test_StateVectorManaged_Param.cpp 
-                                Test_Util.cpp
+                                Test_Bindings.cpp
```
mlxd, line 42: Just curious about the reordering of these? If ordering doesn't matter, I think keeping alphabetical may be a good idea.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -502,6 +502,132 @@ inline auto matrixVecProd(const std::vector<std::complex<T>> mat,
     return v_out;
 }
 
+/**
+ * @brief Calculates transpose of a matrix recursively and Cache-Friendly
+ * using blacking and Cache-optimized techniques.
+ */
+template <class T, size_t BLOCKSIZE = 32> // NOLINT(readability-magic-numbers)
```
mlxd, line 509: Is it possible to add `@param` docstring elements for each argument?

mlxd (changes requested):
Phenomenal work @maliasadi !
I had a few small comments. Also, just wondering if we can add tests that trigger the conditions CodeFactor/CodeCov complain about?

josh146 (changes requested):
Thanks @maliasadi! This is great, I've left some comments throughout. 

Correct me if I am wrong, but currently the built-in Jacobian method only supports gradients technically, since only tapes with a single, scalar, output is supported?

As we iterate the gradient support in PL and lightning, I think it makes sense to 'decouple' the VJP and Jacobian methods at some point, so that a full Jacobian is not reconstructed when the VJP is requested

pennylane_lightning/src/algorithms/AdjointDiff.cpp
maliasadi: it's definitely better to be added into its own file namely `JacobianProd.cpp` in the same directory.

pennylane_lightning/src/tests/CMakeLists.txt
maliasadi: Nice catch! It wasn't intentional :slightly_smiling_face:

pennylane_lightning/src/util/Util.hpp
maliasadi: Added! Please let me know if they are not coherent.

pennylane_lightning/lightning_qubit.py
```diff
@@ -248,6 +251,240 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
+            tape (.QuantumTape): quantum tape to differentiate
+            dy (tensor_like): Gradient-output vector. Must have shape
+                matching the output shape of the corresponding tape.
+            num (int): The length of the flattened ``dy`` argument. This is an
+                optional argument, but can be useful to provide if ``dy`` potentially
+                has no shape (for example, due to tracing or just-in-time compilation).
+            starting_state (array[int]): Starting state indeces.
+            use_device_state (bool): Use device if `True` in case of `starting_state=None`.  
+            vjp_pybind (bool): Use the `VectorJacobianProduct` class in `lighting.qubit` if `True`, 
+                otherwise `adjoint_jacobian` will be called to compute vjp.
```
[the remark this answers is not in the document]
maliasadi: My bad! `vjp_pybind` was only for testing & debugging. This is no longer needed. I re-write this part.

pennylane_lightning/lightning_qubit.py
```diff
@@ -248,6 +251,240 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
+        # Initialization of state
+        if starting_state is not None:
+            ket = np.ravel(starting_state)
+        else:
+            if not use_device_state:
+                self.reset()
+                self.execute(tape)
+            ket = np.ravel(self._pre_rotated_state)
+
+        # adj = AdjointJacobianC128()
```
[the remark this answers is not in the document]
maliasadi: Good catch! It was only used for debugging the code :slightly_smiling_face:

pennylane_lightning/lightning_qubit.py
```diff
@@ -248,6 +251,240 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
+                If ``append``, then the output of the function will be of the form
+                ``List[tensor_like]``, with each element corresponding to the VJP of each
+                input tape. If ``extend``, then the output VJPs will be concatenated.
+            starting_state (array[int]): Starting state indeces.
+            use_device_state (bool): Use device if `True` in case of `starting_state=None`.  
+            vjp_pybind (bool): Use the `VectorJacobianProduct` class in `lighting.qubit` if `True`, 
+                otherwise `adjoint_jacobian` will be called to compute vjp. 
+        Returns:
+            List[tensor_like or None]: list of vector-Jacobian products. ``None`` elements corresponds
+            to tapes with no trainable parameters.
```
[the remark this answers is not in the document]
maliasadi: My bad! I forgot to check the rendered doc. I re-wrote this part.

pennylane_lightning/lightning_qubit.py
```diff
@@ -248,6 +251,240 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
+                return math.convert_like(np.zeros([num_params]), dy)
+        except (AttributeError, TypeError):
+            pass
+
+        if not vjp_pybind:
+            jac = self.adjoint_jacobian(
+                tape, starting_state=starting_state, use_device_state=use_device_state
+            )
+
+            return self._compute_vjp_tensordot(dy, jac, num=num)
```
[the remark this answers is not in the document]
maliasadi: As you mentioned, this is basically the `compute_vjp` method in `qml.gradients`. I wrote this as part of the first draft to help me writing `tensorDot` in `Algorithms::Pennylane::VectorJacobianProduct`. This is now removed from the lightning.

maliasadi:
>

maliasadi:
> just wondering if we can add tests that trigger the conditions CodeFactor/CodeCov complain about?

Thank you @mlxd! I am not sure how to fix CodeFactor complain at `pennylane_lightning/lightning_qubit.py#L254-L356`. Do you have any suggestions?

pennylane_lightning/lightning_qubit.py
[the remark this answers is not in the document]
josh146: got it!

maliasadi:
Thank you @josh146! :slightly_smiling_face: 

> As we iterate the gradient support in PL and lightning, I think it makes sense to 'decouple' the VJP and Jacobian methods at some point, so that a full Jacobian is not reconstructed when the VJP is requested

You're right! However, one can use `compute_vjp` to compute VJP from a pre-calculated Jacobian. I hope this addresses the issue with reconstructing Jacobian in the `vector_jacobian_product` method. Moreover, I updated `vector_jacobian_product` and `batch_vjp` to also return the `adjoint_jacobian` results. Please let me know what you think about these new changes.

mlxd (approved):
Great job putting this together @maliasadi 

As discussed, I think we can safely merge this and build upon it later for the optimal implementation.

josh146 (approved):
Agree with Lee, great work @maliasadi!

PR #185, opened 2021-12-03
Direct Support to probabilities, expectation values and variances in PL-Lightning

AmintorDusko:

**Context:**
Pennylane Lightning doesn't offer direct support for the calculation of probabilities, expectation values and variances.
This PR introduces a C++ class template `Measures` that implements such calculations.
Calculations can be performed in terms of the already defined gates, or with any operator provided as a matrix.
**Benefits:**
A direct implementation in C++ expands the optimization options, and this class allows to perform such calculations for any operator provided in matrix form.
**Possible Drawbacks:**
I'm not aware of.
**Related GitHub Issues:**


pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+using namespace Pennylane;
+StateVector<double> Initializing_StateVector() {
+    // Defining a StateVector in a non-trivial configuration:
+    size_t num_qubits = 3;
+    size_t data_size = std::pow(2, num_qubits);
+
+    std::complex<double> *arr;
+    arr = new complex<double>[data_size]();
+    arr[0] = 1;
```
mlxd, line 19: Use `StateVectorManaged` where possible, or ensure memory is allocated by `std::vector` so that it is freed upon destruction.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -714,8 +714,11 @@ template <class T>
 inline auto operator<<(std::ostream &os, const std::vector<T> &vec)
     -> std::ostream & {
     os << '[';
-    for (size_t i = 0; i < vec.size(); i++) {
-        os << vec[i] << ",";
+    if (!vec.empty()) {
+        for (size_t i = 0; i < vec.size() - 1; i++) {
+            os << vec[i] << ", ";
+        }
+        os << vec.back();
```
mlxd, line 721: Good call

pennylane_lightning/src/util/Util.hpp
```diff
@@ -768,7 +771,7 @@ class NotImplementedException : public std::logic_error {
     /**
      * @brief Construct a NotImplementedException exception object.
      *
-     * @param fname Function name to indicate not imeplemented.
+     * @param fname Function name to indicate not implemented.
```
mlxd, line 774: :100:

mlxd (changes requested):
Thanks for putting this together @AmintorDusko
I have some comments and questions:

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * @return Floating point std::vector with probabilities
+     * in lexicographic order according to wires.
+     */
+    vector<fp_t> probs() {
+        complex<fp_t> *arr_data = measured_statevector.getData();
+        vector<double> basis_probs(measured_statevector.getLength(), 0);
+        for (size_t ind = 0; ind < 8; ind++) {
+            basis_probs[ind] = std::pow(std::abs(arr_data[ind]), 2.0);
+        }
```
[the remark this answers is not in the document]
AmintorDusko: The 8 here is wrong, I will fix that.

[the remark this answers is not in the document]
AmintorDusko: The `basis_prob` can be a member, or maybe calculated on-the-fly for the Probabilities for a subset of the full system.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -0,0 +1,218 @@
...
+        size_t data_size = measured_statevector.getLength();
+        // Copying the original state vector, for the application of the
+        // observable operator.
+        complex<double> *original_data;
+        original_data = new complex<double>[data_size];
+        std::copy(measured_statevector.getData(),
+                  measured_statevector.getData() + data_size, original_data);
+        StateVector<double> operator_statevector(original_data, data_size);
+
+        measured_statevector.applyOperation(operation, wires);
```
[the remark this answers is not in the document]
AmintorDusko: This was better before I performed some last minutes change. Sorry about that, I will fix the notation.

pennylane_lightning/src/tests/Test_Measures.cpp
AmintorDusko: Nice, I will give a look into the `StateVectorManaged`, to better understand it use.

AmintorDusko:
I will work on every request and suggestion. Thanks for your feedback @mlxd .

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * This class performs measurements in the state vector provided to its
+ * constructor. Observables are defined by its operator(matrix) or through a
+ * string-based function dispatch.
+ *
+ * @tparam fp_t Floating point precision of underlying measurements.
+ */
+template <class fp_t = double> class Measures {
+  private:
+    StateVector<fp_t> measured_statevector;
```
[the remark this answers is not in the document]
AmintorDusko: Hi Lee. It was not my intention to create a copy and store the `StateVector` we are measuring, but only to keep a pointer to it. In this way one can keep operating with the `StateVector` and measure again later. 
I'm creating a copy for the measurement of observables. This copy is the one I apply the operators to, and this one I'm discarding after each measure to keep the results consistent with the way is done (with probabilities) in PL. 
Please, let me know your thoughts about that.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * @brief Observable's Measurement Class.
+ *
+ * This class performs measurements in the state vector provided to its
+ * constructor. Observables are defined by its operator(matrix) or through a
+ * string-based function dispatch.
+ *
+ * @tparam fp_t Floating point precision of underlying measurements.
+ */
+template <class fp_t = double> class Measures {
```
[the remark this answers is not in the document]
AmintorDusko: Excelent idea! Its done!

pennylane_lightning/src/simulator/Measures.hpp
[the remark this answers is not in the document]
AmintorDusko:
```
std::transform(arr_data, arr_data+measured_statevector.getLength(), basis_probs.begin(),
                   [](const std::complex<fp_t> &z) -> fp_t { return std::norm(z); } );
```
Beautiful suggestion. Done!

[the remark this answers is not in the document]
AmintorDusko: Regarding making the basis_prob a class member, I gave some think on this and I don't think it is safe to be done. The StateVector we are measuring is not affect by measure by now, but this can change in the future. If we store the basis_probability for future reference, when we come to use it, the data can be out-dated. I mean, it is easy to be implemented, though.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -0,0 +1,218 @@
...
+            measured_statevector.generateBitPatterns(wires);
+        vector<size_t> all_offsets = measured_statevector.generateBitPatterns(
+            measured_statevector.getIndicesAfterExclusion(wires));
+
+        vector<fp_t> probabilities(all_indices.size(), 0);
+
+        size_t ind_probs = 0;
+        for (auto index : all_indices) {
+            for (auto offset : all_offsets) {
+                probabilities[ind_probs] += basis_probs[index + offset];
```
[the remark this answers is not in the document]
AmintorDusko: The [documentation](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.probs.html) says that the probabilities are provided in lexicographic order, that seems to be the order of the computational basis.

AmintorDusko:
Please let me know your thoughts.

AmintorDusko:
Hi @mlxd. Following our discussion, I updated the probabilities calculations.
Please let me know your thoughts.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -915,6 +915,80 @@ auto linspace(T start, T end, size_t num_points) -> std::vector<T> {
...
+ * @brief Determines a new index for a transposed tensor stored linearly.
+ *  This function assumes each axis will have a length of 2 (|0>, |1>).
+ *
+ * @param ind index before.
+ * @param new_axes new axes distribution.
+ * @return unsigned int with the new transposed index.
+ */
+inline auto transposed_state_index (  size_t ind, 
+                                        const std::vector<size_t> &new_axes)
+                                    -> size_t {
```
mlxd: Did clang-format choose this ordering? Looks a little odd on Github.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -915,6 +915,80 @@ auto linspace(T start, T end, size_t num_points) -> std::vector<T> {
...
+    size_t new_index=0;
+    for (std::vector<size_t>::const_reverse_iterator axis = new_axes.rbegin();
```
mlxd:
```suggestion
    for (auto axis = new_axes.rbegin(); 
```

Using `auto` is much cleaner here.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -915,6 +915,80 @@ auto linspace(T start, T end, size_t num_points) -> std::vector<T> {
...
+ * @param ind index before.
+ * @param new_axes new axes distribution.
+ * @return unsigned int with the new transposed index.
+ */
+inline auto transposed_state_index (  size_t ind, 
+                                        const std::vector<size_t> &new_axes)
+                                    -> size_t {
+    size_t new_index=0;
+    for (std::vector<size_t>::const_reverse_iterator axis = new_axes.rbegin(); 
+            axis != new_axes.rend(); ++axis){
```
mlxd:
```suggestion
            axis != new_axes.rend(); axis++){
```

No reason to use pre-increment here.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -915,6 +915,80 @@ auto linspace(T start, T end, size_t num_points) -> std::vector<T> {
...
+        new_index+=(ind%2)*pow(2,(new_axes.size()-1)-(*axis));
```
mlxd:
```suggestion
        new_index += ((ind%2) << ((new_axes.size()-1)-(*axis)));
```

Powers of 2 can be faster (usually) using l-shift operations. I think this should be the same (but please check it for me).

pennylane_lightning/src/util/Util.hpp
```diff
@@ -915,6 +915,80 @@ auto linspace(T start, T end, size_t num_points) -> std::vector<T> {
...
+auto transpose_state_tensor( 
+                            const std::vector<T> &tensor, 
+                            const std::vector<size_t> &new_axes)
+                            -> std::vector<T> {
+    std::vector<T> transposed_tensor(tensor.size());
+    for (size_t ind=0; ind<tensor.size(); ind++){
+        transposed_tensor[ind]=tensor[transposed_state_index(ind, new_axes)];
+    }
+  return transposed_tensor;
+}
```
mlxd: I worry this may be slow for large data sets. Can you run some numbers on this and see how it performs for various numbers of qubits.

mlxd (changes requested):
Thanks for this @AmintorDusko.
I have left some comments and queries.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -61,14 +61,15 @@ class Measures {
      * @brief Probabilities of each computational basis state.
      *
      * @return Floating point std::vector with probabilities
-     * in lexicographic order according to wires.
+     * in lexicographic order.
      */
     vector<fp_t> probs() {
         const complex<fp_t> *arr_data = original_statevector.getData();
         vector<double> basis_probs(original_statevector.getLength(), 0);
```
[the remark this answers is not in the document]
AmintorDusko: Good catch!

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -80,29 +81,39 @@ class Measures {
...
+     * The basis columns are rearranged according to wires.
      */
     vector<fp_t> probs(const vector<size_t> &wires) {
-        const complex<fp_t> *arr_data = original_statevector.getData();
-
+        //Determining index that would sort the vector.
+        //This information is needed later.
+        vector<size_t> sorted_ind_wires=sorting_indexes(wires);
+        //Sorting wires.
+        vector<size_t> sorted_wires(wires.size());
```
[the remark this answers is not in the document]
AmintorDusko: Yes. I was thinking about that. It will probably need some more steps, but I think can be done. I will work on that.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -80,29 +81,39 @@ class Measures {
...
-            original_statevector.getIndicesAfterExclusion(wires));
+            original_statevector.getIndicesAfterExclusion(sorted_wires));
 
         vector<fp_t> probabilities(all_indices.size(), 0);
 
         size_t ind_probs = 0;
         for (auto index : all_indices) {
             for (auto offset : all_offsets) {
-                probabilities[ind_probs] += std::norm(arr_data[index + offset]);
+                probabilities[ind_probs] += std::pow(std::abs(arr_data[index + offset]), 2.0);
```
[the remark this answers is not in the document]
AmintorDusko: norm works perfectly. I thought I had changed that.
I will double check the code.

pennylane_lightning/src/util/Util.hpp
AmintorDusko: I will double check.

pennylane_lightning/src/util/Util.hpp
AmintorDusko: I will. Nice suggestion. Thanks.

pennylane_lightning/src/util/Util.hpp
AmintorDusko: Yes. Sure.

AmintorDusko: I explored some alternative strategies for this transposition, and at least for now, this seems to be the more efficient. In the other hand, It doesn't seem to be too slow.
I checked the performance for 100 runs, up to 22 qubits.
I applied the `transpose_state_tensor` function to vectors with the size of the equivalent state vector (or tensor). This shows exactly the information we are interested in. For 22 qubits, we have a vector with 8,388,608 elements, and the mean time is 0.89 seconds.

image

A loop like that, with no data dependency, is quite the low-hanging fruit for openMP parallelisation, but for the sizes I'm presenting here quite of an overkill. If you think we should proceed with some parallelisation strategy, I think we could impose a threshold based in the number of qubits.

AmintorDusko:
Hi @mlxd. Let me know what you think.

pennylane_lightning/src/util/Util.hpp
mlxd: Thanks Amintor. I do think we should ensure this is fast  for <=30 qubits (for large node usage). We can add parallelism as a next pass though.

AmintorDusko: Hi @mlxd ,
I extended the benchmark to 26 qubits (memory is limiting this range).
I optimised the code to determine the transposed index, and the function to transpose the tensor, accordingly.
The best method so far is still based in transpose the tensor after the calculation of the probabilities.
Anyway, in this new approach for the tensor transposition, I could observe a huge increase in speed, for larger number of qubits. I'm comparing the two methods here:
download
I also plotted the ration between the two times:
download

For 26 qubits, the previous algorithm is taking 18.4 seconds, while the new one is taking 2.3 seconds. The new code is approximately 8 times faster. This qubit state vector has 67,108,864 elements. I'm performing calculation with double precision, and compiling with the -Ofast flag.

We can observe an outlier in 12 qubits, where the previous algorithm seems to outperform the new one. The ration here is 0.85, and this may be because of the number of samples (100). Also, we are talking about small number and this could be strongly affected by fluctuations.

AmintorDusko: In other run, the 12 qubits performed better in the new algorithm. I think it is reasonable to assume that the measure of times in this fast cases is specially prone to fluctuations. It seems that starting in 20 qubits would be a good choice.

pennylane_lightning/src/util/Util.hpp
```diff
@@ -981,7 +981,9 @@ auto transpose_state_tensor(const std::vector<T> &tensor,
     -> std::vector<T> {
     std::vector<T> transposed_tensor(tensor.size());
     for (size_t ind = 0; ind < tensor.size(); ind++) {
-        transposed_tensor[ind] = tensor[transposed_state_index(ind, new_axes)];
+        // transposed_tensor[ind] = tensor[transposed_state_index(ind,
+        // new_axes)];
```
mlxd:
```suggestion
```

mlxd:
Thanks Amintor. I think I am happy to merge this in as the initial pass. We can add the Python bindings in a follow up PR.
Also, I think it best to never use `-Ofast` and favour use of `-O2` or `-O3` instead. We never intend to ship anything with `-Ofast` so we should ensure all numbers taken do not use it.

mlxd (approved):
After the change listed here, I think this is good to go.

pennylane_lightning/src/util/Util.hpp
AmintorDusko: Nice catch! Thanks!

AmintorDusko:
> Thanks Amintor. I think I am happy to merge this in as the initial pass. We can add the Python bindings in a follow up PR. Also, I think it best to never use `-Ofast` and favour use of `-O2` or `-O3` instead. We never intend to ship anything with `-Ofast` so we should ensure all numbers taken do not use it.

Makes sense. Thanks!

Issue #152, opened 2021-10-06
Add wheels for Python 3.10


mlxd:
Given the new release of Python 3.10, Lightning will require updates to the wheelbuilder to support it.

PR #186, opened 2021-12-06
Add Python 3.10 CI builder and updated PL dep

mlxd:
**Context:** Closes https://github.com/PennyLaneAI/pennylane-lightning/issues/152

**Description of the Change:** Adds wheelbuilder support for Python 3.10, and updates the required PL version.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


setup.py
```diff
@@ -106,7 +106,7 @@ def build_extension(self, ext: CMakeExtension):
 requirements = [
     "ninja", 
     "numpy",
-    "pennylane>=0.15",
+    "pennylane>=0.19",
 ]
```
antalszava, line 110: Hi @mlxd, just a fly-by comment: once Lightning supports Python 3.10, this can also be marked on a project level. To do that, just an extra line is required in this file on line 151, to show the Python 3.10 compatibility.

mlxd: Great, thanks for the tip @antal !

.github/CHANGELOG.md
```diff
@@ -76,6 +76,9 @@ Lee J. O'Riordan
 
 ### New features since last release
 
+* Add wheel-builder support for Python 3.10.
+[(#186)](https://github.com/PennyLaneAI/pennylane-lightning/pull/186)
```
antalszava, line 80:
```suggestion
* Add wheel-builder support for Python 3.10.
  [(#186)](https://github.com/PennyLaneAI/pennylane-lightning/pull/186)
```

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -7,14 +7,15 @@ on:
     types: [published]
 
 env:
-  CIBW_BUILD: 'cp37-* cp38-* cp39-*'
+  CIBW_BUILD: 'cp37-* cp38-* cp39-* cp310-*'
+  CIBW_SKIP: "*-musllinux*"
```
antalszava, line 11: For my understanding: what is being skipped here and for what reason?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -7,14 +7,15 @@ on:
...
 
   # Python build settings
   CIBW_BEFORE_BUILD: |
-    pip install pybind11 ninja && yum install -y cmake gcc gcc-c++
+    pip install pybind11 ninja
 
   # Testing of built wheels
-  CIBW_TEST_REQUIRES: numpy==1.19.5 scipy pytest pytest-cov pytest-mock flaky
+  CIBW_TEST_REQUIRES: numpy~=1.21 scipy pytest pytest-cov pytest-mock flaky
```
antalszava, line 18: How come the NumPy version was updated?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -7,14 +7,15 @@ on:
...
 
 env:
-  CIBW_BUILD: 'cp37-* cp38-* cp39-*'
+  CIBW_BUILD: 'cp37-* cp38-* cp39-* cp310-*'
+  CIBW_SKIP: "*-musllinux*"
 
   # Python build settings
   CIBW_BEFORE_BUILD: |
-    pip install pybind11 ninja && yum install -y cmake gcc gcc-c++
+    pip install pybind11 ninja
```
antalszava, line 15: How come some installs were removed?

antalszava (approved):
Looks good :slightly_smiling_face: Just had some questions for my own understanding :+1:

.github/workflows/wheel_linux_aarch64.yml
mlxd: Version 1.19 has/had issues working with Py3.10. The previous choice for 1.19 was due to TF compatibility. With TF 2.7 now out in the wild, there is no more reason to pin to the older version (and I believe the PL repo is 1.21 in `requirements.txt` too https://github.com/PennyLaneAI/pennylane/blob/92e572f0cf8a0ba950ddd0e9440abfc1d9943c27/requirements.txt#L1).

.github/workflows/wheel_linux_aarch64.yml
mlxd: We skip attempting to build MUSL supported wheels (an alternative to the more common glibc linux distributions). This can be enabled a little later when MUSL support is more further along.

.github/workflows/wheel_linux_aarch64.yml
mlxd: GCC/G++ should already be available on these images, so no need to explicitly request anymore.

mlxd: In hindsight, I think it may be better to keep these explicitly in there. Thanks for noticing.

.github/workflows/wheel_linux_aarch64.yml
josh146: @antalszava Alpine Linux is typically used in Docker containers, so this is where MUSL compatibility is usually required

antalszava: Thank you!

Issue #192, opened 2021-12-14
[BUG] Compilation on Windows fails


BastianZim:
#### Issue description

Description of the issue - include code snippets and screenshots here
if relevant. You may use the following template below

* *Expected behavior:* (What you expect to happen)

Compilations succeeds.

* *Actual behavior:* (What actually happens)

Compilation fails on windows with `error C2039: 'all_of': is not a member of 'std'`

* *Reproduces how often:* (What percentage of the time does it reproduce?)

Every time 

* *System information:* (post the output of `import pennylane as qml; qml.about()`)

Azure pipelines

#### Source code and tracebacks

Please include any additional code snippets and error tracebacks related
to the issue here.

Logs: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=428105&view=logs&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=1183ba29-a0b5-5324-8463-2a49ace9e213&l=633

#### Additional information

Any additional information, configuration or data that might be necessary
to reproduce the issue.

This occurs in the context of building the package for conda-forge.

* PR: https://github.com/conda-forge/pennylane-lightning-feedstock/pull/3
* Script: `python -m pip install . -vv`
* Build file: https://github.com/conda-forge/pennylane-lightning-feedstock/blob/f455c7dad8dc571f24fd977eb661fbaf2cf1c209/recipe/meta.yaml

mlxd:
Hi @BastianZim thanks for reporting this.
I suspect this is either a C++17 flag issue for the compiler, or the platform doesn't pull in the header provided for the `std` method. I'll have somebody look at this.

BastianZim:
Thanks @mlxd Let me know if you need more info here from me.

mlxd:
Thanks @BastianZim . We did notice this may be for a Py3.10 build. Due to restrictions in Pennylane's other dependencies we do not currently support Python 3.10.

As for the observed error, we will issue a bugfix release shortly.

BastianZim:
Thanks @mlxd! Pyhton3.10 is not a huge concern, it's just currently passing on the other platforms which is why I had included it. I'll remove Windows for this build and merge it then.

mlxd:
Hi @BastianZim thanks again for reporting this. The problem you found will be fixed in the minor release [v0.20.1](https://github.com/PennyLaneAI/pennylane-lightning/releases/tag/v0.20.1)

BastianZim:
[...]

PR #193, opened 2021-12-14
All_Of Identifier Not Found

maliasadi:
**Context:**
Attempting to compile the `pennylane-lightning` with MSVC2022, `JacobianProd` is failed to complied. The compiler cites C3861 with the description that `error C3861: 'all_of': identifier not found`. This PR addresses this issue by adding the header `algorithm`.
 
**Related GitHub Issues:**
Compilation on Windows fails (#192)


mlxd:
I hard reset to the algorithm header commit. We do not want master to be merged here as it is to match up with the `v0.20.0` tagged release.

PR #201, opened 2021-12-21
Modify Release to RelWithDebInfo for built wheels

mlxd:
**Context:** Debugging with pypi supplied wheels is currently difficult due to the missing debug info. This PR ensure deub info is returned to the dynamic libraries by setting `Release` to `RelWithDebInfo`. This allows use of Lightning through debuggers, such as gdb, and lldb.

**Description of the Change:**  `Release` build type defaults to `RelWithDebInfo`.

**Benefits:** Allows debugging symbols to be included in distributed binaries.

**Possible Drawbacks:** Larger wheel sizes.

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Improvements
 
+* Ensure debug info is built into dynamic libraries.
+[(#201)](https://github.com/PennyLaneAI/pennylane-lightning/pull/201)
+
 ### Documentation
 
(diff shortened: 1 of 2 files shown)


maliasadi (approved):
Very nice improvement! Using `RelWithDebInfo` instead of `Release` increases the wheel size, but the code runs still fast and optimize, and has debug information included; thus a great deal to trace and fix bugs efficiently. Thanks @mlxd!

PR #203, opened 2022-01-05
Bugfix release V0.20.2

mlxd:
**Context:** This PR fixes a bug observed with the adjoint Jacobian method and a missing explicit kernel, where a decomposition was not performed. 

**Description of the Change:** An explicit kernel for the CY gate is added.

**Benefits:** Fixes an issue when using the CY gate as part of the IsingYY operator with the adjoint method.

**Possible Drawbacks:** None

**Related GitHub Issues:**


mlxd:
NOTE: do not update this branch against master!

maliasadi (approved):
It looks good. Thanks @mlxd!

PR #207, opened 2022-01-12
Force the "best" method to point to the "adjoint" method

trevor-vincent:
**Context:**

The "best" method in lightning currently defaults to backprop. This is due to the fact that the _get_best_method function in Pennylane searches for a jacobian function and if it doesn't find one, defaults first to backprop.

Create a jacobian function in lightning_qubit.py that points to adjoint_jacobian and add "provides_jacobian" to capabilities list. 

**Benefits:**

Now lightnings "best" method defaults to adjoint.

**Possible Drawbacks:**

Requires diff_method=parameter_shift to be explicitly specified in cases where the adjoint is not supported.


josh146:
thanks @trevor-vincent for picking this up, this is a really good catch! Should lead to users seeing a significant performance improvement when using lightning with PL 💪

josh146:
We'll need to think about how to safely 'fallback' to parameter-shift if there are operations/measurements in the circuit that are not supported by adjoint 🤔

trevor-vincent:
The only tests that fail are those in pennylane under test_measurements. These can easily be fixed by explicitly adding diff_method="parameter-shift", but do we want to go this route?

josh146:
> These can easily be fixed by explicitly adding diff_method="parameter-shift" in those pennylane tests, but do we want to go this route?

@trevor-vincent what do the tests look like? If they are common user workflows using the main PennyLane UI, probably not, since it means something that currently works fine for users will become broken unless they know to set a particular keyword argument.

trevor-vincent:
> > These can easily be fixed by explicitly adding diff_method="parameter-shift" in those pennylane tests, but do we want to go this route?
> 
> @trevor-vincent what do the tests look like? If they are common user workflows using the main PennyLane UI, probably not, since it means something that currently works fine for users will become broken unless they know to set a particular keyword argument.

Yeah I agree this would be too much, it would require diff_method=parameter-shift even with circuits that don't even require gradients. I think this is something that needs to be done in pennylane as you say, probably something for the core team.

PR #214, opened 2022-01-27
Add bindings to C++ expval, var, probs in Lightning

maliasadi:
**Context:**
Lightning optimally evaluates the expectation value, variance, and probability of a given observable

- [X] Add expval binding 
- [X] Add var binding
- [X] Add probs binding
- [x] Integrate with PL
- [x] Test bindings 

**Benchmark:**
pr214


pennylane_lightning/lightning_qubit.py
```diff
@@ -493,6 +497,153 @@ def batch_vjp(
...
+            shot_range (tuple[int]): 2-tuple of integers specifying the range of samples
+                to use. If not specified, all samples are used.
+            bin_size (int): Divides the shot range into bins of size ``bin_size``, and
+                returns the measurement statistic separately over each bin. If not
+                provided, the entire shot range is treated as a single bin.
+
+        Returns:
+            Expectation value of the observable
+        """
+        if isinstance(observable.name, List) or observable.name in [
```
maliasadi, line 561: Hi @mlxd. Do you think adding a full list of unsupported observables in lightning is a good idea?

mlxd: Hey @maliasadi I think it is better to have an opt in, rather than opt out strategy. Though, I am open to suggestions if you think it makes life easier for maintenance.

AmintorDusko (approved):
Hi @maliasadi . Nice work!

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -185,8 +185,7 @@ class Measures {
     /**
      * @brief Variance of an observable.
      *
-     * @param operation Square matrix in row-major order or string with the
-     * operator name.
```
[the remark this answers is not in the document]
maliasadi: It's fixed. thanks :smile:

chaeyeunpark:
Hi @maliasadi, it seems like the MSVC issue is due to the following line:
https://github.com/PennyLaneAI/pennylane-lightning/blob/4ca7227597a6b0b9fcbbc7d2f68a23f6345f86f8/pennylane_lightning/src/simulator/Measures.hpp#L33

As this affects all subsequent included headers, I think it is better to move this to the inside of each function definition (even though this is not much to do with this PR :slightly_smiling_face: ).

maliasadi:
> Hi @maliasadi, it seems like the MSVC issue is due to the following line:
> 
> https://github.com/PennyLaneAI/pennylane-lightning/blob/4ca7227597a6b0b9fcbbc7d2f68a23f6345f86f8/pennylane_lightning/src/simulator/Measures.hpp#L33

Very nice catch! Thanks @chaeyeunpark. I'll fix this anyway :slightly_smiling_face:

AmintorDusko:
Hi @chaeyeunpark, is it clear to you why this line is causing an issue with MSVC?
I would like to know more about that.

mlxd (approved):

PR #219, opened 2022-01-31
Update clang_format version

maliasadi:
**Context:**
This PR checks `clang-format-12` support instead of 11 (default) as if one tries to do `make format` using `clang-format-11` some of GH formatting checks would be failed. 

Update `DEFAULT_CLANG_FORMAT_VERSION`


mlxd (approved):
Nice one @maliasadi 
Glad it didn't set fire to any of the C++ with the version update!

PR #221, opened 2022-02-02
Ensure adjoint allocates memory for max concurrent observables only

mlxd:

**Context:** The current implementation of the adjoint Jacobian method in Lightning parallelizes over observables, with each observable handled by a separate OpenMP thread. This works fine in practice unless a large number of observables are required, wherein a new statevector memory block is allocated upfront for the total number of observables. This causes OOM errors for large numbers of observables, even for modest qubit counts. The solution is to restrict the requested number of observables into batches of OMP_NUM_THREADS concurrent executions, where the number of statevectors allocated is limited by the number of executing threads. This imposes a small overhead in requiring additional computation across batches, but ensures the user can reach much higher qubit counts for large numbers of concurrent executions.

**Description of the Change:** The adjoint Jacobian calculation is now batched at the maximum number of OpenMP threads, and allocates enough memory for a given batch only.

**Benefits:** Reduces overall memory footprint within a given series of concurrent calculations, allowing larger numbers of qubits in a given calculation, prevents OOM errors encountered for large workflows.

**Possible Drawbacks:** Imposes additional overheads due to repetition between batches.

**Related GitHub Issues:**


tests/test_comparison.py
```diff
@@ -67,21 +68,25 @@ def circuit():
             qml.BasisState(np.array(basis_state), wires=0)
             one_qubit_block(wires=0)
             return qml.expval(qml.PauliZ(0))
+        
+        dev_l = lightning_dev_version(wires)
+        dev_d = lightning_dev_version(wires)
```
AmintorDusko, line 79: Shouldn't the default qubit be called here?
I mean `dev_l` and `dev_d` seem to be the same.

mlxd: Ooops, copy-pasting issue. Thanks @AmintorDusko ! I'll fix for all of the following too

AmintorDusko (approved):
Nice work @mlxd! I'm happy to approve!

doc/devices.rst
```diff
@@ -66,3 +66,35 @@ Supported operations and observables
...
+.. code-block:: python
+
+    # Option 2: Before importing packages
+    import os
+    os.environ["OMP_NUM_THREADS"] = 4
+    import pennylane as qml
+
+Assuming you request multiple expectation values from a QNode, this should automatically parallelize the computation over the requested number of threads. You should ensure that the number of threads does not exceed the available physical cores on your machine.
+
+If you are computating a large number of expectation values, or if you are using a large number of wires on your device, it may be best to limit the number of expectation value calculations to at-most ``OMP_NUM_THREADS`` concurrent executions. This will help save memory, at the cost of additional compute time. To enable this, initialize a ``lightning.qubit`` device with the ``batch_obs=True`` keyword argument, as:
```
maliasadi, line 92:
```suggestion
If you are computing a large number of expectation values, or if you are using a large number of wires on your device, it may be best to limit the number of expectation value calculations to at-most ``OMP_NUM_THREADS`` concurrent executions. This will help save memory, at the cost of additional compute time. To enable this, initialize a ``lightning.qubit`` device with the ``batch_obs=True`` keyword argument, as:
```

doc/devices.rst
```diff
@@ -66,3 +66,35 @@ Supported operations and observables
...
+
+.. code-block:: bash
```
maliasadi, line 94:
```suggestion
.. code-block:: python
```

doc/devices.rst
```diff
@@ -66,3 +66,35 @@ Supported operations and observables
 .. raw:: html
 
     </div>
+
+
+**Parallel adjoint differentiation support:**
+
+The ``lightning.qubit`` device directly supports the `adjoint differentiation method <https://pennylane.ai/qml/demos/tutorial_adjoint_diff.html>`__, and enables parallelization over the requested observables (Linux/MacOS support only).
```
maliasadi, line 73: A silly question: I don't remember what the issue was with getting support for Windows?

doc/devices.rst
```diff
@@ -66,3 +66,35 @@ Supported operations and observables
 .. raw:: html
 
     </div>
+
+
+**Parallel adjoint differentiation support:**
```
maliasadi, line 71: I liked this new section so much! :+1:

pennylane_lightning/lightning_qubit.py
```diff
@@ -295,13 +304,31 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
-            obs_serialized,
-            ops_serialized,
-            tp_shift,
-            tape.num_params,
-        )
+        # If requested batching over observables, chunk into OMP_NUM_THREADS sized chunks.
+        # This will allow use of Lightning with adjoint for large-qubit numbers AND large
+        # numbers of observables, enabling choice between compute time and memory use.
+        if self._batch_obs:
+            requested_threads = int(os.getenv("OMP_NUM_THREADS", "1"))
```
maliasadi, line 311: Wouldn't be a good idea to get `requested_threads` before `if` and so check if `requested_threads==1` jump on `else` instead of the call to `_chunk_iterable`? :thinking:

pennylane_lightning/lightning_qubit.py
```diff
@@ -16,6 +16,8 @@
 interfaces with C++ for fast linear algebra calculations.
 """
 from warnings import warn
+import os
```
maliasadi, line 19: If I'm not mistaken the only use case is with `os.getenv` (like 311). So, 
```suggestion
from os import getenv
```

maliasadi (changes requested):
Awesome work! :rocket: I've had only a few comments...

doc/devices.rst
mlxd: Oops, I was less than careful here!

doc/devices.rst
mlxd: OpenMP isn't natively supported on Windows without linking directly against the MSVC runtime.

mlxd: Or something to that effect

pennylane_lightning/lightning_qubit.py
mlxd: Hmmm, good point. Something like:
```
if self._batch_obs and requested_threads > 1:
```
?

pennylane_lightning/lightning_qubit.py
mlxd: Yep, good point, and well spotted.

doc/devices.rst
maliasadi: Ah! Yessss this is the case :grimacing:

pennylane_lightning/lightning_qubit.py
maliasadi: Yes! something like this.

trevor-vincent (approved):
Nice work! Don't see any problem beyond what Ali pointed out.

maliasadi (approved):

PR #222, opened 2022-02-02
Update VJP and adjoint-jacobian methods

maliasadi:
**Context:**
This PR includes the updates for VJP and AdjDiff in a way that the AdjointJacobian method is more efficient (with reducing the cost of finding trainable params in the list of operations and updating jacobian matrix cache-efficiently) and VJP methods return *processing functions* required to compute the VJPs.  

- [X] Add `JacobianData`  to encapsulate the data required in AdjDiff and VJP methods. 
- [X] Update the implm. of `adjointJacobian` and add `adjointJacobianJD` with a new optimization.
- [X] Update the C++ implm. of `vectorJacobianProduct` using lambda functions.
- [X] Update the python bindings for `adjointJacobian` and VJP methods in `lightning.qubit`.
- [X] Update C++ and Python tests.

**Benchmark:**
The new adjoint-jacobian is up to 6% faster than the current implementation for large enough circuits.


chaeyeunpark (commented):
Hi Ali, nice work :+1:  I here just drop several comments.

pennylane_lightning/lightning_qubit.py
```diff
@@ -441,17 +353,97 @@ def compute_vjp(self, dy, jac, num=None):
         )
         return vjp_tensor
 
+    def vjp(self, tape, dy, starting_state=None, use_device_state=False):
```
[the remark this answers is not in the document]
maliasadi: Hi @chaeyeunpark. Thanks for the review! 

This is a very good question. Here, in lightning, I followed the implementation [here](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/gradients/vjp.py). The VJP method is not still called from PL. However, up to my understanding, the next step through using this in PL efficiently, is the work on getting adjoint working for vector-valued QNodes (PR #215). 

Actually, this PR addresses @josh146 comment on the fact that a desirable VJP method shouldn't call Jacobian itself that leads to saving for vector-valued QNodes and the optimization on `adjoint_jacobian` method. The other work (PR #215) is supposed to check the support for vector-valued QNodes and add the use of VJP from PL efficiently.

pennylane_lightning/src/algorithms/JacobianProd.hpp
```diff
@@ -122,48 +93,47 @@ class VectorJacobianProduct : public AdjointJacobian<T> {
...
+                    -> std::vector<T> { return std::vector<T>(num_params, 0); };
         }
 
-        const bool allzero =
-            std::all_of(dy.cbegin(), dy.cend(), [](T e) { return e == 0; });
-        if (allzero) {
-            vjp.resize(num_params);
-            return;
-        }
+        return [=](const JacobianData<T> &jd) -> std::vector<T> {
```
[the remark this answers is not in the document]
maliasadi: Tbh, I'm worried about this too :slightly_smiling_face: But I thought that as VJP methods are not supposed to call Jacobian. So, this should be true for the C++ VJP too.

pennylane_lightning/lightning_qubit.py
```diff
@@ -441,17 +353,97 @@ def compute_vjp(self, dy, jac, num=None):
...
+        elif dtype == np.complex128:
+            use_csingle = False
+        else:
+            raise TypeError(f"Unsupported complex Type: {dtype}")
+
+        if use_csingle:
+            V = VectorJacobianProductC64()
+        else:
+            V = VectorJacobianProductC128()
+
```
mlxd, line 400: Should we combine these two `if` statements, since `dtype` isn't used outside of this section?

pennylane_lightning/lightning_qubit.py
```diff
@@ -462,36 +454,36 @@ def batch_vjp(
...
+                if vjp is None:
+                    if reduction == "append":
+                        vjps.append(None)
+                    continue
+
+                if isinstance(reduction, str):
+                    getattr(vjps, reduction)(vjp)
+                elif callable(reduction):
+                    reduction(vjps, vjp)
```
mlxd, line 482: I'm a little unsure about the structure here. Can you elaborate a little on the intention of what is to happen?

pennylane_lightning/src/algorithms/AdjointDiff.cpp
```diff
@@ -16,9 +16,4 @@
 
 // explicit instantiation
 template class Pennylane::Algorithms::AdjointJacobian<float>;
-template class Pennylane::Algorithms::AdjointJacobian<double>;
-
-template class Pennylane::Algorithms::ObsDatum<float>;
-template class Pennylane::Algorithms::ObsDatum<double>;
-template class Pennylane::Algorithms::ObsDatum<std::complex<float>>;
-template class Pennylane::Algorithms::ObsDatum<std::complex<double>>;
```
mlxd, line 24: Are these causing issues with compilation? If so, we can declare there existence outside the complaining compilation unit. I think since they are directly defined here I'd like to keep them. Or, we can move them to a separate compilation unit and just link against them (though I understand this may require pulling many threads due to how the templates are currently structured). 

What do you think?

mlxd: Nevermind, I see now you moved everything to a new file.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -722,6 +499,114 @@ template <class T = double> class AdjointJacobian {
...
+     * The resulting row-major ordered `jac` matrix representation will be of
+     * size `jd.getSizeStateVec() * jd.getObservables().size()`. OpenMP is used
+     * to enable independent operations to be offloaded to threads.
+     *
+     * @param jac Preallocated vector for Jacobian data results.
+     * @param jd JacobianData represents the QuantumTape to differentiate
+     * @param apply_operations Indicate whether to apply operations to tape.psi
+     * prior to calculation.
+     */
+    void adjointJacobianJD(std::vector<T> &jac, const JacobianData<T> &jd,
```
mlxd, line 522: Rather the reimplement the internal structure, is it possible for this method to offload to the `adjointJacobian` with the requested datatypes from `jd`? 

That way we can avoid having the same algo implemented twice, and makes testing earier. What do you think?

pennylane_lightning/src/algorithms/JacobianProd.hpp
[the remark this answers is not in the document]
mlxd: Hmmm, this is an interesting one to me. As @chaeyeunpark said, the copy may be an issue. I think this is a situation where a more optimal solution may exist, but until we know how performance is hit by using this, we shouldn't worry if it expresses the right intent for now.
We can either try to craft a more custom solution later if it shows up as a bottleneck on compute or memory use.

mlxd (commented):
Great work on this @maliasadi 
I have some questions and thoughts. I reckon we can figure out the integration testing mentioned earlier and identify how to best examine the use with PL.

Maybe @josh146 would have an idea how to best do some runtime and integration testing with/against the Python-only implementation.

pennylane_lightning/lightning_qubit.py
maliasadi: :100: Fixed it!

pennylane_lightning/src/algorithms/AdjointDiff.hpp
maliasadi: Thanks @mlxd for your review! They are now two `adjointJacobianJD` and `adjointJacobian`... I only kept the other one for testing and comparing. If you are +1 with using a data-structure such as `JacobianData`, I'll replace `adjointJacobian` by `adjointJacobianJD`. The result of this PR must not include two implementations for adjoint-jacobian :slightly_smiling_face:

pennylane_lightning/lightning_qubit.py
maliasadi: I didn't get it when I saw this at first [here](https://github.com/PennyLaneAI/pennylane/blob/5a28983fc7bd950cde8a4014e54261fef4b54293/pennylane/gradients/vjp.py#L341). However, it seems the end user may pass the argument as `list.extend`, `"extend"` or `"append"` (there are tests covering this part [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/e4a21c51379cbe6350042a6fb94c5d8e615f5f31/tests/test_vjp.py#L576))

pennylane_lightning/src/algorithms/JacobianProd.hpp
```diff
@@ -122,48 +93,47 @@ class VectorJacobianProduct : public AdjointJacobian<T> {
...
+    auto vectorJacobianProduct(const std::vector<T> &dy, size_t num_params,
+                               bool apply_operations = false)
+        -> std::function<std::vector<T>(const JacobianData<T> &)> {
+        if (dy.empty() ||
+            std::all_of(dy.cbegin(), dy.cend(), [](T e) { return e == 0; })) {
+            // If the dy vector is zero, then the
+            // corresponding element of the VJP will be zero,
+            // and we can avoid unnecessary computation.
+            return
+                [&num_params]([[maybe_unused]] const JacobianData<T> &jd)
```
chaeyeunpark, line 115: Hi Ali, I also think this function returns a lambda with `num_params` (which is local variable) captured as a reference. As it is just `size_t`, I think you can change the capture to` [=]` or `[num_params = num_params]`.

pennylane_lightning/src/algorithms/JacobianProd.hpp
```diff
@@ -122,48 +93,47 @@ class VectorJacobianProduct : public AdjointJacobian<T> {
...
+            // Compute Jacobian for the input jd using `adjoint` method
+            this->adjointJacobianJD(jac, jd, apply_operations);
 
-        computeVJP(vjp, jac, dy);
+            // Compute VJP
+            computeVJP(vjp, jac, dy, jd.getNumObservables(), num_params);
+            return vjp;
+        };
     }
 }; // class VectorJacobianProduct
```
chaeyeunpark, line 137: I also found this class derives `AdjointJacobian<T>`. Is it necessary? If so, we need to add a virtual destructor in AdjointJacobian<T>.

pennylane_lightning/lightning_qubit.py
josh146: This particular detail in the PennyLane VJP implementation is a bit ugly, and exists only for one reason:

- Autograd and JAX like their VJPs appended to a list, resulting in a nested list
- TensorFlow and Torch like their VJPs extended into a flat list

Being able to specify how the reduction takes place just avoided us duplicating the function.

josh146: But if you don't need this here, I think you can avoid this

pennylane_lightning/lightning_qubit.py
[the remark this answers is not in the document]
josh146: There is no way to call this yet from PennyLane! But we plan to add it ASAP. The idea is that there will be

- `Device.execute([tapes])` which will be called on the forward pass, and
- `Device.vjp([tapes], dy)` which will be called on the backward pass

pennylane_lightning/lightning_qubit.py
maliasadi: Fair enough! Thanks for the clarification. As it is not harmful, I think we can keep it to be consistent with the PL VJP implementation.

pennylane_lightning/src/algorithms/JacobianProd.hpp
maliasadi: Good point! ... DON

pennylane_lightning/src/algorithms/JacobianProd.hpp
maliasadi: Great catch! This is part of my first design decision for this class and the need for `createOpsData` method from `AdjointJacobian<T>`. Both classes are now updated.

chaeyeunpark (approved):
Hi Ali, I just wanna approve this PR. Great job :+1:

mlxd (approved):
Excellent work on this @maliasadi 
Happy to approve too. Once the PL changes come we can do the integration.

PR #232, opened 2022-02-10
Fix extensibility for external packages

### LINKED ISSUE #230: Ambiguous function call for derived class
The following line causes issues with building binaries on derived types, due to an ambiguity in the function name.

https://github.com/PennyLaneAI/pennylane-lightning/blob/8b9d4127831a1bbe18b19fe5029c3bd2707b3a0b/pennylane_lightning/src/simulator/GateOperationsPI.hpp#L550

A quick fix would be to ensure all `IMAG` calls are scoped as `Util::IMAG`, and subsequently for each other `constexpr` function defined in this manner.

### LINKED ISSUE #231: Re-add `add_library(pennylane_lightning INTERFACE)` for derived packages
#### Issue description

The cmake interface to Lightning without the use of Pybind11 is beneficial for packages that depend on Lightning's structure. To ensure this can be supported, we should keep the `add_library(pennylane_lightning INTERFACE)` and subsequent lines in the CMakeLists.txt file for current development. This will be a temporary measure until a future refactor and integration.

mlxd:
Closes #230 and  #231

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,8 @@ add_subdirectory(pennylane_lightning/src)
 #####################################################
 # Maintain for dependent external package development
 #####################################################
+add_library(pennylane_lightning INTERFACE)
+
 target_link_libraries(pennylane_lightning INTERFACE     lightning_utils
                                                         lightning_simulator
                                                         lightning_algorithms)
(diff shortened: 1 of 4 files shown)


chaeyeunpark (approved):
:+1:

PR #234, opened 2022-02-12
Add Docker support

maliasadi:
**Context:**
Add Docker support to setup & test PL-Lightning.

- Add Dockerfile 
- Update README
- Update `cmake/process_options.cmake`
- Update `setup.py`


mlxd (commented):
Hey @maliasadi thanks for putting this together.
Just curious, how does one expect to use the Docker version of lightning?

Will this be strictly used for building wheels, building and running the test-suite, or having a fully capable Python environment with lightning installed from the given branch?

mlxd (approved):
Thanks @maliasadi . I think it will be useful for local testing and dev also. 

I'd suggest one change, placing the `Dockerfile` into a directory called `docker` in the same structure as the PennyLane repo. Happy to approve it and let it be merged after this.

PR #239, opened 2022-02-16
Update quantum tapes serialization and Python tests

maliasadi:
**Context:**
Refactoring operations in PennyLane will deprecation a few operations' features including `matrix`, `eigvals` and `control_wires`, and this breaks serialization of quantum tapes and Python tests in Lightning. This PR addresses all `DeprecationWarning` caused by PR [#2023](https://github.com/PennyLaneAI/pennylane/pull/2023) and also fixes `ComplexWarning` and `UserWarning` issues with tests.


pennylane_lightning/lightning_qubit.py
```diff
@@ -190,9 +190,13 @@ def apply_lightning(self, state, operations, dtype=np.complex128):
...
             if method is None:
-                # Inverse can be set to False since o.matrix is already in inverted form
+                # Inverse can be set to False since o.get_matrix() is already in inverted form
                 method = getattr(sim, "applyMatrix_{}".format(self._kernel_for_ops["Matrix"]))
-                method(o.matrix, wires, False)
+                try:
+                    method(o.get_matrix(), wires, False)
+                except AttributeError:
+                    # To support older versions of PL
+                    method(o.matrix, wires, False)
```
mlxd, line 199: :+1:

tests/test_gates.py
```diff
@@ -63,9 +63,7 @@ def op(op_name):
         "ControlledQubitUnitary": qml.ControlledQubitUnitary(
             np.eye(2) * 1j, wires=[0], control_wires=[2]
         ),
-        "MultiControlledX": qml.MultiControlledX(
-            control_wires=[0, 1], wires=2, control_values="01"
-        ),
+        "MultiControlledX": qml.MultiControlledX(wires=(0, 1, 2), control_values="01"),
```
mlxd, line 66: I always worry about this assumed control vs target wire combination (makes life tricky for certain backend libs).
Not a comment for changing here though, just in general.

tests/test_serialize.py
```diff
@@ -296,8 +296,7 @@ def test_mixed_tensor_return(self, monkeypatch, ObsFunc):
         "ObsStructC128" and "ObsStructC64" not in dir(pennylane_lightning.lightning_qubit_ops),
         reason="ObsStructC128 and ObsStructC64 are required",
     )
-    @pytest.mark.parametrize("ObsFunc", [ObsStructC128, ObsStructC64])
```
mlxd, line 299: Are these separated out now into different tests?

mlxd (commented):
Nice work @maliasadi 
Just a quick question on the parameterized tests, but happy to approve once that is answered.

pennylane_lightning/_serialize.py
```diff
@@ -113,11 +113,11 @@ def _serialize_obs(tape: QuantumTape, wires_map: dict, use_csingle: bool = False
             if is_tensor:
                 for o_ in o.obs:
                     if not _obs_has_kernel(o_):
-                        params.append(o_.matrix.ravel().astype(ctype))
+                        params.append(o_.get_matrix().ravel().astype(ctype))
```
josh146, line 116: @maliasadi no action needed for now, but this is a temporary name that still might change before the next release - I'll update you once I know the final decision!

tests/test_gates.py
maliasadi: Me too. FYI, this is the refactored PL code for initializing `wires` and `control_wires`, 
```
        if control_wires is None:
            if len(wires) > 1:
                control_wires = Wires(wires[:-1])
                wires = Wires(wires[-1])
            else:
                raise ValueError(
                    "MultiControlledX: wrong number of wires. 1 wire given. Need at least 2."
                )
        else:
            wires = Wires(wires)
            control_wires = Wires(control_wires)

            warnings.warn(
                "The control_wires keyword will be removed soon. Use wires = (control_wires,target_wire). See the documentation for more information.",
                category=UserWarning,
            )
```

josh146: I think we plan on changing this! (cc @albi3ro)

tests/test_serialize.py
maliasadi: Yes. There are test cases (`test_integration_c64` and `test_integration_c128`) for each one. I had to remove this to prevent `ComplexWarning` due to some casting complex values to real issues.

pennylane_lightning/_serialize.py
mlxd: Should we go ahead with merging this, or is it best to await on the updated name?

maliasadi: I think that's better to wait until the op refactoring of the coming release gets settled.

maliasadi: There are definitely going to be more changes and so we can address all within this PR.

maliasadi: Hi @josh146. Would it be any further updates on ops requiring changes/fixes in Lightning soon? I want to know when is suitable for this PR to be merged.

josh146: Hey @maliasadi, you can update this now to use `qml.matrix(o_)` 🙂 then it should be good to merge!

tests/test_gates.py
albi3ro: The call signature for `MultiControlledX` was changed recently to bring it in line with other gates like `CNOT`, `Toffoli`, `CRX`, etc.

The new signature will remain, but it will instead be a constructor for an operation of type `Controlled`.

pennylane_lightning/_serialize.py
```diff
@@ -27,6 +27,7 @@
 from pennylane.grouping import is_pauli_word
 from pennylane.operation import Observable, Tensor
 from pennylane.tape import QuantumTape
+import pennylane as qml
```
mlxd, line 30: Should we use the same structure of 
```
from pennylane import matrix
```
or something like
```
from pennylane import matrix as qml_mat
```

to avoid pulling in the full library?

mlxd: Though it is probably fine.

mlxd (approved):
Thanks @maliasadi 
I left one comment, but happy to approve either option :)

pennylane_lightning/_serialize.py
maliasadi: Ah! I fix it now.

maliasadi: Hi @mlxd, would you check if this is completely fixed in both `_serialized` and `lightning_qubit`?

maliasadi: Ok. Tests are failed! Because the `qml.matrix` is new and not supported with current versions of PL. So, we should wait till the next release of PL.

mlxd (approved):
Nice one @maliasadi 
Happy to merge!

PR #241, opened 2022-02-28
Remove QHACK banner

antalszava:


pennylane_lightning/_version.py
```diff
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.22.0-dev10"
+__version__ = "0.22.0-dev11"
```
antalszava, line 19: Hi @chaeyeunpark, just double-checking: this is an automated commit for making sure that the dev number is unique for each commit made to `master`, correct?

dime10 (approved):

pennylane_lightning/_version.py
chaeyeunpark: Hi @antalszava, yes it is correct. As PyPI (and test-PyPI) does not accept the wheel with the same version number, this bot (which now just uses my GH token) automatically updates the version to make it unique.

antalszava: Awesome, thanks!

Issue #243, opened 2022-03-01
Update `pennylane.measure` usage due to recent changes in PennyLane


mlxd:
Recent updates to PennyLane have caused some dependent modules to be moved. These paths should be updated in Lightning. 

```bash
==================================== ERRORS ====================================
___________________ ERROR collecting tests/test_measures.py ____________________
ImportError while importing test module '/home/runner/work/pennylane-lightning/pennylane-lightning/main/tests/test_measures.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_measures.py:19: in <module>
    from pennylane.measure import (
E   ModuleNotFoundError: No module named 'pennylane.measure'
```

antalszava:
Let me open a PR on this :)

PR #244, opened 2022-03-01
Update a test file in light of the `pennylane.measure` `pennylane.measurements` renaming

antalszava:
Closes https://github.com/PennyLaneAI/pennylane-lightning/issues/243.


mlxd (approved):
:100: Done before I had a chance to investigate -- thanks @antal!

PR #245, opened 2022-03-02
Better kernel dispatch and add custom _asarray

chaeyeunpark:
### Before submitting

**Context:** Currently, we have two different gate implementations (kernels) and some more to be added. As it is possible that one kernel outperforms another kernel only for some number of qubits, it is desirable to have a kernel dispatcher mechanism that takes this information into account. In addition, as with AVX kernels, it is necessary to have an array whose memory is managed by `lightning` with a proper alignment.

**Description of the Change:** Lightning now has the `KernelMap` class which returns the kernels to use for a given number of qubits. The statevector classes now manage their own kernel information accordingly. Moreover, `_asarray` is implemented.

**Benefits:** Not huge right now, but with AVX and OpenMP kernels, we should have some advantages.

**Possible Drawbacks:** Slightly slower dispatch time (1 or 2 more `std::unoreder_map` accesses), and constructing a statevector becomes slower as it needs to obtain the best kernels for each operation.

**Related GitHub Issues:**

Still need to fix/add documents.


chaeyeunpark:
As 96 changed files are just insane, I will split this PR and make separate PRs.

chaeyeunpark:
Continue at #291

PR #246, opened 2022-03-06
Ignore snapshot

### LINKED ISSUE #242: Disable `qml.Snapshot` operation
#### Issue description
PennyLane will soon receive an operation to examine the values of the statevector over the course of a given circuit, using `qml.Snapshot` (see https://github.com/PennyLaneAI/pennylane/pull/2233). As a debug feature, this is not required to be currently supported in Lightning and so should be explicitly disabled

chaeyeunpark:
Closes #242

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -72,9 +72,10 @@ def _chunk_iterable(it, num_chunks):
     it = iter(it)
     return iter(lambda: tuple(islice(it, num_chunks)), ())
 
+
 def _remove_snapshot_from_operations(operations):
     operations = operations.copy()
-    operations.remove("Snapshot")
+    operations.discard("Snapshot")
     return operations
 
 
@@ -664,7 +665,6 @@ class LightningQubit(DefaultQubit):
         observable_wires = self.map_wires(observable.wires)
 
         return M.var(observable.name, observable_wires)
-
 
 
 if not CPP_BINARY_AVAILABLE:
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -659,11 +659,12 @@ class LightningQubit(DefaultQubit):
 
         return M.var(observable.name, observable_wires)
 
+
+if CPP_BINARY_AVAILABLE:
     # Remove Snapshot from operations
     LightningQubit.operations.remove("Snapshot")
 
-
-if not CPP_BINARY_AVAILABLE:
+else:
 
     class LightningQubit(DefaultQubit):  # pragma: no cover
         name = "Lightning Qubit PennyLane plugin"
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -190,8 +190,6 @@ class LightningQubit(DefaultQubit):
 
         for o in operations:
             name = o.name.split(".")[0]  # The split is because inverse gates have .inv appended
-            if name == "Snapshot":
-                continue
             if _is_lightning_gate(name):
                 kernel = self._kernel_for_ops[name]
                 method = getattr(sim, f"{name}_{kernel}".format(), None)
@@ -661,6 +659,7 @@ class LightningQubit(DefaultQubit):
 
         return M.var(observable.name, observable_wires)
 
+LightningQubit.operations.remove('Snapshot')
 
 if not CPP_BINARY_AVAILABLE:
 
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -72,6 +72,11 @@ def _chunk_iterable(it, num_chunks):
     it = iter(it)
     return iter(lambda: tuple(islice(it, num_chunks)), ())
 
+def _remove_snapshot_from_operations(operations):
+    operations = operations.copy()
+    operations.remove("Snapshot")
+    return operations
+
 
 class LightningQubit(DefaultQubit):
     """PennyLane Lightning device.
@@ -99,7 +104,7 @@ class LightningQubit(DefaultQubit):
     version = __version__
     author = "Xanadu Inc."
     _CPP_BINARY_AVAILABLE = True
-    operations = DefaultQubit.operations.copy()
+    operations = _remove_snapshot_from_operations(DefaultQubit.operations)
 
     def __init__(self, wires, *, kernel_for_ops=None, shots=None, batch_obs=False):
         self._kernel_for_ops = DEFAULT_KERNEL_FOR_OPS
@@ -660,8 +665,6 @@ class LightningQubit(DefaultQubit):
 
         return M.var(observable.name, observable_wires)
 
-
-LightningQubit.operations.remove("Snapshot")
 
 
 if not CPP_BINARY_AVAILABLE:
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -659,7 +659,8 @@ class LightningQubit(DefaultQubit):
 
         return M.var(observable.name, observable_wires)
 
-LightningQubit.operations.remove('Snapshot')
+
+LightningQubit.operations.remove("Snapshot")
 
 if not CPP_BINARY_AVAILABLE:
 
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -190,6 +190,8 @@ class LightningQubit(DefaultQubit):
 
         for o in operations:
             name = o.name.split(".")[0]  # The split is because inverse gates have .inv appended
+            if name == 'Snapshot':
+                continue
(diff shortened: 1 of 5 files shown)


mlxd (commented):
Thanks @chaeyeunpark 

I think it would also be good to have some tests around encountering this operation (especially when also used in adjoint example.). I may be mistaken, but I think we may also need to adjust the adjoint routines to handle this operation too. In this case, skipping the op if encountered, and adjusting the positions/indices of the params relative to the changed tape (though that may come for free).

dime10:
Hi @chaeyeunpark, thanks for making this change. I think it may be worth checking whether removing `Snapshot` from the list of supported operations is not sufficient, normally this should filter the operation from the tape.

So e.g. adding the following here:

https://github.com/PennyLaneAI/pennylane-lightning/blob/299746f7b99e55dd808c54dde3e798948a263edc/pennylane_lightning/lightning_qubit.py#L103-L117

```py
            self.operations.remove("Snaphshot")
```

and:

```py
delattr(LightningQubit, "debugger")
```

at the bottom of the file.

chaeyeunpark:
Hi @dime10, I think there is a semantic problem in removing `Snapshot` from `operations`. Doing this means that `lightning.qubit` does not support the `Snapshot` operator (so a user just cannot use `Snapshot` with `lightning.qubit`), whereas we are now just ignoring it (a user can run a tape with `Snapshot` but it is ignored). I am not sure which one is better, tho.

dime10:
Yes exactly, the idea is to declare that `lightning.qubit` does not support the `Snapshot` operation. This means that if a user does have a circuit with the `Snapshot` operation, it is decomposed prior to execution on the lightning device.
The only reason any change to `lightning.qubit` is necessary, is because it inherits this list (in this case erroneously) from `default.qubit` for which the op was implement.

mlxd:
I'll try this out with the derived device once merged, and we can feed back any subsequent changes needed in another follow-up PR

dime10:
>I'll try this out with the derived device once merged, and we can feed back any subsequent changes needed in another follow-up PR
Sure thing, I was just thinking that the changes in this PR might be superfluous once the the op is popped from the operations list, what do you think?

chaeyeunpark:
Hi @dime10, when I remove `Snapshot` from operators, I get the following error for `Jacobian`
```python
___________________________________ TestAdjointJacobian.test_ignore_snapshot[complex128] ____________________________________

self = <test_adjoint_jacobian.TestAdjointJacobian object at 0x7fd6dbdc14c0>, tol = 1e-06
dev = <LightningQubit device (wires=2, shots=None) at 0x7fd6dbdc1310>, C = <class 'numpy.complex128'>

    @pytest.mark.skipif(not lq._CPP_BINARY_AVAILABLE, reason="Lightning binary required")
    @pytest.mark.parametrize("C", [np.complex64, np.complex128])
    def test_ignore_snapshot(self, tol, dev, C):
        """Tests provides correct answer when provided starting state."""
        x, y, z = [0.5, 0.3, -0.7]
    
        dev._state = dev._state.astype(C)
    
        with qml.tape.JacobianTape() as tape:
            qml.RX(0.4, wires=[0])
            qml.Snapshot()
            qml.Snapshot().inv()
            qml.Rot(x, y, z, wires=[0])
            qml.RY(-0.2, wires=[0])
            qml.expval(qml.PauliZ(0))
    
        tape.trainable_params = {1, 2, 3}
    
>       grad_F = tape.jacobian(dev, method="numeric")

tests/test_adjoint_jacobian.py:422: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../pennylane/pennylane/tape/jacobian_tape.py:585: in jacobian
    options["y0"] = qml.math.to_numpy(self.execute_device(params_f64, device))
../pennylane/pennylane/tape/tape.py:1376: in execute_device
    res = device.execute(self)
../pennylane/pennylane/_qubit_device.py:198: in execute
    self.check_validity(circuit.operations, circuit.observables)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <LightningQubit device (wires=2, shots=None) at 0x7fd6dbdc1310>
queue = [RX(0.4, wires=[0]), Snapshot(wires=[]), Snapshot.inv(wires=[]), Rot(0.5, 0.3, -0.7, wires=[0]), RY(-0.2, wires=[0])]
observables = [expval(PauliZ(wires=[0]))]

    def check_validity(self, queue, observables):
        """Checks whether the operations and observables in queue are all supported by the device.
        Includes checks for inverse operations.
    
        Args:
            queue (Iterable[~.operation.Operation]): quantum operation objects which are intended
                to be applied on the device
            observables (Iterable[~.operation.Observable]): observables which are intended
                to be evaluated on the device
    
        Raises:
            DeviceError: if there are operations in the queue or observables that the device does
                not support
        """
    
        for o in queue:
    
            operation_name = o.name
    
            if getattr(
                o, "return_type", None
            ) == qml.operation.MidMeasure and not self.capabilities().get(
                "supports_mid_measure", False
            ):
                raise DeviceError(
                    f"Mid-circuit measurements are not natively supported on device {self.short_name}. "
                    "Apply the @qml.defer_measurements decorator to your quantum function to "
                    "simulate the application of mid-circuit measurements on this device."
                )
    
            if o.inverse:
                # TODO: update when all capabilities keys changed to "supports_inverse_operations"
                supports_inv = self.capabilities().get(
                    "supports_inverse_operations", False
                ) or self.capabilities().get("inverse_operations", False)
                if not supports_inv:
                    raise DeviceError(
                        f"The inverse of gates are not supported on device {self.short_name}"
                    )
                operation_name = o.base_name
    
            if not self.supports_operation(operation_name):
>               raise DeviceError(
                    f"Gate {operation_name} not supported on device {self.short_name}"
                )
E               pennylane._device.DeviceError: Gate Snapshot not supported on device lightning.qubit

../pennylane/pennylane/_device.py:904: DeviceError

```

The forward pass works well as expected. As this is a problem from the `_qubit_device.py`, I cannot handle it from `lightning`.

dime10:
Yes this is expected I believe. The way the test is set up there is no decomposition happening, so you would get the same results with any other unsupported operation in the queue, for example with `qml.Displacement`. Do mind telling me what the intension behind this test is?

mlxd:
> Yes this is expected I believe. The way the test is set up there is no decomposition happening, so you would get the same results with any other unsupported operation in the queue, for example with `qml.Displacement`. Do you mind telling me what the intension behind this test is?

Is the plan to explicitly allow failure if a user calls a workflow with `qml.Snapshot` for an unsupported device? In the above test, should it not just silently allow the op to vanish? I guess this is a bigger problem for us than thought if the above tests outright  fail as:

```python
        with qml.tape.JacobianTape() as tape:
            qml.RX(0.4, wires=[0])
            qml.Snapshot()
            qml.Snapshot().inv()
            qml.Rot(x, y, z, wires=[0])
            qml.RY(-0.2, wires=[0])
            qml.expval(qml.PauliZ(0))
    
        tape.trainable_params = {1, 2, 3}
        grad_F = tape.jacobian(dev, method="numeric")
```

are all still public facing functions and should be a valid run. What do you think?

dime10:
I would let this specific instance fail because it is not specific to `qml.Snapshot`, it is the same behaviour as for all unsupported operations (try replacing `qml.Snapshot` with `qml.Displacement` for example).

If these instructions were instead embedded in a QNode (similar to how a user would normally write code), the program would run because unsupported gates are automatically decomposed.

mlxd:
> I would let this specific instance fail because it is not specific to `qml.Snapshot`, it is the same behaviour as for all unsupported operations (try replacing `qml.Snapshot` with `qml.Displacement` for example).
> 
> If these instructions were instead embedded in a QNode (similar to how a user would normally write code), the program would run because unsupported gates are automatically decomposed.

So allow direct failures if explicitly requesting the unsupported op, but if using the QNode mechanism it will succeed due to the decomposition?

pennylane_lightning/lightning_qubit.py
```diff
@@ -660,7 +660,11 @@ def var(self, observable, shot_range=None, bin_size=None):
         return M.var(observable.name, observable_wires)
```
dime10, line 662: I just realized, I think what we need to do is copy the attribute, otherwise `DefaultQubit` is also affected:

```suggestion
    operations = DefaultQubit.operations.copy()
    operations.remove("Snapshot")

```

chaeyeunpark: Yes, that's true. :+1:

dime10 (approved):
Thanks a lot @chaeyeunpark , that should do the trick :)

PR #247, opened 2022-03-07
Add generate_samples() to lightning

trevor-vincent:
**Context:**

generate_samples() performs most of the heavy lifting for the sample() function. Currently there is no lightning version of generate_samples(). 

This pr adds a lightning C++ version of the generate_samples function using the alias sampling method (https://en.wikipedia.org/wiki/Alias_method). This C++ function currently single-threaded only, however a parallel version is straight-forward (e.g. insert a openmp pragma above the for loop that computes samples while being mindful that the parallel random number generation and caching is thread-safe).

**Benefits:**

Lightning now has it's own version of generate_samples(), getting us closer to a complete independent lightning statistics module. For low numbers of shots, the single-threaded version is faster than default.qubit (which uses numpy.random.choice) on my machine: 

bench

This benchmark times the generate_samples() function for a randomly generated 10-qubit circuit (benchmarks were single-threaded for lightning, however numpy may be multi-threaded).

**Possible Drawbacks:**

At higher shots, we should likely switch to a parallel version.

**Related GitHub Issues:**


pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,29 @@ void lightning_class_bindings(py::module &m) {
                 const std::vector<size_t> &wires) {
                  return M.expval(operation, wires);
              })
+        .def("generate_samples",
+             [](Measures<PrecisionT> &M, size_t num_shots) {
+                 auto &&result = M.generate_samples(num_shots);
+                 // int num_samples = num_shots;
+                 // std::vector<int> result(num_samples*wires.size(),0);
+
+                 int num_wires = result.size() / num_shots;
```
mlxd, line 333: If possible, it is better to use `std::size_t` everywhere there is a positive integer, rather than `int`.

```suggestion
                 const size_t num_wires = result.size() / num_shots;
```

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,29 @@ void lightning_class_bindings(py::module &m) {
...
+                 ssize_t ndim = 2;
```
mlxd, line 334: I think same here, if we don't need it signed. Though, ignore if this is a requirement from the binding interface.
```suggestion
                 const size_t ndim = 2;
```

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,29 @@ void lightning_class_bindings(py::module &m) {
...
+                 std::vector<ssize_t> shape = {static_cast<ssize_t>(num_shots),
+					       static_cast<ssize_t>(num_wires)};
+                 std::vector<ssize_t> strides = {static_cast<ssize_t>(sizeof(int) * num_wires),
+						 static_cast<ssize_t>(sizeof(int))};
```
mlxd, line 338:
```suggestion
                 const std::vector<size_t> shape {num_shots, num_wires};
                 const std::vector<size_t> strides {sizeof(size_t) * num_wires, sizeof(size_t)};
```

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,29 @@ void lightning_class_bindings(py::module &m) {
...
+
+                 // return 2-D NumPy array
+                 return py::array(py::buffer_info(
+                     result.data(), /* data as contiguous array  */
+                     sizeof(int),   /* size of one scalar        */
+                     py::format_descriptor<int>::format(), /* data type */
```
mlxd, line 344:
```suggestion
                     py::format_descriptor<size_t>::format(), /* data type */
```

Again, feel free to ignore if `int` are a hard requirement, but if not better to use `size_t`.

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,29 @@ void lightning_class_bindings(py::module &m) {
...
+                 ssize_t ndim = 2;
+                 std::vector<ssize_t> shape = {static_cast<ssize_t>(num_shots),
+					       static_cast<ssize_t>(num_wires)};
+                 std::vector<ssize_t> strides = {static_cast<ssize_t>(sizeof(int) * num_wires),
+						 static_cast<ssize_t>(sizeof(int))};
+
+                 // return 2-D NumPy array
+                 return py::array(py::buffer_info(
+                     result.data(), /* data as contiguous array  */
+                     sizeof(int),   /* size of one scalar        */
```
mlxd, line 343:
```suggestion
                     sizeof(size_t),   /* size of one scalar        */
```

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -256,5 +260,92 @@ class Measures {
...
+    /**
+     * @brief Generate samples using the alias method.
+     * Reference: https://en.wikipedia.org/wiki/Alias_method
+     *
+     * @param num_samples The number of samples to generate.
+     * @return 1-D vector of samples in binary, each sample is
+     * separated by a stride equal to the number of qubits.
+     */
+    std::vector<size_t> generate_samples(size_t num_samples) {
+        size_t num_qubits = original_statevector.getNumQubits();
```
mlxd, line 273:
```suggestion
        const size_t num_qubits = original_statevector.getNumQubits();
```

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -256,5 +260,92 @@ class Measures {
...
+        auto &&probabilities = probs();
+
+        std::vector<size_t> samples(num_samples * num_qubits, 0);
+        std::mt19937 generator(std::random_device{}());
+        std::uniform_real_distribution<fp_t> distribution(0.0, 1.0);
+        std::unordered_map<size_t, size_t> cache;
+
+        size_t N = probabilities.size();
```
mlxd, line 281:
```suggestion
        const size_t N = probabilities.size();
```

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -256,5 +260,92 @@ class Measures {
...
+        }
+
+        // Pick samples
+        for (size_t i = 0; i < num_samples; i++) {
+            fp_t pct = distribution(generator) * N;
+            size_t idx = static_cast<size_t>(pct);
+            if (pct - idx > bucket[idx])
+                idx = bucket_partner[idx];
+
+            // If cached, retrieve sample 1from cache
```
mlxd, line 332:
```suggestion
            // If cached, retrieve sample from cache
```

pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -163,6 +163,55 @@ TEST_CASE("Expected Values", "[Measures]") {
...
+                                             0.001797, 0.180036, 0.003624,
+                                             0.023377, 0.000471};
+
+    size_t num_qubits = 3;
+    size_t N = std::pow(2, num_qubits);
+    size_t num_samples = 10000;
+    auto &&samples = Measurer.generate_samples(num_samples);
+
+    std::vector<int> counts(N, 0);
+    std::vector<int> samples_decimal(num_samples, 0);
```
mlxd, line 190:
```suggestion
    std::vector<size_t> counts(N, 0);
    std::vector<size_t> samples_decimal(num_samples, 0);
```

tests/test_measures.py
```diff
@@ -462,6 +462,59 @@ def circuit2():
...
+        """Tests if the samples returned by sample have
+        the correct values
+        """
+
+        # Explicitly resetting is necessary as the internal
+        # state is set to None in __init__ and only properly
+        # initialized during reset
+        dev = qml.device("lightning.qubit", wires=2, shots=1000)
+
+        dev.apply([qml.RX(1.5708, wires=[0])])
```
mlxd, line 507: I wonder if adding more tests with different param values and operations would be a good idea.

Since code coverage is complaining about several lines, this may help, especially when testing for `float` vs `double` data too.

mlxd (commented):
Thanks @trevor-vincent 
I have a few comments and some suggestions

tests/test_measures.py
trevor-vincent: Sounds good, I'll add some parameters here

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,28 @@ void lightning_class_bindings(py::module &m) {
...
+
+                 // return 2-D NumPy array
+                 return py::array(py::buffer_info(
+                     result.data(),  /* data as contiguous array  */
+                     sizeof(size_t), /* size of one scalar        */
+                     py::format_descriptor<size_t>::format(), /* data type */
+                     ndim,   /* number of dimensions      */
+                     shape,  /* shape of the matrix       */
+                     strides /* strides for each axis     */
+                     ));
```
maliasadi, line 347: :+1:

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,28 @@ void lightning_class_bindings(py::module &m) {
                 const std::vector<size_t> &wires) {
                  return M.expval(operation, wires);
              })
+        .def("generate_samples",
+             [](Measures<PrecisionT> &M, size_t num_shots) {
+                 auto &&result = M.generate_samples(num_shots);
+                 // int num_samples = num_shots;
+                 // std::vector<int> result(num_samples*wires.size(),0);
```
maliasadi, line 331: You may want to remove these lines.

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,28 @@ void lightning_class_bindings(py::module &m) {
...
+
+                 const size_t num_wires = result.size() / num_shots;
```
maliasadi, line 333: Or simply pass the `num_wires` as an argument in the binding function (defined at line 327) to save the division.

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -324,6 +324,28 @@ void lightning_class_bindings(py::module &m) {
...
+                 const size_t ndim = 2;
+                 const std::vector<size_t> shape{num_shots, num_wires};
+                 const std::vector<size_t> strides{sizeof(size_t) * num_wires,
+                                                   sizeof(size_t)};
+
+                 // return 2-D NumPy array
+                 return py::array(py::buffer_info(
+                     result.data(),  /* data as contiguous array  */
+                     sizeof(size_t), /* size of one scalar        */
```
maliasadi, line 342: I would suggest something like:
```suggestion
                 constexpr auto sz = sizeof(size_t); 
                 const std::vector<size_t> strides{sz * num_wires, sz};

                 // return 2-D NumPy array
                 return py::array(py::buffer_info(
                     result.data(),  /* data as contiguous array  */
                     sz, /* size of one scalar        */
```

pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -163,6 +163,55 @@ TEST_CASE("Expected Values", "[Measures]") {
     }
 }
 
+TEST_CASE("Sample", "[Measures]") {
```
mlxd, line 166: I wonder if we could add this as a templated test case for `float` and `double` types. I'd expect them to work fine, but not harm having it.

Something like:

```suggestion
TEMPLATE_TEST_CASE("Sample", "[Measures]", float, double) 
```


and using `TestType;` for the type throughout.

pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -163,6 +163,55 @@ TEST_CASE("Expected Values", "[Measures]") {
...
+    }
+
+    // compare estimated probabilities to real probabilities
+    SECTION("No wires provided:") {
+        REQUIRE_THAT(probabilities,
+                     Catch::Approx(expected_probabilities).margin(.05));
+    }
+}
+
 TEST_CASE("Variances", "[Measures]") {
```
mlxd, line 215: Same again here.

mlxd (changes requested):
Great job Trevor! I have a minor testing suggestion, but happy to approve after that.

tests/test_measures.py
```diff
@@ -462,6 +462,71 @@ def circuit2():
         assert np.allclose(circuit1(), circuit2(), atol=tol)
 
 
+class TestSample:
+    """Tests that samples are properly calculated."""
+
+    @pytest.mark.parametrize("C", [np.complex64, np.complex128])
+    def test_sample_dimensions(self, C):
+
```
maliasadi, line 470: Missing some description for this test def.

tests/test_measures.py
```diff
@@ -462,6 +462,71 @@ def circuit2():
...
+        # Explicitly resetting is necessary as the internal
+        # state is set to None in __init__ and only properly
+        # initialized during reset
+        dev = qml.device("lightning.qubit", wires=2, shots=1000)
```
maliasadi, line 474: You may want to use a fixture for dev,
```Python
    @pytest.fixture
    def dev(self):
        return qml.device("lightning.qubit", wires=2, shots=1000)
```

maliasadi, line 474: You may want to add a fixture for dev as follows and use it for all these tests.
```Python
    @pytest.fixture
    def dev(self):
        return qml.device("lightning.qubit", wires=2, shots=1000)
```

maliasadi (changes requested):
Great work @trevor-vincent! No major comments; I have had only a few suggestions.

pennylane_lightning/src/tests/Test_Measures.cpp
trevor-vincent: Good idea!

pennylane_lightning/src/tests/Test_Measures.cpp
trevor-vincent: Okay will add here too

pennylane_lightning/src/bindings/Bindings.cpp
trevor-vincent: good suggestion

tests/test_measures.py
trevor-vincent: nice will do

pennylane_lightning/src/tests/Test_Measures.cpp
trevor-vincent: This leads to some errors b.c.  apparently Measurer.var expects doubles. Changing Measurer.var too outside the scope of this pr, so I'm going to leave this for another pr.

pennylane_lightning/lightning_qubit.py
```diff
@@ -576,6 +576,33 @@ def probability(self, wires=None, shot_range=None, bin_size=None):
 
         return M.probs(device_wires)
 
+    def generate_samples(self):
+        """Generate samples
+
+        Returns:
+            Array of samples in binary
```
maliasadi, line 583: Is this correct to assume that the return value is as follows?
```suggestion
             array[complex]: array of samples in the shape ``(dev.shots, dev.num_wires)``
```

pennylane_lightning/lightning_qubit.py
```diff
@@ -576,6 +576,33 @@ def probability(self, wires=None, shot_range=None, bin_size=None):
...
+        # Initialization of state
+        ket = np.ravel(self._state)
+
+        if use_csingle:
+            ket = ket.astype(np.complex64)
+
+        state_vector = StateVectorC64(ket) if use_csingle else StateVectorC128(ket)
+        M = MeasuresC64(state_vector) if use_csingle else MeasuresC128(state_vector)
+
+        return M.generate_samples(len(self.wires), self.shots).astype(int)
```
maliasadi, line 604: Sorry! I just noticed `astype(int)`. Shouldn't be `complex` to become compatible with the return value of `generate_samples` method in `QubitDevice`? lmk if I'm missing something.

trevor-vincent: that is a typo in the pennylane documentation, it actually returns an int array if you look at states_to_binary which is returned at the bottom of generate_samples

pennylane_lightning/lightning_qubit.py
trevor-vincent: Changed to array[int] because the pennylane documentation has it wrong. Thank you for spotting this!

maliasadi (approved):
Thanks @trevor-vincent. Happy to approve!

pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -163,39 +164,90 @@ TEST_CASE("Expected Values", "[Measures]") {
...
         REQUIRE(variance == Approx(variances_ref).margin(1e-6));
     }
 
     SECTION("Testing list of operators defined by a matrix:") {
-        vector<std::complex<double>> PauliX = {0, 1, 1, 0};
-        vector<std::complex<double>> PauliY = {0, {0, -1}, {0, 1}, 0};
-        vector<std::complex<double>> PauliZ = {1, 0, 0, -1};
+        vector<std::complex<TestType>> PauliX = {0, 1, 1, 0};
+        vector<std::complex<TestType>> PauliY = {0, {0, -1}, {0, 1}, 0};
+        vector<std::complex<TestType>> PauliZ = {1, 0, 0, -1};
```
mlxd, line 245:
```suggestion
        vector<std::complex<TestType>> PauliX = {{0, 0}, {1, 0}, {1, 0}, {0, 0}};
        vector<std::complex<TestType>> PauliY = {{0, 0}, {0, -1}, {0, 1}, {0, 0}};
        vector<std::complex<TestType>> PauliZ = {{1, 0}, {0, 0}, {0, 0}, {-1, 0}};
```
I think it is best to be explicit here about each complex value field.

mlxd (changes requested):
Great job @trevor-vincent 
One suggestion below, but don't see any blocker otherwise. Happy to approve after this change!

mlxd (approved):

PR #248, opened 2022-03-07
Modify pybind11 classes scope in modules

mlxd:
**Context:** When using Lightning as a library, having globally-exposed methods can cause issues when using the templated types with other external packages (see [here](https://github.com/pybind/pybind11/blob/master/docs/advanced/classes.rst#module-local-class-bindings) for details). In this case, ensuring the classes are exposed locally at the module level only helps avoid such issues as `ImportError: generic_type: type "MyType" is already registered!`. This PR helps to ensure such problems are avoided.

**Description of the Change:** Modifies Lightning bound classes to have local scope to Python modules only.

**Benefits:** Allows easier extensibility of the Lightning functionality for dependent classes and methods.

**Possible Drawbacks:** Disallows direct interoperability and conversion to Python types.

**Related GitHub Issues:**

### DIFF
--- a/pennylane_lightning/_version.py
+++ b/pennylane_lightning/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.22.0-dev12"
+__version__ = "0.22.0-dev13"
--- a/pennylane_lightning/_version.py
+++ b/pennylane_lightning/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.22.0-dev13"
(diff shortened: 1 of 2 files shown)


mlxd:
Note, following the partitioning rules of [here](https://github.com/pybind/pybind11/blob/master/docs/advanced/misc.rst#partitioning-code-over-multiple-extension-modules) may be more sensible in the long run, keeping separate types allows more extensibility while lowering the maintenance burden for disparate repositories.

AmintorDusko (approved):
Very nice! Thanks for that!

chaeyeunpark (approved):
Happy to approve this.

mlxd:
Thanks all.

PR #250, opened 2022-03-08
Install ML libs in CI

chaeyeunpark:


tests/test_vjp.py
```diff
@@ -108,6 +108,18 @@ def test_zero_dy(self, dev, C):
         vjp = dev.compute_vjp(dy, jac)
         assert np.all(vjp == np.zeros([3]))
 
+    @pytest.mark.parametrize("C", [np.complex64, np.complex128])
```
maliasadi, line 111: I have added tests for non_numpy `dy` as part of #239. You may want to remove this here.

.github/workflows/tests.yml
```diff
@@ -82,6 +82,12 @@ jobs:
           pip uninstall pennylane -y
           pip install git+https://github.com/PennyLaneAI/pennylane.git
 
+      - name: Install ML libraries for interfaces
+        run: |
+          pip install --upgrade torch
+          pip install --upgrade "jax[cpu]"
+          pip install --upgrade tensorflow
```
maliasadi, line 89:
```suggestion
          pip install --upgrade "jax[cpu]" jaxlib
          pip install --upgrade tensorflow keras
```

tests/test_vjp.py
```diff
@@ -108,6 +108,18 @@ def test_zero_dy(self, dev, C):
...
+    def test_non_numpy_dy(self, dev, C):
```
mlxd, line 112: Should we also test this for the other frameworks too?

mlxd (commented):
Thanks @chaeyeunpark Just left a comment.
Also, do you have a timing estimate how how much longer the tests take with this PR?

.github/workflows/tests.yml
```diff
@@ -82,6 +82,12 @@ jobs:
           pip uninstall pennylane -y
           pip install git+https://github.com/PennyLaneAI/pennylane.git
 
+      - name: Install ML libraries for interfaces
```
maliasadi, line 85: The only concern here is the fact that we are using the latest versions of `torch`, `jax`, and `tf`. I'd suggest to use some env variables such as `TORCH_VERSION`, `JAX_VERSION`, and `TF_VERSION` to have more control over the ML libs. It would be helpful to catch bugs related to some specific versions in the future too.

maliasadi (changes requested):
Nice update @chaeyeunpark! I only have a few comments...

.github/workflows/tests.yml
mlxd: Maybe we factor tests out into an LTS (ie oldest supported version) and the most recent?

tests/test_vjp.py
maliasadi: I already added tests for all the ML frameworks in PR 239. Updating tests.yml was in TODOs :smile: I am glad there is now a separate PR for it.

.github/workflows/tests.yml
maliasadi: Even better!

tests/test_vjp.py
chaeyeunpark: Aha! Yes, as your version tests all interfaces, it's more reasonable to remove this here.

chaeyeunpark:
Hi @mlxd, this increases the python test time from ~2.5 min to ~4.5 min. As other workflows take around 8-10 min, this will not increase overall time.

.github/workflows/tests.yml
chaeyeunpark: I just found only Pytorch explicitly supports the LTS version. I can use a solution from Ali that is currently used in PL.

mlxd:
> Hi @mlxd, this increases the python test time from ~2.5 min to ~4.5 min. As other workflows take around 8-10 min, this will not increase the overall time.

Great, thanks @chaeyeunpark

chaeyeunpark:
Hi @maliasadi, I added versions for TF and PyTorch, but JAX does not maintain a particular stable version. As PL also does not specify a version of JAX, I think this is a reasonable solution. What do you think?

.github/workflows/tests.yml
```diff
@@ -82,6 +84,12 @@ jobs:
...
+        run: |
+          pip install --upgrade torch==$TORCH_VERSION -f https://download.pytorch.org/whl/cpu/torch_stable.html
+          pip install --upgrade "jax[cpu]"  # This also installs jaxlib
+          pip install --upgrade tensorflow==$TF_VERSION keras
```
maliasadi, line 91: Is that right?
```suggestion
          pip install --upgrade tensorflow==$TF_VERSION keras==$TF_VERSION
```

.github/workflows/tests.yml
```diff
@@ -82,6 +84,12 @@ jobs:
           pip uninstall pennylane -y
           pip install git+https://github.com/PennyLaneAI/pennylane.git
 
+      - name: Install ML libraries for interfaces
+        run: |
+          pip install --upgrade torch==$TORCH_VERSION -f https://download.pytorch.org/whl/cpu/torch_stable.html
+          pip install --upgrade "jax[cpu]"  # This also installs jaxlib
```
maliasadi, line 90: :+1:

maliasadi (approved):

maliasadi:
Fair! Thanks for looking into this. It looks good and happy to approve.

mlxd (approved):
Nice work! Happy to approve too

chaeyeunpark:
Thanks guys for approving this :wink:

PR #252, opened 2022-03-09
Add GPU support to lightning

mlxd:
This adds GPU support to lightning, and is installable using the optional tag 
```bash
pip install pennylane-lightning[gpu]
```

Note: this package works only with Linux on x64


mlxd:
Actually, we can review this for the RC pr instead, to avoid duplicating efforts

PR #255, opened 2022-03-11
Update the `statistics` method

maliasadi:
**Context:**
Lightning currently doesn’t implement the `statistics` method and only offers support for computing probability, expectation value, and variance in analytic mode (`self._shots=None`). A drawback to this implementation is that each call to `probs`, `expval`, or `var` creates a pair of StateVectorC128 (or C64) and MeasuresC128 (or C64) objects. 

A desirable implementation of this method is that of computing the statistics for the list of observables using one single pair of pybind objects. To do so, this PR provides the statistics method in Lightning so that it initializes MeasuresC128 (or C64) only once for the entire computation. In the following, we test the new implementation against master branch for a couple of examples: 

Example 1.
```Python
@qml.qnode(dev)
def circuit():
	for i in range(0, wires-1, 2):
		qml.RX(0.4, wires=[i])
		qml.RY(-0.2, wires=[i+1])

	obs = []
	for i in range(wires):
		obs.append(qml.expval(qml.PauliZ(i)))
		obs.append(qml.var(qml.PauliZ(i)))
	return obs
```
stats_ex1

Example 2.
```Python
@qml.qnode(dev)
def circuit():
	for i in range(0, wires-1, 2):
		qml.RX(0.4, wires=[i])
		qml.RY(-0.2, wires=[i+1])

	obs = []
	for i in range(wires):
		obs.append(qml.probs([*range(i+1)]))
	return obs
```
stats_ex2


pennylane_lightning/lightning_qubit.py
```diff
@@ -528,7 +535,27 @@ def processing_fns(tapes):
...
+        state = self._pre_rotated_state if pre_rotated else self._state
+        ket = np.ravel(state)
+
+        if use_csingle:
+            ket = ket.astype(np.complex64)
+
+        state_vector = StateVectorC64(ket) if use_csingle else StateVectorC128(ket)
+        return MeasuresC64(state_vector) if use_csingle else MeasuresC128(state_vector)
+
+    def probability(self, wires=None, shot_range=None, bin_size=None, mc_obj=None):
```
mlxd, line 558: What is the `mc_obj` in this case? Should it be added to the docstring?

Since we are breaking from the super-class method signature, is this something we want to support externally, or internally only?

Do we expect this to go away with future implementations?

Lastly, would this be better as a class variable/entity and used throughout?

pennylane_lightning/lightning_qubit.py
```diff
@@ -555,27 +582,11 @@ def probability(self, wires=None, shot_range=None, bin_size=None):
...
-            ket = ket.astype(np.complex64)
-
-        state_vector = StateVectorC64(ket) if use_csingle else StateVectorC128(ket)
-        M = MeasuresC64(state_vector) if use_csingle else MeasuresC128(state_vector)
+        M = self.init_measures_object() if mc_obj is None else mc_obj
 
         return M.probs(device_wires)
 
-    def expval(self, observable, shot_range=None, bin_size=None):
+    def expval(self, observable, shot_range=None, bin_size=None, mc_obj=None):
```
mlxd, line 589: Same as above?

pennylane_lightning/lightning_qubit.py
```diff
@@ -604,30 +615,14 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
-        M = MeasuresC64(state_vector) if use_csingle else MeasuresC128(state_vector)
+        M = self.init_measures_object(pre_rotated=True) if mc_obj is None else mc_obj
 
         # translate to wire labels used by device
         observable_wires = self.map_wires(observable.wires)
 
         return M.expval(observable.name, observable_wires)
 
-    def var(self, observable, shot_range=None, bin_size=None):
+    def var(self, observable, shot_range=None, bin_size=None, mc_obj=None):
```
mlxd, line 625: And here?

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -42,12 +42,12 @@ namespace Pennylane {
...
 class Measures {
   private:
-    const SVType &original_statevector;
+    const SVType original_statevector;
     using CFP_t = std::complex<fp_t>;
 
   public:
-    Measures(const SVType &provided_statevector)
-        : original_statevector{provided_statevector} {};
+    Measures(const SVType provided_statevector)
```
mlxd, line 49: Can a const copied element be moved? If so, I am surprised the standard allows it.

mlxd (commented):
Thanks @maliasadi 

I have a few questions:

pennylane_lightning/src/simulator/Measures.hpp
chaeyeunpark: It silently ignores `move`. https://www.youtube.com/watch?v=ZKaoR3dP9uM

mlxd: Huh, I had no idea.

PR #262, opened 2022-03-17
Remove `qml.Snapshot` from fallback device

dime10:
When the CI checks in the main PennyLane repo fail to install a correct wheel of Lightning from the PyPi test repo, some tests that expect the Snapshot operation to be unsupported fail because the operation is now suddenly supported due to the use of the fallback device (default.qubit).

This change makes the Snapshot operation unsupported on the fallback device.

### DIFF
--- a/pennylane_lightning/_version.py
+++ b/pennylane_lightning/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.23.0-dev4"
+__version__ = "0.23.0-dev5"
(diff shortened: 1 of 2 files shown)


mlxd (approved):
Looks good to me, thanks @dime10 
Though, I think we should retrigger the tests (`git commit -m "Trigger CI --allow-empty` is my quick 'n' dirty go-to for this).

mlxd:
> Looks good to me, thanks @dime10 Though, I think we should retrigger the tests (`git commit -m "Trigger CI --allow-empty` is my quick 'n' dirty go-to for this).

NV, just discovered run failing jobs.

dime10:
Thanks @mlxd! You mentioned adding this change in Lightning-GPU as well, but looking in the repo there I don't think it will be required since the device already inherits from `LightningDevice`.

mlxd:
> Thanks @mlxd! You mentioned adding this change in Lightning-GPU as well, but looking in the repo there I don't think it will be required since the device already inherits from `LightningDevice`.

Yep, you are right, this behaviour should carry through. Thanks @dime10

PR #265, opened 2022-03-21
Add C++ coverage to CI

chaeyeunpark:
Add C++ code coverage to GH Action.


chaeyeunpark:
Hi all, this PR adds C++ coverage to CodeCov. Now it only uploads results from Linux and Windows, but macOS can be also added if it is necessary. Still, it seems like the coverage results only count the lines that are visible under the given environment, i.e. `#ifdef __GCC__` block is not added to the denominator of the percentage it calculates when MSVC is used.

.github/workflows/tests_linux.yml
```diff
@@ -98,6 +117,21 @@ jobs:
           check_name: Test Report (C++) on Ubuntu
           files: Build/tests/results/report.xml
 
+      - name: Build and run unit tests for code coverage
+        run: |
+            cmake pennylane_lightning/src -BBuildCov  -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON
```
mlxd, line 122: Do we need to enable `-DENABLE_BLAS=ON` here also?

.github/workflows/tests_linux.yml
```diff
@@ -30,13 +30,15 @@ jobs:
           python-version: '3.7'
 
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2 # for codecov
```
mlxd, line 34: Nice -- wasn't aware this is an option.

mlxd (commented):
Awesome work @chaeyeunpark 
Left a few small comments/questions.

pennylane_lightning/src/util/RuntimeInfo.cpp
```diff
@@ -14,17 +14,15 @@
...
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && defined(_WIN64)
 #include <array>
 #include <intrin.h>
 #endif
-#endif
 
 namespace Pennylane::Util {
-#if defined(__x86_64__) && (defined(__GNUC__) || defined(__clang__))
+#if (defined(__GNUC__) || defined(__clang__)) && defined(__x86_64__)
```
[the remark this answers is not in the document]
chaeyeunpark: Yes but as it also support `cpuid.h` normally, I think it is fine (also tested locally).

.github/workflows/tests_windows.yml
```diff
@@ -0,0 +1,53 @@
...
+          choco install OpenCppCoverage -y
+          echo "C:\Program Files\OpenCppCoverage" >> $env:GITHUB_PATH
+
+      - name: Build and run unit tests for code coverage
+        run: |
+            cmake ./pennylane_lightning/src -BBuild  -DBUILD_TESTS=ON -DENABLE_OPENMP=OFF -DENABLE_WARNINGS=OFF
+            cmake --build ./Build --config Debug
+            mkdir -p ./Build/tests/results
+            .\Build\tests\Debug\runner.exe --order lex --reporter junit --out .\Build\tests\results\report.xml
+            OpenCppCoverage --sources pennylane_lightning\src --export_type cobertura:coverage.xml Build\tests\Debug\runner.exe
```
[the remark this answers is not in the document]
chaeyeunpark: Yes, but as we anyway have an AWS machine for MSVC, it wasn't that challenging :wink:

.github/workflows/tests_linux.yml
chaeyeunpark: Yes, Codecov raises this warning. I think this is the reason why CodeCov sometimes couldn't merge coverage data uploaded from the same commit has.

mlxd (approved):

PR #266, opened 2022-03-21
Fix `cpuid.h` for ppc and aarch64

chaeyeunpark:
After a recent push to master, Linux + PPC and ARM64 fail to build wheels as `cpuid.h` file is not found in those environments (see https://github.com/PennyLaneAI/pennylane-lightning/runs/5633549050?check_suite_focus=true and https://github.com/PennyLaneAI/pennylane-lightning/runs/5633549056?check_suite_focus=true).
This PR fixes this issue.


mlxd (approved):
Nice one @chaeyeunpark Happy to see this fix merged.
I think we can look at this for other arch support later (ARM NEON, etc).

Issue #267, opened 2022-03-21
Follow PennyLane `default.qubit` implementation for identity ops


mlxd:
We should skip over identity operations in Lightning, rather than attempt to apply them. This follows from the PR https://github.com/PennyLaneAI/pennylane/pull/2356/files

PR #268, opened 2022-03-22
Skip over `identity` ops in "lightning.qubit"

maliasadi:
**Context:**
Fix issue #267.

### DIFF
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -15,7 +15,7 @@ r"""
 This module contains the :class:`~.LightningQubit` class, a PennyLane simulator device that
 interfaces with C++ for fast linear algebra calculations.
 """
-from typing import List
+from typing import Iterable, List
 from warnings import warn
 from os import getenv
 from itertools import islice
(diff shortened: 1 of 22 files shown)


pennylane_lightning/lightning_qubit.py
```diff
@@ -203,7 +203,11 @@ def apply_lightning(self, state, operations, dtype=np.complex128):
         else:
             raise TypeError(f"Unsupported complex Type: {dtype}")
 
-        for o in operations:
+        skipped_ops = ["Identity"]
+        ops_iter = operations if isinstance(operations, Iterable) else [operations]
+        ops_generator = (o for o in ops_iter if o not in skipped_ops)
+
+        for o in ops_generator:
```
albi3ro, line 210:
```suggestion
        skipped_ops = ["Identity"]
        for o in operations:
            if op in skipped_ops:
                continue
```

Thought this would've to be done in C++. Nice to know this can just be done from Python.

This might be a more straightforward way to skip over the `"skipped_ops"`, but the way here would work as well.

mlxd: Happy with either approach on my side.
Though, I guess we should keep that `skipped_ops` list small to avoiding adding an additional `o(n)` comparisons when applying each gate.

mlxd (approved):
Nice one @maliasadi 
Just left one comment, but happy to go with whatever you think works.

PR #279, opened 2022-04-18
Create release candidate v0.23

chaeyeunpark:
This PR is for release v0.23

### DIFF
--- a/pennylane_lightning/_version.py
+++ b/pennylane_lightning/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.23.0-dev15"
+__version__ = "0.23.0-rc0"


tests/test_measures.py
```diff
@@ -89,7 +89,30 @@ def circuit():
     @pytest.mark.parametrize(
         "cases",
         [
+            [None, [0.9165164490394898, 0.0, 0.08348355096051052, 0.0]],
```
chaeyeunpark, line 92: Hi @mlxd, I thought I had fixed `qml.probs` errors from `test_measure.py` in https://github.com/PennyLaneAI/pennylane-lightning/pull/280. 

I am also a bit confused about the `qml.probs`. I thought it does not accept `wires=None` but supports `wires=[]` to obtain the probability of all wires, but it seems like `wires=None` also works for this test case....

mlxd: Yea, I thought so too. I think there may have been some additional changes in PL to cause this difference. Are you fine with the updates to the tests here, or do you think we should roll-back and opt for something different?

.github/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* Add `generate_smaples()` to lightning.
```
AmintorDusko, line 5:
```suggestion
* Add `generate_samples()` to lightning.
```

tests/test_measures.py
josh146: Could this be related to https://github.com/PennyLaneAI/pennylane/pull/2432? Maybe it is a bug in PL?

chaeyeunpark: Aha, this was a test case that anyway should be failed. But I think it is anyway better to recover `wires=[]` case for the next test. I will just recover it.

PR #280, opened 2022-04-18
Fix test measures

chaeyeunpark:
[After recent changes in PL](https://github.com/PennyLaneAI/pennylane/pull/2438), `qml.probs` does not accept `wires=None` anymore.


antalszava (approved):
Looks good, thank you for catching this! :tada:

tests/test_measures.py
```diff
@@ -116,15 +115,6 @@ def circuit():
     @pytest.mark.parametrize(
         "cases",
         [
-            [
```
maliasadi, line 119: Thanks @chaeyeunpark! I'm just wondering how this hadn't been caught by `cpp-tests` before :thinking:

PR #281, opened 2022-04-19
Pauli X/Z

multiphaseCFD:
Hi Josh and Lee,

I parallelized the Pauli X/Z code with threads. I also used the #pragma unroll directive
to make the code more concise. As you can see from the results, multithreading can only accelerate the Pauli X/Z operations with large qubits 
number and the speedup for both operations with two threads is around 10%. Due to the overhead caused by creating/destroying threads, 
multithreading for small number of qubits is even slower than the serial program. 

The reason why adding an additional threads can only lead to 10% speedup for the large number qubits operations is because both Pauli X/Z 
operation are memory bounded but not computation bounded. 

I also would like to mention that there are some typos on the website page: https://pennylane-lightning.readthedocs.io/en/latest/devices.html.

Please let me know if you have any questions!

Best,


mlxd:
Thanks @multiphaseCFD 
We will take a look and open a review/discussion on the code.

mlxd:
Just for ease of discussion & review, I have added some of the benchmark data as a plot
image

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    template <class PrecisionT>
+    static inline void PauliX_thread(std::complex<PrecisionT>* data, size_t L, size_t R, size_t wire_parity_inv, size_t wire_parity, size_t rev_wire_shift){
+        
+        size_t n0;
+        size_t i00;
+        size_t i10;
+    
+        for(size_t k=L; k<R;k=k+2){
+            #pragma unroll
```
mlxd, line 53: What advantage is there for adding loop unrolling here?

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+            for(size_t l=0;l<2;l++){
+                n0 = k + l;
+                i00 = (((n0 << 1U) & wire_parity_inv )| (wire_parity & n0));
+                i10 = i00 | rev_wire_shift;
+                std::swap(data[i00],data[i10]);
+            }
```
mlxd, line 59: Can you tell me what is happening in this loop?

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+        using Util::fillLeadingOnes, Util::fillTrailingOnes;
+        assert(wires.size() == 1);
+        const size_t rev_wire = num_qubits - wires[0] - 1;
+        const size_t rev_wire_shift = (static_cast<size_t>(1U) << rev_wire);
+        const size_t wire_parity = fillTrailingOnes(rev_wire);
+        const size_t wire_parity_inv = fillLeadingOnes(rev_wire + 1);
+
+        const size_t N = Util::exp2(num_qubits-1);
+
+        //size_t num_threads = std::thread::hardware_concurrency();//Get the number of threads available
```
mlxd, line 78: Why not use the maximum available hardware threads for this?

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+        size_t num_threads = 2;
+
+        std::vector<std::thread> workers;
+        std::vector<size_t> limits = bounds(num_threads,N);
+
+        //lauch a group of threads
+        for(size_t i=0;i<num_threads;i++){
+            workers.push_back(std::thread(PauliX_thread<PrecisionT>,data, limits[i], limits[i+1],wire_parity_inv, wire_parity, rev_wire_shift));
```
mlxd, line 87: Could there be another way to add threads to a structure such as `std::vector` here?

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    template <class PrecisionT>
+    static inline void PauliZ_thread(std::complex<PrecisionT> *data, size_t L, size_t R, size_t wire_parity_inv, size_t wire_parity, size_t rev_wire_shift){
+     size_t i0;
+     for(size_t k=L; k<R;k=k+2){
+         #pragma unroll
+         for(size_t l=0;l<2;l++){
+             size_t n = k + l;
+            i0 = (((n << 1U) & wire_parity_inv )| (wire_parity & n)) | rev_wire_shift;
+            data[i0]*=-1;
```
mlxd, line 104: Same questions as for PauliX earlier

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+        const size_t wire_parity_inv = fillLeadingOnes(rev_wire + 1);
+
+        for (size_t k = 0; k < Util::exp2(num_qubits - 1); k++) {
+            const size_t i0 = ((k << 1U) & wire_parity_inv) | (wire_parity & k);
+            arr[i0] = std::complex<PrecisionT>{0.0, 0.0};
+        }
+        // NOLINTNEXTLINE(readability-magic-numbers)
+        return static_cast<PrecisionT>(1.0);
+    }
+};
```
mlxd, line 160: Is this needed?

pennylane_lightning/src/benchmarks/CMakeLists.txt
```diff
@@ -53,33 +53,44 @@ target_link_libraries(lightning_benchmarks_dependency INTERFACE lightning_utils
 # Add benchmark_utils exe
```
mlxd, line 53: Do these benchmarks need to be commented out here?

pennylane_lightning/src/gates/MyGateImplementation.hpp
```diff
@@ -0,0 +1,160 @@
...
+            bnd.push_back(N2);
+            N1 = N2;
+        }
+        return bnd;
+    }
+
+
+    template <class PrecisionT>
+    static inline void PauliX_thread(std::complex<PrecisionT>* data, size_t L, size_t R, size_t wire_parity_inv, size_t wire_parity, size_t rev_wire_shift){
+
```
mlxd, line 47: Given the codefactor complain here, we use `clang-tidy` and `clang-format` to ensure compliance of our code style. Running `make format` should automatically format your source to the repository style before committing.

mlxd (commented):
Thanks @multiphaseCFD I left some questions and comments.

Overall, I am curious what your thoughts are for why the kernels are slower for lower qubit counts and on-par for faster counts. Do you think there is any way using your implementation to improve their performance compared to the existing implementations?

pennylane_lightning/src/gates/MyGateImplementation.hpp
mlxd: Also, this fails to be recognised on our CI tests as a valid option; we compile using clang-tidy to test and with warnings enabled as errors.

AmintorDusko:
Hi @multiphaseCFD. Could you please elaborate on your sentence below?
"The reason why adding an additional threads can only lead to 10% speedup for the large number qubits operations is because both Pauli X/Z
operation are memory bounded but not computation bounded."

mlxd:
Hi @multiphaseCFD just wondering why this PR was closed and deleted? We should still be able to discuss here.

pennylane_lightning/src/benchmarks/Bench_ApplyMyImpl.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+        benchmark::DoNotOptimize(sv.getDataVector()[0]);
+        benchmark::DoNotOptimize(sv.getDataVector()[(1 << num_qubits) - 1]);
+    }
+}
+
+BENCHMARK_APPLYOPS(applyOperation_PauliXLM, float, LM, Kernel::LM)
+    ->ArgsProduct({
+        benchmark::CreateDenseRange(6, 30, /*step=*/6), // num_qubits
+        benchmark::CreateDenseRange(2, 4, /*step=*/1),  // num_wires
```
AmintorDusko, line 127: @multiphaseCFD, in the benchmarks you tested the application of the gate to different wires.
What did you expect to observe with this?

AmintorDusko: Also, it would perhaps be interesting to observe the scaling with the number of threads.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: Thanks

multiphaseCFD:
> Hi @multiphaseCFD. Could you please elaborate on your sentence below? "The reason why adding an additional threads can only lead to 10% speedup for the large number qubits operations is because both Pauli X/Z operation are memory bounded but not computation bounded."

The computation ops : bytes number transfer from/to memory can be 4:1, which means theoretically cpu core can do 16 arithmetic operations while reading one floating point data. 
However, for Pauli X operation, it only needs to do 3 arithmetic operations after reading/writing 2 floating point number (4 times memory access). It is also the same for Pauli Z.

pennylane_lightning/src/benchmarks/Bench_ApplyMyImpl.cpp
multiphaseCFD: What I expected is we can get better acceleration when then wire number is less than 4.  Increasing the number of threads would not lead to speed.

pennylane_lightning/src/benchmarks/CMakeLists.txt
multiphaseCFD: No, but I just wanted to compile what I needed.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: Needed to compile it on my system.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: Refer to the answer above.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: There could be a more efficient way to launch thread. But I am not sure because I am also new to multithreading programming.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: The best performance I got was using two threads. On the one hand, using maximum number of threads bring the overhead caused by threads (attached to the same core) switches. On the other hand, as I said before, these operations are memory access limited and the way to speed up the operations is to either hide memory access latency or increase effective memory throughput.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: This is a really nice implementation to use bitwise operation to convert the looping index to the index of the qubit array.

pennylane_lightning/src/gates/MyGateImplementation.hpp
multiphaseCFD: Launching/destroying threads needs time and the time spent on it is much longer than the time reduced on the qubit operation by using threads when the qubit counts is low. When the qubit counts is larger, the overhead caused by launching/destroying threads will be shorter than the time reduced on the qubit operation by using threads when the qubit counts is low.  Thanks @chaeyeunpark

multiphaseCFD:
> Thanks @multiphaseCFD I left some questions and comments.
> 
> Overall, I am curious what your thoughts are for why the kernels are slower for lower qubit counts and on-par for faster counts. Do you think there is any way using your implementation to improve their performance compared to the existing implementations?

Launching/destroying threads needs time and the time spent on it is much longer than the time reduced on the qubit operation by using threads when the qubit counts is low. When the qubit counts is larger, the overhead caused by launching/destroying threads will be shorter than the time reduced on the qubit operation by using threads when the qubit counts is low.

chaeyeunpark (changes requested):
Hi @multiphaseCFD,

as @mlxd addressed lots of questions, I hereby drop a general comment (with just a single change request). 

Is there a specific reason why you consider the speed-up is not achievable due to memory bandwidth (instead of other reasons such as thread overhead/poor cache optimization etc.)?

PR #274, opened 2022-03-24
Splitting matrix operations, refactor dispatcher, add a benchmark suite for all gates/generators/matrices

chaeyeunpark:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:** While working with kernels (two existing and AVX kernels to be added), it is found that implementing the general `applyMatrix` method that works for all different numbers of wires is rather difficult/impractical. Thus this PR splits this method into three (applySingleQubitOp/applyTwoQubitOp/applyMultiQubitOp). In addition, a dispatching procedure is refactored to separate a module for selecting the best kernel (which is now only done in `StateVector`) and calling a kernel function (completely done in the `DynamicDispatcher` class). Benchmark suites are updated accordingly.

**Description of the Change:**

**Benefits:** Faster `applyMatrix` for general single-qubit and two-qubit gates. Easy to add a mechanism to choose the best-performing kernel for the given number of qubits (one of the following PRs). Easy to benchmark all gate/generator/matrix operations.

**Possible Drawbacks:** 

**Related GitHub Issues:**

PR #283, opened 2022-04-20
Implements native support for Sparse Hamiltonians expval() calculation with Kokkos

AmintorDusko:
**Context:**

This PR implements native support for sparse Hamiltonians expval calculation.
Sparse operations are offloaded to Kokkos.

Add native support for the calculation of sparse Hamiltonians' expectation values.

**Benefits:**

Performant calculation with Pennylane-Lightning

**Possible Drawbacks:**

Not that I'm aware of.

**Related GitHub Issues:**

Not that I'm aware of.


chaeyeunpark:
Hi @AmintorDusko, I think Kokkos does not support MSVC. We may need to try clang-cl instead... I will ping you on slack.

AmintorDusko:
> Hi @AmintorDusko, I think Kokkos does not support MSVC. We may need to try clang-cl instead... I will ping you on slack.

Yes. Its looks like it! Thank you very much @chaeyeunpark!
Also, your `add_kokkos` fork was quite useful!
Thank you for looking into it before me!

CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ option(ENABLE_WARNINGS "Enable warnings" ON)
 option(ENABLE_NATIVE "Enable native CPU build tuning" OFF)
 option(ENABLE_AVX "Enable AVX support" OFF)
 option(ENABLE_OPENMP "Enable OpenMP" ON)
+option(ENABLE_THREADS "Enable threads" ON)
```
mlxd, line 53: Is this something we want the user to specify directly, or should we implicitly set it internally based on the build type?

pennylane_lightning/lightning_qubit.py
```diff
@@ -17,8 +17,15 @@
...
-from os import getenv
 from itertools import islice
+import platform
+from os import getenv, environ
+
+if platform.system() == "Linux":
+    # Kokkos ask to set OMP_PROC_BIND=spread and OMP_PLACES=threads
+    # for general best performance with OpenMP 4.0 or later.
+    environ["OMP_PROC_BIND"] = "spread"
+    environ["OMP_PLACES"] = "threads"
```
mlxd, line 28: I do not think we want to set these variables. Those are for the user to decide.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -157,6 +160,35 @@ class Measures {
...
+     * @param values  non-zero elements.
+     * @return fp_t
+     */
+    fp_t expval(const std::vector<Util::index_type> &row_map,
+                const std::vector<Util::index_type> &entries,
+                const std::vector<CFP_t> &values) {
+        if (original_statevector.getLength() != (row_map.size() - 1)) {
+            throw std::invalid_argument(
+                "Statevector and Hamiltonian have incompatible sizes.");
+        }
```
mlxd, line 179: This can be replaced with a `PL_ABORT_IF()`

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -157,6 +160,35 @@ class Measures {
...
+        auto operator_vector = Util::apply_Sparse_Matrix(
+            original_statevector.getData(), original_statevector.getLength(),
+            row_map.data(), row_map.size(), entries.data(), values.data(),
+            values.size());
+
+        CFP_t expected_value = Util::innerProdC(
+            original_statevector.getData(), operator_vector.data(),
+            original_statevector.getLength());
+        return std::real(expected_value);
```
mlxd, line 188: Do other parts of the codebase return real or complex here? Might be worth checking and doing the same as they

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,179 @@
...
+// Implementing Kokkos Sparse operations.
+#include <complex>
+#include <vector>
+
+namespace Pennylane::Util {
+using device_type = typename Kokkos::Device<
+    Kokkos::DefaultExecutionSpace,
+    typename Kokkos::DefaultExecutionSpace::memory_space>;
+
+using index_type = int;
```
mlxd, line 43: Kokkos prefers `int` for indices?

setup.py
```diff
@@ -52,13 +52,24 @@ def build_extension(self, ext: CMakeExtension):
...
             f"-DPYTHON_EXECUTABLE={sys.executable}",
-            f"-DCMAKE_MAKE_PROGRAM={ninja_path}",
             "-DENABLE_WARNINGS=OFF",  # Ignore warnings
         ]
 
+        if platform.system() == "Windows":
+            # As Ninja does not support long path for windows yet:
+            #  (https://github.com/ninja-build/ninja/pull/2056)
+            configure_args += [
+                "-G Visual Studio 17 2022",
```
mlxd, line 64: What does this line do? Does it require an explicit version of visual studio to work?

mlxd (changes requested):
Thanks for this @AmintorDusko. For the testing of the long path on windows, it may be fine to add the powershell command from https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell#tabpanel_1_powershell to [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/4d5232ab1757778df2fca95b0ef9c0b1da81faf6/.github/workflows/wheel_win_x86_64.yml#L17). Try it out and let us know. Also, I think the new tests need to be added to the CMake build, as they do not seem to be called.

The issue here seems to be introducing a new build with the Kokkos backend enabled.

pennylane_lightning/lightning_qubit.py
AmintorDusko: OK. I will remove it. Without it, the user will keep getting a Kokkos warning saying that these variables must be set for better performance.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
AmintorDusko: Internally Kokkos performs loops with int. If kept as unsigned int, or size_t, a compilation performed with warnings as errors will fail.

pennylane_lightning/lightning_qubit.py
mlxd: That is fine. I think it is bad practice of Kokkos to expect this, so we should assume it will go away at some point. Putting the responsibility on the user for this is a better design.

If it becomes problematic later, we can check for existence of such variables and only set if they do not exist, but for now I think keeping this off is fine.

setup.py
AmintorDusko: This is the version we use to build the wheels. It may not be necessary, but it seems to be common practice to set the Visual Studio generator before the clang-cl toolchain.

pennylane_lightning/lightning_qubit.py
AmintorDusko: We can set it back later if we find it a good idea.

setup.py
mlxd: Is there a better way to programmatically determine or set this? I don't think we should hardcode the version in here.

CMakeLists.txt
AmintorDusko: Since these options are usually used by developers, I think it is useful to have this option available here.
As it is now, we can make use of threads implicitly with openBLAS.

pennylane_lightning/lightning_qubit.py
AmintorDusko: > 

Sure! Good idea!

setup.py
AmintorDusko: I will try without it and see if this will default to the right generator.

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -320,10 +320,19 @@ void lightning_class_bindings(py::module &m) {
...
+                 const std::string &, const std::vector<size_t> &)>(
+                 &Measures<PrecisionT>::expval),
+             "Expected value of an operation by name.")
+#ifdef _ENABLE_KOKKOS
+        .def("expval",
+             static_cast<PrecisionT (Measures<PrecisionT>::*)(
+                 const std::vector<Util::index_type> &,
+                 const std::vector<Util::index_type> &,
+                 const std::vector<std::complex<PrecisionT>> &)>(
+                 &Measures<PrecisionT>::expval),
```
chaeyeunpark, line 333: I think a performance issue may be from here (and one in `Measures.hpp` file). If we use `std::vector`, pybind always copies the data from Python to internal storage (as `std::vector` should maintain the data). Probably, changing the arguments here to `py::array_t` and using pointers in real implementation.

CMakeLists.txt
mlxd: If this is disabled, we still get a working build?

pennylane_lightning/src/bindings/Bindings.cpp
chaeyeunpark: You may check [this](https://github.com/chaeyeunpark/UnionFind/blob/1002d0bf2ed2139792b9bad39bc0496dd0e497b7/UnionFindPy/cpp/binding/UnionFindPy.cpp#L62-L87) in my personal project that handles a sparse array with pybind.

AmintorDusko: Thank you @chaeyeunpark! I will check!

CMakeLists.txt
AmintorDusko: Yes, we get it. Do you want me to remove it?

pennylane_lightning/src/bindings/Bindings.cpp
AmintorDusko: @chaeyeunpark, could you please clarify what you mean with "one in Measures.hpp file"?

CMakeLists.txt
AmintorDusko: Just to expand into this. I tested in Linux both Threads and OpenMP implementations and both yielded a working build.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -157,6 +160,35 @@ class Measures {
...
+     * @param row_map row_map array.
+     *                The j element encodes the number of non-zeros above row j.
+     * @param entries column indices of the non-zero elements.
+     * @param values  non-zero elements.
+     * @return fp_t
+     */
+    fp_t expval(const std::vector<Util::index_type> &row_map,
+                const std::vector<Util::index_type> &entries,
+                const std::vector<CFP_t> &values) {
+        if (original_statevector.getLength() != (row_map.size() - 1)) {
```
chaeyeunpark, line 176: Hi @AmintorDusko, to prevent data copy, you may need to change the arguments of this function to pointers (`std::span` is more appropriate but we don't have C++20 support yet).

pennylane_lightning/src/bindings/Bindings.cpp
chaeyeunpark: Hi Amintor, I left a comment in a function that may need to be changed.

chaeyeunpark:
Hi Amintor, I also suspect that [these lines](https://github.com/PennyLaneAI/pennylane-lightning/blob/13e1887bba4d47b148c104b63c2267a00ee31544/.github/workflows/wheel_win_x86_64.yml#L58-L61) in a workflow may change the shell to `cmd` from `PowerShell`. As we are using `clang-cl` for Windows, the lines may not be necessary. Could you try after removing these lines?

AmintorDusko:
> Hi Amintor, I also suspect that [these lines](https://github.com/PennyLaneAI/pennylane-lightning/blob/13e1887bba4d47b148c104b63c2267a00ee31544/.github/workflows/wheel_win_x86_64.yml#L58-L61) in a workflow may change the shell to `cmd` from `PowerShell`. As we are using `clang-cl` for Windows, the lines may not be necessary. Could you try after removing these lines?

Sure. I will try.

cmake/process_options.cmake
```diff
@@ -1,12 +1,18 @@
...
+# options and produces interface libraries
 # lightning_compile_options and lightning_external_libs.
 ##############################################################################
 
 # Include this file only once
 include_guard()
 
+if (WIN32)
+    # Increasing maximum full-path length allowed.
+  set(CMAKE_OBJECT_PATH_MAX 512)
```
mlxd, line 13:
```suggestion
  set(CMAKE_OBJECT_PATH_MAX 259)
```

AmintorDusko: Done

setup.py
AmintorDusko: Done!

cmake/process_options.cmake
```diff
@@ -117,3 +123,68 @@ if(ENABLE_BLAS)
...
+    # If no backend is specified, KOKKOS will enable the serial backend by default.
+
+    option(Kokkos_ENABLE_COMPLEX_ALIGN "Enable complex alignment in memory" OFF)
+
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+    include(FetchContent)
+
+    FetchContent_Declare(kokkos
+                         GIT_REPOSITORY https://github.com/kokkos/kokkos.git
+                         GIT_TAG        3.5.00
```
mlxd, line 169: I assume the new release tag 3.6.00 does not improve things here?

chaeyeunpark:
When tested locally, I found that running `pip install -e .` without cibuildwheel already raises FTK2009 (path too long) error. I could remove this error by adding `shell=True`, but other errors started to appear (even in Linux). Let me check this more tomorrow.

AmintorDusko:
> When tested locally, I found that running `pip install -e .` without cibuildwheel already raises FTK2009 (path too long) error. I could remove this error by adding `shell=True`, but other errors started to appear (even in Linux). Let me check this more tomorrow.

Sure! Thank you @chaeyeunpark!

cmake/process_options.cmake
AmintorDusko: So, I checked it before, as soon as it was released, but I wanted to double-check because something could have changed.
It builds, but with a cascade of deprecated warnings from functions called internally by Kokkos.
We may find some improvements, but nothing evident for now.
I suggest keeping it as `3.5.00`, for now.

AmintorDusko: I will change it to 3.6.00.

pennylane_lightning/src/simulator/Measures.hpp
AmintorDusko: Done.

.github/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ dev2 = qml.device('lightning.qubit', wires=4, c_dtype=np.complex128) # for doubl
 * Split matrix operations, refactor dispatch mechanisms, and add a benchmark suits.
 [(#274)](https://github.com/PennyLaneAI/pennylane-lightning/pull/274)
 
+* Add native support for the calculation of sparse Hamiltonians' expectation values. This implementation largely decreases the calculation time. Sparse operations are offloaded to [Kokkos](https://github.com/kokkos/kokkos) and [Kokkos-Kernels](https://github.com/kokkos/kokkos-kernels).
```
mlxd, line 20:
```suggestion
* Add native support for the calculation of sparse Hamiltonians' expectation values. Sparse operations are offloaded to [Kokkos](https://github.com/kokkos/kokkos) and [Kokkos-Kernels](https://github.com/kokkos/kokkos-kernels).
```

I think we can keep the details objective for now, and explore performance in a following write-up

.github/workflows/wheel_win_x86_64.yml
```diff
@@ -14,7 +14,7 @@ env:
 
   # Python build settings
   CIBW_BEFORE_BUILD: |
-    pip install pybind11 ninja cmake
+    pip install pybind11 ninja cmake && powershell.exe New-ItemProperty -Path "HKLM:\\SYSTEM\\CurrentControlSet\\Control\\FileSystem" -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
```
mlxd, line 17: I'll keep working on this in #293  and merge onto this PR when ready,

CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ option(ENABLE_WARNINGS "Enable warnings" ON)
...
+option(ENABLE_KOKKOS "Enable Kokkos" ON) # ON by default during development.
```
mlxd, line 54: Once we decide to release, this should be off by default unless explicitly stated.

Makefile
```diff
@@ -92,6 +92,25 @@ test-cpp-omp:
...
+
+test-cpp-kokkos-serial:
+	rm -rf ./BuildTests
+	cmake $(LIGHTNING_CPP_DIR) -BBuildTests -DBUILD_TESTS=ON -DENABLE_KOKKOS=ON -DENABLE_BLAS=OFF -DENABLE_THREADS=OFF -DENABLE_OPENMP=OFF
+	cmake --build ./BuildTests --target runner
+	cmake --build ./BuildTests --target test
+
+.PHONY: benchmark
+benchmark:
+	cmake --build BuildBench --target clean || true
```
mlxd, line 109: Should this be a `||` or `|`?

cmake/process_options.cmake
```diff
@@ -117,3 +124,38 @@ if(ENABLE_BLAS)
...
+    include(FetchContent)
+
+    FetchContent_Declare(kokkos
+                         GIT_REPOSITORY https://github.com/kokkos/kokkos.git
+                         GIT_TAG        3.6.00
+    )
+  
+    FetchContent_MakeAvailable(kokkos)
+
+    get_target_property(_kokkos_INC_DIR kokkos INTERFACE_INCLUDE_DIRECTORIES)
```
mlxd, line 145: Why use the underscore here?

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,179 @@
...
+ */
+template <class fp_precision>
+std::vector<std::complex<fp_precision>> apply_Sparse_Matrix(
+    const std::complex<fp_precision> *vector_ptr, const index_type vector_size,
+    const index_type *row_map_ptr, const index_type row_map_size,
+    const index_type *entries_ptr, const std::complex<fp_precision> *values_ptr,
+    const index_type numNNZ) {
+
+    std::vector<std::complex<fp_precision>> result;
+    Kokkos::initialize();
```
mlxd, line 132: What kinda of setup happens in the init here? Is there much overhead? Should we move the init to the class ctor and the finalise to the destructor?

mlxd (commented):
Great work @AmintorDusko 
I have a few queries, and some comments.

We can hopefully get the Windows fixes in soon to see this merged.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
AmintorDusko: I made some tests a while ago and it was really light.
I'm not against moving it to an upper layer, though.

CMakeLists.txt
AmintorDusko: For simplicity, in the most recent changes, I removed the ENABLE_THREADS option.
The ENABLE_KOKKOS is probably better to be OFF by default.
I intend to do it before merging. 
It is ON now just for testing, as you know.

.github/workflows/wheel_win_x86_64.yml
AmintorDusko: Sure! Thank you, Lee!

.github/CHANGELOG.md
AmintorDusko: Makes sense! Thank you!

cmake/process_options.cmake
AmintorDusko: It is not necessary. I will remove it.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
mlxd: I think if there is nothing heavy in here, we can keep it in place for now

AmintorDusko:
Hi, @mlxd. I brought what I considered to be the essential changes in your [PR](https://github.com/PennyLaneAI/pennylane-lightning/pull/293). 
These changes are been incorporated as commits starting with `win64_path_fix: `.
Windows wheels are building nicely now! Nice job on that!

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,179 @@
...
+                                        entries_ptr, values_ptr, numNNZ);
+
+        const data_type<fp_precision> alpha(1.0);
+        const data_type<fp_precision> beta;
+        KokkosSparse::spmv("N", alpha, myMatrix, vector_view, beta,
+                           result_view);
+
+        result = std::move(std::vector<std::complex<fp_precision>>(
+            reinterpret_cast<std::complex<fp_precision> *>(result_view.data()),
+            reinterpret_cast<std::complex<fp_precision> *>(result_view.data()) +
```
mlxd, line 150: I assume this reinterpret is due to the Kokkos complex vs std::complex issues we discussed previously?

mlxd (changes requested):
Thanks @AmintorDusko 
I have some comments and suggestions here.

cmake/process_options.cmake
```diff
@@ -117,3 +124,38 @@ if(ENABLE_BLAS)
...
+    set_target_properties(kokkos PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${_kokkos_INC_DIR}")
+
+    FetchContent_Declare(kokkoskernels
+                         GIT_REPOSITORY https://github.com/kokkos/kokkos-kernels.git
+                         GIT_TAG        3.6.00
```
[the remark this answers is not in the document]
AmintorDusko: I removed the underscore but somehow it came back. Magic, I guess.
Regarding clang-tidy: Yes! Several deprecation warnings.

pennylane_lightning/lightning_qubit.py
```diff
@@ -576,6 +575,14 @@ def expval(self, observable, shot_range=None, bin_size=None):
 
         state_vector = StateVectorC64(ket) if self.use_csingle else StateVectorC128(ket)
         M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+        if observable.name == "SparseHamiltonian":
+            # converting COO to CRS sparse representation.
+            CSR_SparseHamiltonian = observable.data[0].tocsr(copy=False)
```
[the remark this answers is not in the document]
AmintorDusko: Sorry. My bad! I'm trying to keep it consistent with CSR, this is the one used by Kokkos.

cmake/process_options.cmake
AmintorDusko: It came back somehow. I will remove it, again.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,179 @@
...
+#ifdef __has_include
+#if __has_include(                                                             \
+    <Kokkos_Core.hpp>) && __has_include (<KokkosKernels_default_types.hpp>)
+#include "KokkosKernels_default_types.hpp"
+#include "KokkosSparse_CrsMatrix.hpp"
+#include "KokkosSparse_spmv.hpp"
+#include "Kokkos_Core.hpp"
+#define _ENABLE_KOKKOS
+#endif
+#endif
```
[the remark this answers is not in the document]
AmintorDusko: Good catch! Thanks, will fix it!

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,179 @@
...
+    const_index_view_type entries(entries_ptr, numNNZ);
+    const_data_view_type<fp_precision> values(
+        reinterpret_cast<const Kokkos::complex<fp_precision> *>(values_ptr),
+        numNNZ);
+
+    const_graph_type<fp_precision> myGraph(entries, row_map);
+    const_crs_matrix_type<fp_precision> SparseMatrix("matrix", numRows, values,
+                                                     myGraph);
+    return SparseMatrix;
+}
```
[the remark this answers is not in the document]
chaeyeunpark: Hi @Lee, I assume that this is from my request: https://github.com/PennyLaneAI/pennylane-lightning/pull/283#discussion_r867057984

A sparse Hamiltonian can be generally huge and we may not want to make a copy of this. In this case, using the raw pointer from numpy might be an appropriate solution until we have `std::span` from C++20. 

We probably need to revisit all C++20 related issues next Q

pennylane_lightning/src/bindings/Bindings.cpp
```diff
@@ -321,10 +327,25 @@ void lightning_class_bindings(py::module &m) {
...
+                return M.expval(
+                    static_cast<Util::index_type *>(row_map.request().ptr),
+                    static_cast<Util::index_type>(row_map.request().size),
+                    static_cast<Util::index_type *>(entries.request().ptr),
+                    static_cast<std::complex<PrecisionT> *>(
+                        values.request().ptr),
+                    static_cast<Util::index_type>(values.request().size));
+            },
+            "Expected value of a sparse Hamiltonian.")
+#endif
```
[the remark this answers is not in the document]
AmintorDusko: I'm refactoring this part of the code.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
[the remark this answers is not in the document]
mlxd: He @chaeyeunpark I wasn't aware of your comment, but thanks for the linking.
If this seems the most sensible option, then I am fine with it for now (as you say, span should help improve on this).

pennylane_lightning/src/tests/Test_Measures.cpp
```diff
@@ -45,6 +48,64 @@ StateVectorManaged<T> Initializing_StateVector() {
     return Measured_StateVector;
 }
 
+#ifdef _ENABLE_KOKKOS
```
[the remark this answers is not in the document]
AmintorDusko: I agree.

tests/test_measures.py
```diff
@@ -227,6 +228,28 @@ def circuit():
...
+        @qml.qnode(dev)
+        def circuit():
+            qml.RX(0.4, wires=[0])
+            qml.RY(-0.2, wires=[1])
+            return qml.expval(qml.SparseHamiltonian(coo_matrix(qml.matrix(cases[0])), wires=[0, 1]))
+
+        assert np.allclose(circuit(), cases[1], atol=tol, rtol=0)
+
     def test_value(self, dev, tol):
         """Test that the expval interface works"""
```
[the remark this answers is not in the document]
AmintorDusko: I will expand the tests. Just as a note, this is the only place that will need to come back later if/when my PR in PennyLane is merged.

trevor-vincent:
Awesome work Amintor! I see that you have enabled the serial device. Any chance we can get benchmarks for the OPENMP device? We could potentially see great speedups here.

AmintorDusko:
> Awesome work Amintor! I see that you have enabled the serial device. Any chance we can get benchmarks for the OPENMP device? We could potentially see great speedups here.

Hi @trevor-vincent! I made several benchmarks comparing Serial, OpenMP and Threads devices.
Because the calculation of a single expval for Sparse matrix is very fast (even in a single thread), adding any type of multi-threading only incorporates overhead.
The total time is sizeable, because the single expval is calculated hundreds of times, and because of other device operations.

AmintorDusko:
> Just one more question as I am not sure how Kokkos works: Will this functionality also use GPU when we set Kokkos to use GPU?

Hi @chaeyeunpark. In theory, functionalities implemented with Kokkos are device-independent.
Having said that, maybe we would need to add some extra steps (memory-related) to accomplish this. 
Without access to GPUs, it is difficult to be 100% sure.

mlxd:
> > Just one more question as I am not sure how Kokkos works: Will this functionality also use GPU when we set Kokkos to use GPU?
> 
> Hi @chaeyeunpark. In theory, functionalities implemented with Kokkos are device-independent. Having said that, maybe we would need to add some extra steps (memory-related) to accomplish this. Without access to GPUs, it is difficult to be 100% sure.

I think this is a good one to come back to later once @trevor-vincent has finished his work with Kokkos. It may be straight forward to take advantage of any other devices, since we can technically target an entire collection of backends (CUDA, HIP, SYCL, HPX, OpenMP and C++ threads). We can try to target SYCL as the backend and it may pick-up Intel GPUs for some testing locally.

chaeyeunpark:
Thanks, @mlxd @AmintorDusko, for the answers. However, I was asking as I wondered whether it is possible to use GPU backends for quantum gate operations but CPU for sparse operations (or vice versa). I was thinking about the strategy to integrate Kokkos kernels (for gates) with the current codebase. Of course, as @mlxd said, we should discuss it more when the Kokkos kernel is available, but we may need to decide who is responsible for choosing the device for Kokkos (we or users?).

AmintorDusko:
> Thanks, @mlxd @AmintorDusko, for the answers. However, I was asking as I wondered whether it is possible to use GPU backends for quantum gate operations but CPU for sparse operations (or vice versa). I was thinking about the strategy to integrate Kokkos Kernels (for gates) with the current codebase. Of course, as @mlxd said, we should discuss it more when the Kokkos kernel is available, but we may need to decide who is responsible for choosing the device for Kokkos (we or users?).

Hi @chaeyeunpark, thanks for your clarification.
Yes, one can design the code so that part of it is in the CPU (host), and the other part is in the GPU. 
You can define more than one device. Maybe toasters in the future (@trevor-vincent ?)

.github/workflows/tests_linux.yml
```diff
@@ -132,6 +132,66 @@ jobs:
...
+          files: Build/tests/results/report.xml
+
+      - name: Build and run unit tests for code coverage
+        run: |
+            cmake pennylane_lightning/src -BBuildCov  -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON -DENABLE_KOKKOS=ON
+            cmake --build ./BuildCov
+            cd ./BuildCov
+            ./tests/runner
+            lcov --directory . -b ../pennylane_lightning/src --capture --output-file coverage.info
+            lcov --remove coverage.info '/usr/*' --output-file coverage.info
```
mlxd, line 187: In hindight, we should be able to do this as part of the previous test, rather than compile and run the suite twice. Also, why do we need the debug flag here?

If it is too hard to combine the two, then maybe creating this as a separate test suite to run in parallel is a good idea.

mlxd (commented):
Hey Amintor, I have a few more comments/questions.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,198 @@
...
+void apply_Sparse_Matrix_Kokkos(
+    [[maybe_unused]] const std::complex<fp_precision> *vector_ptr,
+    [[maybe_unused]] const index_type vector_size,
+    [[maybe_unused]] const index_type *row_map_ptr,
+    [[maybe_unused]] const index_type row_map_size,
+    [[maybe_unused]] const index_type *entries_ptr,
+    [[maybe_unused]] const std::complex<fp_precision> *values_ptr,
+    [[maybe_unused]] const index_type numNNZ,
+    [[maybe_unused]] std::vector<std::complex<fp_precision>> &result) {
+    PL_ABORT("Executing the product of a Sparse matrix and a vector needs "
```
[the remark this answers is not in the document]
AmintorDusko: This exception is only tested in C++. Because in Python we call the `default.qubit` method if not compiled with Kokkos.

.github/workflows/tests_linux.yml
```diff
@@ -132,12 +132,66 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get -y -q install cmake gcc g++ gcovr lcov
+
+      - name: Build and run unit tests for code coverage
+        run: |
+            cmake pennylane_lightning/src -BBuildCov  -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON -DENABLE_KOKKOS=ON
```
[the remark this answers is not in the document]
AmintorDusko: Sure!

pennylane_lightning/src/util/Kokkos_Sparse.hpp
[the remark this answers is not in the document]
mlxd: We should be able to explicitly import the module though and call this method, right? That should get rid of this error.

mlxd (changes requested):
Awesome work on this Amintor! One last thing we need is to update the license statement at the end of the Lightning README to indicate we are using Kokkos under the BSD license (see Jet's statement [here](https://github.com/XanaduAI/jet#license) for an example).

mlxd (approved):
:100:

Makefile
```diff
@@ -92,6 +92,19 @@ test-cpp-omp:
...
+	rm -rf ./BuildBench/CMakeCache.txt ./BuildBench/compiler_info.txt ./BuildBench/run_gate_benchmark.sh
+	cmake $(LIGHTNING_CPP_DIR) -BBuildBench -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -DENABLE_AVX=ON
+	cmake --build ./BuildBench
+
```
maliasadi, line 107: This part is already removed from Makefile (#274). There is no `run_gate_benchmark.sh` file anymore.

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -157,6 +159,37 @@ class Measures {
...
+     *                      The j element encodes the number of non-zeros above
+     * row j.
+     * @param row_map_size  row_map array size.
+     * @param entries_ptr   pointer to an array with column indices of the
+     * non-zero elements.
+     * @param values_ptr    pointer to an array with the non-zero elements.
+     * @param numNNZ        number of non-zero elements.
+     * @return fp_t
+     */
+    template <class index_type>
```
maliasadi, line 175: Could you please add a brief info about `index_type` using `@tparam`?

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -157,6 +159,37 @@ class Measures {
...
+    fp_t expval(const index_type *row_map_ptr, const index_type row_map_size,
+                const index_type *entries_ptr, const CFP_t *values_ptr,
+                const index_type numNNZ) {
```
maliasadi, line 178: Not sure if this is a good idea as I don't know how much you need to use such long lists of inputs. But, do you think that encapsulating these inputs (e.g., in a struct) would make this function signature cleaner? It not, forget my comment :slightly_smiling_face:

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.16)
```
maliasadi, line 1: You may also want to update this version number in other CMakeLists.txt files.

pennylane_lightning/src/tests/TestHelpers.hpp
```diff
@@ -290,6 +291,100 @@ auto createParams(Gates::GateOperation op) -> std::vector<PrecisionT> {
...
+    return Measured_StateVector;
+}
+
+/**
+ * @brief Fills the empty vectors with the CSR (Compressed Sparse Row) sparse
+ * matrix representation for a tridiagonal + periodic boundary conditions
+ * Hamiltonian.
+ *
+ * @tparam fp_precision
+ * @tparam index_type
```
maliasadi, line 337: Would you complete adding some info for these along with other missing `@tparam`?

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,198 @@
...
+ * @param vector_ptr    pointer to the vector.
+ * @param vector_size   size of the vector.
+ * @param row_map_ptr   pointer to the row_map.
+ *                      The j element encodes the number of non-zeros above row
+ * j.
+ * @param row_map_size  number of elements in the row_map.
+ * @param entries_ptr   pointer to the column indices of the non-zero elements.
+ * @param values_ptr    non-zero elements.
+ * @param numNNZ        number of non-zero elements.
+ * @return std::vector<data_type<fp_precision>>
```
maliasadi, line 121: `result` is missing here and other places.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -0,0 +1,198 @@
...
+    [[maybe_unused]] const index_type *entries_ptr,
+    [[maybe_unused]] const std::complex<fp_precision> *values_ptr,
+    [[maybe_unused]] const index_type numNNZ,
+    [[maybe_unused]] std::vector<std::complex<fp_precision>> &result,
+    bool throw_exception = true) {
+    if (throw_exception) {
+        PL_ABORT("Executing the product of a Sparse matrix and a vector needs "
+                 "Kokkos and Kokkos Kernels installation.");
+    }
+};
```
maliasadi, line 180: how does this function calculate the product? I maybe missing something, but it seems that it doesn't do anything.

maliasadi (commented):
Hey @AmintorDusko. great job! :100:  Sorry for this late review; I was thinking this one is still work-in-progress :slightly_frowning_face: As this one already merged and you also merged the follow-up PR (#296). I'll go with a few general comments to be applied for later...

PR #288, opened 2022-04-28
Fix the issue with using available clang-format version in format

maliasadi:
`make format` (and `make format-cpp`) now accepts the version number in case `clang-format` is not installed or not satisfying the default version -- that is currently `12`. This update avoids inconsistencies between different clang-format aliases and eases trying different versions when needed.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -74,4 +74,4 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Run formatter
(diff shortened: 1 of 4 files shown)


mlxd (commented):
Awesome work @maliasadi 

I re-enabled the failed tests. Once They pass I'll approve

chaeyeunpark (approved):

mlxd (approved):
:100:

PR #291, opened 2022-05-07
New kernel dispatch mechanism

chaeyeunpark:
### Before submitting

Context: Currently, we have two different gate implementations (kernels) and some more to be added. As it is possible that one kernel outperforms other kernels only for particular numbers of qubits, it is desirable to have a kernel dispatcher mechanism that takes this information into account. In addition, as with AVX kernels, it is necessary to have an array whose memory is managed by lightning with a proper alignment.

Description of the Change: Lightning now has the KernelMap class which returns the kernels to use for a given number of qubits. The statevector classes now manage their own kernel information accordingly. Moreover, custom `_asarray` is implemented.

Benefits: Not huge right now, but with AVX and OpenMP kernels, we should have some advantages.

Possible Drawbacks: Slightly slower dispatch time (1 or 2 more std::unoreder_map accesses), and constructing a statevector becomes slower as it needs to obtain the best kernels for each operation.

Related GitHub Issues:

Note that I have added  `StateVectorCPU` class as a parent class of `StateVectorManaged` and `StateVectorRaw`, which are now renamed to `StateVectorManagedPU` and `StateVectorRawCPU`, respectively. This renaming resulted in lots of changed files.


pennylane_lightning/src/bindings/Bindings.hpp
```diff
@@ -89,12 +93,90 @@ auto createManaged(
...
+ * @brief Create an aligned numpy array for a given type, memory model and array
+ * size.
+ *
+ * @tparam T Datatype of numpy array to create
+ * @param memory_model Memory model to use
+ * @param size Size of the array to create
+ * @return Numpy array
+ */
+template <typename T>
+auto alignedNumpyArray(CPUMemoryModel memory_model, size_t size)
```
mlxd, line 136: Just for my own understanding, numpy will maintain ownership of the data returned from this array?
Would it be possible to use a unique_ptr here, or does that break the memory model?

pennylane_lightning/src/simulator/StateVectorCPU.hpp
```diff
@@ -0,0 +1,166 @@
...
+ * @tparam Derived Derived class for CRTP.
+ */
+template <class PrecisionT, class Derived>
+class StateVectorCPU : public StateVectorBase<PrecisionT, Derived> {
+  public:
+    using ComplexPrecisionT = std::complex<PrecisionT>;
+
+  protected:
+    const Threading threading_;
+    const CPUMemoryModel memory_model_;
```
mlxd, line 37: :100: Great to get these in here

pennylane_lightning/src/util/BitUtil.hpp
```diff
@@ -171,6 +172,42 @@ inline auto log2PerfectPower(unsigned long val) -> size_t {
...
+ */
+constexpr auto constLog2PerfectPower(size_t value) -> size_t {
+    if (value == 0) {
+        return 0; // not well defined. TODO: Raise an exception instead in
+                  // a later version.
+    }
+    size_t n = 0;
+    while ((value & 1U) == 0U) {
+        value >>= 1U;
+        ++n;
```
mlxd, line 188:
```suggestion
        n++;
```

pennylane_lightning/src/util/IntegerInterval.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <cassert>
+#include <limits>
+#include <type_traits>
+
+namespace Pennylane::Util {
+
+/**
+ * @brief Define integer interval [min_, max_)
+ */
+template <typename IntegerType> class IntegerInterval {
```
mlxd, line 27: I like this. Good idea.

pennylane_lightning/src/util/Memory.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * Note that alignment must be larger than max_align_t. Otherwise, the behavior
+ * is undefined.
+ *
+ * @param alignment Alignment value we want for the data pointer
+ * @param bytes Number of bytes to allocate
+ * @return Memory pointer
+ */
+inline auto alignedAlloc(uint32_t alignment, size_t bytes) -> void * {
+#if defined(__clang__)
+    /* Apple clang does not support std::aligned_alloc in Mac 10.14.
```
mlxd, line 39: I wonder if we should just deprecate MacOS 10.14 and opt for a 10.15 minimum.
Good call on catching this anyway.

pennylane_lightning/src/util/Memory.hpp
```diff
@@ -0,0 +1,188 @@
...
+     * Thus we use Posix function instead. */
+    void *p;
+    posix_memalign(&p, alignment, bytes);
+    return p;
+#elif defined(_MSC_VER)
+    return _aligned_malloc(bytes, alignment);
```
mlxd, line 45: Interesting too that Windows does not default to cpp standard here.

pennylane_lightning/src/util/Memory.hpp
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @brief Constructor of AlignedAllocator class
+     *
+     * @param alignment Memory alignment we want.
+     */
+    constexpr explicit AlignedAllocator(uint32_t alignment)
+        : alignment_{alignment} {
+        // We do not check input now as it doesn't allow the constructor to be
+        // a constexpr.
+        // TODO: Using exception is allowed in GCC>=10
```
mlxd, line 84: Should we bump the GCC builder image? Ubuntu 22.04 should have GCC 11 by default.

pennylane_lightning/src/util/TypeTraits.hpp
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+/**
+ * @file
+ * Contains type traits
+ */
+#pragma once
+#include <complex>
+#include <type_traits>
+namespace Pennylane::Util {
+// Enable until C++20 support is explicitly allowed
```
mlxd, line 22: If we are bumping GCC and the Ubuntu builder, should we allow this? I assume we may need to revisit MacOs support if so.

mlxd (commented):
Awesome work @chaeyeunpark 
I have a few questions and comments below:

pennylane_lightning/src/bindings/Bindings.hpp
chaeyeunpark: Yes, one can assign a custom function as a destructor of a NumPy array. A problem with `std::unique_ptr` is that NumPy does not know how to treat it as an array.

pennylane_lightning/src/bindings/Bindings.hpp
```diff
@@ -158,6 +239,40 @@ void registerGatesForStateVector(PyClass &pyclass) {
     });
 }
 
+/**
+ * @brief Get a gate kernel map for a statevector
+ */
+template <class PrecisionT>
+auto svKernelMap(const StateVectorRawCPU<PrecisionT> &sv) -> pybind11::dict {
+    pybind11::dict res_map;
```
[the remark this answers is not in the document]
chaeyeunpark: Using `std::unordered_map` also makes sense. I just thought that using `py::dict` can be more efficient as it does not convert `unordered_map` to python type (but construction/assignment may be slower). I also don't know which one is better.

pennylane_lightning/src/util/TypeTraits.hpp
chaeyeunpark: Yes, we may move all other build systems to GCC>=10. As the latest LTS version of Ubuntu has GCC11 as default (and we need to move C++20 somewhen), I think we need to do this in the not-so-far future. Still, the real problem is `manylinux2014` image we use for `cibuildwheel` (https://github.com/pypa/cibuildwheel/discussions/984).

pennylane_lightning/src/util/Memory.hpp
chaeyeunpark: Yes, that's ture. I will try to use std::aligned_malloc in MacOS 10.15

pennylane_lightning/src/util/Memory.hpp
chaeyeunpark: Yes, the underlying reason is that the standard requires that even aligned memory can be deallocated with `std:free`, but this is impossible on Windows. The alignment business is quite tricky yet.

pennylane_lightning/src/simulator/KernelMap.cpp
```diff
@@ -0,0 +1,191 @@
...
+    instance.assignKernelForOp(GateOperation::Toffoli, all_threading,
+                               all_memory_model, all_qubit_numbers,
+                               Gates::KernelType::PI);
+    instance.assignKernelForOp(GateOperation::CSWAP, all_threading,
+                               all_memory_model, all_qubit_numbers,
+                               Gates::KernelType::PI);
+    instance.assignKernelForOp(GateOperation::MultiRZ, all_threading,
+                               all_memory_model, all_qubit_numbers,
+                               Gates::KernelType::LM);
+    return 1;
```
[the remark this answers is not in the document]
chaeyeunpark: No special reason. The return value is anyway unused, but ensure this code is executed before the main (when the library is loaded).

pennylane_lightning/src/simulator/KernelMap.hpp
```diff
@@ -0,0 +1,323 @@
...
+  public:
+    [[nodiscard]] bool
+    conflict(uint32_t test_priority,
+             const Util::IntegerInterval<size_t> &test_interval) const {
+        const auto test_elt = DispatchElement{test_priority, test_interval,
+                                              Gates::KernelType::None};
+        const auto [b, e] =
+            std::equal_range(ordered_vec_.begin(), ordered_vec_.end(), test_elt,
+                             higher_priority);
+        for (auto iter = b; iter != e; ++iter) {
```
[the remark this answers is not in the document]
chaeyeunpark: Here, as `iter` is an iterator, `++iter` and `iter++` produce a different machine code (they just call different functions). Thus I would like to use the prefix version here.

pennylane_lightning/src/tests/CompareVector.hpp
```diff
@@ -0,0 +1,18 @@
...
+    if (lhs.size() != rhs.size()) {
+        return false;
+    }
+    for (size_t idx = 0; idx < lhs.size(); idx++) {
+        if (lhs[idx] != rhs[idx]) {
+            return false;
+        }
+    }
+    return true;
+}
```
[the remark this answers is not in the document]
chaeyeunpark: This is only used inside the test code. To test approximation in test code, `PLApprox` already exists.

pennylane_lightning/src/util/IntegerInterval.hpp
chaeyeunpark: :+1:

pennylane_lightning/src/util/Memory.hpp
chaeyeunpark: As I mention below, there will be a subtle issue in `cibuildwheel`.

pennylane_lightning/src/simulator/KernelMap.hpp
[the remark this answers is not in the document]
mlxd: Ah, I missed that. Sure, makes sense here then

pennylane_lightning/src/util/TypeTraits.hpp
mlxd: Hmmm, yea, always the issue with the build-wheel. I wonder if it makes sense for us to roll our own wheel-builder based on this image. We already do this in part for LightningGPU, so it wouldn't be too far a stretch to add one with a new toolchain if we need to.

pennylane_lightning/src/util/BitUtil.hpp
chaeyeunpark: Updated

pennylane_lightning/src/util/Memory.hpp
chaeyeunpark: Done

pennylane_lightning/src/simulator/Threading.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "Macros.hpp"
+
+#include <cstdint>
+
+#ifdef PL_USE_OMP
+#include <omp.h>
+#endif
+
+namespace Pennylane {
+enum class Threading : uint8_t {
```
maliasadi, line 30: :100: Great idea to have this!

pennylane_lightning/src/simulator/Threading.hpp
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @return Threading
+ */
+inline auto bestThreading() -> Threading {
+#ifdef PL_USE_OMP
+    if (omp_in_parallel() != 0) {
+        // We are already inside of the openmp parallel region (e.g.
+        // inside adjoint diff).
+        return Threading::SingleThread;
+    }
```
maliasadi, line 60: For my understanding, why the adjoing diff would break without this condition? To me this return value is a bit misleading.

pennylane_lightning/src/simulator/StateVectorManagedCPU.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include "Threading.hpp"
+#include "Util.hpp"
+
+namespace Pennylane {
+
+/**
+ * @brief StateVector class where data resides in CPU memory. Memory ownership
+ * resides within class.
+ *
+ * @tparam PrecisionT
```
maliasadi, line 29:
```suggestion
 * @tparam PrecisionT Precision data type
```

pennylane_lightning/src/simulator/StateVectorManagedCPU.hpp
```diff
@@ -0,0 +1,149 @@
...
+        : BaseType{num_qubits, threading, memory_model},
+          data_{Util::exp2(num_qubits), ComplexPrecisionT{0.0, 0.0},
+                getAllocator<ComplexPrecisionT>(this->memory_model_)} {
+        data_[0] = {1, 0};
+    }
+
+    /**
+     * @brief Construct a statevector from another statevector
+     *
+     * @param other Another statevector to construct the statevector from
```
maliasadi, line 63: Please add some description for template params here and other places in this file...

maliasadi (changes requested):
Great work @chaeyeunpark! I have a few comments up to the tests files.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
```diff
@@ -26,7 +26,7 @@
 #include "Error.hpp"
 #include "JacobianTape.hpp"
 #include "LinearAlgebra.hpp"
-#include "StateVectorManaged.hpp"
+#include "StateVectorManagedCPU.hpp"
```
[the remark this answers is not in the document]
chaeyeunpark: Yes, but files in the `examples` directory is absolute. It remains there as it is from @isschoch's contribution. If it is fine for you @isschoch, can we delete this directory? All benchmarks are now available with GoogleBenchmark in the `benchmarks` subdirectory (partially base on your code).

pennylane_lightning/src/simulator/Threading.hpp
chaeyeunpark: :heart_eyes:

pennylane_lightning/lightning_qubit.py
```diff
@@ -123,6 +127,21 @@ def __init__(self, wires, *, c_dtype=np.complex128, shots=None, batch_obs=False)
...
+    @staticmethod
+    def _asarray(arr, dtype=None):
+        arr = np.asarray(arr)  # arr is not copied
+        if not dtype:
+            dtype = arr.dtype
+
+        # We allocate a new aligned memory and copy data to there if alignment or dtype mismatches
+        # Note that get_alignment does not neccsarily returns CPUMemoryModel(Unaligned) even for
+        # numpy allocated memory as the memory location happens to be aligend.
+        if int(get_alignment(arr)) < int(best_alignment()) or arr.dtype != dtype:
```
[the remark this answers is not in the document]
chaeyeunpark: Thanks!

pennylane_lightning/src/util/TypeTraits.hpp
chaeyeunpark: I looked for some more information and it seems like moving to GCC10 is not quite difficult. Do you think is it reasonable to go to GCC10 now? Then I will try this.

mlxd: Yea, I think we can go for this, at least for a trial. Maybe create a separate PR so we can compare how the changes influence Win & MacOS builds too.

pennylane_lightning/src/algorithms/AdjointDiff.hpp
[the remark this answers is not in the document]
isschoch: Works for me @chaeyeunpark

pennylane_lightning/src/bindings/Bindings.hpp
```diff
@@ -89,12 +93,90 @@ auto createManaged(
...
+/**
+ * @brief Create numpy array view for the underlying data of
+ * `%StateVectorManagedCPU` object.
+ *
+ * @tparam PrecisionT Floating point data type
+ * @param sv `%StateVectorManagedCPU` object
+ * @return A numpy array
+ */
+template <class PrecisionT = double>
+auto toNumpyArray(const StateVectorManagedCPU<PrecisionT> &sv)
```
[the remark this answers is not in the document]
maliasadi: Great! Thanks @chaeyeunpark

pennylane_lightning/src/simulator/StateVectorManagedCPU.hpp
chaeyeunpark: Added!

doc/add_kernel.rst
```diff
@@ -13,9 +13,14 @@ We discuss how one can add another gate implementation in this document. Assume
...
         constexpr static std::string_view = "MyGateImpl"; // Name of your kernel
 
+        /* This defines the required alignment for this kernel. If there is no special requirement, 
+           using std::alignment_of_v is sufficient. */
+        template <typename PrecisionT>
+        constexpr static size_t required_alignment = std::alignment_of_v<PrecisionT>;
+
         template <class PrecisionT>
         static void applyPauliX(std::complex<PrecisionT>* data,
                                 size_t num_qubits,
```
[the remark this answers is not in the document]
chaeyeunpark: Thanks! Corrected.

pennylane_lightning/src/util/BitUtil.hpp
```diff
@@ -171,6 +172,42 @@ inline auto log2PerfectPower(unsigned long val) -> size_t {
...
+}
+
+/**
+ * @brief Fill ones from LSB to nbits. Runnable in a compile-time and for any
+ * integer type.
+ *
+ * @tparam IntegerType Integer type to use
+ * @param nbits Number of bits to fill
+ */
+template <class IntegerType = size_t>
```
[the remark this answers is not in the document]
chaeyeunpark: Good suggestion. Done.

pennylane_lightning/lightning_qubit.py
```diff
@@ -54,6 +54,10 @@
         StateVectorC128,
         AdjointJacobianC128,
         VectorJacobianProductC128,
+        allocate_aligned_array,
+        get_alignment,
+        best_alignment,
+        CPUMemoryModel,
```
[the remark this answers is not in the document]
chaeyeunpark: You are right. I removed this line.

Makefile
```diff
@@ -102,13 +102,13 @@ test-cpp-kokkos:
...
 	rm -rf ./BuildBench/CMakeCache.txt ./BuildBench/compiler_info.txt ./BuildBench/run_gate_benchmark.sh
-	cmake $(LIGHTNING_CPP_DIR) -BBuildBench -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -DENABLE_AVX=ON
+	cmake $(LIGHTNING_CPP_DIR) -BBuildBench -DCMAKE_BUILD_TYPE=Release -DENABLE_AVX=ON
 	cmake --build ./BuildBench
 
 .PHONY: gbenchmark
 gbenchmark:
 	rm -rf ./BuildGBench
-	cmake $(LIGHTNING_CPP_DIR) -BBuildGBench -DBUILD_BENCHMARKS=ON -DENABLE_OPENMP=ON -DENABLE_BLAS=ON -DCMAKE_BUILD_TYPE=Release
+	cmake $(LIGHTNING_CPP_DIR) -BBuildGBench -DBUILD_BENCHMARKS=ON -DENABLE_OPENMP=ON -DENABLE_BLAS=ON -DCMAKE_BUILD_TYPE=Release -DBLA_VENDOR=OpenBLAS
```
mlxd, line 111: I never understood why they approved `DBLA_VENDOR` as the flag for BLAS.

pennylane_lightning/src/simulator/KernelMap.hpp
```diff
@@ -0,0 +1,322 @@
...
+inline bool higher_priority(const DispatchElement &lhs,
+                            const DispatchElement &rhs) {
+    return lhs.priority > rhs.priority;
+}
+
+/**
+ * @brief Maintain dispatch element using a vector decreasingly-ordered by
+ * priority.
+ */
+class PriorityDispatchSet {
```
mlxd, line 73: Should we use `https://en.cppreference.com/w/cpp/algorithm/make_heap` here to handle the priority ordering after insertion?

pennylane_lightning/src/simulator/KernelMap.hpp
```diff
@@ -0,0 +1,322 @@
...
+        ordered_vec_.erase(begin, end);
+    }
+};
+///@endcond
+
+struct AllThreading {};
+struct AllMemoryModel {};
+
+constexpr static AllThreading all_threading{};
+constexpr static AllMemoryModel all_memory_model{};
```
mlxd, line 132: May be worth adding some info for these structs if we are declaring them within the Doxygen generation scope

pennylane_lightning/src/simulator/KernelMap.hpp
```diff
@@ -0,0 +1,322 @@
...
+     * @param kernel Kernel to assign
+     */
+    void assignKernelForOp(Operation op, Threading threading,
+                           CPUMemoryModel memory_model, uint32_t priority,
+                           const Util::IntegerInterval<size_t> &interval,
+                           Gates::KernelType kernel) {
+        if (std::find(allowed_kernels_.at(memory_model).cbegin(),
+                      allowed_kernels_.at(memory_model).cend(),
+                      kernel) == allowed_kernels_.at(memory_model).cend()) {
+            PL_ABORT("The given kernel is now allowed for "
```
mlxd, line 203:
```suggestion
            PL_ABORT("The given kernel is not allowed for "
```
?

pennylane_lightning/src/util/Memory.hpp
```diff
@@ -0,0 +1,197 @@
...
+ * Note that alignment must be larger than max_align_t. Otherwise, the behavior
+ * is undefined.
+ *
+ * @param alignment Alignment value we want for the data pointer
+ * @param bytes Number of bytes to allocate
+ * @return Pointer to the allocated memory
+ */
+inline auto alignedAlloc(uint32_t alignment, size_t bytes) -> void * {
+#if defined(__clang__) && defined(__APPLE__)
+    /* Apple clang does not support std::aligned_alloc in Mac 10.14.
```
mlxd, line 40: Since we removed MacOS 10.14 support, should we be able to remove this?

mlxd (commented):
Awesome work again @chaeyeunpark 
I have a few small questions/comments:

Makefile
mlxd: Ah, good to know. No problem having this here from my side.

pennylane_lightning/src/util/Memory.hpp
mlxd: Huh, that's really odd. Sure, keeping this is fine.
If the level of support ever becomes a problem we can build MacOS wheels from GCC or clang mainline compilers rather than Apple.

chaeyeunpark: Yes, I would prefer GCC as it supports new C++ functionalities better and will reduce the maintenance cost (by using the same compiler for Linux and MacOS). I may try this after this PR.

pennylane_lightning/src/simulator/KernelMap.hpp
chaeyeunpark: Heap (priority queue) works when we only want the highest priority object. Here I want to pick up the object with the highest priority among the object with the given pair of (number of wires, threading option). So I think it is not straightforward to use a heap (and using the sorted array has the same time complexity). Still, I expect that the kernel map mechanism may be rewritten when we have more kernels (AVX, OpenMP, and maybe Kokkos).

chaeyeunpark:
Hi @mlxd, thanks for the detailed (and prompt) reviews! As I also changed the name of workflows for Window wheels (added Python versions after `windows-latest`), we also need to update the names of the required workflows.

pennylane_lightning/src/tests/Test_AdjDiff.cpp
```diff
@@ -11,13 +11,12 @@
...
 #include "AdjointDiff.hpp"
-#include "StateVectorManaged.hpp"
-#include "StateVectorRaw.hpp"
+#include "StateVectorRawCPU.hpp"
 #include "Util.hpp"
 
 #include "TestHelpers.hpp"
 
-#ifndef _USE_MATH_DEFINES
+#if !defined(_USE_MATH_DEFINES)
```
AmintorDusko, line 19: Is there any difference in how these two preprocessor directives are compiled?

pennylane_lightning/src/tests/Test_KernelMap.cpp
```diff
@@ -0,0 +1,140 @@
...
+    SECTION("Test conflict") {
+        /* If two elements has the same priority but integer intervals overlap,
+         * they conflict. */
+        REQUIRE(pds.conflict(10U, Util::IntegerInterval<size_t>(19, 23)));
+    }
+
+    SECTION("Get Kernel") {
+        REQUIRE(pds.getKernel(15) == Gates::KernelType::PI);
+        REQUIRE_THROWS_WITH(pds.getKernel(30),
+                            Contains("Cannot find a kernel"));
```
AmintorDusko, line 27: We have a new macro that you can use to REQUIRE_THROWS_AS and REQUIRE_THROWS_WITH in a single line:
`PL_REQUIRE_THROWS_MATCHES(expr, type, message_match)`
With this, you can check for the exception type and match part of the message.
We have a similar one for CHECK:
`PL_CHECK_THROWS_MATCHES(expr, type, message_match)`

pennylane_lightning/src/tests/Test_Measures_Sparse.cpp
```diff
@@ -28,12 +28,11 @@ using std::vector;
 TEMPLATE_TEST_CASE("Expected Values - Sparse Hamiltonian [Kokkos]",
                    "[Measures]", float, double) {
     // Defining the State Vector that will be measured.
-    StateVectorManaged<TestType> Measured_StateVector =
-        Initializing_StateVector<TestType>();
+    auto Measured_StateVector = Initializing_StateVector<TestType>();
 
     // Initializing the measures class.
     // This object attachs to the statevector allowing several measures.
```
AmintorDusko, line 34:
```suggestion
    // This object attaches to the statevector allowing several measures.
```

pennylane_lightning/src/util/BitUtil.hpp
```diff
@@ -37,10 +38,9 @@ inline auto constexpr log2PerfectPower(uint64_t val) -> size_t {
 }
 
 /**
- * @brief Check if there is a positive integer n such that value == 2^n.
+ * @brief Compute log2 of value in a compile-time.
```
AmintorDusko, line 41: I'm not following the change in `@brief`.
It looks like we are still making a check here.

pennylane_lightning/src/util/IntegerInterval.hpp
```diff
@@ -0,0 +1,110 @@
...
+    return IntegerInterval<IntegerType>{
+        0, std::numeric_limits<IntegerType>::max()};
+}
+
+/**
+ * @brief
+ * @rst
+ * Test if :math:`I_1 \cap I_2 = \phi`.
+ * @endrst
+ */
```
AmintorDusko, line 103: Could you please complete this?

pennylane_lightning/src/tests/Test_AdjDiff.cpp
chaeyeunpark: I don't remember why I changed this... Both work exactly the same, but using `defined()` is better some cases as one cand add additional conditions.

pennylane_lightning/src/simulator/KernelMap.hpp
mlxd: Ah, ok, I see now. Sure, that's fine.

pennylane_lightning/src/util/IntegerInterval.hpp
chaeyeunpark: Done!

chaeyeunpark:
While comparing benchmark results from this PR vs `master`, I found an effect from a recent compiler update.

I observed with the previous code with GCC9 that the PI kernel implementation of **IsingXX** outperforms the LM kernel for some number of qubits. This is the set-up I used in this PR, i.e., we use the PI kernel for IsingXX for some number of qubits. However, I found that this assumption is no more valid with GCC10. With GCC10, the LM kernel always outperforms the PI kernel (for all number of qubits and all gates). Below is the benchmark results comparing the performance of the IsingXX gate for different kernels/compilers. 

compare-compilers

Because of this, this PR only will improve the performance with additional kernels. Fortunately, it does not decrease performance either in a benchmark (I will add a plot).

pennylane_lightning/src/tests/CompareVector.hpp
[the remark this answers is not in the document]
chaeyeunpark: @AmintorDusko This is used somewhere inside 'Test_GateImplementations_Nonparam.cpp' file. Adding this definition inside of `TestHelper.hpp` didn't work due to some complex relationship between macros and operator overloading.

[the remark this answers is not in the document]
AmintorDusko: Can we have this overload template scope limited to integers, restricted to the `Test_GateImplementations_Nonparam.cpp` file, or both?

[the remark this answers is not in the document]
chaeyeunpark: May I ask why? We already use this function for float and double types and different allocators. I also think I may use it in other files when needed.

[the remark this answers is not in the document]
AmintorDusko: This is just my opinion, please disregard it if this doesn't make sense.
1. If we have a header called `CompareVector.hpp` I expect all comparisons between vectors to be there.
2. If someone inadvertently includes this header and tries to compare two floating-point vectors this will simplify return false. This can a headache to debug, especially for someone not so close to all this development.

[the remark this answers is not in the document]
chaeyeunpark: I note that there is already `operator==` defined for `std::vector` in the standard library (see https://en.cppreference.com/w/cpp/container/vector/operator_cmp and https://godbolt.org/z/6Kbjhvjas). This is just an extension of it for different allocators.

mlxd (approved):
Awesome work @chaeyeunpark 
I'd recommend adding an entry to the changelog as the last step. I will approve this in the meantime, so feel free to merge when this is done.

pennylane_lightning/src/tests/CompareVector.hpp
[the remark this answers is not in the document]
AmintorDusko: Yes. You are right about that. Thanks for the clarification.

maliasadi (approved):
Great work @chaeyeunpark! Looking forward to seeing this new dispatch mechanism merge soon. :rocket:

chaeyeunpark:
Thanks, @maliasadi, @mlxd, @AmintorDusko for your careful reviews!

co9olguy:
:tada:

PR #292, opened 2022-05-09
CIbuildwheel and benchmark CIs use PL master for tests

chaeyeunpark:
Update CIs to use `master` branch of PL instead of the latest release.


chaeyeunpark:
> @chaeyeunpark the changes are looking +1 What is the reason for making them?

@antalszava We want to utilize the latest feature of PL (in this case, `float32` support). Currently, some Lightning CIs use the latest release, so raises errors when I want to make a newly added feature of PL (only existing in the `master` branch).

mlxd (approved):
Looks good to me. Thanks @chaeyeunpark

PR #293, opened 2022-05-10
Winx64 path fix test

mlxd:
CI-activation PR check only.


mlxd:
Some combination of Kokkos, CMake, Windows, Pybind11 and setuptools is creating an additional `Debug` directory for wheels on Winx64, that does not appear on other platforms, where the extension modules are automatically placed. This is despite the release target set by CMake.
I have not been able to track down what generates this, but have decided for now to simply unpack the built wheel, moved the compiled libraries and repack it.

AmintorDusko:
Hi @mlxd, excellent work! This seems like a good approach for now.
Please be aware that, for whatever reason, GH considers this branch to be up-to-date with `feature/Kokkos-sparse`, but it isn't!
The merging to the `feature/Kokkos-sparse` branch will probably need to be done manually.

AmintorDusko:
The file `cmake/process_options.cmake` is not up-to-date, but `.github/workflows/wheel_win_x86_64.yml` seems to be OK to merge.

AmintorDusko (approved):

PR #295, opened 2022-05-19
Update GCC version and move to C++20

chaeyeunpark:
### Before submitting

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


chaeyeunpark:
Moving to GCC10 is smooth (but going to 11 must be harder due to `manylinux2014` image used by `cibuildwheel`). As GCC10 already supports lots of new C++20 features, this PR also tests some of them (bit utility functions such as `std::popcount` and some more `constexpr` functions) and it works (even with `appleclang` and MSVC).

.github/workflows/benchmarks.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           make clean || true
           pip uninstall pennylane-lightning -y
           cd ../pr
-          pip install -e .
+          CXX=g++-10 pip install -e .
```
mlxd, line 64: Do we need to explicitly list g++-10 here explicitly? Would it be safer to specific this as a minimum supported version in the CMake, and assume that the system has >=10 available?

Since Ubuntu 22.04 defaults to GCC 11 and manylinux2014 defaults to GCC 10, I think that would likely be a better option.

.github/workflows/format.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Run clang-tidy compilation
         run: |
-          cmake -BBuild -DENABLE_CLANG_TIDY=ON -DCLANG_TIDY_BINARY=clang-tidy-12 -DBUILD_EXAMPLES=ON -DBUILD_TESTS=on -DENABLE_WARNINGS=on .
+          CXX=g++-10 cmake -BBuild -DENABLE_CLANG_TIDY=ON -DCLANG_TIDY_BINARY=clang-tidy-12 -DBUILD_EXAMPLES=ON -DBUILD_TESTS=on -DENABLE_WARNINGS=on .
```
mlxd, line 57: If this is necessary, after the above, maybe we should do this as `-DCMAKE_CXX_COMPILER=g++-10`

CMakeLists.txt
```diff
@@ -7,12 +7,9 @@ set(LOGO [=[
 ]=])
 message(${LOGO})
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.14" CACHE STRING "Minimum OS X deployment version")
-
-project(pennylane_lightning)
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X deployment version")
```
mlxd, line 10: Since we are deprecating 10.14 support, we should also make sure to mention this as a breaking change in the changelog.

CMakeLists.txt
```diff
@@ -35,8 +32,6 @@ set_pennylane_lightning_version(${PROJECT_SOURCE_DIR}/pennylane_lightning/_versi
 message(STATUS "pennylane_lightning version ${VERSION_STRING}")
 set(PROJECT_VERSION ${VERSION_STRING})
 
-set(CMAKE_CXX_STANDARD 17) # At least C++17 is required
```
mlxd, line 38: Why not 
```
 set(CMAKE_CXX_STANDARD 20) # At least C++20 is required
```
?

pennylane_lightning/src/algorithms/JacobianProd.hpp
```diff
@@ -117,7 +117,7 @@ template <class T = double> class VectorJacobianProduct {
                     -> std::vector<T> { return std::vector<T>(num_params, 0); };
         }
 
-        return [=](const JacobianData<T> &jd) -> std::vector<T> {
+        return [=, this](const JacobianData<T> &jd) -> std::vector<T> {
```
mlxd, line 120: Does the `this` capture change in C++20?

pennylane_lightning/src/benchmarks/Bench_BitUtil.cpp
```diff
@@ -32,18 +33,18 @@ static void generate_uniform_random_number(benchmark::State &state) {
...
  */
-static void naive_popcount(benchmark::State &state) {
+static void std_popcount(benchmark::State &state) {
     std::random_device rd;
     std::mt19937_64 eng(rd());
     std::uniform_int_distribution<unsigned long> distr;
     for (auto _ : state) {
         auto val = static_cast<unsigned long>(distr(eng));
-        benchmark::DoNotOptimize(Pennylane::Util::Internal::countBit1(val));
+        benchmark::DoNotOptimize(std::popcount(val));
```
mlxd, line 44: :100:

pennylane_lightning/src/benchmarks/CMakeLists.txt
```diff
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 3.14)
 
 project(pennylane_lightning_benchmarks)
 
-set(CMAKE_CXX_STANDARD 17)
```
mlxd, line 5: Same here as earlier?

pennylane_lightning/src/tests/Test_CompilerSupport.cpp
```diff
@@ -0,0 +1,35 @@
...
+    return t;
+}
+
+TEST_CASE("Test exception in a constexpr constructor") {
+    [[maybe_unused]] constexpr static auto t =
+        TestExceptionInConstexprConstructor(1);
+
+    // The following line must not be compiled
+    //[[maybe_unused]] constexpr static auto t1 =
+    // TestExceptionInConstexprConstructor(0);
```
mlxd, line 26: IS this to be uncommented?

pennylane_lightning/src/util/BitUtil.hpp
```diff
@@ -16,169 +16,38 @@
...
 namespace Pennylane::Util {
-/**
- * @brief Define popcount for multiple compilers as well as different types.
- *
- * TODO: change to std::popcount in C++20
- */
-///@{
-#if defined(_MSC_VER)
-inline auto popcount(uint64_t val) -> size_t {
-    return static_cast<size_t>(__popcnt64(val));
```
mlxd, line 137: Nice that this is all gone!

pennylane_lightning/src/util/Util.hpp
```diff
@@ -31,6 +31,14 @@
 #include <type_traits>
 #include <vector>
 
+#if __has_include(<version>)
+#include <version>
+#endif
+
+#if __cpp_lib_math_constants >= 201907L
```
mlxd, line 38: Are these values compiler specific, or language version spec specific?
Should msvc/clang be affected by these values?

mlxd (commented):
Thanks for this @chaeyeunpark 
I have a few questions/comments.

pennylane_lightning/src/algorithms/JacobianProd.hpp
chaeyeunpark: Yes, C++20 deprecated the implicit capture of *this.

pennylane_lightning/src/benchmarks/CMakeLists.txt
chaeyeunpark: It seems like using `target_compile_feature` is recommended in modern CMake (see e.g. https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html). Still, if you think setting `CMAKE_CXX_STANDARD` is better for backward compatibility, I can recover it.

pennylane_lightning/src/tests/Test_CompilerSupport.cpp
chaeyeunpark: Yes, those lines must not be compiled (so it is failed to be compiled when uncommented). Still, it may be possible to add a SNIFAE-based solution to test whether an expression is compiled. Let me try it.

.github/workflows/format.yml
chaeyeunpark: Updated workflows to use `DCMAKE_CXX_COMPILER` instead of the `CXX` env variable.

pennylane_lightning/src/benchmarks/CMakeLists.txt
mlxd: I think the `set(CMAKE_CXX_STANDARD 17)` version is (to my brain) at least easier to see and parse at the start of a file. Plus, we can define this for the entire project up there, and easily swap in newer versions as needed.

While I like the newer option, I think I'd prefer this way for now (we can always update this later).

pennylane_lightning/src/util/Util.hpp
mlxd: Ah, I see. So once we get full C++20 support across the compilers we can get rid of these types of checks?

pennylane_lightning/src/tests/Test_Util.cpp
```diff
@@ -299,6 +258,11 @@ TEST_CASE("Test utility functions for constants", "[Util][ConstantUtil]") {
...
         REQUIRE(Util::count_unique(test_arr1) == 5);
         REQUIRE(Util::count_unique(test_arr2) == 5);
+
+        REQUIRE(Util::count_unique(std::array{nullptr, nullptr, nullptr}) == 1);
+        REQUIRE(Util::count_unique(std::array{0, 0, 0}) == 1);
+        REQUIRE(Util::count_unique(std::array{0, 1, 1}) == 2);
+        REQUIRE(Util::count_unique(std::array{0, 1, 2}) == 3);
     }
 
     SECTION("static_lookup") {
```
maliasadi, line 268: Does this one also need an update with replacing `Util::static_lookup` by `Util::lookup`?

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.hpp
```diff
@@ -24,6 +24,7 @@
 #include "LinearAlgebra.hpp"
 #include "PauliGenerator.hpp"
 
+#include <bit>
```
maliasadi, line 27: I think we don't need to include this here or even in `GateImplementationsPI.hpp`. right?
```suggestion
```

maliasadi, line 27: I think we don't need to include this here or even in `GateImplementationsPI.hpp` as `BitUtil.hpp` is already included. right? 
```suggestion
```

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
```diff
@@ -32,6 +32,7 @@
 #include "LinearAlgebra.hpp"
 #include "PauliGenerator.hpp"
 
+#include <bit>
```
maliasadi, line 35: Not sure, just suggesting :slightly_smiling_face: 
```suggestion
```

pennylane_lightning/src/benchmarks/Bench_BitUtil.cpp
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#include <bit>
```
maliasadi, line 14: No need for this one if it's already included in `BitUtil.hpp` I guess.
```suggestion
```

maliasadi (changes requested):
Awesome work @chaeyeunpark! Are you going to commit these changes (move to C++20 and gcc-10) for the coming release?

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.hpp
mlxd: Better to be explicit about any needed headers. This can cause havoc if we ever refactor and the header goes away in the other file. Best to keep this in wherever used.

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
mlxd: As above

pennylane_lightning/src/benchmarks/Bench_BitUtil.cpp
mlxd: As above

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.hpp
maliasadi: Excellent point! Thanks @mlxd

pennylane_lightning/src/tests/Test_DynamicDispatcher.cpp
```diff
@@ -62,7 +62,7 @@ struct testDispatchForKernel<
         // and compare it to the dynamic dispatcher
         auto test_st = ini_st;
         const auto gate_name =
-            std::string(Util::static_lookup<gate_op>(Constant::gate_names));
+            std::string(Util::lookup(Constant::gate_names, gate_op));
```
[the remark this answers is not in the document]
chaeyeunpark: Nice catch! Updated.

pennylane_lightning/src/tests/Test_Util.cpp
```diff
@@ -135,27 +135,6 @@ size_t ctz_slow(uint64_t x) {
 }
```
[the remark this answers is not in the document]
chaeyeunpark: Thanks!

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -66,11 +64,12 @@ endif()
 ################################################################################
 
 add_executable(compile_time_tests compile_time_tests.cpp)
-target_link_libraries(compile_time_tests lightning_gates lightning_utils lightning_simulator)
+target_link_libraries(compile_time_tests lightning_compile_options lightning_gates lightning_utils lightning_simulator)
 
 set(TEST_SOURCES CreateAllWires.cpp
                  Test_AdjDiff.cpp
 #                 Test_Bindings.cpp
```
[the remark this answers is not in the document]
chaeyeunpark: Because `Test_Bindings.cpp` is empty now. We do not test any binding functions now... I think most of the binding functionalities are now tested on Python side, but we may add something in this file later if there are some essential functionalities to test.

pennylane_lightning/src/tests/Test_CompilerSupport.cpp
chaeyeunpark: I made several trials but failed to make it works. The problem is that this type of compilation failure is not appropriately reflected in the type system, e.g. when I use `decltype(testAssertInConstexprFunction(0))`, it just gives `int`. So I think we need to keep these lines.

.github/workflows/benchmarks.yml
chaeyeunpark: I set `cmake` to raise an error if the GCC version is less than 9 (i.e. >= 10). Still, as it is not straightforward to use the latest installed version in a workflow (and I think it is safer to fix the GCC version for both tests and build wheels), I just set

pennylane_lightning/src/benchmarks/CMakeLists.txt
chaeyeunpark: I reverted some `set(CMAKE_CXX_STANDARD 20)`.

pennylane_lightning/src/tests/CMakeLists.txt
```diff
@@ -66,11 +64,12 @@ endif()
 ################################################################################
 
 add_executable(compile_time_tests compile_time_tests.cpp)
-target_link_libraries(compile_time_tests lightning_gates lightning_utils lightning_simulator)
+target_link_libraries(compile_time_tests lightning_compile_options lightning_gates lightning_utils lightning_simulator)
 
 set(TEST_SOURCES CreateAllWires.cpp
```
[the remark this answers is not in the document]
chaeyeunpark: Nice catch. Thanks!

CMakeLists.txt
chaeyeunpark: @maliasadi As compilers on our target OSs support C++20 (even very partially), the purpose of the current PR is to move to C++20 experimentally. As we serve compiled wheels to all users, I think it is okayish to move C++20. We may roll back this requirement when a fatal problem is discovered.

pennylane_lightning/src/algorithms/JacobianProd.hpp
chaeyeunpark: https://www.nextptr.com/tutorial/ta1430524603/capture-this-in-lambda-expression-timeline-of-change says "The value-capture-default can implicitly capture this or &(*this). However, the implicit capture of this through [=] is deprecated in C++20."

.github/workflows/tests_linux.yml
```diff
@@ -310,7 +312,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext --define="ENABLE_KOKKOS=ON"
+          python setup.py build_ext -i --define="ENABLE_KOKKOS=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```
mlxd, line 315: Nice, I didn't know we could cascade them like this!

CMakeLists.txt
mlxd: Agreed with Chae-Yeun. It helps simplify many parts of the codebase and remove some utilities we can defer to the standard, so provided those all work, I think we are good.

cmake/process_options.cmake
```diff
@@ -14,10 +14,21 @@ if (WIN32)
   set(CMAKE_OBJECT_PATH_MAX 249)
 endif ()
 
+# Check GCC version
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.0)
+        message(FATAL_ERROR "GCC version must be at least 10.0")
```
mlxd, line 20: :+1:

mlxd (approved):
Awesome work @chaeyeunpark 
I've nothing to add. I'm happy to approve.

CMakeLists.txt
maliasadi: Thanks guys! Sure it's a good move then.

chaeyeunpark:
Thanks @mlxd and @maliasadi! I am still sure that there are some more parts in the codebase that need to be converted into C++20, which may be addressed in subsequent PRs.

PR #316, opened 2022-07-21
Caching Kokkos build for wheel-builds

AmintorDusko:
**Context:**
This PR introduces the use of caching for Kokkos' libraries, to decrease wheel-build time.

The cache GH action is used for caching the Kokkos libraries.

**Benefits:**
Faster build times.

**Possible Drawbacks:**

**Related GitHub Issues:**


chaeyeunpark:
Nice work @AmintorDusko, but it seems like there is a typo in yml file for MacOS: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/2870428237 Could you fix it?

AmintorDusko:
> Nice work @AmintorDusko, but it seems like there is a typo in yml file for MacOS: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/2870428237 Could you fix it?

Hi @chaeyeunpark, this PR is a WIP. Only the Linux wheel is supposed to work at the moment.
Thanks for the fix for the CMAKE kernels libraries file.

chaeyeunpark:
> > Nice work @AmintorDusko, but it seems like there is a typo in yml file for MacOS: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/2870428237 Could you fix it?
> 
> Hi @chaeyeunpark, this PR is a WIP. Only the Linux wheel is supposed to work at the moment. Thanks for the fix for the CMAKE kernels libraries file.

Aha, I see. Still, speed-ups for Linux wheels seem great!

AmintorDusko:
> > > Nice work @AmintorDusko, but it seems like there is a typo in yml file for MacOS: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/2870428237 Could you fix it?
> > 
> > 
> > Hi @chaeyeunpark, this PR is a WIP. Only the Linux wheel is supposed to work at the moment. Thanks for the fix for the CMAKE kernels libraries file.
> 
> Aha, I see. Still, speed-ups for Linux wheels seem great!

Yes! I'm pretty happy with that!

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -6,6 +6,7 @@ name: Wheel::Linux::ARM
 # **Who does it impact**: Wheels to be uploaded to PyPI.
 
 on:
+  pull_request: #Just for testing. Needs to be removed before merging.
```
AmintorDusko, line 9:
```suggestion
```
I'm just testing in the PR.

.github/workflows/wheel_linux_ppc64le.yml
```diff
@@ -1,11 +1,12 @@
 name: Wheel::Linux::PowerPC
 
-# **What it does**: Build python wheels for Linux (ubuntu-latest) architecture PowerPC 64 and store it as artifacts.
+# **What it does**: Builds python wheels for Linux (ubuntu-latest) architecture PowerPC 64 and store it as artifacts.
 #                   Python versions: 3.7, 3.8, 3.9, 3.10.
 # **Why we have it**: To build wheels for pennylane-lightning installation.
 # **Who does it impact**: Wheels to be uploaded to PyPI.
 
 on:
+  pull_request: #Just for testing. Needs to be removed before merging.
```
AmintorDusko, line 9:
```suggestion
```
Just for testing.

AmintorDusko:
Here follows a rough estimative of increase in workflow performance.
Data was extracted from one of the most recent runs (Python 3.7).

Wheel | timing without caching | timing now
-- | -- | --
Wheel::Linux::ARM | 49 | 14
Wheel::Linux::PowerPC | 56 | 13
Wheel::Linux::x86_64 | 10 | 5
Wheel::MacOS::Intel | 10 | 5

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -62,41 +166,49 @@ jobs:
...
+          CIBW_ARCHS_LINUX: ${{matrix.arch}}
+
           CIBW_BUILD: ${{ matrix.cibw_build }}
+
           CIBW_SKIP: "*-musllinux*"
 
           # Python build settings
           CIBW_BEFORE_BUILD: |
-            pip install pybind11 ninja cmake && yum install -y gcc gcc-c++
+            cat /etc/yum.conf | sed "s/\[main\]/\[main\]\ntimeout=5/g" > /etc/yum.conf
```
chaeyeunpark, line 198: Do we need this line?

chaeyeunpark (commented):
Great work @AmintorDusko! I just have one question.

.github/workflows/wheel_linux_aarch64.yml
mlxd: This was added as the CI image would spend ~30s trying to hit remote URLs that were no longer in service. This line ensures that timeout is 5s only, and saved us a lot of deadtime over the course of setup and execution.

.github/workflows/wheel_linux_aarch64.yml
mlxd, line 9:
```suggestion
```

mlxd (approved):
Thanks @AmintorDusko 
Happy for the Linux/MacOS versions to go through. We can revisit Windows at a later date.

cmake/process_options.cmake
```diff
@@ -117,70 +117,30 @@ endif()
...
-            ${CMAKE_SOURCE_DIR}/lib64
-            ENV KOKKOS_LIB
-            ENV LD_LIBRARY_PATH
-    )
-    find_file( KOKKOS_CORE_INC
-    NAMES   Kokkos_Core.hpp
-    HINTS   /usr/include
-            /usr/local/include
+
+    find_package(Kokkos
```
mlxd, line 121: Can we guarantee the existence of the compiled library with this? Not sure if this is a better approach.

AmintorDusko: I believe that we can. Even in Windows CMake was able to detect the libraries.

AmintorDusko: Could you please elaborate on the reason why this is not a good approach?

AmintorDusko: I mean, the reason why you are not sure if this is a good approach.

chaeyeunpark (approved):
Thanks @AmintorDusko for a nice work again! It seems like Linux wheels work without problem.

cmake/process_options.cmake
mlxd: Do we not need a separate `FindKokkos.cmake` file for this functionality to work, wherein we specify how it finds the required libraries and headers?

AmintorDusko: I don't think so. According to Kokkos, they provide everything that is needed.
This is their suggested method of incorporating the already installed Kokkos and Kokkos Kernels packages into a third-party building process.

mlxd: Are there examples of where this happens in other repos? Just curious how this works, as I expected the [docs](https://cmake.org/cmake/help/latest/command/find_package.html) seem to mention the path it investigates for the .cmake file.

Though, if it just works, that is fine by me.

AmintorDusko: Yes. But we are not supposed to make a file for that. In the same way, we are not supposed to make a file for finding BLAS, or OpenMP. 
Because the installation is not in a usual place, in the code, I hint to CMake about the Kokkos installation directory and CMake finds the files it needs there.

mlxd: The reason we don't for BLAS or OpenMP is because https://github.com/Kitware/CMake/blob/8d8957e56d8a58fd3966618eff742def509cfd39/Modules/FindBLAS.cmake and https://github.com/Kitware/CMake/blob/ddfccc36de642c96990d33a48dade0b00609169c/Modules/FindOpenMP.cmake exist natively to CMake.

But if it works like this, fine by me.

mlxd (approved):
Thanks @AmintorDusko

cmake/process_options.cmake
AmintorDusko: Makes sense. Regarding Kokkos, by providing a hint CMake is able to find it.

.github/workflows/wheel_linux_x86_64.yml
```diff
@@ -42,45 +43,161 @@ jobs:
...
+      - name: Install dependencies (CentOS)
+        if: ${{ (matrix.container_img == 'quay.io/pypa/manylinux2014_x86_64') }}
+        run: |
+          # Reduce wait time for repos not responding
+          cat /etc/yum.conf | sed "s/\[main\]/\[main\]\ntimeout=5/g" > /etc/yum.conf
+          yum update -y && yum install -y docker
 
       - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==2.3.0
+        run: python3.7 -m pip install cibuildwheel==2.8.1
```
mlxd, line 187:
```suggestion
        run: python3 -m pip install cibuildwheel==2.8.1
```

We probably don't need to pin to 3.7

.github/workflows/wheel_linux_x86_64.yml
```diff
@@ -92,9 +209,10 @@ jobs:
             pl-device-test --device=lightning.qubit --skip-ops -x --tb=short --no-flaky-report
 
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-          CIBW_ARCHS_LINUX: ${{matrix.arch}}
 
-        run: python -m cibuildwheel --output-dir wheelhouse
+          CIBW_BUILD_VERBOSITY: 3
+
+        run: python3.7 -m cibuildwheel --output-dir wheelhouse
```
mlxd, line 215:
```suggestion
        run: python3 -m cibuildwheel --output-dir wheelhouse
```

setup.py
```diff
@@ -93,20 +93,19 @@ def build_extension(self, ext: CMakeExtension):
                 configure_args += []
             else:
                 configure_args += ["-DENABLE_OPENMP=OFF"]
-        elif platform.system() == "Linux":
-            if platform.machine() == "x86_64":
-                configure_args += ["-DENABLE_AVX=ON"]  # Enable AVX if x64 on Linux
         elif platform.system() == "Windows":
             configure_args += ["-DENABLE_OPENMP=OFF", "-DENABLE_BLAS=OFF"]
+            if not debug:
+                build_args += ["--config", "RelWithDebInfo"]
```
mlxd, line 99: Should we set this at the outer package level too? As in, make it default behaviour for all platforms?

mlxd (commented):
Nice work @AmintorDusko 
Just a few follow-up comments.

We may also want to backport the Windows fix to a fork of the `v0.25.0` branch for the `v0.25.1` release candidate.

setup.py
AmintorDusko: We don't need it, but I can do it.

AmintorDusko:
> Nice work @AmintorDusko Just a few follow-up comments.
> 
> We may also want to backport the Windows fix to a fork of the `v0.25.0` branch for the `v0.25.1` release candidate.

Are we pushing it to a proper PR or directly to the `v0.25.1` rc?

mlxd:
> > Nice work @AmintorDusko Just a few follow-up comments.
> > We may also want to backport the Windows fix to a fork of the `v0.25.0` branch for the `v0.25.1` release candidate.
> 
> Are we pushing it to a proper PR or directly to the `v0.25.1` rc?

I think since this only affects the CI, and hence the build of the wheels we have, we can probably offshoot this into a new v0.25.1 branch with 0.25.0 as the base. This should also make sure nothing else major was added between the 0.25 and master branches though, otherwise we will need to do a git cherrypick of these specific commits.

.github/workflows/wheel_linux_x86_64.yml
AmintorDusko: In fact, we need, since we are building inside a container.

.github/workflows/wheel_linux_x86_64.yml
AmintorDusko: Same as before.

PR #322, opened 2022-07-29
[Do not merge] Avx kernel fix macos

chaeyeunpark:
Try to fix MacOS errors from PR #313


chaeyeunpark:
I think the problem is that compilation units for avx2/avx512 kernels also use the LM kernel. Thus there are several template instantiation of the LM kernels (with different compile options), and the linker does not choose the basic one when they are all linked together. Explicit instantiation of the LM kernel may be helpful, but let me check again.

PR #324, opened 2022-07-30
Explicit instantiate default kernel templates

chaeyeunpark:
### Before submitting

**Context:** While adding AVX kernels, it is found that the LM kernels are instantiated several times with different compile options. This confuses the linker, resulting in an illegal instruction error. This PR aims to fix this issue as well as reduce the compile time by explicitly instantiating all kernel functions.

**Benefits:** Slightly faster compile time and better integration of AVX2/512 kernels

**Possible Drawbacks:** Explicit instantiation must also be added when adding each kernel function for default kernels.

Some other changes:
* Update the default clang-tidy version to 13
* Fix possible errors in `alignedAlloc`


pennylane_lightning/src/gates/cpu_kernels/QChemGateImplementations.hpp
```diff
@@ -1,425 +0,0 @@
...
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-/**
- * @file
- * Defines qchem kernel functions with generators
- */
-#pragma once
```
chaeyeunpark, line 18: Contents of this file are distributed to each cpp file.

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
```diff
@@ -63,12 +63,11 @@ class GateImplementationsPI : public PauliGenerator<GateImplementationsPI> {
         GateOperation::Hadamard,
         GateOperation::S,
         GateOperation::T,
+        GateOperation::PhaseShift,
```
chaeyeunpark, line 66: Gate operations are reordered to have the same ordering as `GateOperations.hpp`

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.hpp
```diff
@@ -83,41 +83,41 @@ class GateImplementationsLM : public PauliGenerator<GateImplementationsLM> {
         GateOperation::Hadamard,
         GateOperation::S,
         GateOperation::T,
+        GateOperation::PhaseShift,
```
chaeyeunpark, line 86: Gate operations are reordered to have the same ordering as `GateOperations.hpp`

pennylane_lightning/src/gates/DynamicDispatcher.hpp
```diff
@@ -33,19 +33,12 @@
 #include <functional>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <variant>
 #include <vector>
 
 /// @cond DEV
 namespace Pennylane::Internal {
```
chaeyeunpark, line 41: There are some changes in this file following the changes in the PR #313

chaeyeunpark:
As each gate operation is a member function template, explicit instantiations become ridiculously large... We may change these back to template class member functions (note that [template, member function] $\neq$ 0).

pennylane_lightning/src/gates/CMakeLists.txt
```diff
@@ -1,11 +1,18 @@
 project(lightning_gates)
 
-set(GATES_FILES GateUtil.cpp DynamicDispatcher.cpp CACHE INTERNAL "" FORCE)
+# We make a static library for default kernels using explicit template instantiations.
+# We thus ensure that only a single instantiation exists during the build process.
```
mlxd, line 4: Yep, definitely needed this

pennylane_lightning/src/gates/DynamicDispatcher.hpp
```diff
@@ -137,6 +121,89 @@ template <typename PrecisionT> class DynamicDispatcher {
         return singleton;
```
mlxd, line 121: Assuming we are using multiple different instantiations across multiple threads, will this be fine? Or do we need to add any locking behaviour to anything in the given singleton?

pennylane_lightning/src/gates/DynamicDispatcher.hpp
```diff
@@ -165,8 +232,8 @@ template <typename PrecisionT> class DynamicDispatcher {
     void registerGateOperation(Gates::GateOperation gate_op,
                                Gates::KernelType kernel, FunctionType &&func) {
         // TODO: Add mutex when we go to multithreading
```
mlxd, line 234: Ah, yea, I see you commented on the above here. +1

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.cpp
```diff
@@ -0,0 +1,531 @@
...
+        const size_t i11 = i00 | rev_wire0_shift | rev_wire1_shift;
+
+        arr[i00] = std::complex<PrecisionT>{};
+        arr[i01] *= Util::IMAG<PrecisionT>();
+        arr[i10] *= -Util::IMAG<PrecisionT>();
+        arr[i11] = std::complex<PrecisionT>{};
+
+        std::swap(arr[i10], arr[i01]);
+    }
+    // NOLINTNEXTLINE(readability-magic-numbers)
```
mlxd, line 138: Should we define a `HALF` templated function, and add the above linter details there?

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.hpp
```diff
@@ -1355,4 +1357,294 @@ class GateImplementationsLM : public PauliGenerator<GateImplementationsLM> {
...
+    -> double;
+extern template auto GateImplementationsLM::applyGeneratorSingleExcitationPlus(
+    std::complex<float> *, size_t, const std::vector<size_t> &, bool) -> float;
+
+extern template auto
+GateImplementationsLM::applyGeneratorMultiRZ(std::complex<double> *, size_t,
+                                             const std::vector<size_t> &, bool)
+    -> double;
+extern template auto GateImplementationsLM::applyGeneratorMultiRZ(
+    std::complex<float> *, size_t, const std::vector<size_t> &, bool) -> float;
```
mlxd, line 1649: Given the size of this file, should we maybe move the above off to a separate file to be included?

mlxd: We can also later see if there is a way to use a parameter pack to forward declare both the `float` and `double` templates to reduce the length.

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
```diff
@@ -89,21 +89,21 @@ class GateImplementationsPI : public PauliGenerator<GateImplementationsPI> {
...
-        GeneratorOperation::PhaseShift,
         GeneratorOperation::IsingXX,
         GeneratorOperation::IsingXY,
         GeneratorOperation::IsingYY,
         GeneratorOperation::IsingZZ,
         GeneratorOperation::CRX,
         GeneratorOperation::CRY,
         GeneratorOperation::CRZ,
+        GeneratorOperation::ControlledPhaseShift,
         GeneratorOperation::DoubleExcitation,
```
mlxd, line 104: Do we also have the single excitation gates for here?

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
```diff
@@ -1059,4 +1059,285 @@ class GateImplementationsPI : public PauliGenerator<GateImplementationsPI> {
...
+extern template auto
+GateImplementationsPI::applyGeneratorCRZ(std::complex<double> *, size_t,
+                                         const std::vector<size_t> &, bool)
+    -> double;
+
+extern template auto GateImplementationsPI::applyGeneratorControlledPhaseShift(
+    std::complex<float> *, size_t, const std::vector<size_t> &, bool) -> float;
+extern template auto GateImplementationsPI::applyGeneratorControlledPhaseShift(
+    std::complex<double> *, size_t, const std::vector<size_t> &, bool)
+    -> double;
```
mlxd, line 1342: As above with potentially moving these into a separate file for inclusion here.

mlxd (commented):
Thanks @chaeyeunpark 
Just a few questions/comments.

pennylane_lightning/src/gates/DynamicDispatcher.hpp
chaeyeunpark: Always the same instance is used. Currently, we add all registrations when the program is loaded (in a single thread) and only use them afterward. As all other member functions besides registration functions in this class are thread-safe (they are const), we don't need mutex here unless there is any further design changes. I will add this comment to the description of this class.

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsLM.cpp
chaeyeunpark: Maybe? I am still not so sure whether it improves readability. It is possible that someone gets confused and wants to check the definition of `HALF`.

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
chaeyeunpark: We only have implementations of those gates in the LM kernel. We can add these implementations (and check which one is faster).

pennylane_lightning/src/gates/cpu_kernels/GateImplementationsPI.hpp
chaeyeunpark: As `extern` explicitly force the compiler not to instantiate the functions, I concern that separating these statements to another file fails to achieve this objective. One happens to include only the class definition and does not aware he/she made another instantiation. Unfortunately, I couldn't find a way to make the compiler warns or raises an error when this happens (besides moving all member function definitions to cpp files, only remaining declarations).

chaeyeunpark: Another solution I can think is one from this comment: https://github.com/PennyLaneAI/pennylane-lightning/pull/324#issuecomment-1200073065

mlxd: Hmm, good point. Since we need this to work now, probably best to just power ahead and keep this for now. That ok with you? We can revisit then potentially later once the new kernels are through.

chaeyeunpark: :+1:

mlxd: Great, nothing else to add from my side. If you are happy too, I will approve.

mlxd (approved):

maliasadi (approved):
Great job @chaeyeunpark!

multiphaseCFD:
Nice work @chaeyeunpark! No comments and will approve it.

multiphaseCFD (approved):

chaeyeunpark:
Thanks, everyone, for approving this! I will definitely revisit this after release.

PR #326, opened 2022-08-08
remove op arithmetic observable support

albi3ro:
`DefaultQubit` supports observable measurement of the new `Sum`, `Prod`, and `SProd` classes. Lightning does not yet support these observables, so we need to remove them from the list of supported observables. 

In the longer term, we should determine what is necessary to support these general observables.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,9 @@
 * We explicitly disable support for PennyLane's parameter broadcasting.
 [#317](https://github.com/PennyLaneAI/pennylane-lightning/pull/317)
 
+* We explicitly remove support for PennyLane's `Sum`, `SProd` and `Prod`
+  as observables.
+
 ### Improvements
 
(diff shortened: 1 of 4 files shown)


.github/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 * We explicitly disable support for PennyLane's parameter broadcasting.
 [#317](https://github.com/PennyLaneAI/pennylane-lightning/pull/317)
 
+* We explicitly remove support for PennyLane's `Sum`, `SProd` and `Prod`
+  as observables.
+
```
albi3ro, line 12:
```suggestion
  [(#326)](https://github.com/PennyLaneAI/pennylane-lightning/pull/326)

```

mlxd (approved):
Looks good to me. Thanks @albi3ro

antalszava (approved):
Looks good :+1:

PR #327, opened 2022-08-08
Updated ReadTheDocs runner version to Ubuntu 22.04

rashidnhm:
**Context:** Updated runner version of ReadTheDocs to Ubuntu 22.04

This pull request updates the runner that ReadTheDocs would use to build the pennylane-lightning docs. The reason for this update is to get the updates system packages (gcc, cmake, etc) available out of the box.

Installing and switching the default version of gcc in the ubuntu 20.04 is more cumbersome and require more "hacky" workarounds.

**Benefits:**
The pennylane-lightning docs build would pass for both RTD community and business plan. As is stands even though the docs are being built on the community runners, it is failing on the business plan runner.

**Possible Drawbacks:**
N/A. The build is passing on the business runners with the changes proposed in this Pull request.

**Related GitHub Issues:**
N/A.


mlxd (approved):
Nice one Rashid. Approving.

PR #330, opened 2022-08-12
Create a PennyLane-lightning requirements-dev.txt and update the Dockerfile

AmintorDusko:
**Context:**
PennyLane-Lightning needs a requirements file for development purposes.
Also, the dockerfile should be updated to reflect Lightning's dependence on GCC 10.

The requirements-dev file will point to the PennyLane master.
The dockerfile was updated accordingly.

**Benefits:**
Improved environment for developers.

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane-lightning/issues/329

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 ninja
 flaky
 numpy
-pennylane~=0.26
+pennylane~=0.25
 pybind11
 pytest
 pytest-cov
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 ninja
 flaky
 numpy
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git@master
 pybind11
 pytest
 pytest-cov
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -23,6 +23,8 @@
 [(#313)](https://github.com/PennyLaneAI/pennylane-lightning/pull/313)
 
 ### Documentation
+* Updated ReadTheDocs runner version from Ubuntu 20.04 to 22.04 
+[(#327)](https://github.com/PennyLaneAI/pennylane-lightning/pull/327)
 
 ### Bug fixes
 
@@ -700,4 +702,4 @@ Initial release.
 
 This release contains contributions from (in alphabetical order):
 
-Tom Bromley, Josh Izaac, Nathan Killoran, Antal Száva
+Tom Bromley, Josh Izaac, Nathan Killoran, Rashid N H M, Antal Száva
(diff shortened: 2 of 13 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 ninja
 flaky
 numpy
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git@master
```
chaeyeunpark, line 4: I am not so sure whether this will yield the behavior we want. This means `pip install pennylane-lightning` will install the `master` version (after removing the installed PennyLane).

chaeyeunpark (commented):
Hi @AmintorDusko, thanks for the nice work. Just one small question.

requirements.txt
mlxd: I think this is what we want, as the only use-case for this docker image right now is to work with the latest Pennylane, and using the master branch of lightning with the non master of PennyLane is more likely to cause issues than both at the same tag. We will likely deprecate this builder in favour of the PL Dockerfiles in a release or two, as it seems to provide a better end-to-end ecosystem support.

mlxd: Oh no, you are right, this was not a patch for the Dockerfile. Ignore my previous comment. @chaeyeunpark is 100% correct.

mlxd: Can we move this into the Dockerfile and directly patch the requirements.txt from there instead of making the change outright?

AmintorDusko: Hi @chaeyeunpark, this change yielded exactly the change I wanted. 
After some discussions with @mlxd, we decided to move it to a new requirements-dev.txt, though.
Thank you all for the quick reviews.

requirements.txt
```diff
@@ -1,7 +1,7 @@
 ninja
 flaky
 numpy
-pennylane>=0.15
+pennylane~=0.26
```
mlxd, line 4:
```suggestion
pennylane~=0.25
```

mlxd: Pip will not pick up on dev or rc releases nicely, so best to keep this fixed to the most current release.

mlxd: We can revisit this once 0.25 goes live.

AmintorDusko: Thank you! I intended it to be 0.25

chaeyeunpark (approved):
Nice update @AmintorDusko! Happy to approve this.

mlxd (approved):
Thanks @AmintorDusko

AmintorDusko:
Thank you @mlxd and @chaeyeunpark for your nice reviews.

PR #346, opened 2022-08-26
Quick fix for upload artifact for win

AmintorDusko:
**Context:** The path in the upload artifact action needs to be updated.

**Description of the Change:** Update the path.

**Benefits:** upload-pypi will work

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/wheel_win_x86_64.yml
+++ b/.github/workflows/wheel_win_x86_64.yml
@@ -204,7 +204,7 @@ jobs:
           regex: '.*[0-9]+.[0-9]+.[0-9]+[-_]?rc[0-9]+'
 
       - uses: actions/upload-artifact@v2
-        # if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master' || steps.rc_build.outputs.match != ''}}
+        if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master' || steps.rc_build.outputs.match != ''}}
         with:
           name: ${{ runner.os }}-wheels-${{ matrix.arch }}.zip
           path: ./wheelhouse/*.whl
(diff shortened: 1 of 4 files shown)


.github/workflows/wheel_win_x86_64.yml
```diff
@@ -206,22 +206,22 @@ jobs:
...
-          name: ${{ runner.os }}-wheels.zip
-          path: Z:\dist\*.whl
+          name: ${{ runner.os }}-wheels-${{ matrix.arch }}.zip
+          path: ./wheelhouse/*.whl
 
   upload-pypi:
     needs: win-wheels
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master'}}
+    # if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master'}}
```
mlxd, line 215: I guess the main issue was the wheels built already with the release version, and uploaded as part of the test workflow, due to the above line being commented out. I think this is fine to uncomment the lines here and improve our process for testing these in future to avoid the duplicate wheel uploads causing the failure.

mlxd (commented):
I think we can close this issue, as it was the result of our testing the artifacts that auto uploaded to PyPI with the release version tags. Best to just introduce a better process for this going forward.

.github/workflows/wheel_win_x86_64.yml
mlxd: Or maybe I am misunderstanding the given issue -- is it a path issue to the zip file then?

AmintorDusko: We had a path issue, this comment is for testing. I will let you know when it is in its final form.

mlxd: Maybe best to keep in draft form then for now.

mlxd (approved):
Nice one, nothing else from me.

chaeyeunpark (approved):
Thanks, @AmintorDusko, for the quick fix!

PR #349, opened 2022-09-02
Defer op-arithmetic to default qubit

albi3ro:
```
dev = qml.device('lightning.qubit', wires=2)

@qml.qnode(dev, diff_method=None)
def circuit():
    obs = qml.s_prod(2.1, qml.PauliZ(0)) + qml.op_sum(qml.PauliX(0), qml.PauliZ(1))
    return qml.expval(obs)
```

This PR adds support for operator arithmetic observables to lightning qubit by deferring handling of them to `default.qubit`.

While we should examine to natively support them in lightning, this fix at least gets support in.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -15,6 +15,17 @@
 * Implements caching for Kokkos installation.
 [(#316)](https://github.com/PennyLaneAI/pennylane-lightning/pull/316)
 
+* Supports measurements of operator arithmetic classes such as `Sum`, `Prod`,
+  and `SProd` by deferring handling of them to `DefaultQubit`.
+  [(#349)](https://github.com/PennyLaneAI/pennylane-lightning/pull/349)
+
+```
+@qml.qnode(qml.device('lightning.qubit', wires=2))
+def circuit():
+    obs = qml.s_prod(2.1, qml.PauliZ(0)) + qml.op_sum(qml.PauliX(0), qml.PauliZ(1))
+    return qml.expval(obs)
+```
+
 ### Documentation
 
 ### Bug fixes
@@ -32,7 +43,7 @@
 
 This release contains contributions from (in alphabetical order):
 
-Amintor Dusko, Chae-Yeun Park
+Amintor Dusko, Christina Lee, Chae-Yeun Park
 
 ---
 
--- a/pennylane_lightning/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit.py
@@ -608,11 +608,15 @@ class LightningQubit(DefaultQubit):
         Returns:
             Expectation value of the observable
         """
-        if (observable.arithmetic_depth > 0) or isinstance(observable.name, List) or observable.name in [
-            "Identity",
-            "Projector",
-
-        ]:
+        if (
+            (observable.arithmetic_depth > 0)
+            or isinstance(observable.name, List)
+            or observable.name
+            in [
+                "Identity",
+                "Projector",
+            ]
+        ):
             return super().expval(observable, shot_range=shot_range, bin_size=bin_size)
 
         if self.shots is not None:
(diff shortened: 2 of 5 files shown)


pennylane_lightning/lightning_qubit.py
```diff
@@ -617,10 +608,15 @@ def expval(self, observable, shot_range=None, bin_size=None):
         Returns:
             Expectation value of the observable
         """
-        if isinstance(observable.name, List) or observable.name in [
-            "Identity",
-            "Projector",
-        ]:
+        if (
+            (observable.arithmetic_depth > 0)
```
mlxd, line 612: What is the `arithmetic_depth` of an observable?

.github/CHANGELOG.md
```diff
@@ -15,6 +15,17 @@
...
 [(#316)](https://github.com/PennyLaneAI/pennylane-lightning/pull/316)
 
+* Supports measurements of operator arithmetic classes such as `Sum`, `Prod`,
+  and `SProd` by deferring handling of them to `DefaultQubit`.
+  [(#349)](https://github.com/PennyLaneAI/pennylane-lightning/pull/349)
+
+```
+@qml.qnode(qml.device('lightning.qubit', wires=2))
+def circuit():
+    obs = qml.s_prod(2.1, qml.PauliZ(0)) + qml.op_sum(qml.PauliX(0), qml.PauliZ(1))
```
mlxd, line 25: Given the structure of these, we could obtain a sizeable speed-up by lowering them to the appropriate C++ calls we already have in place. Though, I think that can remain a v0.27 priority.

tests/test_expval.py
```diff
@@ -110,6 +110,72 @@ def test_hadamard_expectation(self, theta, phi, qubit_device_3_wires, tol):
...
+            return qml.expval(obs)
+
+        x = np.array(0.654)
+        y = np.array(-0.634)
+
+        res = circuit(x, y)
+        expected = 2.3 * np.cos(x) + 0.5 * np.sin(x) * np.cos(y)
+        assert qml.math.allclose(res, expected)
+
+
```
mlxd, line 178: Since `adjoint_jacobian` can side-step the `expval` pipeline, it would be good to put in checks there too to verify it works as expected.

mlxd (commented):
Thanks @albi3ro 
A few comments and suggestions

pennylane_lightning/lightning_qubit.py
albi3ro: Arithmetic depth is the level of nesting. So `Sum(qml.PauliX(0), qml.PauliY(0))` will have an arithmetic depth of 1.  `Sum(adjoint(qml.PauliX(0)), qml.PauliY(0) ))` will have an arithmetic depth of 2.  Plain elementary ops will have an arithmetic depth of `0`.

.github/CHANGELOG.md
albi3ro: Definitely, though doing so will be outside my C++ abilities.

chaeyeunpark: @mlxd Yeap. I will write a simple C++ code (in fact, it already exists) constructing a sparse matrix from a sum of products of Pauli operators and want to compare its performance with an existing Python code in PennyLane.

pennylane_lightning/lightning_qubit.py
mlxd: I assume then this is pulled as the height of the tree representing the operation?

While not an issue here, just clarifying so that we can provide a way to port this to CPP later.

albi3ro: Yes, it is defined for all `Operator` objects.  And yes, it is just the height of the tree.

tests/test_expval.py
albi3ro: added gradient tests.

chaeyeunpark (approved):
Looks good to me! Happy to approve.

AmintorDusko (approved):
Thanks for that @albi3ro!

Issue #352, opened 2022-09-14
Improve tests and checks workflows


AmintorDusko:
#### Issue description

After the recent improvements to build wheels workflows its run time decreased a lot.
Now, tests and checks are the slowest part of the CI/CD pipeline.

* *Expected behavior:* (What you expect to happen)
Tests and checks are to be performed in ~< 10 minutes.

* *Actual behavior:* (What actually happens)
Some cases can be observed where the run time is > 20 minutes.

* *Reproduces how often:* (What percentage of the time does it reproduce?)
Every time a commit is pushed to a PR or a PR is merged. All the time.

* *System information:* (post the output of `import pennylane as qml; qml.about()`)
GH runners

PR #351, opened 2022-09-13
Update clang-tools version

chaeyeunpark:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #353, opened 2022-09-14
Improve tests and checks workflows

AmintorDusko:
**Context:**
Checks and tests are, at the moment, slowing down the CI/CD pipeline.
Some have a running time greater than 20 minutes.

Introduction of Kokkos caching for tests and checks, and removal of redundancies.

**Benefits:**
The expected benefit is to have tests and checks running in less than 10 minutes.

**Possible Drawbacks:**
Not aware of any.

**Related GitHub Issues:**
#352

### DIFF
--- a/pennylane_lightning/_version.py
+++ b/pennylane_lightning/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.26.0-dev12"
+__version__ = "0.27.0-dev1"


AmintorDusko:
[sc 25948]

AmintorDusko:
I intend to fill the changelog and merge this PR only after the Release.

.github/workflows/benchmarks.yml
```diff
@@ -1,83 +1,45 @@
...
-          python .github/workflows/benchmarks/run_bench.py lightning.qubit lightning_master.json
+        run: sudo apt-get update && sudo apt-get -y -q install cmake gcc-10 g++-10 libopenblas-dev
 
-      - name: Benchmark default qubit device
+      - name: Build GBenchmark
         run: |
-          cd main
-          python .github/workflows/benchmarks/run_bench.py default.qubit default_qubit.json
+            cmake pennylane_lightning/src/ -BBuildGBench -DBUILD_BENCHMARKS=ON -DENABLE_BLAS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER="$(which g++-$GCC_VERSION)"
+            cmake --build ./BuildGBench  --parallel 2
```
mlxd, line 40: We can just use ninja as the build target here `-GNinja`, and there will be no reason to specify the parallel build. This will allow us to autoscale if/when Github actions runners offer more cores.

Ninja may also need to be installed at line 35.

.github/workflows/benchmarks.yml
```diff
@@ -1,83 +1,45 @@
...
-        run: |
-          cd main
-          python .github/workflows/benchmarks/plot_results.py
-
-      - uses: actions/upload-artifact@v2
-        with:
-          name: gates.png
-          path: ./main/gates.png
+            ./BuildGBench/benchmarks/utils
+            ./BuildGBench/benchmarks/apply_operations
```
mlxd, line 45: As @chaeyeunpark suggested, we can use a regex here to select a subset of the tests to ensure things work.

.github/workflows/format.yml
```diff
@@ -39,39 +39,64 @@ jobs:
...
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.7
-
       - name: Install dependencies
-        run: sudo apt update && sudo apt -y install clang-tidy-13 cmake g++-10 libomp-13-dev
-        env:
-          DEBIAN_FRONTEND: noninteractive
+        run: sudo apt update && sudo apt -y install clang-format-13 python3
```
mlxd, line 43:
```suggestion
        run: sudo apt update && sudo apt -y install clang-format-13 python3
```
I expect this will need updating when #351 merges

.github/workflows/tests_linux.yml
```diff
@@ -14,124 +14,120 @@ env:
...
-          name: ubuntu-test-report
-          path: Build/tests/results/report.xml
-
-      - name: Build and run unit tests for code Coverage
+      - name: Build and run unit tests with code coverage
         run: |
-            cmake pennylane_lightning/src -BBuildCov  -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON -DCMAKE_CXX_COMPILER="$(which g++-$GCC_VERSION)"
-            cmake --build ./BuildCov
+            cmake pennylane_lightning/src -BBuildCov -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON -DCMAKE_CXX_COMPILER="$(which g++-$GCC_VERSION)"
+            cmake --build ./BuildCov --parallel 2
```
mlxd, line 46: Same here with ninja

.github/workflows/tests_linux.yml
```diff
@@ -14,124 +14,120 @@ env:
...
-            cmake pennylane_lightning/src -BBuildCov  -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON -DENABLE_COVERAGE=ON -DENABLE_BLAS=ON -DCMAKE_CXX_COMPILER="$(which g++-$GCC_VERSION)"
-            cmake --build ./BuildCov
-            cd ./BuildCov
-            ./tests/runner
-            lcov --directory . -b ../pennylane_lightning/src --capture --output-file coverage.info
-            lcov --remove coverage.info '/usr/*' --output-file coverage.info
-
+          pip install --upgrade torch==$TORCH_VERSION -f https://download.pytorch.org/whl/cpu/torch_stable.html
+          pip install --upgrade "jax[cpu]"  # This also installs jaxlib
+          pip install --upgrade tensorflow~=$TF_VERSION keras~=$TF_VERSION
```
mlxd, line 98: We should check which versions are tested by default in PennyLane now, as there was a recent update to the packages

mlxd (commented):
Thanks @AmintorDusko 
A few comments below. Also, it may be worth updating all the runner images from 20.04 to 22.04.

.github/workflows/tests_linux.yml
```diff
@@ -14,124 +14,120 @@ env:
...
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ubuntu-tests-reports
+          path: BuildCov/tests/results/report_cpptests.xml
+
+  pythontests:
     strategy:
       matrix:
         os: [ubuntu-20.04]
```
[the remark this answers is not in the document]
AmintorDusko: Yes. It was my intention to update that, but our test requirements are set to 20.04.
When the time comes, you will need to update the repository requirements.

.github/workflows/format.yml
AmintorDusko: Yes. I'm waiting for that.

.github/workflows/tests_linux.yml
AmintorDusko: Makes sense, I will do that.

.github/workflows/tests_linux.yml
```diff
@@ -1,137 +1,133 @@
...
+        os: [ubuntu-22.04]
+
+    name: Python tests
+    runs-on: ${{ matrix.os }}
+
     steps:
-      - name: Cancel previous runs
-        uses: styfle/cancel-workflow-action@0.10.0
+      - name: Checkout PennyLane-Lightning
+        uses: actions/checkout@v2
```
multiphaseCFD, line 76: Do we need to change v2 to v3 here?

.github/workflows/tests_linux.yml
```diff
@@ -1,137 +1,133 @@
...
         with:
-          access_token: ${{ github.token }}
+          path: main
+          fetch-depth: 2
 
       - uses: actions/setup-python@v2
         name: Install Python
```
multiphaseCFD, line 82: v2->v4?

.github/workflows/tests_linux.yml
```diff
@@ -141,88 +137,198 @@ jobs:
...
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Checkout PennyLane-Lightning
+        uses: actions/checkout@v2
+        with:
+          path: main
+          fetch-depth: 2
+
+      - uses: actions/setup-python@v2
```
multiphaseCFD, line 185: Do we need to change v2 to v3 here also?

.github/workflows/build_and_cache_Kokkos_linux.yml
```diff
@@ -0,0 +1,112 @@
...
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.10.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Kokkos execution strategy
+        id: exec_model
+        run: echo "::set-output name=exec_model::[\"SERIAL\"]" # We may also adopt [OPENMP, THREADS] in later iterations
```
chaeyeunpark, line 30: Can't we just set this directly in matrix? I think it is easier to see what's going on.

.github/workflows/build_and_cache_Kokkos_linux.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Kokkos version
+        id: kokkos_version
+        run: echo "::set-output name=kokkos_version::[\"3.6.00\"]"
```
chaeyeunpark, line 34: Same for this.

.github/workflows/build_and_cache_Kokkos_linux.yml
```diff
@@ -0,0 +1,112 @@
...
+          git clone https://github.com/kokkos/kokkos-kernels.git
+          cd kokkos-kernels
+          git checkout ${{ matrix.kokkos_version }}
+          cd -
+
+      - name: Install dependencies (Ubuntu)
+        if: ${{ (((inputs.os == 'ubuntu-latest') || (inputs.os == 'ubuntu-20.04') || (inputs.os == 'ubuntu-22.04')) && (steps.kokkos-cache.outputs.cache-hit != 'true')) }}
+        run: |
+          sudo apt-get update && sudo apt-get -y -q install cmake gcc-10 g++-10 ninja-build git
+          echo "COMPILER=g++-10" >> $GITHUB_ENV
```
chaeyeunpark, line 80: I think it is better to set the compiler outside of this script. Such as
```
env:
   GCC_VERSION: 10
```
and use it e.g.
```
sudo apt-get update && sudo apt-get -y -q install cmake gcc-$GCC_VERSION g++-$GCC_VERSION ninja-build git
```

chaeyeunpark (changes requested):
Nice job @AmintorDusko! Just few comments.

.github/workflows/build_and_cache_Kokkos_linux.yml
AmintorDusko: These parameters are set here and used as an output of this job.
This way, we can have consistent access to this information around our CI/CD pipelines.

.github/workflows/build_and_cache_Kokkos_linux.yml
AmintorDusko: Same as before.

.github/workflows/build_and_cache_Kokkos_linux.yml
mlxd: Unfortunately this is the recommended way to build these types of workflows where one depends on another.
The evaluation order means that the matrix cannot be used for these types of jobs.

.github/workflows/build_and_cache_Kokkos_linux.yml
AmintorDusko: Good idea. Thanks.

.github/workflows/build_and_cache_Kokkos_linux.yml
```diff
@@ -0,0 +1,112 @@
...
+                          -DCMAKE_CXX_COMPILER=${{ env.COMPILER }} \
+                          -DCMAKE_CXX_STANDARD=20 \
+                          -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+                          -G Ninja
+          cmake --build ./Build --verbose
+          cmake --install ./Build
+          cd -
+
+      - name: Build Kokkos kernels library (Ubuntu)
+        if: ${{ (((inputs.os == 'ubuntu-latest') || (inputs.os == 'ubuntu-20.04') || (inputs.os == 'ubuntu-22.04')) && (steps.kokkos-cache.outputs.cache-hit != 'true')) }}
```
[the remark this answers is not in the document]
chaeyeunpark: Aha, I see. So this workflow is called from another workflow.

.github/workflows/build_and_cache_Kokkos_linux.yml
chaeyeunpark: I understand why it is required for python versions (it distinguishes pull_request and `master` merge), but I wasn't sure it is also the case for this workflow. Still, if it is the simplest way to do that, I am okay.

multiphaseCFD:
Thanks @AmintorDusko for the nice work! No more comments from my side.

multiphaseCFD (approved):

.github/workflows/benchmarks.yml
```diff
@@ -1,83 +1,45 @@
...
```
chaeyeunpark, line 45:
```suggestion
            ./BuildGBench/benchmarks/apply_operations --benchmark_filter="^applyOperations_RandOps<double>/LM_all/32/[0-9]+$"
```

.github/workflows/benchmarks.yml
```diff
@@ -1,83 +1,45 @@
...
-      - name: Plot results
-        run: |
-          cd main
-          python .github/workflows/benchmarks/plot_results.py
-
-      - uses: actions/upload-artifact@v2
-        with:
-          name: gates.png
-          path: ./main/gates.png
+            ./BuildGBench/benchmarks/utils
```
chaeyeunpark, line 44:
```suggestion
            ./BuildGBench/benchmarks/utils --benchmark_filter="^[a-z]+_innerProd_cmplx<double>/[0-9]+$"
```

chaeyeunpark (approved):
Thanks for the work! I just have suggestions on the benchmark.

chaeyeunpark (approved):
Thanks again @AmintorDusko! I think @mlxd needs to change the required CIs in the settings (ubuntu-20.24 to ubuntu-22.04).

AmintorDusko:
> Thanks again @AmintorDusko! I think @mlxd needs to change the required CI names in the settings (ubuntu-20.24 to ubuntu-22.04).

@mlxd, I summon thee!

mlxd:
Should we be concerned about the coverage drop?

AmintorDusko:
> Should we be concerned about the coverage drop?

I don't think so. This PR did nothing directly regarding coverage.
This might be a case of proportional change and not an effective change on coverage.
We increase the amount of tests only.

AmintorDusko:
> Should we be concerned about the coverage drop?

This might be work for other PR, though. The report shows that the drop in coverage is connected with cpu_kernels.

PR #365, opened 2022-10-03
Change Lightning to inherit from QubitDevice instead of DefaultQubit

AmintorDusko:
**Context:** Lightning qubit device should inherit from `QubitDevice` instead of `DefaultQubit` now.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


AmintorDusko:
[sc-26718]

pennylane_lightning/lightning_qubit.py
```diff
@@ -70,17 +72,86 @@ def _chunk_iterable(it, num_chunks):
...
+    "SingleExcitationPlus",
+    "SingleExcitationMinus",
+    "DoubleExcitation",
+    "DoubleExcitationPlus",
+    "DoubleExcitationMinus",
+    "QubitCarry",
+    "QubitSum",
+    "OrbitalRotation",
+    "QFT",
+    "ECR",
```
mlxd, line 131: For the ops without direct C++ support, do they still decompose or have a fall-back ?

pennylane_lightning/lightning_qubit.py
```diff
@@ -113,9 +185,40 @@ def __init__(self, wires, *, c_dtype=np.complex128, shots=None, batch_obs=False)
...
+            index (int): integer representing the computational basis state
+        Returns:
+            array[complex]: complex array of shape ``[2]*self.num_wires``
+            representing the statevector of the basis state
+        Note: This function does not support broadcasted inputs yet.
+        """
+        state = np.zeros(2**self.num_wires, dtype=np.complex128)
+        state[index] = 1
+        state = self._asarray(state, dtype=self.C_DTYPE)
+        return self._reshape(state, [2] * self.num_wires)
```
mlxd, line 208: I assume for functions like these we need to replicate what is in default.qubit?
Probably makes more sense at a later time to move such static functions to a separate module in PL, and we can depend upon them at that time.

mlxd (commented):
Amazing effort @AmintorDusko 
Just a few questions mostly.

pennylane_lightning/lightning_qubit.py
```diff
@@ -127,56 +230,122 @@ def _asarray(arr, dtype=None):
...
+        self._state = self._create_basis_state(num)
+
+    # To be able to validate the adjoint method [_validate_adjoint_method(device)],
+    #  the qnode requires the definition of:
+    # ["_apply_operation", "_apply_unitary", "adjoint_jacobian"]
+    def _apply_operation():
+        pass
+
+    def _apply_unitary():
+        pass
```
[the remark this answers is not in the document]
AmintorDusko: I agree that this is annoying. :joy: 
As you said, for now, this should be fine.

pennylane_lightning/lightning_qubit.py
```diff
@@ -226,6 +394,33 @@ def apply_lightning(self, state, operations):
...
+                    "applied on a {} device.".format(operation.name, self.short_name)
+                )
+
+        if operations:
+            self._pre_rotated_state = self.apply_lightning(self._state, operations)
+        else:
+            self._pre_rotated_state = self._state
+
+        if rotations:
+            self._state = self.apply_lightning(np.copy(self._pre_rotated_state), rotations)
```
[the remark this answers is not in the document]
AmintorDusko: I tried to get rid of the copy during development, but in the way that `apply_lightning` is coded we need to provide a copy of `_pre_rotated_state`.
In `apply_lightning`, the inputted state and the outputted state_vector share the same pointer.
[Here](https://github.com/PennyLaneAI/pennylane-lightning/blob/device-inheritance/pennylane_lightning/lightning_qubit.py#L362) is where it happens. 
I didn't think too much about that, because this is the way that was before.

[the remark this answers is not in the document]
AmintorDusko: For now, I see no problem with how `apply_lightning` is coded, though.

pennylane_lightning/lightning_qubit.py
```diff
@@ -127,56 +230,122 @@ def _asarray(arr, dtype=None):
...
+
+        return None
+
+    @property
+    def state(self):
+        dim = 2**self.num_wires
+        batch_size = self._get_batch_size(self._pre_rotated_state, (2,) * self.num_wires, dim)
+        # Do not flatten the state completely but leave the broadcasting dimension if there is one
+        shape = (batch_size, dim) if batch_size is not None else (dim,)
+        return self._reshape(self._pre_rotated_state, shape)
```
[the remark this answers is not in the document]
AmintorDusko: It was fine to keep. 
But as you said we do not support it, so I think the most sensible action is to remove it for now.

tests/test_adjoint_jacobian.py
```diff
@@ -819,56 +819,18 @@ def f(params1, params2):
...
     qml.RZ(params[0], wires=wires[3])
     qml.CRX(params[3], wires=[wires[3], wires[0]])
     qml.PhaseShift(params[4], wires=wires[2])
     qml.CRY(params[5], wires=[wires[2], wires[1]])
-    qml.CRZ(params[5], wires=[wires[0], wires[3]]).inv()
-    qml.PhaseShift(params[6], wires=wires[0]).inv()
+    qml.adjoint(qml.CRZ(params[5], wires=[wires[0], wires[3]]))
+    qml.adjoint(qml.PhaseShift(params[6], wires=wires[0]))
     qml.Rot(params[6], params[7], params[8], wires=wires[0])
     # #     qml.Rot(params[8], params[8], params[9], wires=wires[1]).inv()
```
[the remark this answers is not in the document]
AmintorDusko: Good call! We can enable all gates in the circuit ansatz, now.

pennylane_lightning/lightning_qubit.py
AmintorDusko: The ops without direct C++ support fall back to the `applyMatrix` method. 
We apply the full matrix. Let's call this a first-order approach.

pennylane_lightning/lightning_qubit.py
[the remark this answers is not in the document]
mlxd: Ok, let's call this good for now. I suspect we can always try to improve things at a later date if necessary.

pennylane_lightning/lightning_qubit.py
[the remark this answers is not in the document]
mlxd: Whatever you think is the best plan.

pennylane_lightning/lightning_qubit.py
```diff
@@ -38,8 +37,11 @@
 from pennylane.measurements import MeasurementProcess, Expectation, State
 from pennylane.wires import Wires
 
-# Remove after the next release of PL
-# Add from pennylane import matrix
+ABC_ARRAY = np.array(list(ABC))
+
+# tolerance for numerical errors
+tolerance = 1e-10
```
multiphaseCFD, line 43: I'm curious if there is any reason to set the tolerance as `1e-10`. The minimal positive normal value that fp32 supports if I remember correctly is around `1e-38` or so.

AmintorDusko: At this first moment, we are following [PennyLane](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/devices/default_qubit.py#L37) lead here.
In the future, we can adjust this if needed.

If you are wondering why they choose this number, what I have to say is that this is a reasonably small number, that will be used to check the normalization of the statevector ([here](https://github.com/PennyLaneAI/pennylane-lightning/blob/96cff21e3df9a182eb1c283618683c3c26b3aa23/pennylane_lightning/lightning_qubit.py#L271)). 
In a case like that, we are summing, sometimes, millions of numbers and we can have a propagation of errors inconsistent with the fp32 precision.

multiphaseCFD: Thanks @AmintorDusko for the explanation and the nice work!

pennylane_lightning/lightning_qubit.py
```diff
@@ -18,17 +18,16 @@
 from typing import List
 from warnings import warn
 from os import getenv
-from itertools import islice
+from itertools import islice, product
+from string import ascii_letters as ABC
```
mlxd, line 22: Since `ABC` in Python is normally meant for abstract base classes as `from abc import ABC` we should probably rename this to something else

pennylane_lightning/lightning_qubit.py
```diff
@@ -38,8 +37,11 @@
 from pennylane.measurements import MeasurementProcess, Expectation, State
 from pennylane.wires import Wires
 
-# Remove after the next release of PL
-# Add from pennylane import matrix
+ABC_ARRAY = np.array(list(ABC))
```
mlxd, line 40: Where do we use this?

pennylane_lightning/lightning_qubit.py
```diff
@@ -127,56 +230,103 @@ def _asarray(arr, dtype=None):
...
+
+        if n_basis_state != len(device_wires):
+            raise ValueError("BasisState parameter and wires must be of equal length.")
+
+        # get computational basis state number
+        basis_states = 2 ** (self.num_wires - 1 - np.array(device_wires))
+        basis_states = qml.math.convert_like(basis_states, state)
+        num = int(qml.math.dot(state, basis_states))
+
+        self._state = self._create_basis_state(num)
```
mlxd, line 320: While we do not need this now, we should move this behaviour over to C++

mlxd (commented):
Great work @AmintorDusko 
I'll try to test this new branch as the `lightning.gpu` fallback later. Will approve if all goes well.

pennylane_lightning/lightning_qubit.py
AmintorDusko: We don't. Once again, nice catch!

pennylane_lightning/lightning_qubit.py
AmintorDusko: I completely agree. This is a good thing to be added to our to-do list.

pennylane_lightning/lightning_qubit.py
```diff
@@ -127,56 +227,103 @@ def _asarray(arr, dtype=None):
...
+            norm = qml.math.linalg.norm(state, axis=-1, ord=2)
+            if not qml.math.allclose(norm, 1.0, atol=tolerance):
+                raise ValueError("Sum of amplitudes-squared does not equal one.")
+
+        if len(device_wires) == self.num_wires and Wires(sorted(device_wires)) == device_wires:
+            # Initialize the entire device state with the input state
+            self._state = self._reshape(state, output_shape)
+            return
+
+        # generate basis states on subset of qubits via the cartesian product
```
multiphaseCFD, line 276: Hi @AmintorDusko , I'm wondering if operation in batch needs to be supported here?

AmintorDusko: I don't understand. Could you please expand on your question?

multiphaseCFD: For example, lines [here](https://github.com/PennyLaneAI/pennylane/blob/37a4024bdf12781dced0c847cdc80edfc321f4b6/pennylane/devices/default_qubit.py#L680-L690)

AmintorDusko: Ah... OK! Following a discussion with @mlxd, since we do not support batching yet, we decide to remove it for now.

mlxd: For context, we need CPP threading support to properly enable this, so we will keep it disabled to allow a safe fallback for now.

mlxd (approved):
No more comments from me. Thanks @AmintorDusko

maliasadi (approved):
Thanks @AmintorDusko for the nice work :tada:

albi3ro (approved):
Looks good to me 👍

AmintorDusko:
Thank you, @mlxd, @albi3ro, @multiphaseCFD and @maliasadi!

AmintorDusko:
Some QFT benchmarks, for the full operator matrix and decomposition threshold implemented with the last commits, as a reference:
image

AmintorDusko:
Some benchmarks for the QTF operation illustrate the full matrix decomposition threshold for the operation implemented in the last commits.
image

PR #384, opened 2022-11-08
Add mcmc sampler

trevor-vincent:


pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -309,6 +310,76 @@ class Measures {
...
+        auto s1_qratio = tk->operator()(s1);
+
+        // transition kernel outputs these two
+        auto &s2 = s1_qratio.first;
+        auto &qratio = s1_qratio.second;
+
+        auto s2_plog =
+            log((sv.getData()[s2] * std::conj(sv.getData()[s2])).real());
+
+        auto alph = std::min(1., qratio * exp(s2_plog - s1_plog));
```
mlxd, line 329:
```suggestion
        auto alph = std::min<fp_t>(1., qratio * exp(s2_plog - s1_plog));
```

The failures seem to be related to problems with argument type deduction. Explicitly giving the templated type here most likely will fix the Mac and Windows failures.

mlxd (commented):
Just a quick one for the failures. Will do a more in-depth review later.

pennylane_lightning/src/simulator/Measures.hpp
trevor-vincent: Thank you Lee for the suggestion

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -309,6 +310,76 @@ class Measures {
...
+
+        auto s2_plog =
+            log((sv.getData()[s2] * std::conj(sv.getData()[s2])).real());
+
+        auto alph = std::min<fp_t>(1., qratio * exp(s2_plog - s1_plog));
+        auto ran = distrib(gen);
+
+        if (ran < alph) {
+            return s2;
+        } else {
```
mlxd, line 334: Is it possible to get some tests to cover both branches here?

pennylane_lightning/src/simulator/TransitionKernels.hpp
```diff
@@ -0,0 +1,112 @@
...
+    std::uniform_int_distribution<size_t> distrib_binary_;
+    size_t num_qubits_;
+
+  public:
+    LocalTransitionKernel(size_t num_qubits) {
+        num_qubits_ = num_qubits;
+        gen_ = std::mt19937(rd_());
+        distrib_num_qubits_ =
+            std::uniform_int_distribution<size_t>(0, num_qubits - 1);
+        distrib_binary_ = std::uniform_int_distribution<size_t>(0, 1);
```
mlxd, line 41: Can we assign these values as initializer list arguments, rather than in the constructor body? 
It should also alleviate the coverage warning.

pennylane_lightning/src/simulator/TransitionKernels.hpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    size_t init_state() { return 0; }
```
mlxd, line 44: What does this function do? Seems to just return 0?

pennylane_lightning/src/simulator/TransitionKernels.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    std::pair<size_t, fp_t> operator()(size_t s1) {
```
mlxd, line 46: Again there are some coverage complaints here. Not sure if the c++ tests need to be more explicit, or we need to add some additional information to ensure the coverage is correctly reported.

pennylane_lightning/src/simulator/TransitionKernels.hpp
```diff
@@ -0,0 +1,112 @@
...
+    // : public TransitionKernel<fp_t> {
+  private:
+    std::random_device rd_;
+    std::mt19937 gen_;
+    std::uniform_int_distribution<size_t> distrib_;
+    size_t sv_length_;
+    std::vector<size_t> non_zeros_;
+
+  public:
+    size_t init_state() { return 0; }
```
mlxd, line 72: Why does this function exist again?

mlxd (changes requested):
Thanks @trevor-vincent 
Some suggestions regarding the test coverage. Also, can we add a docstring to each class and associated method?

pennylane_lightning/src/simulator/TransitionKernels.hpp
trevor-vincent: This is a parent class method that needs to be defined in the child classes, even if it is "useless" per say. Let me know if there is a classier way to do this.

pennylane_lightning/src/simulator/TransitionKernels.hpp
trevor-vincent: This is a parent class method that needs to be defined in the child classes, even if it is "useless" per say. Let me know if there is a classier way to do this.

pennylane_lightning/src/simulator/TransitionKernels.hpp
trevor-vincent: Will do

pennylane_lightning/src/simulator/Measures.hpp
trevor-vincent: This should already be covered by the C++ test

multiphaseCFD: let me check.

pennylane_lightning/src/simulator/TransitionKernels.hpp
multiphaseCFD: removed

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -309,6 +310,98 @@ class Measures {
...
+     *
+     * @param transition_kernel User-defined functor for producing transitions
+     * between metropolis states.
+     * @param num_burnin Number of Metropolis burn-in steps.
+     * @param num_samples The number of samples to generate.
+     * @return 1-D vector of samples in binary, each sample is
+     * separated by a stride equal to the number of qubits.
+     */
+    std::vector<size_t>
+    generate_samples_metropolis(const TransitionKernelType &transition_kernel,
```
AmintorDusko, line 361: Is it possible to make the transition_kernel name the input here?
In this way, we could perform the transition_kernel selection inside this method instead of doing it in the bindings as it is now. I left a comment there too.

AmintorDusko (changes requested):
Hey @multiphaseCFD, I left some comments and suggestions.
Nice work so far!

AmintorDusko:
Also, could you please fill in the description for this PR?

pennylane_lightning/lightning_qubit.py
```diff
@@ -782,6 +782,32 @@ def generate_samples(self):
...
         return M.generate_samples(len(self.wires), self.shots).astype(int, copy=False)
 
+    def generate_mcmc_samples(self, kernel_name="Local", num_burnin=100):
+        """Generate Markov chain Monte Carlo(MCMC) samples
+        Args:
+            kernel_name (str): name of transition kernel.
+            num_burnin (int): number of steps that will be dropped.
+
+        Returns:
+            array[int]: array of samples in binary representation with shape ``(dev.shots, dev.num_wires)``
```
mlxd, line 792: Can we add some additional info to the dosctring on the supported kernels for `kernel_name`?

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -309,6 +310,98 @@ class Measures {
...
 
+    /**
+     * @brief Complete a single Metropolis-Hastings step.
+     *
+     * @param sv state vector
+     * @param tk User-defined functor for producing transitions
+     * between metropolis states.
+     * @param gen Random number generator.
+     * @param distrib Random number distribution.
+     * @param s1 initial state
```
mlxd, line 321: I assume `s1` is the basis state index? Or something else? I'd prefer a more descriptive name for this value that is tied to its type and utility.

pennylane_lightning/src/simulator/Measures.hpp
mlxd: Yep, I'd also like to see this.

mlxd (commented):
Thanks @trevor-vincent  and @multiphaseCFD 
Just some follow-up comments

tests/test_measures.py
```diff
@@ -560,6 +560,118 @@ def test_sample_values(self, dev, tol):
...
 
 
+class TestMCMCSample:
+    """Tests that samples are properly calculated."""
+
+    @pytest.fixture(params=[np.complex64, np.complex128])
+    def dev(self, request):
+        return qml.device("lightning.qubit", wires=2, shots=1000, c_dtype=request.param)
+
+    def test_mcmc_sample_dimensions(self, dev):
```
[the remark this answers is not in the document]
multiphaseCFD: good idea!

tests/test_measures.py
```diff
@@ -560,6 +560,118 @@ def test_sample_values(self, dev, tol):
...
+        """Tests if the samples returned by sample have
+        the correct dimensions
+        """
+
+        dev.apply([qml.RX(1.5708, wires=[0]), qml.RX(1.5708, wires=[1])])
+
+        kernel = "Local"
+        num_burnin = 10
+        dev.shots = 10
```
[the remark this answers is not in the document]
multiphaseCFD: sure

pennylane_lightning/src/simulator/Measures.hpp
multiphaseCFD: good point!

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -318,32 +318,35 @@ class Measures {
      * between metropolis states.
      * @param gen Random number generator.
      * @param distrib Random number distribution.
-     * @param s1 initial state
+     * @param init_idx Init index of basis state.
      */
     size_t metropolis_step(const SVType &sv,
                            std::unique_ptr<TransitionKernel<fp_t>> &tk,
```
mlxd, line 324: Since this is coming as a unique_ptr, it doesn't need to be a reference.

pennylane_lightning/src/simulator/TransitionKernels.hpp
```diff
@@ -59,18 +59,19 @@ class LocalTransitionKernel : public TransitionKernel<fp_t> {
               std::uniform_int_distribution<size_t>(0, num_qubits - 1)),
           distrib_binary_(std::uniform_int_distribution<size_t>(0, 1)) {}
 
-    std::pair<size_t, fp_t> operator()(size_t s1) final {
+    std::pair<size_t, fp_t> operator()(size_t init_idx) final {
         size_t qubit_site = distrib_num_qubits_(gen_);
         size_t qubit_value = distrib_binary_(gen_);
-        size_t current_bit = ((unsigned)s1 >> (unsigned)qubit_site) & 1U;
+        size_t current_bit = ((unsigned)init_idx >> (unsigned)qubit_site) & 1U;
```
mlxd, line 65: No C-style casting. Can you use `static_cast` instead?

mlxd (commented):
Thanks @multiphaseCFD 
Some follow-up comments

pennylane_lightning/src/simulator/Measures.hpp
multiphaseCFD: Good spot! Thanks @mlxd !

multiphaseCFD: Do you mean to use `std::move` transfer ownership of `tk` to `metropolis_step`?

mlxd: Yep, assuming it is safe to do so. 

Though, in hindsight, I am now wondering if that would be a good idea. I think not, so forgot I suggested this. Sorry for the confusion.

mlxd: Fine to keep as is, but maybe make it a `const` reference since it won't be modified.

multiphaseCFD: yea, I will make it a const reference. need to pass tk for many iterations, so will keep it as it is.

trbromley (commented):
Thanks @multiphaseCFD @trevor-vincent!

This looks great! How will a PL user enable the MCMC sampling mode? Aside from directly calling the device methods (which may not be known to non-expert users), it would be nice for users to be able to opt-in to MCMC sampling easily. For example, when instantiating the device `qml.device("lightning.qubit", shots=100, mcmc=True)` - though I'm curious if there has been any prior discussion on this?

multiphaseCFD:
> Thanks @multiphaseCFD @trevor-vincent!
> 
> This looks great! How will a PL user enable the MCMC sampling mode? Aside from directly calling the device methods (which may not be known to non-expert users), it would be nice for users to be able to opt-in to MCMC sampling easily. For example, when instantiating the device `qml.device("lightning.qubit", shots=100, mcmc=True)` - though I'm curious if there has been any prior discussion on this?

Hi @trbromley , That's a good point! Let me work on it and discuss with @mlxd , then get back to you.

pennylane_lightning/lightning_qubit.py
```diff
@@ -785,6 +785,36 @@ def generate_samples(self):
...
+        Args:
+            kernel_name (str): name of transition kernel. Current version supports two kernels:
+            ``Local`` and ``NonZeroRandom``. The ``Local`` kernel conducts a 'SpinFlip' local transition
+            between states. The ``Local`` kernel generates a random qubit site and then generates
+            a random number to determine the new bit at that qubit site. The ``NonZeroRandom`` randomly transits
+            between states that have nonzero probability
+            num_burnin (int): number of steps that will be dropped.
+
+        Returns:
+            array[int]: array of samples in binary representation with shape ``(dev.shots, dev.num_wires)``
```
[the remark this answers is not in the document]
multiphaseCFD: Thanks @trbromley ! Done.

pennylane_lightning/lightning_qubit.py
```diff
@@ -785,6 +785,36 @@ def generate_samples(self):
 
         return M.generate_samples(len(self.wires), self.shots).astype(int, copy=False)
 
+    def generate_mcmc_samples(self, kernel_name="Local", num_burnin=100):
+        """Generate Markov chain Monte Carlo(MCMC) samples
```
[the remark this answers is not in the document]
multiphaseCFD: Done!

pennylane_lightning/lightning_qubit.py
```diff
@@ -173,7 +173,18 @@ class LightningQubit(QubitDevice):
...
-    def __init__(self, wires, *, c_dtype=np.complex128, shots=None, batch_obs=False, analytic=None):
+    def __init__(
+        self,
+        wires,
+        *,
+        c_dtype=np.complex128,
+        shots=None,
+        mcmc=False,
+        kernel_name="Local",
+        num_burnin=100,
```
mlxd, line 184: We may need to document these somewhere, otherwise it will be lost into the void. Should the device documentation suffice for these, while we aren't expecting them to be a full-on multi-device implementation? Tagging @trbromley  for thoughts here.

multiphaseCFD: Hi @trbromley, how do you think?

trbromley: Yes would be good to mention the new arguments here: https://docs.pennylane.ai/projects/lightning/en/stable/code/api/pennylane_lightning.LightningQubit.html

and potentially at the bottom of here: https://docs.pennylane.ai/projects/lightning/en/stable/devices.html
Similar to how we have some additional info on parallel adjoint.

trbromley: So if `mcmc=False` then the values for `kernel_name` and `num_burnin` are ignored? I wonder if there's a way to have fewer than 3 arguments.

multiphaseCFD: Here is the timing benchmark results from the python layer bench_ghz_10000_True_True_mcmc
bench_ghz_100000_True_True_mcmc

multiphaseCFD: Here are the accuracy comparison:
 
bench_ghz_20_False_False_mcmc
bench_ghz_25_False_False_mcmc

mlxd: Thanks Shuli --- for the relative error plots, are these averaged over multiple runs, or a single run?
They look very noisy, so we should ideally try this a few times and see if the results are consistent.

multiphaseCFD: Hi @mlxd, It's a single run for multiple shots. Let me run many times and update it then.

multiphaseCFD: > So if `mcmc=False` then the values for `kernel_name` and `num_burnin` are ignored? I wonder if there's a way to have fewer than 3 arguments.

Hi @trbromley, per few arguments when loading a device , for sure, we can reduce the number of arguments at `ctor` by moving the `kernel_name` and `num_burnin` to the method `generate_samples(kernel_name, num_burnin)`. But I would like to suggest to put these arguments at the `ctor` since it will raise error if `kernel_name` and `num_burnin` are not correctly set before running circuit. What do you think?

trbromley: > Hi @trbromley, per few arguments when loading a device , one way that we can reduce the number of arguments at ctor is to move the kernel_name and num_burnin to the method generate_samples(kernel_name, num_burnin). But I would like to suggest to put these arguments at the ctor since it will raise error if kernel_name and num_burnin are not correctly set before running circuit. What do you think?

Ok sounds good, happy to keep the current approach :+1:

mlxd: I'm happy to resolve this issue now

tests/test_measures.py
```diff
@@ -611,6 +611,81 @@ def test_sample_values(self, qubit_device, tol):
...
+        ("Local", 10, 110, [1], qml.PauliZ(wires=[1]), 110),
+        ("Local", 10, 120, [0, 1], qml.PauliX(0) @ qml.PauliZ(1), 120),
+        ("NonZeroRandom", 10, 100, [0], qml.PauliZ(wires=[0]), 100),
+        ("NonZeroRandom", 10, 110, [1], qml.PauliZ(wires=[1]), 110),
+        ("NonZeroRandom", 10, 120, [0, 1], qml.PauliX(0) @ qml.PauliZ(1), 120),
+    ]
+
+    @pytest.mark.parametrize(
+        "kernel,num_burnin,num_shots,measured_wires,operation,shape", test_data_no_parameters
+    )
```
vincentmr, line 632:
```suggestion
    test_data_no_parameters = [
        (10, 100, [0], qml.PauliZ(wires=[0]), 100),
        (10, 110, [1], qml.PauliZ(wires=[1]), 110),
        (10, 120, [0, 1], qml.PauliX(0) @ qml.PauliZ(1), 120),
    ]
    @pytest.mark.parametrize("kernel", ["Local", "NonZeroRandom"])
    @pytest.mark.parametrize(
            "num_burnin,num_shots,measured_wires,operation,shape", test_data_no_parameters
        )
    def test_mcmc_sample_dimensions(
        self, dev, kernel, num_burnin, num_shots, measured_wires, operation, shape
    ):
```

vincentmr (approved):
Just made a couple suggestions. Happy to approve!

pennylane_lightning/src/simulator/Measures.hpp
```diff
@@ -331,6 +332,106 @@ class Measures {
...
+     * @param tk User-defined functor for producing transitions
+     * between metropolis states.
+     * @param gen Random number generator.
+     * @param distrib Random number distribution.
+     * @param init_idx Init index of basis state.
+     */
+    size_t metropolis_step(const SVType &sv,
+                           const std::unique_ptr<TransitionKernel<fp_t>> &tk,
+                           std::mt19937 &gen,
+                           std::uniform_real_distribution<fp_t> &distrib,
```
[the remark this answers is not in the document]
multiphaseCFD: Thanks Vincent! Let's keep as it is now and revisit it next time.

doc/devices.rst
```diff
@@ -98,3 +98,26 @@ If you are computing a large number of expectation values, or if you are using a
...
     import pennylane as qml
     dev = qml.device("lightning.qubit", wires=2, batch_obs=True)
+
+.. raw:: html
+
+    </div>
+
+**Markov Chain Monte Carlo sampling support:**
+
+The ``lightning.qubit`` device allows users to use Markov Chain Monte Carlo (MCMC) sampling method to generate samples. To enable MCMC sampling method for sample generation, initialize a ``lightning.qubit`` device with the ``mcmc=True`` keyword argument, as:
```
trbromley, line 108:
```suggestion
The ``lightning.qubit`` device allows users to use the Markov Chain Monte Carlo (MCMC) sampling method to generate approximate samples. To enable the MCMC sampling method for sample generation, initialize a ``lightning.qubit`` device with the ``mcmc=True`` keyword argument, as:
```

doc/devices.rst
```diff
@@ -98,3 +98,26 @@ If you are computing a large number of expectation values, or if you are using a
...
+
+.. code-block:: python
+
+    import pennylane as qml
+    dev = qml.device("lightning.qubit", wires=2, shots=1000, mcmc=True)
+
+By default, the ``kernel_name`` is ``Local`` and ``num_burnin`` is ``100``. The ``Local`` kernel conducts a ``SpinFlip`` local transition between states. The ``Local`` kernel generates a random qubit site and then generates a random number to determine  the new bit at that qubit site.
```
trbromley, line 115:
```suggestion
By default, the ``kernel_name`` is ``"Local"`` and ``num_burnin`` is ``100``. The local kernel conducts a bit-flip local transition between states. The local kernel generates a random qubit site and then generates a random number to determine  the new bit at that qubit site. 
```
I changed `SpinFlip` to bit-flip, is that correct?

doc/devices.rst
```diff
@@ -98,3 +98,26 @@ If you are computing a large number of expectation values, or if you are using a
...
+The ``lightning.qubit`` device allows users to use Markov Chain Monte Carlo (MCMC) sampling method to generate samples. To enable MCMC sampling method for sample generation, initialize a ``lightning.qubit`` device with the ``mcmc=True`` keyword argument, as:
+
+.. code-block:: python
+
+    import pennylane as qml
+    dev = qml.device("lightning.qubit", wires=2, shots=1000, mcmc=True)
+
+By default, the ``kernel_name`` is ``Local`` and ``num_burnin`` is ``100``. The ``Local`` kernel conducts a ``SpinFlip`` local transition between states. The ``Local`` kernel generates a random qubit site and then generates a random number to determine  the new bit at that qubit site. 
+
+Current version also supports the ``NonZeroRandom`` kernel. The ``NonZeroRandom`` randomly transits between states that have nonzero probability. To enable the ``NonZeroRandom`` kernel and set ``num_burnin``, initialize a ``lightning.qubit`` device as:
```
trbromley, line 117:
```suggestion
The ``lightning.qubit`` device also supports a ``"NonZeroRandom"`` kernel. This kernel randomly transits between states that have nonzero probability. It can be enabled by initializing the device as:
```

pennylane_lightning/lightning_qubit.py
```diff
@@ -160,6 +160,12 @@ class LightningQubit(QubitDevice):
             the expectation values. Defaults to ``None`` if not specified. Setting
             to ``None`` results in computing statistics like expectation values and
             variances analytically.
+        mcmc (bool): Determine whether we use Markov Chain Monte Carlo sampling method when generating samples.
```
trbromley, line 163:
```suggestion
        mcmc (bool): Determine whether to use the approximate Markov Chain Monte Carlo sampling method when generating samples.
```
I want to emphasize that this is an approximate method.

pennylane_lightning/lightning_qubit.py
```diff
@@ -160,6 +160,12 @@ class LightningQubit(QubitDevice):
...
+        kernel_name (str): name of transition kernel. Current version supports two kernels: ``Local`` and ``NonZeroRandom``.
+            The ``Local`` kernel conducts a ``SpinFlip`` local transition between states. The ``Local`` kernel generates a
+            random qubit site and then generates a random number to determine the new bit at that qubit site. The ``NonZeroRandom``
```
trbromley, line 166:
```suggestion
        kernel_name (str): name of transition kernel. The current version supports two kernels: ``"Local"`` and ``"NonZeroRandom"``.
            The local kernel conducts a bit-flip local transition between states. The local kernel generates a
            random qubit site and then generates a random number to determine the new bit at that qubit site. The ``"NonZeroRandom"`` kernel
```

pennylane_lightning/lightning_qubit.py
```diff
@@ -160,6 +160,12 @@ class LightningQubit(QubitDevice):
...
+            randomly transits between states that have nonzero probability.
+        num_burnin (int): number of steps that will be dropped.
```
trbromley, line 168:
```suggestion
        num_burnin (int): number of steps that will be dropped. Increasing this value will
        result in a closer approximation but increased runtime.
```
Is this true?

trbromley (approved):
Thanks @multiphaseCFD! I've left a few suggestions on the documentation side, but otherwise looks great to me!

doc/devices.rst
multiphaseCFD: Correct!

pennylane_lightning/lightning_qubit.py
mlxd: This is a good point --- I think "may" is a more accurate wording here.
@multiphaseCFD can you make a follow-up PR?

mlxd: Unless this is already resolved?

PR #389, opened 2022-12-02
Native implementation of parameter broadcasting

AmintorDusko:
**Context:**
Native implementation of parameter broadcasting on `lightning.qubit`.
This allows the device to accept single circuits with N-dimensional parameter arrays from PennyLane, and execute them on a single tape.

Lightning was updated to accept N-dimensional parameters (parameter broadcasting).
Tests are also introduced, to keep track of changes.

**Breaking Change:**
Deprecate support to python 3.7.


AmintorDusko:
[sc-26720]

AmintorDusko:
The support for parameter broadcasting with the adjoint method is under consideration.

.gitignore
```diff
@@ -29,3 +29,4 @@ cpptests
 /pennylane_lightning/*.pyd
 /pennylane_lightning/src/Kokkos/
 /pennylane_lightning/src/GBenchmarks/
+prototypes/
```
dwierichs, line 32: Is this supposed to be pushed to the repository? Is there a convention about this/the name `prototypes` agreed upon for local prototypes? :)

pennylane_lightning/lightning_qubit.py
```diff
@@ -263,22 +273,35 @@ def reset(self):
...
+        # Do not flatten the state completely but leave the broadcasting dimension if there is one
+        shape = (batch_size, dim) if batch_size is not None else (dim,)
         return self._reshape(self._pre_rotated_state, shape)
 
     def _apply_state_vector(self, state, device_wires):
         """Initialize the internal state vector in a specified state.
+
         Args:
             state (array[complex]): normalized input state of length ``2**len(wires)``
-                or broadcasted state of shape ``(batch_size, 2**len(wires))``
```
dwierichs, line 273: I'm slightly confused here:
1. Why was this line here before?
2. Why is it removed?

pennylane_lightning/lightning_qubit.py
```diff
@@ -263,22 +273,35 @@ def reset(self):
...
+        dim = 1 << len(device_wires)
+        batch_size = self._get_batch_size(state, (dim,), dim)
+
         output_shape = [2] * self.num_wires
+        if batch_size is not None:
+            output_shape.insert(0, batch_size)
+
+        # Lightning doesn't support broadcasted state vector initialization.
+        if batch_size:
+            raise ValueError("Lightning doesn't support broadcasted state vector initialization")
```
dwierichs, line 304: Should we move this up, before the creation of `output_shape`? Does not make a true difference, but conceptually it makes more sense not to create `output_shape` for the case `batch_size is not None`. Also, we can skip the check `if batch_size is not None` in that case (which is effectively done twice currently).

pennylane_lightning/lightning_qubit.py
```diff
@@ -342,6 +365,37 @@ def _apply_operation():
...
+
+        Args:
+            state (array[complex]): the input state tensor
+
+        Returns:
+            The batch dimension (None, if no batching is in action.),
+            and the state with an extra dimension for processing if not batching.
+        """
+        dim = 1 << self.num_wires
+        batch_size = self._get_batch_size(state, (dim,) * self.num_wires, dim)
```
dwierichs, line 395: I'm not sure I understand the `expected_shape` `(dim,)*self.num_wires` here. Why should the state on a device with 3 wires have shape `(8, 8, 8)`?

pennylane_lightning/lightning_qubit.py
```diff
@@ -354,40 +408,61 @@ def apply_lightning(self, state, operations):
         Returns:
             array[complex]: the output state tensor
         """
-        state_vector = np.ravel(state)
+        batch_size_op = self.operations_batch_size(operations)
```
dwierichs, line 411: Note that a `QuantumTape` comes with a pre-processed `batch_size` attribute. Can't we reuse that one, instead of implementing `operations_batch_size`? :)

pennylane_lightning/lightning_qubit.py
```diff
@@ -354,40 +408,61 @@ def apply_lightning(self, state, operations):
...
 
-        if self.use_csingle:
-            # use_csingle
-            sim = StateVectorC64(state_vector)
+        dim = 1 << self.num_wires
+        batch_size_state = self._get_batch_size(state, (dim,) * self.num_wires, dim)
```
dwierichs, line 414: Again I am not entirely sure about the expected shape here.

pennylane_lightning/lightning_qubit.py
```diff
@@ -354,40 +408,61 @@ def apply_lightning(self, state, operations):
...
+                name = o.name.split(".")[0]  # The split is because inverse gates have .inv appended
+                method = getattr(sim, name, None)
+
+                wires = self.wires.indices(o.wires)
+                if method is None:
+                    # Inverse can be set to False since qml.matrix(o) is already in inverted form
+                    method = getattr(sim, "applyMatrix")
+
+                    try:
+                        _matrix = qml.matrix(o)[batch_ind] if o.batch_size else qml.matrix(o)
```
dwierichs, line 452: This means that we compute the batched matrix in each iteration, but then only use one matrix of the batch, right? Can we improve on that, e.g. by first iterating over the operations and then over the batch dimension? This way, for each operation we may compute the batch of matrices, and cache them for one loop over the batch dimension.

pennylane_lightning/lightning_qubit.py
```diff
@@ -354,40 +408,61 @@ def apply_lightning(self, state, operations):
...
+                        _matrix = o.matrix()[batch_ind] if o.batch_size else o.matrix()
+
+                    method(_matrix, wires, False)
+                else:
+                    inv = o.inverse
+                    param = np.transpose(np.array(o.parameters, copy=False))
+                    param = param[batch_ind] if o.batch_size else param
+                    method(wires, inv, param)
+            batched_states[batch_ind, ...] = np.reshape(state_vector, state.shape)
+        batched_states = batched_states if batch_size_op else batched_states[0, ...]
```
dwierichs, line 464: As the whole differentiation between broadcasted states and non-broadcasted states happens within this function, would it maybe be better to 
1. not artificially create a "dummy" batch dimension and extract the first state at the end, but
2. outsource the application code itself and call it once for non-broadcasted cases and repeatedly for broadcasted cases?

pennylane_lightning/lightning_qubit.py
```diff
@@ -760,12 +835,19 @@ def probability(self, wires=None, shot_range=None, bin_size=None):
...
-        return M.probs(device_wires)
+        batched_probs = []
+
+        batch_size, batched_states = self.check_and_guarantee_batched_states_shape(self._state)
+
+        for state in batched_states:
+            state = np.ravel(state)
+            state_vector = StateVectorC64(state) if self.use_csingle else StateVectorC128(state)
+            M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+            batched_probs += [
```
dwierichs, line 846: Should we be using `append` here, rather than `list` addition? :)

pennylane_lightning/lightning_qubit.py
```diff
@@ -774,13 +856,21 @@ def generate_samples(self):
...
 
-        state_vector = StateVectorC64(ket) if self.use_csingle else StateVectorC128(ket)
-        M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+        for state in batched_states:
+            state = np.ravel(state)
+            state_vector = StateVectorC64(state) if self.use_csingle else StateVectorC128(state)
+            M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+            samples = M.generate_samples(len(self.wires), self.shots).astype(int, copy=False)
+
+            batched_samples += [
```
dwierichs, line 868: See above re `list` addition.

pennylane_lightning/lightning_qubit.py
```diff
@@ -813,32 +903,50 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
+        for state in batched_states:
+            state = np.ravel(state)
+
+            state_vector = StateVectorC64(state) if self.use_csingle else StateVectorC128(state)
+            M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+            if observable.name == "SparseHamiltonian":
+                if Kokkos_info()["USE_KOKKOS"] == True:
+                    # converting COO to CSR sparse representation.
+                    CSR_SparseHamiltonian = observable.data[0].tocsr(copy=False)
+                    batched_expval += [
```
dwierichs, line 920: See above re `list` addition

pennylane_lightning/lightning_qubit.py
```diff
@@ -813,32 +903,50 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
-            )
 
-        if observable.name in ["Hamiltonian", "Hermitian"]:
-            ob_serialized = _serialize_ob(observable, self.wire_map, use_csingle=self.use_csingle)
-            return M.expval(ob_serialized)
+            if observable.name in ["Hamiltonian", "Hermitian"]:
+                ob_serialized = _serialize_ob(
+                    observable, self.wire_map, use_csingle=self.use_csingle
+                )
+                batched_expval += [
```
dwierichs, line 936: See above re `list` addition

pennylane_lightning/lightning_qubit.py
```diff
@@ -813,32 +903,50 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
+                    M.expval(ob_serialized),
+                ]
+                continue
+
+            # translate to wire labels used by device
+            observable_wires = self.map_wires(observable.wires)
 
-        # translate to wire labels used by device
-        observable_wires = self.map_wires(observable.wires)
+            batched_expval += [
```
dwierichs, line 944: See above re `list` addition

pennylane_lightning/lightning_qubit.py
```diff
@@ -867,16 +975,24 @@ def var(self, observable, shot_range=None, bin_size=None):
...
-        state_vector = StateVectorC64(ket) if self.use_csingle else StateVectorC128(ket)
-        M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
+            state_vector = StateVectorC64(state) if self.use_csingle else StateVectorC128(state)
+            M = MeasuresC64(state_vector) if self.use_csingle else MeasuresC128(state_vector)
 
-        # translate to wire labels used by device
-        observable_wires = self.map_wires(observable.wires)
+            # translate to wire labels used by device
+            observable_wires = self.map_wires(observable.wires)
+            batched_var += [
```
dwierichs, line 990: See above re `list` addition

pennylane_lightning/lightning_qubit.py
```diff
@@ -867,16 +975,24 @@ def var(self, observable, shot_range=None, bin_size=None):
             samples = self.sample(observable, shot_range=shot_range, bin_size=bin_size)
             return np.squeeze(np.var(samples, axis=0))
 
-        # Initialization of state
-        ket = np.ravel(self._pre_rotated_state)
+        batched_var = []
```
dwierichs, line 978: Maybe out of scope for this PR, and mostly out of curiosity: Why does `var` not require the distinction between `SparseHamiltonian`, `Hamiltonian/Hermitian` and others? :)

tests/test_apply.py
```diff
@@ -586,71 +586,12 @@ def circuit():
...
-
 class TestLightningQubitIntegration:
-    """Integration tests for lightning.qubit. This test ensures it integrates
+    """Integration tests for lightning.qubit. These tests ensure it integrates
     properly with the PennyLane interface, in particular QNode."""
 
-    def test_load_default_qubit_device(self):
-        """Test that the default plugin loads correctly"""
+    def test_load_lightning_qubit_device(self):
+        """Test that the lightning plugin loads correctly"""
```
dwierichs, line 594: Nice catch!

dwierichs (commented):
Thanks @AmintorDusko, this looks great!
Just have a few questions, mostly regarding the usage of `expected_shape` in `_get_batch_size`, the iteration order in `apply_lightning` and minor performance aspects like `list` addition vs. `append`.
I'll review the tests in a second PR.

.gitignore
AmintorDusko: Yes. It is. We were lacking a place to implement prototypes, tests and benchmarks. This is for that.

pennylane_lightning/lightning_qubit.py
AmintorDusko: This line was here because of a previous attempt at broadcasting. I'm removing it now because in a first approach we are not supporting initializing a state vector in a broadcasted shape.

pennylane_lightning/lightning_qubit.py
AmintorDusko: Nice suggestion. Thank you.

pennylane_lightning/lightning_qubit.py
AmintorDusko: Oops. Thank you for catching that.

pennylane_lightning/lightning_qubit.py
AmintorDusko: The apply method receives a list of operations to be applied, not a tape.

.gitignore
dwierichs: Brilliant, thanks! :)

pennylane_lightning/lightning_qubit.py
AmintorDusko: As a matter of fact, this wouldn't improve the calculation. Lightning needs to initialize a StateVector instance (a simulator) for each batch, moving it to an inner loop would increase the number of times the python and C++ layers would need to communicate data. It is more performant to instantiate the simulator one time and go through the operations, even if you need to call for the same matrix a number of times. We would expect that, for lightning, in normal operating conditions the chance of having a large state vector to manipulate is larger than having:  1) a large number of parameters to broadcast and 2) to build several matrices.

pennylane_lightning/lightning_qubit.py
dwierichs: Thanks for clarifying, makes sense :)

pennylane_lightning/lightning_qubit.py
dwierichs: Okay. I suppose as we do not recompute the `batch_size` but just retrieve it from the operations -- or from the first one that has a `batch_size!=None` -- it is not very expensive :)

pennylane_lightning/lightning_qubit.py
dwierichs: Ohhh, okay, I see.
I suppose for the usual use case of lightning, the computation time of `qml.matrix(o)` indeed is negligible in many cases. Would there be any use in precomputing matrices nonetheless, e.g. for rare cases of very large (~Hilbert space dimension size) operation matrices? :thinking: Or is this a prohibitive regime anyways?
Sorry in case this is a stupid question, I'm not too used to the lightning mindset :D

pennylane_lightning/lightning_qubit.py
AmintorDusko: Neither lightning nor default qubits support the calculation of var for Hamiltonians.

pennylane_lightning/lightning_qubit.py
AmintorDusko: I thought about that, but this would increase the maintenance burden and ask for a more significant refactor.
More work and no apparent increase in performance.

pennylane_lightning/lightning_qubit.py
dwierichs: Ah, thank you a lot for the clarification, I somehow missed that this happens in the `method is None` branch when considering the repeated `qml.matrix` call! :grimacing: Sounds good to me!

pennylane_lightning/lightning_qubit.py
dwierichs: Okay, fair enough. I think I personally would find the other structure a bit easier to digest, but that's certainly a personal preference. The benefit of course would not be in performance, that's right :)

pennylane_lightning/lightning_qubit.py
dwierichs: ...fair point :laughing:

AmintorDusko:
> Thank you @AmintorDusko for the clarifications!
> 
> The tests look 100 Just to make sure: Is the branch for unsupported custom operations explored in the `IsingAB` tests, or do decompositions avoid that `apply_lightning` actually encounters operations for which no method is known to the lightning device?

Hi @dwierichs, thank you for your nice review! Answering your question, the branch where `lightning` identifies that there is no method is tested by the `QubitUnitary` operations, for example.

mlxd:
Thanks @AmintorDusko 

One last check, can we do some before/after runtime comparision against v0.27.0 for a workflow that should be broadcastable?

pennylane_lightning/lightning_qubit.py
```diff
@@ -354,40 +406,61 @@ def apply_lightning(self, state, operations):
...
             array[complex]: the output state tensor
         """
-        state_vector = np.ravel(state)
+        batch_size_op = self.operations_batch_size(operations)
 
-        if self.use_csingle:
-            # use_csingle
-            sim = StateVectorC64(state_vector)
+        dim = 1 << self.num_wires
+        batch_size_state = self._get_batch_size(state, (2,) * self.num_wires, dim)
```
multiphaseCFD, line 412: Is it possible to change 'batch_size_state' to 'batch_size' to make it consistent with previous text (like #L393)?

pennylane_lightning/lightning_qubit.py
```diff
@@ -813,32 +896,46 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
+                ob_serialized = _serialize_ob(
+                    observable, self.wire_map, use_csingle=self.use_csingle
+                )
+                batched_expval.append(M.expval(ob_serialized))
+                continue
 
-        # translate to wire labels used by device
-        observable_wires = self.map_wires(observable.wires)
+            # translate to wire labels used by device
+            observable_wires = self.map_wires(observable.wires)
```
multiphaseCFD, line 933: Does this mean all the states should be observed with the same wires now? If yes, do we need to explictly mention it in docstring?

pennylane_lightning/lightning_qubit.py
AmintorDusko: My intention with the suffix was to have it on the same foot as `batch_size_op`.
But I can change it. Not a problem, at all.

pennylane_lightning/lightning_qubit.py
AmintorDusko: Sorry, I'm not exactly getting what you are asking. We, in principle, have the same behaviour for broadcasted parameters + measurements as before. 
Does that answer your question?

multiphaseCFD: Thanks! Aha, What I mentioned above should be a parameter "scatter" behaviour.

multiphaseCFD:
Thanks @AmintorDusko for the nice work!

pennylane_lightning/lightning_qubit.py
multiphaseCFD: Just take the one better in your mind.

AmintorDusko:
I present some benchmarks comparing the native parameter broadcasting support and the release v0.27.0. 
In all benchmarks, I consider the same circuit:
```python
x = np.linspace(0.11, 1, num_param)
y = np.linspace(0.21, 1, num_param)

dev = qml.device("lightning.qubit", wires=wires_total)
@qml.qnode(dev, diff_method='parameter-shift')
def circuit(x, y):
    for nn in range(wires_total):
        qml.RX(x, wires=0)
        qml.RY(y, wires=0)
```
So the free parameters here are the number of qubits (`wires_total`), and the number of broadcasted parameters (`num_param`).

image

image

image

image

As we can see there is a performance threshold between the native (this PR) and v0.27 approaches. When increasing the number of broadcasted parameters, the threshold moves to the right increasing the windows where the native approach is more performant. These results suggest that the user might benefit from having a switch to control when using the v0.27 approach, with broadcast_expand and multiple tapes or not.

tests/test_broadcasting.py
```diff
@@ -0,0 +1,780 @@
...
+    )
+    def test_apply_operation_two_wires_with_parameters(
+        self, qubit_device, tol, operation, input, expected_output, phi, theta, omega
+    ):
+        """Tests that applying an operation yields the expected output state for two wire
+        operations that have parameters."""
+
+        dev = qubit_device(wires=2)
+        dev._state = np.array(input).reshape(2 * [2]).astype(dev.C_DTYPE)
+        dev.apply([operation(np.array(phi), np.array(theta), np.array(omega), wires=[0, 1])])
```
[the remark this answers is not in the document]
AmintorDusko: There is no change in the user interface with this implementation. 
Everything should work as before.

multiphaseCFD (approved):

PR #391, opened 2022-12-07
Deprecate 3.7 wheels build

AmintorDusko:
**Context:**
Disable 3.7 wheels building for lightning and update tests for measures.

**Breaking Change:**
Deprecate support to python 3.7.


mlxd (approved):
:100:

PR #394, opened 2022-12-13
Remove use of inverse and base name properties

albi3ro:
Tied to PennyLane [PR #3518](https://github.com/PennyLaneAI/pennylane/pull/3518).

We are now removing use of in-place inversion.

To do this, PennyLane is removing:

* `Operation._inverse`
* `Operation.inverse`, both getter and setter
* `Operation.inv()`
* `Operation.base_name`: only existed because inversion modified `Operation.name`

Instead, we want to encourage use of `qml.adjoint(op)` or `qml.pow(op, -1)`.  Most PennyLane operations have default behaviour that will decompose those operations into things natively supported by devices. Like:

```
qml.adjoint(qml.PauliX(0)) -> qml.PauliX(0)
qml.adjoint(qml.RX(1.234, 0)) -> qml.RX(-1.234, 0)
```

All Lightning must do is not support `"Adjoint(PauliX)"` and it will receive `PauliX` instead.

The only special cases that don't have decompositions are `S`, `T`, `SISWAP`, and `ISWAP`.  Lightning already has support for the adjoint of these gates.

Once we remove in-place inversion, we have cut down on the mutability of Operators. It also removes the responsibility for handling inversion from the base class `Operation` to a wrapper class that can modify any `Operator`.


chaeyeunpark (commented):
Thanks for adding the context. Just wondering whether we should add tests for `qml.adjoint`.

albi3ro:
@chaeyeunpark  In `tests/test_gates.py`, `test_inverse_unitary_correct`  currently uses `qml.adjoint` instead of `op.inv()`.

https://github.com/PennyLaneAI/pennylane-lightning/blob/e790f9212b609798820b3b3aa0b619f1cb5cc81b/tests/test_gates.py#L115

Any other specific cases you can think of adding?

AmintorDusko:
> @chaeyeunpark In `tests/test_gates.py`, `test_inverse_unitary_correct` currently uses `qml.adjoint` instead of `op.inv()`.
> 
> https://github.com/PennyLaneAI/pennylane-lightning/blob/e790f9212b609798820b3b3aa0b619f1cb5cc81b/tests/test_gates.py#L115
> 
> Any other specific cases you can think of adding?

Yes. I have been migrating `op.inv()` to `qml.adjoint` in the last few PRs, to smooth out the transition.

AmintorDusko (approved):
Thanks for that @albi3ro!

chaeyeunpark (approved):
Thanks @albi3ro!

PR #403, opened 2023-01-23
Improve installation supports with external builds

mlxd:
**Context:** The CMake build system, and associated targets, do not fully conform to a CMake installation step. As such, to allow better reproducibility across builds, and integration with platforms such as spack, we need to make some adjustments to best support this.

**Description of the Change:** Change benchmark and test target names, macro variable usage and header visibility. Running `cmake --install ./build` following a compilation step will now install the headers and compiled binaries into a predefined target location with `bin` `include` and `lib` directories.

**Benefits:** Enables better support for installation of lightning targets, headers, and spack integration.

**Possible Drawbacks:** 

**Related GitHub Issues:**


pennylane_lightning/src/benchmarks/Bench_LinearAlgebra.cpp
```diff
@@ -42,12 +42,12 @@ static void create_random_cmplx_vector(benchmark::State &state) {
     }
 }
 BENCHMARK(create_random_cmplx_vector<float>)
-    ->RangeMultiplier(1l << 3)
```
chaeyeunpark, line 45: It is strange that `clang-tidy` didn't complain about this. Or don't we run `clang-tidy` for benchmark code?

pennylane_lightning/src/benchmarks/Bench_LinearAlgebra.cpp
```diff
@@ -111,18 +111,20 @@ template <class T> static void omp_innerProd_cmplx(benchmark::State &state) {
 
     for (auto _ : state) {
         std::complex<T> res(.0, .0);
+        // Create indirection to avoid GCC issue with AVX512 compilation
```
chaeyeunpark, line 114: Nice catch!

chaeyeunpark (commented):
First iteration. Looks good so far.

pennylane_lightning/src/benchmarks/Bench_LinearAlgebra.cpp
mlxd: I think we never ran it in the original submissions. I would've thought we'd see the complaints earlier too.

pennylane_lightning/src/benchmarks/Bench_LinearAlgebra.cpp
mlxd: The simplest fix for a very obtuse error

multiphaseCFD: +1

pennylane_lightning/src/util/LinearAlgebra.hpp
```diff
@@ -81,16 +81,12 @@ omp_innerProd(const std::complex<T> *v1, const std::complex<T> *v2,
 #pragma omp declare \
             reduction (sm:std::complex<T>:omp_out=ConstSum(omp_out, omp_in)) \
             initializer(omp_priv=std::complex<T> {0, 0})
-#endif
```
AmintorDusko, line 84: I loved what you did here. This always bugged me.

multiphaseCFD (commented):
Thanks @mlxd for the nice work and looks nice for me.

CMakeLists.txt
```diff
@@ -108,6 +108,12 @@ install(TARGETS pennylane_lightning
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib
         INCLUDES DESTINATION include
+        PUBLIC_HEADER DESTINATION include
+)
+
+install(DIRECTORY
```
AmintorDusko, line 114: I don't think I saw any occurrence of  `cmake --install ./build` in our workflows.
I like the idea of using such a feature in some of our workflows to test and serve as examples.
Should we also change the [README.rst](https://github.com/PennyLaneAI/pennylane-lightning#readme)?

AmintorDusko (commented):
Nice work so far!

CMakeLists.txt
mlxd: I think while that's a good plan, I'd like to test the waters first with Spack, and see if this is something useful. If so, we can support it through CLI installations and document it as such. I expect we will be updating this based on end-user feedback so rather than moving a lot, we can leave as is for now.

vincentmr:
Hi @mlxd, looks good to me. I'm wondering, why did the benchmarks and tests got renamed with the `pennylane_lightning_` prefix?

mlxd:
> Hi @mlxd, looks good to me. I'm wondering, why did the benchmarks and tests got renamed with the `pennylane_lightning_` prefix?

Thanks @vincentmr. With this update, we have the option to install the built binaries into a given prefix path from CMake, and for our updates to support [spack](spack.readthedocs.io/). If we installed our test runner (previously named `runner`) into a `/usr/local/bin` directory, it would be very ambiguous what it is doing (the same with the benchmarking binary). The naming update was to avoid conflicting with any other packages that have similiarly named binaries, by prepending our package name to the binary names.

multiphaseCFD (approved):

AmintorDusko (approved):
Thank you for that!

AmintorDusko:
Hey @mlxd, black just released its new version. Before merging, please update black locally and run:
`black -l 100 ./pennylane_lightning/ ./tests`

mlxd:
> `black -l 100 ./pennylane_lightning/ ./tests`

Done. Will merge after the CI passes

mlxd:
> `black -l 100 ./pennylane_lightning/ ./tests`

Done. Will merge after the CI passes

mlxd:
> `black -l 100 ./pennylane_lightning/ ./tests`

Done. Will merge after the CI passes

PR #410, opened 2023-01-31
Downgrade Numpy to avoid conflicts

AmintorDusko:
**Context:** Numpy last minor version 1.24 has some conflicts with Autograd.

**Description of the Change:** For now we'll limit installation to Numpy < 1.24

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/wheel_win_x86_64.yml
+++ b/.github/workflows/wheel_win_x86_64.yml
@@ -148,7 +148,7 @@ jobs:
             pip install pybind11 cmake~=3.24.0 build
 
           # Testing of built wheels
-          CIBW_TEST_REQUIRES: numpy~=1.21 scipy pytest pytest-cov pytest-mock flaky
+          CIBW_TEST_REQUIRES: numpy<1.24 scipy pytest pytest-cov pytest-mock flaky
 
           CIBW_BEFORE_TEST: |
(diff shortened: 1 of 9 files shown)


.github/workflows/wheel_win_x86_64.yml
```diff
@@ -148,7 +148,7 @@ jobs:
             pip install pybind11 cmake~=3.24.0 build
 
           # Testing of built wheels
-          CIBW_TEST_REQUIRES: numpy~=1.21 scipy pytest pytest-cov pytest-mock flaky
+          CIBW_TEST_REQUIRES: numpy~=1.23 scipy pytest pytest-cov pytest-mock flaky
```
mlxd, line 151: Is there any reason for using this instead of the <1.24 used in the other workflow files?

multiphaseCFD, line 151: why not setting numpy version <1.24?

mlxd (approved):
Thanks @AmintorDusko 
One quick question from me. Will approve.

.github/workflows/wheel_win_x86_64.yml
AmintorDusko: Windows had a problem with that.

multiphaseCFD (approved):
Thanks @AmintorDusko! Approved.

PR #424, opened 2023-03-01
Serialize obs with pauli reps into hamiltonians

timmysilv:
**Context:**
Arithmetic operators are the latest craze in PennyLane, and lightning should support them natively as well! If an observable (that isn't a Tensor or Hamiltonian) has a representation in the Pauli basis, it will be saved on the Operator. Those representations can easily be converted into what is passed to the existing C++ bindings for PennyLane Hamiltonians.

- Add serializing helpers that convert `PauliSentence` objects into `HamiltonianCxx` lightning classes
- Use this serialization for any observable that meets 2 criteria:
  1. not a Tensor, Hamiltonian, or otherwise already has some kernel (Hadamard or Pauli op)
  2. has a non-None `_pauli_rep` assigned to it
- Clean up the serialization decision tree: `_obs_has_kernel` did some consideration for Hamiltonians and Tensors that wasn't used because it was only called on things we knew were neither of those. The new flow should be simpler and more readable, and it lends itself better to op-arithmetic support

**Benefits:**
- `qml.prod` and other arithmetic operators were previously defaulting to `qml.matrix(op)`, and that is not efficient. With this PR, they are being serialized (via `Hamiltonian` for now) and computed with that C++ kernel. This change supports all operators with a `_pauli_rep` attribute.
- I did some basic benchmarking measuring the expectation of large Sum ops (also in a QNode with adjoint-diff) and it takes the same time as Hamiltonians - see latest graph below

**Possible Drawbacks:**
This won't cover all cases. For example, consider the following:
```pycon
>>> pauli_hadamard = qml.s_prod((1/np.sqrt(2)), qml.sum(qml.PauliX(1), qml.PauliZ(1)))
>>> all_pauli = qml.prod(qml.PauliX(0), pauli_hadamard)
>>> uses_hadamard = qml.prod(qml.PauliX(0), qml.Hadamard(1))
>>> print(all_pauli._pauli_rep is None, uses_hadamard._pauli_rep is None)
(False, True)
```

`all_pauli` and `uses_hadamard` are the same, and are valid observables, but only the former will be serialized. This is most unfortunate because there are kernels for Pauli ops _and_ Hadamard, but I don't cover that with this PR. ~Note that as a sideways fix to this, I propose a Pauli representation for Hadamard in another PR (linked below).~ EDIT: this is a larger problem with op-arithmetic (should also consider `qml.Hermitian`) and needs more work on the PL side.


tests/test_serialize.py
```diff
@@ -176,7 +176,7 @@ def test_hermitian_return(self, use_csingle):
             ),
             [0, 1],
         )
-        s[0] == s_expected
+        assert s[0] == s_expected
```
AmintorDusko, line 179: Nice catch!

AmintorDusko (commented):
Thank you, @timmysilv. Nice job here!
Could you please add some benchmarks to this PR showing how these changes impact performance?
Also, you need to update the changelog.

timmysilv:
I generated random Pauli sentences for 1-8 wires with a reasonable number of terms in each, then converted each into an old-style Hamiltonian, as well as a new-style arithmetic op (with a pauli rep). I also downloaded Hamiltonians for H6, H2O and BH3 (12, 14, 16 qubits, respectively) and stuck those into my benchmark data. I ran `dev.expval(obs)` 100 times for both types, and I ran a basic QNode with `diff_method=adjoint` that does `RX(np.array(1.1), 0)` then returns `dev.expval(obs)` 20 times for both types.

The two observable types are at near-parity, with some solid boosts using the new operator type! I'll note that occasionally, I saw the old-school observable winning by microseconds in very small systems (1-2 qubits, a handful of terms) but I think that isn't very significant.

Please note that the data shown below is plotted with a log scale, meaning that the new observable type does significantly better in larger systems.

image

pennylane_lightning/_serialize.py
```diff
@@ -87,6 +87,8 @@ def _serialize_named_obs(ob, wires_map: dict, use_csingle: bool):
     """Serializes a Named observable"""
     named_obs = NamedObsC64 if use_csingle else NamedObsC128
     wires = [wires_map[w] for w in ob.wires]
+    if ob.name == "Identity":
+        wires = wires[:1]
```
timmysilv, line 91: new arithmetic operators might decompose into a new Identity with multiple wires, but the C++ hard-codes Identity to only have 1 wire. @mlxd I remember there was some concern about this new Identity feature - is this a bad thing to do here? It feels not great.

AmintorDusko:
> I generated random Pauli sentences for 1-8 wires with a reasonable number of terms in each, then converted each into an old-style Hamiltonian, as well as a new-style arithmetic op (with a pauli rep). I also downloaded Hamiltonians for H6, H2O and BH3 (12, 14, 16 qubits, respectively) and stuck those into my benchmark data. I ran `dev.expval(obs)` 100 times for both types, and I ran a basic QNode with `diff_method=adjoint` that does `RX(np.array(1.1), 0)` then returns `dev.expval(obs)` 20 times for both types.
> 
> The two observable types are at near-parity, with some solid boosts using the new operator type! I'll note that occasionally, I saw the old-school observable winning by microseconds in very small systems (1-2 qubits, a handful of terms) but I think that isn't very significant.
> 
> Please note that the data shown below is plotted with a log scale, meaning that the new observable type does significantly better in larger systems.
> 
> Also, the performance boost is likely due to pre-processing. Hamiltonians will use [hamiltonian_expand(group=False)](https://github.com/PennyLaneAI/pennylane/blob/f8bbd7f42ef29f1b95090896c0f27f16137a7193/pennylane/_device.py#L748) to split tapes, whereas Sum observables will use [sum_expand](https://github.com/PennyLaneAI/pennylane/blob/f8bbd7f42ef29f1b95090896c0f27f16137a7193/pennylane/_device.py#L754). Given that the observables should be identical other than type/python representation, this makes the most sense. Example with my code (16-qubit Hamiltonian):
> 
> ```
> >>> ham_circuit = qml.tape.QuantumScript(measurements=[qml.expval(hams[-1])])
> >>> op_circuit = qml.tape.QuantumScript(measurements=[qml.expval(ops[-1])])
> >>> ham_tapes, _ = qml.transforms.hamiltonian_expand(ham_circuit, group=False)
> >>> op_tapes, _ = qml.transforms.sum_expand(op_circuit)
> >>> len(ham_tapes), len(op_tapes)
> (771, 433)
> ```
> 
> image
> 
> PS: I converted my notebook to [a python file](https://gist.github.com/timmysilv/eaf26a09e6cade0d073bfdd3a7cf3d3c) for your viewing (if you're curious about what I did). I can't upload .py or .ipynb files to a PR, so I made a gist!

Hi @timmysilv. Thank you for the benchmarks! I'm happy to see the performance improvements.

pennylane_lightning/_serialize.py
timmysilv: ok I at least figured out why I only sometimes saw the issue. In non-adjoint expectation measurements, this is [handled by QubitDevice](https://github.com/PennyLaneAI/pennylane-lightning/blob/08dfbe0f8e7f962f2034b032e63ad2fa4eed0e1e/pennylane_lightning/lightning_qubit.py#L803), thereby never calling `_serialize_ob(servables)` on it. However, adjoint-diff QNodes don't do this check while running `_process_jacobian_tape`, and will [blindly call](https://github.com/PennyLaneAI/pennylane-lightning/blob/08dfbe0f8e7f962f2034b032e63ad2fa4eed0e1e/pennylane_lightning/lightning_qubit.py#L486) `_serialize_observables` on it. Is there a strong reason we do the pre-processing and defaulting to super in only non-adjoint computations?

timmysilv:
This needs more work. I just ran some validation and it makes an actual difference with results. I'm not sure why, but this big slow molecule demonstrates it:
```python
import pennylane as qml
from timeit import default_timer as timer

ham = qml.data.load("qchem", molname="BH3", basis="STO-3G", bondlength=0.86)[0].hamiltonian
op = qml.pauli.pauli_sentence(ham).operation()

dev = qml.device("lightning.qubit", op.wires)
@qml.qnode(dev)
def c_ham():
    return qml.expval(ham)

@qml.qnode(dev)
def c_op():
    return qml.expval(op)
```
expectation with hamiltonian is 10.295583513349388, but with op it's 9.790384688804833. This shouldn't be related to the Identity change because that isn't used in non-adjoint expectations. How to convince myself that the observables are indeed equivalent... will try on Monday

AmintorDusko:
> This needs more work. I just ran some validation and it makes an actual difference with results. I'm not sure why, but this big slow molecule demonstrates it:
> 
> ```python
> import pennylane as qml
> from timeit import default_timer as timer
> 
> ham = qml.data.load("qchem", molname="BH3", basis="STO-3G", bondlength=0.86)[0].hamiltonian
> op = qml.pauli.pauli_sentence(ham).operation()
> 
> dev = qml.device("lightning.qubit", op.wires)
> @qml.qnode(dev)
> def c_ham():
>     return qml.expval(ham)
> 
> @qml.qnode(dev)
> def c_op():
>     return qml.expval(op)
> ```
> 
> expectation with hamiltonian is 10.295583513349388, but with op it's 9.790384688804833. This shouldn't be related to the Identity change because that isn't used in non-adjoint expectations. How to convince myself that the observables are indeed equivalent... will try on Monday.
> 
> EDIT: it's related to how they are split:
> 
> ```
> >>> dev.expval(ham), dev.expval(op)
> (10.295583513349388, 10.295583513349388)
> ```

Hmm. You are right. There is something weird here.
I tested some smaller Hamiltonians, with your script.
I observed that we have some matches and some mismatches.
`molname="H2", basis="STO-3G", bondlength=1.1` - match
`molname="HeH+", basis="STO-3G", bondlength=1.1` - mismatch
`molname="H3+", basis="STO-3G", bondlength=1.1` - mismatch
`molname="H4", basis="STO-3G", bondlength=1.1` - match
`molname="LiH", basis="STO-3G", bondlength=1.11` - mismatch

I hope this can help you map out the problem.

mlxd:
> ok, I've done a bunch of little changes locally, and this is now looking a lot better. I will re-open this as officially ready for review when the relevant fixes are in PennyLane itself, but the associated problems have been identified and are being addressed. New results:
> 
> image

Thanks @timmysilv 
One question --- does the performance order invert beyond the 16 qubit count, or do they follow each other along the same curve? If this slows down beyond the number of samples taken, we should run a larger sim and push things into the 20 qubit regime to be sure.

timmysilv:
I was wondering the same thing, so fortunately I have the answer for you! I just got Utkarsh to give me a 20-qubit, 2951-term Hamiltonian for Li2 and they remain neck and neck even up there (the Sum op remains the teeniest bit faster). I also tried with a 24-qubit, 30k-term Hamiltonian but my poor old MacBook Air just couldn't do it, so you'd need to ask a more powerful computer to confirm what happens beyond 20.

timmysilv:
PRs (and follow-up work) needed for this to be ready:
- PennyLaneAI/pennylane#3912
- PennyLaneAI/pennylane#3915
  - follow-up: also need to add the device capability to lightning in this PR
- PennyLaneAI/pennylane#3899
  - follow-up: also need the replacement of `Unwrap(tapes)` in interfaces

mlxd:
> I was wondering the same thing, so fortunately I have the answer for you! I just got Utkarsh to give me a 20-qubit, 2951-term Hamiltonian for Li2 and they remain neck and neck even up there (ran a few times locally, they seem pretty much equal. Sums win by up to 5%). I also tried with a 24-qubit, 30k-term Hamiltonian but my poor old MacBook Air just couldn't do it, so you'd need to ask a more powerful computer to confirm what happens beyond 20.

Thanks @timmysilv . I think given the agreement at 20, the bottleneck is likely no longer the routines modified here, and may instead be other parts of the pipeline. I'd say that's fine for now.

timmysilv:
Chatted with Christina a couple times about this. We've ultimately decided to not make these hot-fix changes to `tape.get_parameters`, `tape.set_parameters`, `UnwrapTape`, or `tape.diagonalizing_gates` (with Sums) because these are all symptoms of greater problems unrelated to Lightning. `Sum` will be far less performant at first, but it will technically work. The big performance hits will come from tape unwrapping and trying to diagonalize (ie. compute the matrix of) large Sums, but those are pre-processing things in PennyLane itself, and they will be addressed one day.

Some quick timing evaluating a QNode returning `expval(obs)` with `diff_method="adjoint"`:
num wires, num terms | Hamiltonian obs [ms] | Sum obs [ms]
-- | -- | --
7, 64 | 2.96621045 | 34.45110835
8, 185 | 6.2983354 | 196.15018125
12, 919 | 53.59318335 | 14743.9977625

Even the Hamiltonians are slower than expected on current master.

tl;dr - performance will _not_ be what we saw in earlier graphs, but this PR is ready to go.

albi3ro:
Ok so these times are worryingly problematic. 

This is occuring because `Sum` defines diagonalizing gates, and `QubitDevice` assumes that if the diagonalizing gates are defined, we want to compute and use them.

I *don't* want to remove the diagonalizing gates from `Sum` because we do want to offer them.  The problem is in `QubitDevice` always assuming that we want to compute them.

Since `QubitDevice` now has a limited lifetime, I'm OK with putting in a quick patch in `QubitDevice.execute` where if the device is lightning, we don't compute the diagonalizing gates of any sum classes. Something like:
```
if self.short_name == 'lightning.qubit':
       # patch for not computing the diagonalizing gates of Sum
       # see QuantumScript.diagonalizing_gates
        rotation_gates = []
        for observable in circuit.observables:
           if not isinstance(observable, qml.ops.op_math.Sum):
                with contextlib.suppress(qml.operation.DiagGatesUndefinedError):
                    rotation_gates.extend(observable.diagonalizing_gates())
else:
    rotation_gates = circuit.diagonalizing_gates
```

Ugly, but gets the job done for now. This is why I want the new device interface.

timmysilv:
so my average timing were nonsense because the matrix is cached... the first execution (ran with snakeviz) of the 12-qubit Sum just took me 250 seconds (indeed, mostly running `diagonalizing_gates`), and the second took 17 seconds. This means that _not_ getting rid of `diagonalizing_gates` would be disastrous, but even if we _do_ avoid it for lightning, a `Sum` of this size would still take ~15 seconds in `tape.get/set_parameters` from unwrapping.

timmysilv:
re: the above conversation, I've changed QubitDevice in PennyLane to provide a hook called `_get_diagonalizing_gates`, and I am now overriding it in this PR to filter out any Sum observables. This means that the above example which took 4 minutes computing diagonalizing gates and 15 seconds getting/setting parameters will now only take 15 seconds! Sure, the Hamiltonian equivalent still only takes 0.05 seconds, but we're getting there.

pennylane_lightning/_serialize.py
```diff
@@ -121,13 +132,50 @@ def _serialize_hamiltonian(ob, wires_map: dict, use_csingle: bool):
...
+
 def _serialize_ob(ob, wires_map, use_csingle):
     if isinstance(ob, Tensor):
         return _serialize_tensor_ob(ob, wires_map, use_csingle)
     elif ob.name == "Hamiltonian":
         return _serialize_hamiltonian(ob, wires_map, use_csingle)
+    elif _obs_has_kernel(ob):
+        return _serialize_named_obs(ob, wires_map, use_csingle)
+    elif ob._pauli_rep is not None:
+        return _serialize_pauli_sentence(ob._pauli_rep, wires_map, use_csingle)
```
albi3ro, line 176: I think we need to switch the order of these two.

AmintorDusko: Hi @timmysilv, I transferred the logic you are introducing here to [my PR](https://github.com/PennyLaneAI/pennylane-lightning/pull/425).
What I observed, and discussed with @albi3ro, was that the operator S_Prod was reaching the method `_serialize_named_obs`, changing the order really helped with that.
It is possible that I made a mistake in the process, so let me know your thoughts on that.

timmysilv: I switched the order locally to demonstrate what I'm trying to avoid:
```pycon
>>> import pennylane as qml
>>> from pennylane_lightning._serialize import _serialize_ob
>>> _serialize_ob(qml.PauliZ(0), wires_map={0: 0}, use_csingle=True)
Hamiltonian: { 'coeffs' : [1], 'observables' : [PauliZ[0]]}
```

I am surprised that an SProd returned true for `obs_has_kernel`, but it is kinda the thing I was scared of. I checked locally, and it returns `True` when you import `is_pauli_word` from `pennylane.pauli`, but lightning is currently importing the deprecated one from `pennylane.grouping`. Anyway, this clearly needs some fixing because both approaches cause problems. I'll separate things out and make it better

AmintorDusko: I came here to say what you said in your edit :sweat_smile:.
But yes, Lightning is still importing `is_pauli_word` from `pennylane.grouping`, but in my PR I already update it to the `pennylane.pauli`. 
Thank you for your help with that. 
As soon as you have a more definite solution I will update it in my PR.

timmysilv:
this is ready for review again! my latest benchmarks show the old and new operators taking the same amount of time

.github/CHANGELOG.md
```diff
@@ -75,7 +79,7 @@ array([-0.87758256, -0.47942554])
 
 This release contains contributions from (in alphabetical order):
 
-Chae-Yeun Park, Christina Lee, Lee James O'Riordan, Vincent Michaud-Rioux
+Chae-Yeun Park, Christina Lee, Lee James O'Riordan, Vincent Michaud-Rioux, Matthew Silverman
```
AmintorDusko, line 82: The last names should be in alphabetical order.
```suggestion
Christina Lee, Vincent Michaud-Rioux, Lee James O'Riordan, Chae-Yeun Park, Matthew Silverman
```

AmintorDusko (approved):
Nothing more to add from my side.
Thank you for that!

pennylane_lightning/_serialize.py
```diff
@@ -54,45 +55,35 @@
 except ImportError:
     pass
 
+pauli_name_map = {
+    "I": "Identity",
+    "X": "PauliX",
+    "Y": "PauliY",
+    "Z": "PauliZ",
+}
```
mlxd, line 63: Is there an equivalent map to this defined elsewhere in PL that we can just use here? If not, I think it may be worth adding as this will need to be replicated for LightningGPU and LightningKokkos too.

pennylane_lightning/_serialize.py
mlxd: Does this solve the identity situation as mentioned earlier? Does stripping the wires remove Identities as @chaeyeunpark mentioned?

pennylane_lightning/_serialize.py
```diff
@@ -121,13 +112,50 @@ def _serialize_hamiltonian(ob, wires_map: dict, use_csingle: bool):
     return hamiltonian_obs(coeffs, terms)
 
 
+def _serialize_pauli_word(ob, wires_map: dict, use_csingle: bool):
+    """Serialize a PauliWord into a Named or Tensor."""
```
mlxd, line 116:
```suggestion
    """Serialize a :class:`pennylane.pauli.PauliWord` into a Named or Tensor observable."""
```

pennylane_lightning/_serialize.py
```diff
@@ -121,13 +112,50 @@ def _serialize_hamiltonian(ob, wires_map: dict, use_csingle: bool):
...
+        wire, pauli = list(ob.items())[0]
+        return named_obs(pauli_name_map[pauli], [wires_map[wire]])
+
+    return tensor_obs(
+        [named_obs(pauli_name_map[pauli], [wires_map[wire]]) for wire, pauli in ob.items()]
+    )
+
+
+def _serialize_pauli_sentence(ob, wires_map: dict, use_csingle: bool):
+    """Serialize a PauliSentence into a Hamiltonian."""
```
mlxd, line 134:
```suggestion
    """Serialize a :class:`pennylane.pauli.PauliSentence` into a Hamiltonian."""
```

pennylane_lightning/_serialize.py
```diff
@@ -121,13 +112,50 @@ def _serialize_hamiltonian(ob, wires_map: dict, use_csingle: bool):
...
+    if use_csingle:
+        rtype = np.float32
+        hamiltonian_obs = HamiltonianC64
+    else:
+        rtype = np.float64
+        hamiltonian_obs = HamiltonianC128
+
+    pwords, coeffs = zip(*ob.items())
+    terms = [_serialize_pauli_word(pw, wires_map, use_csingle) for pw in pwords]
+    coeffs = np.array(coeffs).astype(rtype)
```
mlxd, line 144: Do we need to cast this? Shouldn't the data already be in the appropriate precision?

pennylane_lightning/_serialize.py
```diff
@@ -121,13 +112,50 @@ def _serialize_hamiltonian(ob, wires_map: dict, use_csingle: bool):
...
+
+
 def _serialize_ob(ob, wires_map, use_csingle):
     if isinstance(ob, Tensor):
         return _serialize_tensor_ob(ob, wires_map, use_csingle)
     elif ob.name == "Hamiltonian":
         return _serialize_hamiltonian(ob, wires_map, use_csingle)
+    elif isinstance(ob, (PauliX, PauliY, PauliZ, Identity, Hadamard)):
+        return _serialize_named_obs(ob, wires_map, use_csingle)
+    elif ob._pauli_rep is not None:
```
mlxd, line 155: We shouldn't depend on private attributes from external types. Is there a better way to handle this?

pennylane_lightning/lightning_qubit.py
```diff
@@ -964,6 +964,16 @@ def var(self, observable, shot_range=None, bin_size=None):
 
         return M.var(observable.name, observable_wires)
 
+    def _get_diagonalizing_gates(self, circuit: qml.tape.QuantumTape) -> List[Operation]:
+        meas_filtered = [
+            m
+            for m in circuit.measurements
+            if m.obs is None
+            or not isinstance(m.obs, qml.Hamiltonian)
+            and (not isinstance(m.obs, qml.ops.Sum) or m.obs._pauli_rep is None)
```
mlxd, line 973: Can this logical statement be tidied up a bit? I think there are could be a better way to express the same intent.
Also, worth noting the order here, and relying on the most likely to fail candidate up-front to avoid evaluating the other terms, and replacing up-front `or` with `and` if it makes sense to do so. Maybe applying de-Morgan's rules here can help reshape it as such.

Also, this still depends on that private attribute. Same as above, if this can be handled another way we should.

mlxd (changes requested):
Thanks @timmysilv 
I have some questions and some suggested updates.

Also, since this work will likely need to be ported to L-GPU and L-Kokkos also, it may be worth planning that out too so that the packages stay in-line.

timmysilv (commented):
i'll prepare the changes for kokkos and gpu

pennylane_lightning/_serialize.py
mlxd: I always had the assumption that params were the defining precision for the resulting simulation type. Feel free to keep this now, and I can remove it later when doing the FP32/FP64 normalization across the ecosystem.

pennylane_lightning/lightning_qubit.py
mlxd: Yea, I think having a lambda, or even a separate function could be a good call here.

tests/test_serialize.py
```diff
@@ -43,52 +43,36 @@
...
+            qml.PauliZ(0) @ qml.Hermitian(np.eye(2), wires=1) @ qml.Projector([0], wires=2),
+            TensorProdObsC64,
+        ),
+        (qml.Projector([0], wires=0), HermitianObsC64),
+        (qml.Hamiltonian([1], [qml.PauliZ(0)]), HamiltonianC64),
+    ],
+)
+def test_obs_returns_expected_type(obs, obs_type):
+    """Tests that observables get serialized to the expected type."""
+    assert isinstance(_serialize_ob(obs, dict(enumerate(obs.wires)), True), obs_type)
```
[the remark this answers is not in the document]
mlxd: That should be fine. Thanks for the clarification.

pennylane_lightning/_serialize.py
mlxd: Thanks @timmysilv and fine by me. You ok with this @chaeyeunpark ?

chaeyeunpark: Yeap, it looks good.

pennylane_lightning/_serialize.py
albi3ro: https://app.shortcut.com/xanaduai/story/37494/the-pauli-representation-can-be-computed-and-cached-efficiently-in-a-public-manner

pennylane_lightning/_serialize.py
mlxd: Is this resolved?

mlxd (approved):
Thanks for all the work @timmysilv 
Happy to approve.

pennylane_lightning/_serialize.py
timmysilv: I believe so. I've summarized the relevant changes [in this comment](https://github.com/PennyLaneAI/pennylane-lightning/pull/424#discussion_r1171848775), and added test cases to ensure things work as expected

PR #428, opened 2023-04-06
Update MacOS builder to favour native compiler on macos-12 runner

mlxd:
### Before submitting

**Context:** Due to an update in strictness requirements checks on brew install clang (now clang-16), Lightning fails to build due to static check failures and issues with use of allocators with the STL datatypes. As we aim to support libomp on MacOS (at least for x86_64) we aim to continue supporting builds that include this. However, as the brew clang version can be downgraded from clang-16 to clang-15 (`brew install llvm@15`) the libomp version is always at the latest released version, and fails to build against the clang-15 version. In this instance, the goal is to update the MacOS runner from macos-11 to macos-12, allowing us to access the native clang-14 compiler, which has the minimum required supports for the C++20 library features. With this version, we can support libomp 16, and build the wheels again.

* Update the x86_64 MacOS runner to macos-12 from macos-11
* Place the brew-installed clang compiler support and version behind new env vars (`USE_BREW_LLVM` and `BREW_LLVM_VERSION`)

**Benefits:** This will fix the current issues with clang-16 builds failing.

**Possible Drawbacks:** To ensure we can fall-back to brew installed clang later, keeping the environment variable control may introduce some clutter until we can fully commit to one direction or another.

**Related GitHub Issues:** Fixes current failures encountered in #427

### DIFF
--- a/.github/workflows/wheel_macos_arm64.yml
+++ b/.github/workflows/wheel_macos_arm64.yml
@@ -102,8 +102,6 @@ jobs:
 
           CIBW_ARCHS_MACOS: ${{ matrix.arch }}
 
-          USE_OMP: 1
-
         run: |
           python -m cibuildwheel --output-dir wheelhouse
 
--- a/.github/workflows/wheel_macos_arm64.yml
+++ b/.github/workflows/wheel_macos_arm64.yml
@@ -88,7 +88,7 @@ jobs:
 
           # Python build settings
           CIBW_BEFORE_BUILD: |
-            pip install pybind11 ninja cmake~=3.24.0
+            pip install pybind11 ninja cmake~=3.24.0 setuptools
 
           # Testing of built wheels
           CIBW_TEST_REQUIRES: pytest pytest-cov pytest-mock flaky
@@ -101,6 +101,8 @@ jobs:
           CIBW_BUILD_VERBOSITY: 1
 
           CIBW_ARCHS_MACOS: ${{ matrix.arch }}
+
+          USE_OMP: 1
 
         run: |
           python -m cibuildwheel --output-dir wheelhouse
--- a/.github/workflows/wheel_macos_x86_64.yml
+++ b/.github/workflows/wheel_macos_x86_64.yml
@@ -82,7 +82,7 @@ jobs:
                           -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
                           -DCMAKE_CXX_STANDARD=20 \
                           -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
-                          -DCMAKE_CXX_COMPILER=$(brew --prefix llvm)/bin/clang++ \
+                          -DCMAKE_CXX_COMPILER=$(brew --prefix llvm@15)/bin/clang \
                           -G Ninja
           cmake --build ./Build --verbose
           cmake --install ./Build
@@ -100,7 +100,7 @@ jobs:
                           -DCMAKE_CXX_STANDARD=20 \
                           -DCMAKE_PREFIX_PATH=${{ github.workspace}}/Kokkos_install/${{ matrix.exec_model }} \
                           -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
-                          -DCMAKE_CXX_COMPILER=$(brew --prefix llvm)/bin/clang++ \
+                          -DCMAKE_CXX_COMPILER=$(brew --prefix llvm@15)/bin/clang \ 
                           -G Ninja
           cmake --build ./Build --verbose
           cmake --install ./Build
@@ -156,7 +156,7 @@ jobs:
           CIBW_BEFORE_ALL_MACOS: |
             brew uninstall --force oclint
             brew install libomp
-            brew install llvm
+            brew install llvm@15
 
           # Python build settings
           CIBW_BEFORE_BUILD: |
--- a/.github/workflows/wheel_macos_x86_64.yml
+++ b/.github/workflows/wheel_macos_x86_64.yml
@@ -50,7 +50,6 @@ jobs:
 
       - name: Install clang
         run: |
-          brew install llvm@15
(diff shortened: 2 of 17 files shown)


.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -82,7 +83,7 @@ jobs:
                           -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
                           -DCMAKE_CXX_STANDARD=20 \
                           -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
-                          -DCMAKE_CXX_COMPILER=$(brew --prefix llvm)/bin/clang++ \
+                          -DCMAKE_CXX_COMPILER=g++ \
```
vincentmr, line 86: So we build with `g++` now?

.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -49,7 +49,8 @@ jobs:
           key: ${{ matrix.os }}-kokkos${{ matrix.kokkos_version }}-${{ matrix.exec_model }}
 
       - name: Install clang
-        run: brew install llvm
+        run: |
+          brew install libomp
```
vincentmr, line 53: Unlike `llvm`, there is no need to install g++, right?

setup.py
```diff
@@ -91,19 +91,46 @@ def build_extension(self, ext: CMakeExtension):
                     "-DCMAKE_SYSTEM_PROCESSOR=ARM64",
                 ]
             else:  # X64 arch
-                if shutil.which("brew"):
+                if os.getenv("USE_BREW_LLVM") and shutil.which("brew"):
+                    brew_llvm_version = os.getenv("BREW_LLVM_VERSION")
```
vincentmr, line 95: If `BREW_LLVM_VERSION` is not a common variable, maybe we could simply check whether `BREW_LLVM_VERSION` is set and let go of `USE_BREW_LLVM`.

vincentmr (approved):
Thanks for fixing this. This will be useful moving forward in [PR#48](https://github.com/PennyLaneAI/pennylane-lightning-kokkos/pull/48).

.github/workflows/wheel_macos_arm64.yml
```diff
@@ -88,7 +88,7 @@ jobs:
 
           # Python build settings
           CIBW_BEFORE_BUILD: |
-            pip install pybind11 ninja cmake~=3.24.0
+            pip install pybind11 ninja cmake~=3.24.0 setuptools
```
AmintorDusko, line 91: Do we need to install setuptools on ARM64 now?

.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -156,16 +157,15 @@ jobs:
           CIBW_BEFORE_ALL_MACOS: |
             brew uninstall --force oclint
             brew install libomp
-            brew install llvm
 
           # Python build settings
           CIBW_BEFORE_BUILD: |
-            pip install pybind11 ninja cmake~=3.24.0
+            python -m pip install pybind11 ninja cmake~=3.24.0 setuptools
```
AmintorDusko, line 163: I'm aware that the error log was pointing to some problem with setuptools.
But do we still need to install it after changing the compiler and os?

AmintorDusko (commented):
Hi @mlxd, good job here. 
I left some questions, but from my side, this is in a good shape to be approved.

.github/workflows/wheel_macos_x86_64.yml
mlxd: `g++` on MacOS maps to the internal clang compiler. I set it here to avoid `clang` being captured by the on-path brew installed version.

.github/workflows/wheel_macos_x86_64.yml
mlxd: Nope, as above, maps to the MacOS clang compiler internally.

setup.py
mlxd: Yep, that's a good idea, and much cleaner.

.github/workflows/wheel_macos_x86_64.yml
mlxd: I believe it is worth including. It should be autoinstalled, but this seemed to not be the case. I can try removing it and validating again.

setup.py
vincentmr: Then I think we just need to replace `USE_BREW_LLVM` by `BREW_LLVM_VERSION`.
```suggestion
                if os.getenv("BREW_LLVM_VERSION") and shutil.which("brew"):
                    brew_llvm_version = os.getenv("BREW_LLVM_VERSION")
```

setup.py
```diff
@@ -91,19 +91,46 @@ def build_extension(self, ext: CMakeExtension):
...
+                            "--prefix",
+                            "libomp",
+                        ],
+                        check=False,
+                        capture_output=True,
+                        text=True,
+                    )
+                    .stdout
+                    .strip()
+                )
```
vincentmr, line 127: This will cause issues on conda-forge, because `brew` is not installed.
I suggest conditioning `ENABLE_OPENMP` on `USE_OMP`. Then change 
```suggestion
                libomp_path = (
                    (
                        subprocess.run(
                            [
                                "brew",
                                "--prefix",
                                "libomp",
                            ],
                            check=False,
                            capture_output=True,
                            text=True,
                        ).stdout.strip()
                    )
                    if shutil.which("brew")
                    else None
                )
```
Finally, if `libomp_path` is not `None`, let
```
configure_args += [f"-DOpenMP_ROOT={libomp_path}/"]
```

mlxd: I moved this around a little more to more aptly keep things contained. While I am not a fan of the current highly nested if-else structure, it will unblock the other builds for now.

AmintorDusko (approved):
Nice job!

PR #448, opened 2023-05-04
Update tests for PL 0.31 to remove deprecated functions

mlxd:
**Context:** PennyLane v0.30.0-dev has removed support for previously deprecated functions, as defined in https://docs.pennylane.ai/en/stable/development/deprecations.html. This PR updates those functions to the new supported syntax.

**Description of the Change:** As above. In addition, due to an ongoing issue with RTD builds using unaligned versions of setuptools and pip, we create a subtle change to the RTD builder to fix the package versions until https://github.com/readthedocs/readthedocs.org/pull/10287 becomes usable. In addition, to avoid CI issues, CYBot is now GHBot and will auto-update the package versions in the CI.

**Benefits:** Fixes currently failing tests.

**Possible Drawbacks:**

**Related GitHub Issues:**


.readthedocs.yml
```diff
@@ -21,3 +21,4 @@ build:
     - libopenblas-base
     - libopenblas-dev
     - graphviz
+    - libffi-dev
```
rashidnhm, line 24:
```suggestion
  jobs:
    pre_install:
      - pip install --force-reinstall --no-cache-dir pip<={some version}
```

vincentmr (approved):
Looks good to me. Thanks @mlxd

AmintorDusko:
Nice work, short and sweet! I'm just waiting for the CIs to conclude successfully and will approve!

AmintorDusko (approved):

PR #450, opened 2023-05-18
Refactor setup.py to align with CF.

vincentmr:
### Before submitting

**Context:**
The way `setup.py` is written, it tries to find binaries in the environment (for MacOS in particular) and it may override some CMake decisions which would actually make sense in some build environments (e.g. on Conda-Forge). This is trying to apply [this patch](https://github.com/conda-forge/pennylane-feedstock/blob/main/recipe/cross_compilation.diff) and a bit more. 

Remove (most of ) CMake argument parsing from `setup.py`. Adapt CI files and use `env:` when possible to set environment variables. 

**Benefits:**
Where necessary, libraries and default build tools are defined declaratively in the Actions files. `setup.py` is simplified and any CMake argument can be passed in `CMAKE_ARGS` if required. This complies better with Conda-Forge, and hence we'll be able to remove some patches there.

**Possible Drawbacks:**
Building from source may change a bit for users used to do so.

**Related GitHub Issues:**
[PR 58](https://github.com/PennyLaneAI/pennylane-lightning-kokkos/pull/58)


.github/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 
 ### Improvements
 
+* Remove logic from `setup.py` and transfer paths and env variable definitions into workflow files.
+[(#450)](https://github.com/PennyLaneAI/pennylane-lightning/pull/450)
```
AmintorDusko, line 16:
```suggestion
* Remove logic from `setup.py` and transfer paths and env variable definitions into workflow files.
  [(#450)](https://github.com/PennyLaneAI/pennylane-lightning/pull/450)
```

.github/workflows/tests_linux.yml
```diff
@@ -227,7 +227,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext -i --define="ENABLE_BLAS=ON;ENABLE_KOKKOS=OFF;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
+          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);ENABLE_BLAS=ON;ENABLE_KOKKOS=OFF;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```
AmintorDusko, line 230: Python is enabled by default.
```suggestion
          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);ENABLE_BLAS=ON;ENABLE_KOKKOS=OFF;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```

.github/workflows/tests_linux.yml
```diff
@@ -382,7 +382,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext -i --define="CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;ENABLE_KOKKOS=ON;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
+          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;ENABLE_KOKKOS=ON;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```
AmintorDusko, line 385: Kokkos and Python are enabled by default.
```suggestion
          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```

.github/workflows/tests_linux.yml
```diff
@@ -531,7 +531,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext -i --define="ENABLE_BLAS=ON;ENABLE_KOKKOS=ON;DENABLE_PYTHON=ON;CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
+          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);ENABLE_BLAS=ON;ENABLE_KOKKOS=ON;DENABLE_PYTHON=ON;CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```
AmintorDusko, line 534: Kokkos and Python are enabled by default:
```suggestion
          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);ENABLE_BLAS=ON;CMAKE_PREFIX_PATH=${{ github.workspace }}/Kokkos;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```

setup.py
```diff
@@ -39,26 +38,29 @@ class CMakeBuild(build_ext):
...
+        if self.verbosity != "":
+            self.verbosity = "--verbose"
 
         super().finalize_options()
 
     def build_extension(self, ext: CMakeExtension):
         extdir = str(Path(self.get_ext_fullpath(ext.name)).parent.absolute())
-
         debug = int(os.environ.get("DEBUG", 0)) if self.debug is None else self.debug
+        cfg = "Debug" if debug else "RelWithDebInfo"
```
AmintorDusko, line 55: Could you please change this variable name to a more meaningful one?

AmintorDusko (commented):
Some suggestions.

.github/workflows/tests_linux.yml
```diff
@@ -109,7 +109,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext -i --define="ENABLE_KOKKOS=OFF;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
+          python setup.py build_ext -i --define="PYTHON_EXECUTABLE=$(which python);ENABLE_KOKKOS=OFF;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
```
mlxd, line 112: Do we need the `PYTHON_EXECUTABLE` here and below? I'd assume since the calling python will be on path, it should find it first.

.github/workflows/wheel_win_x86_64.yml
```diff
@@ -149,6 +149,9 @@ jobs:
           CIBW_BEFORE_BUILD: |
             pip install pybind11 cmake~=3.24.0 build
 
+          CIBW_ENVIRONMENT: |
+            CMAKE_ARGS="-DPython_EXECUTABLE=$(which python)"
```
mlxd, line 153: Since this is Windows, we shouldn't use Linux specific syntax in commands if we can opt for pwsh versions instead.

.github/workflows/tests_linux.yml
vincentmr: I'm not sure about this, but it used to be forced in `setup.py`. I can remove it and see what happens.

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -161,6 +161,9 @@ jobs:
             cat /etc/yum.conf | sed "s/\[main\]/\[main\]\ntimeout=5/g" > /etc/yum.conf
             pip install ninja cmake~=3.24.0
 
+          CIBW_ENVIRONMENT: |
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python)"
```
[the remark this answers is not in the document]
vincentmr: I already documented it in L-Kokkos. I'll add it here, but it is not as crucial since `--define` still works.

.github/workflows/wheel_linux_x86_64.yml
```diff
@@ -184,6 +185,8 @@ jobs:
 
           CIBW_BUILD_VERBOSITY: 3
 
+          OMP_PROC_BIND: false
```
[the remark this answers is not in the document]
vincentmr: This is just to avoid a warning that `OMP_PROC_BIND` is better set to `false` for unit tests.

.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -162,6 +162,9 @@ jobs:
           CIBW_BEFORE_BUILD: |
             python -m pip install pybind11 ninja cmake~=3.24.0 setuptools
 
+          CIBW_ENVIRONMENT: |
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python) -DCMAKE_CXX_COMPILER=$(which clang++) -DCMAKE_LINKER=$(dirname $(which clang++))/lld -DOpenMP_ROOT=$(brew --prefix libomp)"
```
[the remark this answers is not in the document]
vincentmr: I had a bit of trouble with this one. I transferred what was in the `setup.py`. It was not taking the `brew` branch, as far as I could tell, but just falling back to the `which clang++` branch that we originally introduced for CF. So I kept the workflow as is and I think it's using MacOS clang.

.github/workflows/wheel_win_x86_64.yml
vincentmr: Will find replacement.

.github/workflows/wheel_linux_x86_64.yml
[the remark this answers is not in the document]
mlxd: Isn't that warning exclusive to the Kokkos OMP backend? Since we only build serial Kokkos here, it shouldn't be an issue, no?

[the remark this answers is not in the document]
vincentmr: Right, I'm testing locally with my default Kokkos which has OMP and I thought that was a generic warning. I think it doesn't hurt, but I can remove it until we distribute OMP-enabled Kokkos wheels. What do you prefer?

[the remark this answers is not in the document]
mlxd: I think we can remove it from here, since it is not appearing in our build/test process for Lightning. We can revisit later if we distribute OMP based Kokkos builds.

.github/workflows/wheel_win_x86_64.yml
vincentmr: In my powershell the following works
```
Set-Variable -Name CMAKE_ARGS -Value "-DPython_EXECUTABLE=$(@(where.exe <file pattern> 2>$null)[0])"
```
but it fails here because `where` [does not exists](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/5059095262/jobs/9080155441?pr=450#step:7:62). By Linux specific syntax, were you thinking `VAR=VAL` or `which`?

.github/workflows/tests_linux.yml
vincentmr: You were right, not required.

.github/workflows/wheel_linux_x86_64.yml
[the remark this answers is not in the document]
vincentmr: Removed.

.github/workflows/wheel_win_x86_64.yml
mlxd: What about something like capturing in an earlier step and reusing with
```pwsh
$py_exe = Get-Command python | Select-Object -ExpandProperty Definition
```

Or just embedding that line into the string builder:
```pwsh
CMAKE_ARGS="-DPython_EXECUTABLE=$(Get-Command python | Select-Object -ExpandProperty Definition)"
```

Though, if the executable isn't explicitly needed, we can probably drop it.

vincentmr: If I just remove that, as on Linux and MacOS, the build passes, but I don't think we target the same Python. If we specify it, we have [paths to Python 3.7](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/5059416984/jobs/9080925170#step:7:669) and otherwise we get [paths to Python 3.11](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/5060342821/jobs/9083098290?pr=450#step:7:668). If we look at another branch, say [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/5058922324/jobs/9079823976?pr=452#step:7:664), we have something similar to the later, but we have a path in `C:\Users\runneradmin\AppData` instead of `C:/hostedtoolcache/windows`.  Is that ok? Just not familiar with these Windows envs.

mlxd: We definitely don't want 3.7 for the 3.11 build.

So, if we specify `CMAKE_ARGS="-DPython_EXECUTABLE=$(Get-Command python | Select-Object -ExpandProperty Definition)"` we get 3.11, and if we do not, we get 3.7, or did I misunderstand? I'd be less concerned about the paths, as long as 3.11 is used.

vincentmr: No sorry. So 
```
CMAKE_ARGS="-DPython_EXECUTABLE=$(Get-Command python | Select-Object -ExpandProperty Definition)"
```
[just failed](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/5060148817/jobs/9082661374#step:7:146). But removing `which` was a good call, because even though the build worked out, CMake reports things like 
```
Found PythonLibs: C:/hostedtoolcache/windows/Python/3.7.9/x64/libs/python37.lib
```
Simply removing `-DPythonEXECUTABLE` altogether, CMake resolves 
```
Found Python: C:/hostedtoolcache/windows/Python/3.11.3/x64/python3.exe
```
and this work too.

.github/workflows/wheel_macos_arm64.yml
```diff
@@ -89,6 +89,9 @@ jobs:
           CIBW_BEFORE_BUILD: |
             pip install pybind11 ninja cmake~=3.24.0 setuptools
 
+          CIBW_ENVIRONMENT: |
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python) -DCMAKE_CXX_COMPILER_TARGET=arm64-apple-macos11 -DCMAKE_SYSTEM_NAME=Darwin -DCMAKE_SYSTEM_PROCESSOR=ARM64 -DENABLE_OPENMP=OFF"
```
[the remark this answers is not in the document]
vincentmr: @multiphaseCFD Could you please give it a shot?

.github/workflows/wheel_win_x86_64.yml
mlxd: Ah, got it. In this case we can simply ignore the statement and keep the 3.11 builder, since that is appropriately picked up.

.github/workflows/tests_linux.yml
```diff
@@ -227,8 +226,7 @@ jobs:
       - name: Install lightning.qubit device
         run: |
           cd main
-          python setup.py build_ext -i --define="ENABLE_BLAS=ON;ENABLE_KOKKOS=OFF;ENABLE_PYTHON=ON;CMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)"
-          pip install -e .
+          CMAKE_ARGS="-DENABLE_BLAS=ON -DENABLE_KOKKOS=OFF -DCMAKE_CXX_COMPILER=$(which g++-$GCC_VERSION)" pip install -e . -vv
```
mlxd, line 229: Just to clarify, if we don't set the CMAKE_ARGS, the default options chosen will still be the same as we have currently?

As in, a user can still compile everything as it was before by running `python -m pip install -e .` ?

mlxd (commented):
Another final q

.github/workflows/tests_linux.yml
vincentmr: Good point clarifying. So a user will likely compile everything as before, however some default options changed. `setup.py` contains some platform dependent logic that was (at least partly) introduced to fix some CI issues. Some default options were forced which caused some issues in certain environments (e.g. Conda-Forge). For example, `Python_EXECUTABLE` and `PYTHON_EXECUTABLE` were always specified (but didn't need to apparently). In this PR, we try to keep `setup.py`  to a minimum, moving defaults either to CMake lists or specifying more environment dependent options in the Github actions files. This way, if say CF or spack have `CXX` defined in some env, we let CMake discover and deal with it. If we want to build with Brew's Clang on MacOS, we directly pass the option in `CMAKE_ARGS` so we don't have to check if `brew` or `clang++` exists in `setup.py` and specify some fallback.

mlxd: Thanks Vincent. I'd rather the pip install maintains the same defaults as before if not specified. Ideally we should be able to reproduce the released wheel build options without providing CMAKE_ARGS up front. If. that isn't the case, I think we should revert the changes to this.

vincentmr: `CMAKE_ARGS` is not required here, the changes could be reverted to be passed by`--define` as well, it would just take two lines instead of one. 
As far as wheels are concerned, the only place defaults might have changed is in the MacOS wheel builders below. For x86 for example, they are 
```
CMAKE_ARGS="-DCMAKE_CXX_COMPILER=$(which clang++) -DCMAKE_LINKER=$(dirname $(which clang++))/lld -DOpenMP_ROOT=$(brew --prefix libomp)"
```
Since these options depend on the build environment and how we want to build the wheel, I thought they would be better put explicitly here instead of `setup.py` where they may override good default options in other contexts and on other systems (e.g. those without `brew` or `clang++`). But if you don't agree, I can put those back.

vincentmr: I'm actually going to check whether these are required anyways.

vincentmr: After discussion, I'll keep some default, I'll move others. Let's close it for now.

mlxd (approved):
Thanks @vincentmr

AmintorDusko (approved):
Looks good to me!

PR #451, opened 2023-05-23
Only upload to TestPyPI on release

timmysilv:
**Context:**
Suggested by Lee in PennyLaneAI/pennylane#4147, to ensure a built version of lightning is more available on TestPyPI.

Only upload to TestPyPI on release. Before, we would also upload on merge to master.

**Benefits:**
This is to help ensure PL CI runs fail less often in ways addressed by the above linked PR (ie. try not to run PL CI with the fallback default.qubit device)

**Possible Drawbacks:**
Recent intermediate versions of Lightning will be less available.

**Related GitHub Issues:**
PennyLaneAI/pennylane#4147


mlxd (approved):
Quick 'n' easy --- thanks @timmysilv

.github/workflows/wheel_noarch.yml
```diff
@@ -56,7 +56,7 @@ jobs:
   upload-pypi:
     needs: build-pure-python-wheel
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/master'}}
+    if: ${{ github.event_name == 'release' }}
```
vincentmr, line 59: To be sure, so here we do not want to upload to PyPI unless we're doing a release, right?

AmintorDusko (approved):
Thank you for this fix.

.github/workflows/wheel_noarch.yml
mlxd: This is for TestPyPI, and yep, that is correct.
We always manually upload wheel to PyPI to avoid any fallout of mistakes.

vincentmr (approved):
Looks good to me.

vincentmr (approved):
Looks good to me.

timmysilv:
this [happened again](https://github.com/PennyLaneAI/pennylane/actions/runs/5335215567/jobs/9668124437?pr=4269) on a PR against the current RC branch. I guess that's expected because it's not master? Just posting here for posterity since we can re-run if needed in those cases, but I had to remind myself what was happening when I saw it

mlxd:
> RC

It looks like it installed the correct lightning version though, so not sure why those failures appeared. The no-binary wheel was not pulled as part of the CI install. 
image

Not sure in this case what happened.

timmysilv:
[cognitive dissonance](https://github.com/PennyLaneAI/pennylane/actions/runs/5335215567/jobs/9668124437?pr=4269#step:12:189)

PR #465, opened 2023-06-21
Add SPVM routine for when kokkos not available.

vincentmr:
### Before submitting

**Context:**
Kokkos & Kokkos-kernels (serial backends) are used only for a single sparse matrix/dense vector operation. This is a  

Add simple spvm implementation when `ENABLE_KOKKOS=OFF`.

**Benefits:**
No need to pull large dependencies Kokkos & Kokkos-kernels. SparseHamiltonian always supported.

**Possible Drawbacks:**
It could be slower, although the (serial) code is so simple that there is little room for optimization and that is unlikely. 

**Related GitHub Issues:**


pennylane_lightning/_version.py
```diff
@@ -17,4 +17,4 @@
 """
 
 
-__version__ = "0.31.0-dev10"
+__version__ = "0.31.0-dev11"
```
mlxd, line 20: Since this won't be hitting the 30.31 release, we can keep this in the 0.32 cycle

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -185,16 +185,21 @@ namespace Pennylane::Util {
...
-    [[maybe_unused]] const std::complex<fp_precision> *values_ptr,
+    const index_type *entries_ptr, const std::complex<fp_precision> *values_ptr,
     [[maybe_unused]] const index_type numNNZ,
-    [[maybe_unused]] std::vector<std::complex<fp_precision>> &result) {
-    PL_ABORT("Executing the product of a Sparse matrix and a vector needs "
-             "Kokkos and Kokkos Kernels installation.");
+    std::vector<std::complex<fp_precision>> &result) {
+    result.resize(vector_size);
+    size_t count = 0;
+    for (index_type i = 0; i < vector_size; ++i) {
```
mlxd, line 196: We always use post-increment in Lightning, unless pre is required for functionality.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -185,16 +185,21 @@ namespace Pennylane::Util {
...
+    const index_type *row_map_ptr,
     [[maybe_unused]] const index_type row_map_size,
-    [[maybe_unused]] const index_type *entries_ptr,
-    [[maybe_unused]] const std::complex<fp_precision> *values_ptr,
+    const index_type *entries_ptr, const std::complex<fp_precision> *values_ptr,
     [[maybe_unused]] const index_type numNNZ,
-    [[maybe_unused]] std::vector<std::complex<fp_precision>> &result) {
-    PL_ABORT("Executing the product of a Sparse matrix and a vector needs "
-             "Kokkos and Kokkos Kernels installation.");
+    std::vector<std::complex<fp_precision>> &result) {
```
mlxd, line 193: The formatting here looks odd to me.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -185,16 +185,21 @@ namespace Pennylane::Util {
...
+    result.resize(vector_size);
+    size_t count = 0;
+    for (index_type i = 0; i < vector_size; ++i) {
+        result[i] = 0.0;
+        for (index_type j = 0; j < row_map_ptr[i + 1] - row_map_ptr[i]; ++j) {
+            result[i] += values_ptr[count] * vector_ptr[entries_ptr[count]];
+            ++count;
```
mlxd, line 200: Same here about post-increment

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -185,16 +185,21 @@ namespace Pennylane::Util {
...
-    [[maybe_unused]] std::vector<std::complex<fp_precision>> &result) {
-    PL_ABORT("Executing the product of a Sparse matrix and a vector needs "
-             "Kokkos and Kokkos Kernels installation.");
+    std::vector<std::complex<fp_precision>> &result) {
+    result.resize(vector_size);
+    size_t count = 0;
+    for (index_type i = 0; i < vector_size; ++i) {
+        result[i] = 0.0;
+        for (index_type j = 0; j < row_map_ptr[i + 1] - row_map_ptr[i]; ++j) {
+            result[i] += values_ptr[count] * vector_ptr[entries_ptr[count]];
```
mlxd, line 199: does the `j` index need to be used here somewhere?

tests/test_measures_sparse.py
```diff
@@ -35,41 +35,17 @@ class TestSparseExpval:
...
-            [qml.PauliY(0) @ qml.Identity(1), -0.38941834230865050],
-            [qml.Identity(0) @ qml.PauliY(1), 0.00000000000000000],
-            [qml.PauliZ(0) @ qml.Identity(1), 0.92106099400288520],
-            [qml.Identity(0) @ qml.PauliZ(1), 0.98006657784124170],
+            [qml.PauliX(0) @ qml.Identity(1), 0.00000000000000000, 1.000000000000000000],
+            [qml.Identity(0) @ qml.PauliX(1), -0.19866933079506122, 0.960530638694763184],
+            [qml.PauliY(0) @ qml.Identity(1), -0.38941834230865050, 0.848353326320648193],
+            [qml.Identity(0) @ qml.PauliY(1), 0.00000000000000000, 1.000000119209289551],
+            [qml.PauliZ(0) @ qml.Identity(1), 0.92106099400288520, 0.151646673679351807],
+            [qml.Identity(0) @ qml.PauliZ(1), 0.98006657784124170, 0.039469480514526367],
```
mlxd, line 46: Still wondering about the old values here. Maybe we should double check this analytically.

mlxd (commented):
Some quick comments for the next week.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
vincentmr: No, just a counter going through `row_map_ptr[i + 1] - row_map_ptr[i]`.

tests/test_measures_sparse.py
vincentmr: We could, but the numbers matched when directly inputting say `qml.PauliY(0) @ qml.Identity(1)` (no `Hamiltonian`, `SparseHamiltonian`). I can check it again though. To be sure, by "old values" you mean the second (expval) column, right?

.github/CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ This release contains contributions from (in alphabetical order):
 * Update Kokkos support to 4.0.01.
   [(#439)] (https://github.com/PennyLaneAI/pennylane-lightning/pull/439)
 
+* Lightning supports SparseHamiltonian natively (without Kokkos & Kokkos-kernels).
```
AmintorDusko, line 26: I consider that our support, in terms of KokkosKernels, for SparseHamiltonian is already native.
I propose something like this:
```suggestion
* Add native support to sparse Hamiltonians in the absence of Kokkos & Kokkos-kernels.
```

AmintorDusko (commented):
So far looks good to me. I will check in more depth later.

AmintorDusko:
Hi, @vincentmr. It appears that the CI may have become stuck, so you may want to try re-triggering it.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
mlxd: That's probably it

tests/test_measures_sparse.py
mlxd: Yep, that's right. Fine by me with it being defined anyway that seems sensible.

pennylane_lightning/src/util/Kokkos_Sparse.hpp
```diff
@@ -185,16 +185,21 @@ namespace Pennylane::Util {
  */
 template <class fp_precision, class index_type>
 void apply_Sparse_Matrix_Kokkos(
```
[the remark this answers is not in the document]
vincentmr: Yes, how about simply `apply_Sparse_Matrix`?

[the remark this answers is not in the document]
vincentmr: I've only kept `apply_Sparse_Matrix` and moved the module to `SparseLinAlg.hpp`.

pennylane_lightning/src/util/SparseLinAlg.hpp
```diff
@@ -160,16 +160,15 @@ void apply_Sparse_Matrix_Kokkos(
             });
         }
     }
+    return result;
```
mlxd, line 163: Is a return here better than passing in a pre-allocated buffer ?

vincentmr: We used to have 
```
std::vector<std::complex<fp_precision>> apply_Sparse_Matrix
```
which declared `result` and then called `apply_Sparse_Matrix_Kokkos` (which `resize`s for robustness). To maintain the interface and functionality I just changed the return type of both `void` functions to `std::vector<std::complex<fp_precision>>` and renamed to `apply_Sparse_Matrix`. If we used this function in places where there is a preallocated buffer of the correct size, it would be better to pass it in and bypass `resize`, but I think it is not the case now.

mlxd: Thanks. I think as long as we can ensure we have RVO/copy elision (https://en.cppreference.com/w/cpp/language/copy_elision) with this, it should be fine. In this case, I suspect this design may be better capable of reasoning that, though it may become a problem if the SV is large, no?

vincentmr: > In this case, I suspect this design may be better capable of reasoning that

I'm unsure about this. My intuition is most compilers would treat the previous and current designs the same and probably use RVO. 

> though it may become a problem if the SV is large, no?

You can't really do a general sparse matrix product in-place, so we (or Kokkos) need at least a temporary buffer the size of SV. I think the current design chooses not to keep this buffer around to save memory. I guess we can run many circuits without that buffer, but we could accelerate (likely not by much) some circuits such as those with sparse Hamiltonians.

mlxd (approved):
Thanks Vincent --- no more thoughts from me. Happy to approve.

pennylane_lightning/src/util/SparseLinAlg.hpp
```diff
@@ -181,52 +180,27 @@ namespace Pennylane::Util {
...
+apply_Sparse_Matrix(const std::complex<fp_precision> *vector_ptr,
+                    const index_type vector_size, const index_type *row_map_ptr,
+                    [[maybe_unused]] const index_type row_map_size,
+                    const index_type *entries_ptr,
+                    const std::complex<fp_precision> *values_ptr,
+                    [[maybe_unused]] const index_type numNNZ) {
+    std::vector<std::complex<fp_precision>> result;
+    result.resize(vector_size);
+    size_t count = 0;
+    for (index_type i = 0; i < vector_size; i++) {
```
multiphaseCFD, line 196: Is it possible to add OpenMP support here when ENABLE_OPENMP is set as ON?

vincentmr: Yes, I think the best way would be to split the `vector_size` loop into `n_threads` blocks. Since Kokkos is always built only with a SERIAL backend and I just wanted to reproduce this functionality, I didn't add OpenMP here.

multiphaseCFD (approved):
Nice job @vincentmr !

PR #477, opened 2023-08-21
Add CUDA-12 capability 

vincentmr:
# Add CUDA-12 capability 

### Before submitting

**Context:**
The Kokkos backend could not be built or executed on devices. A few minors fixes are necessary. Notably, `JacobianData` works with a data pointer which is not accessible if the data lives on a device.

Fix few bugs, and get the state vector data from the device to the host, into `JacobianData`. 

**Benefits:**
This code builds with CUDA-12 and executes on capable NVidia devices.
 
**Possible Drawbacks:**
This is a quick fix, but we should think about a more perennial solution. For example, we could store the state vector in `JacobianData` or pass the state vector to `adjointJacobian` as follows `adjointJacobian(std::span{jac}, jd, sv)`. `sv` could be `[[maybe_unused]]` in LQubit and used in device-executing backends.  

**Related GitHub Issues:**


requirements-dev.txt
```diff
@@ -6,4 +6,5 @@ pybind11
 pytest
 pytest-cov
 pytest-mock
-black==23.7.0
\ No newline at end of file
+black==23.7.0
+clang-format==14
```
AmintorDusko, line 10: Do we need that? Or maybe you forgot it here?

pennylane_lightning/core/src/bindings/Bindings.hpp
```diff
@@ -462,15 +462,20 @@ auto registerAdjointJacobian(
     using PrecisionT = typename StateVectorT::PrecisionT;
     std::vector<PrecisionT> jac(observables.size() * trainableParams.size(),
                                 PrecisionT{0.0});
+#ifdef _ENABLE_PLQUBIT
+    auto svdata = sv.getData();
+#elif _ENABLE_PLKOKKOS == 1
+    std::vector<Kokkos::complex<PrecisionT>> datavec(sv.getLength());
+    sv.DeviceToHost(datavec.data(), datavec.size());
+    auto svdata = datavec.data();
+#endif
```
AmintorDusko, line 471: We can probably avoid the use of compiler preprocessor directives, by using unique pointers.
A simpler version of what we have [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_qubit/algorithms/AdjointJacobianLQubit.hpp#L253-L263) should suffice.

requirements-dev.txt
vincentmr: I think it's good to have it since these are the versions used by the CI and pre-commit hooks.

pennylane_lightning/core/src/bindings/Bindings.hpp
vincentmr: I'm not sure that works because it requires branching depending on the state vector type, but the first condition may not be defined depending on the backend.

mlxd: Since the data type differs, using a u_ptr may not work here without some effort. We'd likely get this working with some combination of `if constexpr` but in the interest of time let's accept this as is.

AmintorDusko: OK.

mlxd (approved):
Happy to see this GPU fix go in.

.github/CHANGELOG.md
```diff
@@ -26,7 +26,10 @@
 
 * Update the CMake internal references to enable sub-project compilation with affecting the parent package.
   [(#478)](https://github.com/PennyLaneAI/pennylane-lightning/pull/478)
-
+  
+* Modify `registerAdjointJacobian` and LKokkos' `applyMatrix` method to support device execution (with CUDA-12)
+  [(#477)](https://github.com/PennyLaneAI/pennylane-lightning/pull/477)
```
AmintorDusko, line 31: I believe this needs to be updated to reflect the most recent changes.

AmintorDusko (approved):
Thank you for that!

AmintorDusko:
Hey @vincentmr, you may need to re-trigger CIs here.

PR #485, opened 2023-08-25
Add test coverage for LKokkos

vincentmr:
### Before submitting

**Context:**
The LKokkos backend has lower coverage than PennyLane and LQubit. Codecov oftentimes fails as LKokkos' coverage does not meet the target.

- Add tests, parametrization with `GENERATE` (gate `inverse` in particular to target the `inverse` branch of functors), etc. 
- Enums defined in `GateOperation.hpp` start at `1` (previously `0`). `::BEGIN` is introduced in a few places where it was assumed `0` accordingly. 
- Enable pre-commit hooks to format all Python files and linting of all Python source files.

**Benefits:**
Better test coverage.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,4 +1,22 @@
-# Release 0.32.0-dev
+# Release 0.33.0-dev
+
+### New features since last release
+
+### Breaking changes
+
+### Improvements
+
+### Documentation
+
+### Bug fixes
+
+### Contributors
+
+This release contains contributions from (in alphabetical order):
+
+---
+
+# Release 0.32.0
 
 ### New features since last release
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -15,6 +15,9 @@
 * Add tests to increase LKokkos coverage.
   [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
 
+* Add memory locality tag reporting and adjoint diff dispatch for `lightning.qubit` statevector classes.
+  [(#492)](https://github.com/PennyLaneAI/pennylane-lightning/pull/492)
+
 ### Documentation
 
 ### Bug fixes
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Improvements
 
+* Add tests to increase LKokkos coverage.
+  [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
+
 ### Documentation
 
 ### Bug fixes
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -25,6 +25,9 @@
 
 ### Bug fixes
 
+* Ensure PennyLane has an `active_return` attribute before calling it.
+ [(#483)] (https://github.com/PennyLaneAI/pennylane-lightning/pull/483)
+
 * Do no import `sqrt2_v` from `<numbers>` in `Util.hpp` to resolve issue with Lightning-GPU builds.
   [(#479)](https://github.com/PennyLaneAI/pennylane-lightning/pull/479)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Breaking changes
 
+* Enable pre-commit hooks to format all Python files and linting of all Python source files.
+  [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
+
 ### Improvements
 
 * Add tests to increase LKokkos coverage.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -16,9 +16,14 @@
 
 ### Bug fixes
 
+* Fix RTD builds by removing unsupported `sytem_packages` configuration option.
+  [(#491)](https://github.com/PennyLaneAI/pennylane-lightning/pull/491)
+
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
+
+Lee J. O'Riordan
 
 ---
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Breaking changes
 
+* Enums defined in `GateOperation.hpp` start at `1` (previously `0`). `::BEGIN` is introduced in a few places where it was assumed `0` accordingly.
+  [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
+
 * Enable pre-commit hooks to format all Python files and linting of all Python source files.
   [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 This release contains contributions from (in alphabetical order):
 
-Lee J. O'Riordan
+Vincent Michaud-Rioux, Lee J. O'Riordan
 
 ---
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -16,6 +16,9 @@
 
 ### Breaking changes
 
+* Rename `QubitStateVector` to `StatePrep` in the `LightningQubit` and `LightningKokkos` classes.
+  [(#486)](https://github.com/PennyLaneAI/pennylane-lightning/pull/486)
+
 * Modify `adjointJacobian` methods to accept a (maybe unused) reference `StateVectorT`, allowing device-backed simulators to directly access state vector data for adjoint differentiation instead of copying it back-and-forth into `JacobianData` (host memory).
   [(#477)](https://github.com/PennyLaneAI/pennylane-lightning/pull/477)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* The `lightning_kokkos` backend supports Nvidia GPU execution (with Kokkos v4 and CUDA v12). 
+  [(#477)](https://github.com/PennyLaneAI/pennylane-lightning/pull/477)
+
 * Complete overhaul of repository structure to facilitates integration of multiple backends. Refactoring efforts we directed to improve development performance, code reuse and decrease overall overhead to propagate changes through backends. New C++ modular build strategy allows for faster test builds restricted to a module. Update CI/CD actions concurrency strategy. Change minimal Python version to 3.9.
   [(#472)] (https://github.com/PennyLaneAI/pennylane-lightning/pull/472)
 
@@ -13,7 +16,13 @@
 
 ### Breaking changes
 
+* Modify `adjointJacobian` methods to accept a (maybe unused) reference `StateVectorT`, allowing device-backed simulators to directly access state vector data for adjoint differentiation instead of copying it back-and-forth into `JacobianData` (host memory).
+  [(#477)](https://github.com/PennyLaneAI/pennylane-lightning/pull/477)
+
 ### Improvements
+
+* Refactor LKokkos `Measurements` class to use (fast) specialized functors whenever possible.
+  [(#481)] (https://github.com/PennyLaneAI/pennylane-lightning/pull/481)
 
 * Merge Lightning Qubit and Lightning Kokkos backends in the new repository.
   [(#472)] (https://github.com/PennyLaneAI/pennylane-lightning/pull/472)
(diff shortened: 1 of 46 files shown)


.github/workflows/tests_without_binary.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-22.04]
-        pl_backend: ["lightning_qubit", "lightning_kokkos"]
```
AmintorDusko, line 22: Hey Vincent, why are you removing ``lightning_kokkos`` from the tests without binary?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -622,7 +624,9 @@ class StateVectorKokkos final
         case GeneratorOperation::MultiRZ:
             return applyGeneratorMultiRZ(wires, inverse, params);
         default:
+            // LCOV_EXCL_START
             PL_ABORT(std::string("Generator does not exist for ") + opName);
+            // LCOV_EXCL_END
```
AmintorDusko, line 629: I wonder why we are excluding it from the coverage checks.

AmintorDusko (commented):
I notice that you are excluding some parts from the coverage checks.
Could you please elaborate on why?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: It is excluded to increase the % coverage (acting on the denominator).

The `default` case is there just in case (for the future one might say), but it is not covered by the tests first because `applyNamedOperation` is never called directly, second because `gates_indices_[opName]` only registers valid operations and last because `opName` is usually validated before hitting that function.

.github/workflows/tests_without_binary.yml
AmintorDusko: Yes, we need to install lightning_qubit to ensure package compatibility.
We shouldn't have any problems with the CI, could you please provide more context for that?
We want to keep testing lightning_kokkos without binaries.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
AmintorDusko: Even when we have a default case, for a just-in-case situation, it is better to test it.
We want to mitigate the number of variables when we are potentially hitting a bug, especially if it is in an unusual situation.

pennylane_lightning/core/src/gates/GateOperation.hpp
```diff
@@ -25,9 +25,9 @@ namespace Pennylane::Gates {
  * @brief Enum class for all gate operations
  */
 enum class GateOperation : uint32_t {
-    BEGIN = 0,
+    BEGIN = 1,
     /* Single-qubit gates */
-    Identity = 0,
+    Identity = 1,
```
mlxd, line 30: This has always bothered me -- I'd like to figure out how to get rid of this begin/end thing here and duplicate values, but for now I'll ask --- does this change cause any headaches for us, or is it overall beneficial?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Generator.cpp
```diff
@@ -695,6 +708,7 @@ TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applyGeneratorCRZ",
 
 TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applyGeneratorMultiRZ",
                    "[StateVectorKokkosManaged_Generator]", float, double) {
+    const bool inverse = GENERATE(true, false);
```
mlxd, line 711: What is the `GENERATE` call for? I assume since the next section is scoped it creates two test variants at this point?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
```diff
@@ -978,10 +992,11 @@ TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applySingleExcitationPlus",
...
             StateVectorKokkos<TestType> kokkos_sv{num_qubits};
             std::vector<ComplexT> result_sv(kokkos_sv.getLength(), {0, 0});
-
+            const TestType angle =
+                (inverse) ? -0.267030328057308 : 0.267030328057308;
             kokkos_sv.HostToDevice(ini_st.data(), ini_st.size());
-            kokkos_sv.applyOperation("SingleExcitationPlus", {0, 2}, false,
-                                     {0.267030328057308});
+            kokkos_sv.applyOperation("SingleExcitationPlus", {0, 2}, inverse,
+                                     {angle});
```
mlxd, line 999: If we negate the value and also pass the inverse value, should these cancel one another? Is it better to use one or the other?

pennylane_lightning/core/src/simulators/lightning_qubit/gates/tests/Test_OpToMemberFuncPtr.cpp
```diff
@@ -38,25 +38,28 @@ template <typename EnumClass> constexpr auto allGateOps() {
...
 }
-template <class PrecisionT, class ParamT, class GateImplemenation>
-constexpr bool testAllGatesImplemeted() {
-    return testAllGatesImplementedIter<PrecisionT, ParamT, GateImplemenation,
-                                       0>();
+template <class PrecisionT, class ParamT, class GateImplementation>
+constexpr bool testAllGatesImplemented() {
+    return testAllGatesImplementedIter<PrecisionT, ParamT, GateImplementation,
+                                       static_cast<uint32_t>(
+                                           GateOperation::BEGIN)>();
```
mlxd, line 62: Wait, so these numbers were just raw passed in previously, rather than being referenced from their origin? Good catch on tracking these things.

mlxd (commented):
Thanks @vincentmr 

Just a few comments

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Generator.cpp
vincentmr: `GENERATE` is akin to `pytest.mark.parametrize`. Here is a [link](https://github.com/catchorg/Catch2/blob/devel/docs/generators.md) to the doc. Most gate functors are templated on `inverse`, but the `inverse=true` branches are almost never tested as pointed by Codecov. For example, in the previous `multiQubitOpFunctor` implementation, the [inverse branch](https://github.com/PennyLaneAI/pennylane-lightning-kokkos/pull/75/files#diff-f158dd60dba477dda099bef0bf80cdd6d7606d3e66aa035ddf294470e6618dfbL228) had `const std::size_t n_wires = dim;` instead of `const std::size_t n_wires = wires.size();` (the former is `2**wires.size()` and the latter is correct). I thus introduced parametrization over `inverse` to cover all `inverse` branches, as much as possible.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
vincentmr: The goal here is to test both `inverse` implementations of the functor. I could have two set of reference values, but here I just flipped the angle instead.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/tests/Test_OpToMemberFuncPtr.cpp
vincentmr: > Wait, so these numbers were just raw passed in previously, rather than being referenced from their origin? 

Yep :)

pennylane_lightning/core/src/gates/GateOperation.hpp
vincentmr: The amount of error messages popping after changing this to `1` was frightening at first and makes you want to change this back to `0` and forget about it. They do not appear in many places in the end. They are used in a few places to size the enum or iterate over it. 

> does this change cause any headaches for us, or is it overall beneficial?

I put that in following @AmintorDusko 's [comment](https://github.com/PennyLaneAI/pennylane-lightning/pull/485#discussion_r1309090301). I don't think it causes any headaches. It allowed us to catch a few lurking magic `0`'s, so I guess it makes the code more robust. The immediate benefit is I can add coverage to the `default` cases (`PL_ABORT`) for named gates and generators.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Generator.cpp
AmintorDusko: The use of Catch2 data generators is a nice addition to our test stack.

.github/CHANGELOG.md
```diff
@@ -4,8 +4,14 @@
...
 ### Breaking changes
 
+* Enable pre-commit hooks to format all Python files and linting of all Python source files.
+  [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
+
 ### Improvements
 
+* Add tests to increase LKokkos coverage.
+  [(#485)](https://github.com/PennyLaneAI/pennylane-lightning/pull/485)
+
```
AmintorDusko, line 14: Don't forget to add your name to the changelog.

pennylane_lightning/core/src/gates/GateOperation.hpp
AmintorDusko: Thank you for that, @vincentmr.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
```diff
@@ -256,6 +256,7 @@ template <class PrecisionT> struct getExpectationValueTwoQubitOpFunctor {
     }
 };
 
+// LCOV_EXCL_START
```
AmintorDusko, line 259: We are also excluding this part of the code from coverage. Is there something we can do to include it? Are you planning a follow-up PR?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -198,10 +198,12 @@ class Measurements final
         case ExpValFunc::Hadamard:
             return applyExpValNamedFunctor<getExpectationValueHadamardFunctor,
                                            1>(wires);
+            // LCOV_EXCL_START
```
AmintorDusko, line 201: We are also excluding this part of the code from coverage. Is there something we can do to include it? Are you planning a follow-up PR?

tests/test_adjoint_jacobian.py
```diff
@@ -30,6 +31,15 @@
...
 )
 
+kokkos_args = [None]
+if device_name == "lightning.kokkos" and ld._CPP_BINARY_AVAILABLE:
+    from pennylane_lightning.lightning_kokkos_ops import InitializationSettings
+
+    kokkos_args += [InitializationSettings().set_num_threads(2)]
+
+fixture_params = itertools.product([np.complex64, np.complex128], kokkos_args)
+# fixture_params = [np.complex64, np.complex128]
```
AmintorDusko, line 41:
```suggestion
```

AmintorDusko (commented):
You did a nice work so far.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
vincentmr: > We are also excluding this part of the code from coverage. Is there something we can do to include it? 

Yes.

> Are you planning a follow-up PR?

But it will be much more straightforward in a follow-up. We are still using the "inner-product" scheme in `expval` instead of `getExpectationValueMultiQubitOpFunctor` on `master`. When we merge #489 , we can simply add an `expval` test to cover it.

AmintorDusko: If by removing the ``LCOV_EXCL`` zone we still can pass the coverage checks, I think it would be preferable.

AmintorDusko (approved):
Good work on that! I will approve it now, but please check if, without the coverage exclusion zone, we still pass the coverage checks. I think it is better not to add it, if not needed.

mlxd (approved):
Thanks @vincentmr

PR #481, opened 2023-08-23
Accel/expval

vincentmr:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [ ] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**
In the LKokkos backend, there are two ways to implement expectation values:

1. Copy the statevector, apply the observable to the copy, compute the expectation value with a BLAS-like inner product.
2. Accumulate the expectation value on-the-fly, applying the observable to a portion of the statevector.

The first method is currently in use, but is wasteful in a couple ways:

- The statevector copy can require a lot of memory, which means running out with one or two fewer qubits depending on the system.
- Computational efficiency is asymptotically memory-bound and reduce operations can achieve roughly twice the bandwidth of for loop operations on Nvidia devices, for example, which favors on-the-fly implementations.

A simple benchmarking script like the one below yields, for 29 qubits (on Perlmutter's A100)

1. 0.330 sec.
2. 0.012 sec.

i.e. a 27.5x speed-up.

```
import pennylane as qml
import time

n_wires = 29
n_repeat = 100

dev = qml.device("lightning.kokkos", wires=n_wires)

@qml.qnode(dev)
def circuit():
    return qml.expval(qml.PauliZ(n_wires//2))

t0 = time.time()
for _ in range(n_repeat):
    circuit()
dt = (time.time() - t0)/n_repeat
print(f"{dt}")
```
 
**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #489, opened 2023-08-28
Template/expval

vincentmr:
### Before submitting

**Context:**
This PR is a follow-up on #481. In the last PR, it appeared that reducing `expval` on the fly is generally faster than using inner products. Another factor is the computation of the observable-statevector product and the parallelization scheme used to do it. The general scheme uses three layers of parallelism with team policies. This introduces several parameters which should be tuned for optimal performance, but are currently left to Kokkos' heuristics to decide. On the other hand, the straightforward range policy-based scheme of the 1- and 2-qubit kernels outperforms the general scheme significantly. 

Since this discrepancy does not appear explainable by the flop intensity increase between 2- and 3+-qubit kernels, I introduce specialized 3- to 5-qubit kernels. I draw the following conclusions:

- On-the-fly expval kernels are generally faster.
- Range-policy kernels are faster than the team-policy one up to 4-qubits on the OPENMP and HIP backends and up to 5-qubits on CUDA.

The following figures show timings to get the expectation value of a `Hermitian` observable for OPENMP, CUDA and HIP respectively.

benchmarks_CPU
benchmarks_GPU
benchmarks_HIP

Introduce specialized 3- to 5-qubit kernels. Refactor `getExpValMatrix` wrapper in `MeasurementsKokkos.hpp`. Add few tests.

**Benefits:**
Faster `expval` on all platforms, especially for 3+-qubit observables. 

**Possible Drawbacks:**
None

**Related GitHub Issues:**
#481


pennylane_lightning/core/src/simulators/lightning_kokkos/gates/README.md
AmintorDusko, line 1: This part is a little bit confusing. You are saying that you generated the C++ code with Python, and after that copy-and-paste here? I remember something about exploring the use of macros, what is the status of that?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
```diff
@@ -355,4 +265,370 @@ template <class PrecisionT> struct getExpectationValueSparseFunctor {
     }
 };
 
+template <class PrecisionT> struct getExpVal1QubitOpFunctor {
```
AmintorDusko, line 268: Here start the code generated with Python, right?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/README.md
AmintorDusko: Python-generated C++ code bound to be called from Python.

AmintorDusko (commented):
I left only a few comments for now.
I see that we still have some work to do in terms of coverage.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/README.md
vincentmr: I kept it there for context and just in case, but if you think the source code is clear enough I'll remove it.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
vincentmr: It was, but I replace the boilerplate with macro calls. You may think 64 macro calls for the 5Qubit functor is boilerplaty, but it used to be ~1000 lines for that functor alone.

AmintorDusko: Oh my!

vincentmr:
> I left only a few comments for now. I see that we still have some work to do in terms of coverage.

I would like to merge #485 first to assess the coverage situation.

AmintorDusko:
> > I left only a few comments for now. I see that we still have some work to do in terms of coverage.
> 
> I would like to merge #485 first to assess the coverage situation.

Absolutely, I think it is only sensible to do so.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/tests/Test_StateVectorKokkos_Expval.cpp
```diff
@@ -379,6 +381,158 @@ TEMPLATE_TEST_CASE("Test expectation value of TensorProdObs",
...
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  -u, z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  -u, z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  u, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, -u, z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  u,  z,  z,  z, z, z,  z, z,  z, z, z,
+            z,  z, z,  z, z,  z,  z, z,  z,  z,  z,  z, z, z,  z, z,  z};
```
mlxd, line 521: This is a big one

mlxd (approved):
Nothing more to add --- thanks a bunch @vincentmr 
I'm happy with the macro-approach for now, but we can revisit later to see if it can become some compile-time generated parameter-packed solution.

PR #490, opened 2023-08-29
Accelerate multi-qubit gates

vincentmr:
### Before submitting

**Context:**
This PR is a follow-up on #489 . The general scheme for multi-qubit gates uses three layers of parallelism with team policies. This introduces several parameters which should be tuned for optimal performance, but are currently left to Kokkos' heuristics to decide. On the other hand, the straightforward range policy-based scheme of the 1- and 2-qubit kernels outperforms the general scheme significantly. 

I introduce specialized 3- to 5-qubit kernels. I draw the following conclusion:

- Range-policy kernels have the same performance as the team-policy one up to 4-qubits on the OPENMP backend and are then slower.
- Range-policy kernels are faster than the team-policy one up to at least 5-qubits on the CUDA and HIP backends.

The following figures show timings to apply a `QubitUnitary` for OPENMP, CUDA and HIP respectively.

benchmarks_CPU
benchmarks_CUDA
benchmarks_HIP

Introduce specialized 3- to 5-qubit unitary gate kernels. Refactor `applyMultiQubitOp` wrapper in `StateVectorKokkos.hpp`. Functors are not templated on `inverse` anymore, taking the conjugate-transpose once and for all upon entering `applyMultiQubitOp` instead of on-the-fly for each element of the for loop. Add few tests.

**Benefits:**
Faster `QubitUnitary`, especially for 3+-qubit observables on GPU-devices. 

**Possible Drawbacks:**
None

**Related GitHub Issues:**


AmintorDusko (commented):
The content here is similar to the other PR, I will be back when the coverage situation improves.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -257,96 +257,62 @@ class StateVectorKokkos final
...
+                           const std::vector<std::size_t> &wires,
+                           bool inverse = false) {
         auto &&num_qubits = this->getNumQubits();
-        if (!inverse) {
+        std::size_t two2N = std::exp2(num_qubits - wires.size());
+        std::size_t dim = std::exp2(wires.size());
+        KokkosVector matrix_trans("matrix_trans", matrix.size());
+
+        if (inverse) {
+            Kokkos::MDRangePolicy<Kokkos::Rank<2>> policy_2d({0, 0},
```
multiphaseCFD, line 275: Can we use a const variable instead of `2` here?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -257,96 +257,62 @@ class StateVectorKokkos final
...
+                two2N, apply3QubitOpFunctor<fp_t>(*data_, num_qubits,
+                                                  matrix_trans, wires));
+            break;
+        case 4:
+            Kokkos::parallel_for(
+                two2N, apply4QubitOpFunctor<fp_t>(*data_, num_qubits,
+                                                  matrix_trans, wires));
+            break;
+        default:
             std::size_t scratch_size = ScratchViewComplex::shmem_size(dim) +
```
multiphaseCFD, line 307: Wondering if larger `shmem_size` can deliver better performance?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
```diff
@@ -1687,4 +1483,426 @@ template <class PrecisionT, bool inverse = false> struct rotFunctor {
...
+template <class PrecisionT> struct apply1QubitOpFunctor {
+
+    using ComplexT = Kokkos::complex<PrecisionT>;
+    using KokkosComplexVector = Kokkos::View<ComplexT *>;
+    using KokkosIntVector = Kokkos::View<std::size_t *>;
+
+    KokkosComplexVector arr;
+    KokkosComplexVector matrix;
+    const std::size_t n_wires = 1;
+    const std::size_t dim = 1U << n_wires;
```
multiphaseCFD, line 1495: Is there implicit type conversion from `U` to `size_t` here?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
```diff
@@ -1687,4 +1483,426 @@ template <class PrecisionT, bool inverse = false> struct rotFunctor {
...
+        arr(i010) = GATESUM3(0B010);
+        arr(i011) = GATESUM3(0B011);
+        arr(i100) = GATESUM3(0B100);
+        arr(i101) = GATESUM3(0B101);
+        arr(i110) = GATESUM3(0B110);
+        arr(i111) = GATESUM3(0B111);
+    }
+};
+
+#define GATEENTRY4(xx, yy) xx << 4 | yy
```
multiphaseCFD, line 1667: could you add more tests for the code generated by preprocessor?

multiphaseCFD (commented):
Thanks @vincentmr for the nice work! Just leave a few comments and thoughts.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: This is replacing a double `parallel_for` with `RangePolicy`s, so we could (maybe) have a constant like `constexpr int double_loop_rank=2`, but I don't see the benefit in this case.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: AFAIK the size doesn't impact performance, it just has to be enough to store the `ScratchView`.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
vincentmr: I don't think it does (or maybe depends on the compiler). Nice catch! Will change all similar casts.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
multiphaseCFD: It would be nice to name it as `double_loop_rank` for the maintenance purpose.

multiphaseCFD (commented):
Will come back to have a look once the GPU CI is added.

vincentmr:
> Will come back to have a look once the GPU CI is added.

I'm unsure when we'll have a runner with CUDA-12 (and not sure we'll have any runner with HIP-capable devices any time soon), so could we move forward with this PR nevertheless?

multiphaseCFD:
> > Will come back to have a look once the GPU CI is added.
> 
> I'm unsure when we'll have a runner with CUDA-12 (and not sure we'll have any runner with HIP-capable devices any time soon), so could we move forward with this PR nevertheless?

yea, sounds good to me!

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
```diff
@@ -1337,3 +1346,142 @@ TEMPLATE_TEST_CASE("Sample", "[StateVectorKokkosManaged_Param]", float,
                      Catch::Approx(expected_probabilities).margin(.05));
     }
 }
+
+TEMPLATE_TEST_CASE("Test NQubit gate versus expectation value",
+                   "[StateVectorKokkosManaged_Param]", float, double) {
+    using ComplexT = StateVectorKokkos<TestType>::ComplexT;
+    // using VectorT = TestVector<std::complex<TestType>>;
```
mlxd, line 1353:
```suggestion
```

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -257,96 +258,61 @@ class StateVectorKokkos final
...
 
     /**
-     * @brief Apply a single qubit operator to the state vector using a matrix
+     * @brief Apply a multi qubit operator to the state vector using a matrix
      *
      * @param matrix Kokkos gate matrix in the device space
      * @param wires Wires to apply gate to.
      * @param inverse Indicates whether to use adjoint of gate.
      */
-    void applySingleQubitOp(const KokkosVector &matrix,
```
mlxd, line 266: Just for my recollection, the `applySingleQubitOp` function is now replaced by something else?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -257,96 +258,61 @@ class StateVectorKokkos final
...
-                        .set_scratch_size(0, Kokkos::PerTeam(scratch_size)),
-                    multiQubitOpFunctor<PrecisionT, true>(*data_, num_qubits,
-                                                          matrix, wires_view));
-            }
+            Kokkos::parallel_for(
+                "multiQubitOpFunctor",
+                TeamPolicy(two2N, Kokkos::AUTO, dim)
+                    .set_scratch_size(0, Kokkos::PerTeam(scratch_size)),
+                multiQubitOpFunctor<PrecisionT>(*data_, num_qubits,
+                                                matrix_trans, wires));
```
mlxd, line 314: We don't need to worry about the inverse at this point anymore?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: Everything is under `applyMultiQubitOp`, and `singleQubitOpFunctor`, `twoQubitOpFunctor` are replaced respectively with `apply1QubitOpFunctor`, `apply2QubitOpFunctor` to have a more consistent naming (we also have `apply3QubitOpFunctor`, `apply4QubitOpFunctor`, etc.)

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: I'm assuming unitary gates, and hence the inverse is obtained with the `parallel_for` above taking the transpose conjugate. We have to do this once, on a relatively small array. This way, we don't need to template each functor on `inverse` and (almost) duplicate the body (see [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/490/files#diff-5e7d16242d2c553f5c54d49701599fb90cf2157979006d8e3bd29c845035f9cbL139) for example)). I'm not sure whether the compiler used to optimize that away, but now we don't need to take the conjugate transpose of the matrix on the fly, and hence we always get contiguous access during the gate application.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
```diff
@@ -1687,4 +1483,421 @@ template <class PrecisionT, bool inverse = false> struct rotFunctor {
...
+        arr = arr_;
+        matrix = matrix_;
+        num_qubits = num_qubits_;
+    }
+
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const std::size_t k) const {
+        const std::size_t kdim = k * dim;
+        INDEX(i0000, 0B0000);
+        ComplexT v0000 = arr(i0000);
```
AmintorDusko, line 1705: Is there a simple way to make it a single-line statement like:`` ComplexT v0001 = INDEX(0B0001);``?
I think this could improve the reading here.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
```diff
@@ -1687,4 +1483,421 @@ template <class PrecisionT, bool inverse = false> struct rotFunctor {
...
+        arr = arr_;
+        matrix = matrix_;
+        num_qubits = num_qubits_;
+    }
+
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const std::size_t k) const {
+        const std::size_t kdim = k * dim;
+        INDEX(i00000, 0B00000);
+        ComplexT v00000 = arr(i00000);
```
AmintorDusko, line 1804: Same here, regarding making it a single-line statement. (I will not point out all of them).

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
vincentmr: INDEX is a macro that declares and computes `std::size_t i0000`. `i0000` is used to create a copy `v000` of `arr(i000)` and then used to reassign `arr(i000)` as `GATESUM3(0B000)`. I could modify the macro to have lines like
```
std::size_t i000 = INDEX(i000, 0B000);
```
Is that better?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
vincentmr: Keep discussion [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/490#discussion_r1319014974).

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/GateFunctorsParam.hpp
AmintorDusko: Yes, I got how the macros work, and I think that if you can change it to make these statements single-line would be great.
Thank you for that.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
mlxd: Got it, thanks!

AmintorDusko (dismissed):
Hi Vincent, Nice work!
Thank you for that!

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
```diff
@@ -414,14 +416,14 @@ template <class PrecisionT> struct getExpVal3QubitOpFunctor {
...
-        INDEX(i110, 0B110);
-        INDEX(i111, 0B111);
+        std::size_t i000 = INDEX(i000, 0B000);
+        std::size_t i001 = INDEX(i001, 0B001);
+        std::size_t i010 = INDEX(i010, 0B010);
+        std::size_t i011 = INDEX(i011, 0B011);
+        std::size_t i100 = INDEX(i100, 0B100);
+        std::size_t i101 = INDEX(i101, 0B101);
+        std::size_t i110 = INDEX(i110, 0B110);
+        std::size_t i111 = INDEX(i111, 0B111);
```
AmintorDusko, line 426: I understand that you need to store each one of these index values for further reuse.
I wonder if you could instead of a block of code like that (here and in other places) have a for loop and store the values in an array.

AmintorDusko: Here for example you are essentially going from 0 to 7.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
```diff
@@ -563,38 +566,38 @@ template <class PrecisionT> struct getExpVal5QubitOpFunctor {
...
+        std::size_t i10110 = INDEX(i10110, 0B10110);
+        std::size_t i10111 = INDEX(i10111, 0B10111);
+        std::size_t i11000 = INDEX(i11000, 0B11000);
+        std::size_t i11001 = INDEX(i11001, 0B11001);
+        std::size_t i11010 = INDEX(i11010, 0B11010);
+        std::size_t i11011 = INDEX(i11011, 0B11011);
+        std::size_t i11100 = INDEX(i11100, 0B11100);
+        std::size_t i11101 = INDEX(i11101, 0B11101);
+        std::size_t i11110 = INDEX(i11110, 0B11110);
+        std::size_t i11111 = INDEX(i11111, 0B11111);
```
AmintorDusko, line 600: If you could go with a loop here and store the elements in an array, we could greatly reduce code here, and below.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
vincentmr: > I understand that you need to store each one of these index values for further reuse. I wonder if you could instead of a block of code like that (here and in other places) have a for loop and store the values in an array.

This is similar to the previous (bugged) implementation was doing. The problem is declaring an array, vector or a Kokkos::view is a host function, so it breaks device-backended Kokkos. Now what was done to circumvent that was having a Kokkos::view attribute in the functor. This will compile but since this attribute is shared across threads, they all try writing to it which lead to the bug preventing evaluation of multi-qubit gates. The way to solve this (in the last release) was introducing the three-level team implementation which allows declaring Kokkos::ScratchView's (which are not shared across threads). But this implementation (at least with the default heuristics) is significantly slower than the one proposed here in many cases. 

So long story short, I don't see this working because of Kokkos-related constraints unfortunately.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
mlxd: The above vectorises nicely though using GCC 11, from avx all the way to avx512f. I'd like to see if using a loop would give the same output before suggesting the change here.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp
mlxd: So the only way we could do this is to define a host-side Kokkos view, then pass that in somehow and use that for the updates? Would that be possible, or preferred? If not, at least we know the above generates performant code for now.

mlxd: I assume also these variables are private-level in CUDA, rather than global, so each would be thread-local?

vincentmr: Beside `Kokkos::ScratchView`, I don't know how this could work. You could maybe allocate and pass a couple `Kokkos::View`s that have `nthreads*2**nwires` entries and try to read/write only parts of it, but I'm not sure this will work or be more elegant down the line.

vincentmr: > I assume also these variables are private-level in CUDA, rather than global, so each would be thread-local?

Yes.

AmintorDusko (approved):
Once again, thank you for the nice job!

mlxd (approved):
Thanks @vincentmr

PR #496, opened 2023-09-08
Feature/docker

vincentmr:
### Before submitting

**Context:**
Some systems have messy environments which makes on-site deployment complicated. Docker is a solution to this issue.

Add Docker container utilities.

**Benefits:**
Easier deployment, visibility on DockerHub.

**Possible Drawbacks:**

**Related GitHub Issues:**


docker/README.md
```diff
@@ -0,0 +1,68 @@
+# PennyLane and the Lightning plugins
+
+PennyLane is an open-source software framework for quantum machine learning, quantum chemistry, and quantum computing, with the ability to run on all hardware. Maintained with ❤️ by Xanadu.
+
+## Documentation
+
+For more information on PennyLane, including the demos, APIs and development guide, visit the [PennyLane documentation site](https://pennylane.ai/).
```
AmintorDusko, line 7: It would be great if we had direct links to Lightning and Lightning-GPU docs.

docker/Dockerfile
AmintorDusko, line 1: I think we could have more comments in this Dockerfile explaining the intentions behind each block of commands.
This will help a user to better understand what is happening and to keep this file under control.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* Add `Dockerfile` in `docker` and `make docker` workflow in `Makefile`. The Docker images and documentation are available on [DockerHub](https://hub.docker.com/r/vincentmichaudrioux/pennylane).
+  [(#496)](https://github.com/PennyLaneAI/pennylane-lightning/pull/496)
```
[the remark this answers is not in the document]
vincentmr: After discussion with @smokingroosters , we'll go with the [Github container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry), which is hosted by `pennylaneai`. I'll therefore update the docker repo links and all to resolve this comment.

docker/README.md
```diff
@@ -0,0 +1,75 @@
...
+Listed below are links to the Lightning-plugins documentation:
+
+- [pennylane-lightning doc](https://docs.pennylane.ai/projects/lightning/en/stable/)
+- [pennylane-lightning-gpu doc](https://docs.pennylane.ai/projects/lightning-gpu/en/stable/)
+- [pennylane-lightning-kokkos doc](https://docs.pennylane.ai/projects/lightning-kokkos/en/latest/)
```
AmintorDusko, line 12:
```suggestion
You can also check the Lightning-plugins documentation:

- [pennylane-lightning](https://docs.pennylane.ai/projects/lightning/en/stable/)
- [pennylane-lightning-gpu](https://docs.pennylane.ai/projects/lightning-gpu/en/stable/)
- [pennylane-lightning-kokkos](https://docs.pennylane.ai/projects/lightning-kokkos/en/latest/)
```

AmintorDusko (approved):
I had just a small suggestion, but overall looks pretty good!

Makefile
```diff
@@ -125,3 +125,25 @@ docs:
...
 	$(MAKE) -C doc clean
+
+.PHONY : docker
+ifdef target
+    TARGET := $(target)
+else
+    TARGET := lightning-qubit
+endif
+docker-build:
+	docker build -f docker/Dockerfile --tag=pennylaneai/pennylane:$(TARGET) --target wheel-$(TARGET) .
```
mlxd, line 136: Would it be better to use `--tag=pennylaneai/pennylane-$(TARGET):${TAG}` or equivalent, so that we can version the images? Use `latest` for master/main and by default, and then 0.32.0, etc otherwise for fixed releases?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
-FROM ubuntu:focal AS base
+# Create basic runtime environment base on Ubuntu 22.04 (jammy)
+# Define build arguments
+# Create and activate runtime virtual environment
+FROM ubuntu:jammy AS base-runtime
+ARG KOKKOS_VERSION=4.1.00
+ARG LIGHTNING_VERSION=0.32.0
+ARG CUDA_VERSION=12.2
+ARG CUDA_ARCH=AMPERE80
+ARG AMD_ARCH=AMD_GFX90A
```
mlxd, line 23: I assume these are all user-configurable?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
-    && update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100 --slave /usr/bin/g++ g++ /usr/bin/g++-11 --slave /usr/bin/gcov gcov /usr/bin/gcov-10 \
-    && /usr/sbin/update-ccache-symlinks \
-    && mkdir /opt/ccache \
-    && ccache --set-config=cache_dir=/opt/ccache \
-    && python3 -m venv /opt/venv
+    && rm -rf /var/lib/apt/lists/*
+RUN update-alternatives \
+    --install /usr/bin/gcc gcc /usr/bin/gcc-11 100 \
+    --slave /usr/bin/g++ g++ /usr/bin/g++-11 \
+    --slave /usr/bin/gcov gcov /usr/bin/gcov-10
```
mlxd, line 57: Should this be 11 too?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+RUN /usr/sbin/update-ccache-symlinks
+RUN mkdir /opt/ccache
+RUN ccache --set-config=cache_dir=/opt/ccache
+
+# Create and activate build virtual environment
+# Install Lightning dev requirements
+FROM base-build AS base-build-python
+ENV VIRTUAL_ENV=/opt/venv-build
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
mlxd, line 67: Do we need LD_LIBRARY_PATH here too, such as for custatevec and others?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
-    && ccache --set-config=cache_dir=/opt/ccache \
-    && python3 -m venv /opt/venv
+    && rm -rf /var/lib/apt/lists/*
+RUN update-alternatives \
+    --install /usr/bin/gcc gcc /usr/bin/gcc-11 100 \
+    --slave /usr/bin/g++ g++ /usr/bin/g++-11 \
+    --slave /usr/bin/gcov gcov /usr/bin/gcov-10
+RUN /usr/sbin/update-ccache-symlinks
+RUN mkdir /opt/ccache
+RUN ccache --set-config=cache_dir=/opt/ccache
```
mlxd, line 60: Is it worth combining these run commands to help reduce the base-build image size? Or is that no longer required?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+    pip install --no-cache-dir -r requirements-doc.txt
+
+# Download Lightning release and build lightning-qubit backend
+FROM base-build-python AS build-wheel-lightning-qubit
+WORKDIR /opt/pennylane-lightning
+RUN wget --progress=dot:giga https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v${LIGHTNING_VERSION}.tar.gz
+RUN tar -zxf v${LIGHTNING_VERSION}.tar.gz && mv pennylane-lightning-${LIGHTNING_VERSION}/* /opt/pennylane-lightning
+RUN pip uninstall -y pennylane-lightning
+RUN python setup.py build_ext
+RUN python setup.py bdist_wheel
```
mlxd, line 81: As above re: combining run commands?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+RUN pip install --force-reinstall --no-cache-dir PennyLane_Lightning*.whl
+
+# Install CUDA-12 in build venv image
+FROM base-build-python as base-build-cuda
+WORKDIR /opt/cuda-build
+RUN curl -o cuda-12_2.run https://developer.download.nvidia.com/compute/cuda/12.2.2/local_installers/cuda_12.2.2_535.104.05_linux.run
+RUN chmod a+x cuda-12_2.run
+RUN ./cuda-12_2.run --silent --toolkit --toolkitpath=/usr/local/cuda-12.2
+ENV PATH=/usr/local/cuda/bin:${PATH}
+ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64:${LD_LIBRARY_PATH}
```
mlxd, line 132: Just confirming we don't ship these with nvidia libs contained? Or is that needed?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+
+# Download Kokkos release and build CUDA backend
+FROM base-build-cuda as build-kokkos-cuda
+WORKDIR /opt/kokkos-${KOKKOS_VERSION}
+# RUN wget --progress=dot:giga https://github.com/kokkos/kokkos/archive/refs/tags/${KOKKOS_VERSION}.tar.gz
+# RUN tar -zxf ${KOKKOS_VERSION}.tar.gz && mv /opt/kokkos-${KOKKOS_VERSION}/kokkos-${KOKKOS_VERSION}/* /opt/kokkos-${KOKKOS_VERSION}
```
mlxd, line 138: Are these needed?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+RUN git clone --branch fix-rtld-deepbind https://github.com/cz4rs/kokkos.git /opt/kokkos-${KOKKOS_VERSION}
```
mlxd, line 139: What is this version for? Can this be replaced with the official GH repo?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+ENV CMAKE_PREFIX_PATH=/opt/kokkos:$CMAKE_PREFIX_PATH
+RUN wget --progress=dot:giga https://github.com/PennyLaneAI/pennylane-lightning/archive/refs/tags/v${LIGHTNING_VERSION}.tar.gz
+RUN tar -zxf v${LIGHTNING_VERSION}.tar.gz && mv pennylane-lightning-${LIGHTNING_VERSION}/* /opt/pennylane-lightning
+RUN pip uninstall -y pennylane-lightning
+RUN PL_BACKEND=lightning_kokkos python setup.py build_ext
+RUN PL_BACKEND=lightning_kokkos python setup.py bdist_wheel
+
+# Install python3 and setup runtime virtual env in CUDA-12-runtime image (includes CUDA runtime and math libraries)
+# Install lightning-kokkos CUDA backend
+FROM nvidia/cuda:12.2.0-base-ubuntu22.04 AS wheel-lightning-kokkos-cuda
```
mlxd, line 170: Ah, I think this answer my earlier question.

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN pip install --no-cache-dir custatevec-cu12
+COPY --from=build-wheel-lightning-gpu /opt/pennylane-lightning/dist/ /
+RUN pip install --no-cache-dir --force-reinstall PennyLane_Lightning*.whl
+
+# Install ROCm in build venv image
+FROM base-build-python as base-build-rocm
+RUN wget --progress=dot:giga https://repo.radeon.com/amdgpu-install/5.3/ubuntu/jammy/amdgpu-install_5.3.50300-1_all.deb
```
mlxd, line 221: Is there a ROCm container available? Since we effectively pin the ROCm version here, I wonder if another way is available to build with a later/more up to date release, or allow control of the build version.

This thought could also be applied retrospectively to the CUDA builds

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y \
+    "linux-headers-$(uname -r)" "linux-modules-extra-$(uname -r)" \
+    ./amdgpu-install_5.3.50300-1_all.deb \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+RUN amdgpu-install -y --accept-eula --usecase=rocm,hiplibsdk --no-dkms
```
mlxd, line 228: I don't think we want to accept EULAs in builds like this --- is it required? Is there another way?

docker/Dockerfile
```diff
@@ -12,43 +12,267 @@
...
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y \
+    "linux-headers-$(uname -r)" "linux-modules-extra-$(uname -r)" \
+    ./amdgpu-install_5.3.50300-1_all.deb \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+RUN amdgpu-install -y --accept-eula --usecase=hip --no-dkms
+COPY --from=build-kokkos-rocm /opt/kokkos /opt/kokkos
+COPY --from=build-wheel-lightning-kokkos-rocm /opt/pennylane-lightning/dist/ /
+RUN pip install --force-reinstall --no-cache-dir PennyLane_Lightning*.whl
```
mlxd, line 278: Given the size of this file, I wonder if it is possible to split the multi-stage builds across files for each respective architecture and backend/target

mlxd (commented):
Some thoughts

Makefile
vincentmr: The scope and specs on this story were not clear at the beginning, so I went for `stable` and L, L-GPU and L-Kokkos-OMP first. Since L-GPU worked I added L-Kokkos-CUDA, and since I had L-Kokkos-CUDA I thought I'll try adding L-Kokkos-ROCm to have all single-device backends. Now, I didn't know if the goal was to provide recipes that users can modify to their needs, or have a complete set of official images. If we go for the later, then your suggestion makes a lot of sense. Otherwise, still makes sense, but does not matter as much. I think maybe we should ask @trbromley what he thinks about the scope at this point.

docker/Dockerfile
vincentmr: Yes. For example,
```
docker build --build-arg="LIGHTNING_VERSION=0.31.0" -f docker/Dockerfile --tag=pennylaneai/pennylane:ligthning-qubit --target wheel-ligthning-qubit .
```
should build v0.31.

docker/Dockerfile
vincentmr: Yes, and I think it could be a build arg too.

docker/Dockerfile
vincentmr: That's only for building so we don't ship.

docker/Dockerfile
vincentmr: They will be once the branch I'm building from is merged into Kokkos and Kokkos produces a new release.

docker/Dockerfile
vincentmr: > Can this be replaced with the official GH repo?

No, this branch has a bugfix which allows correct execution on CUDA and HIP-supporting devices when building with CUDA-12. I filed an [issue](https://github.com/kokkos/kokkos/issues/6366) about it.

docker/Dockerfile
vincentmr: This version is a bit old indeed. I'll try to update it while adding control over the version.

docker/Dockerfile
vincentmr: It is required for a non-interactive install. There is a [note](https://amdgpu-install.readthedocs.io/en/latest/install-overview.html) about it in the doc, saying "this option will prevent the proprietary packages or updates from being accessible directly with the system package manager". I could start from other ROCm containers to go around it, but I wanted to build all packages on the same footing as much as possible.

docker/Dockerfile
vincentmr: There is certainly a way, let me check.

vincentmr: Seems like everything needs to be in one file for Docker to manage the multi-stage building process by itself. If we want to split the Dockerfile, we'll need a script to deal with it ourselves. I think it's not that bad yet :D

docker/Dockerfile
vincentmr: I think Kokkos' develop branch now has a fix. I'll double check it works on with both CUDA and HIP and we can switch back to the official Kokkos repo.

docker/Dockerfile
vincentmr: Can ship from `rocm/dev-ubuntu-22.04`.

docker/Dockerfile
vincentmr: So the develop commit that works in the CI tests breaks the CUDA/HIP containers. I thus suggest we keep it as `--branch fix-rtld-deepbind https://github.com/cz4rs/kokkos.git` for now.

mlxd (approved):

PR #498, opened 2023-09-11
Add compat workflows.

vincentmr:
### Before submitting

**Context:**
SC-43411 requires compatibility tests between PennyLane and Lightning's most recent stable releases and development versions. 

This PR adds `compat`-workflows calling the Linux and no-binary test workflows with `lightning-version` and `pennylane-version` inputs. These inputs are then used to switch between `stable` and `latest` versions.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/plugin-test-matrix/pull/50


.github/workflows/compat-check-latest-latest.yml
```diff
@@ -0,0 +1,20 @@
+name: Compat Check w/PL - latest/latest
+
+on:
+  schedule:
+    - cron: "0 3 * * 1-5"  # Run daily at 3am Mon-Fri
```
mlxd, line 5: Any reason 3am is used? Or is it arbitrary?

.github/workflows/tests_linux.yml
```diff
@@ -97,7 +108,16 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: main
-          fetch-depth: 2
+          fetch-depth: ${{ inputs.lightning-version == 'stable' && 0 || 2 }}
```
mlxd, line 111: What is `&& 0 || 2` selecting for?

.github/workflows/tests_linux.yml
```diff
@@ -235,6 +271,12 @@ jobs:
           cd main
           python -m pip install -r requirements-dev.txt
 
+      - name: Install Stable PennyLane
+        if: inputs.pennylane-version == 'stable'
+        run: |
+          cd main
+          python -m pip uninstall -y pennylane && python -m pip install -U pennylane
```
mlxd, line 278: Do we need to uninstall first?

mlxd (commented):
Just a few queries

.github/workflows/compat-check-latest-latest.yml
vincentmr: It's mostly arbitrary, but plugin-test-matrix workflows are triggered at midnight, so I'm thinking we should let them some time to finish. At 3 o'clock Actions should be least busy too I reckon.

.github/workflows/tests_linux.yml
vincentmr: Yes, because we install pennylane@master when we install the dev requirements above.

.github/workflows/tests_linux.yml
vincentmr: I'm not 100% sure, I took that from @rashidnhm . I think the logic is: if latest fetch depth == 2, if stable fetch depth == 0 (i.e. everything) to get all tags. We could instead use the default fetch depth (1) everywhere and use the `fetch-tags` option to get tags where needed.

mlxd (approved):
Thanks @vincentmr

AmintorDusko (approved):
Just a small suggestion. Thank you for that.

PR #502, opened 2023-09-15
Fix wheels build for aarch64

AmintorDusko:
**Context:** Wheels for aarch64 were not being tested with PennyLane master.

**Description of the Change:** Install Pennylane master before testing.

**Benefits:** Tests will work properly.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/wheel_linux_aarch64.yml
+++ b/.github/workflows/wheel_linux_aarch64.yml
@@ -6,7 +6,6 @@ name: Wheel::Linux::ARM
 # **Who does it impact**: Wheels to be uploaded to PyPI.
 
 on:
-  pull_request:
   push:
     branches:
       - master
--- a/.github/workflows/wheel_linux_aarch64.yml
+++ b/.github/workflows/wheel_linux_aarch64.yml
@@ -6,6 +6,7 @@ name: Wheel::Linux::ARM
 # **Who does it impact**: Wheels to be uploaded to PyPI.
 
 on:
+  pull_request:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -6,6 +6,7 @@ name: Wheel::Linux::ARM
 # **Who does it impact**: Wheels to be uploaded to PyPI.
 
 on:
+  pull_request:
```
AmintorDusko, line 9:
```suggestion
```
To be removed after review.

vincentmr (approved):
Looks good to me!

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -142,7 +142,9 @@ jobs:
 
           CIBW_TEST_REQUIRES: pytest pytest-cov pytest-mock flaky
 
-          CIBW_BEFORE_TEST: if ${{ matrix.pl_backend == 'lightning_kokkos'}}; then SKIP_COMPILATION=True PL_BACKEND="lightning_qubit" pip install -e . -vv; fi
+          CIBW_BEFORE_TEST: |
+            python -m pip install git+https://github.com/PennyLaneAI/pennylane.git@master
+            if ${{ matrix.pl_backend == 'lightning_kokkos'}}; then SKIP_COMPILATION=True PL_BACKEND="lightning_qubit" pip install -e . -vv; fi
```
mlxd, line 147: So if the backend installed in LK, then we must also install LQ but not build the module? 
This seems a little convoluted to me. I wonder if there is a clearer way.

Should we always ensure LK installs LQ as well, so that we never have a scenario where the package is installed without its required dependencies?

AmintorDusko: Hi @mlxd, this was not a new addition. This is the patch we have in place to run the LKokkos tests and wheels.
A more definite solution will come when we have a proper installation for LKokkos.
The strategy I have in mind to install LKokkos will actually be moving only the LKokkos-specific files and having as a pre-requisite the LQubit installation (with core files) targeting the same version.

mlxd (approved):

PR #503, opened 2023-09-19
Lower timeout-minutes.

vincentmr:
### Before submitting

**Context:**
Windows workflows sometimes get stuck at some step and times out after 6 hours.

Lower timeout on all workflows to avoid this situation.  

**Benefits:**
Fail faster.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ### Breaking changes
 
+* Introduce `timeout-minutes` in various workflows, mainly to avoid Windows builds hanging for several hours.
+  [(#503)](https://github.com/PennyLaneAI/pennylane-lightning/pull/503)
+
 * Cast integral-valued arrays to the device's complex type on entry in `_preprocess_state_vector` to ensure the state is correctly represented with floating-point numbers.  
   [(#501)](https://github.com/PennyLaneAI/pennylane-lightning/pull/501)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Introduce `timeout-minutes` in various workflows, mainly to avoid Windows builds hanging for several hours.
   [(#503)](https://github.com/PennyLaneAI/pennylane-lightning/pull/503)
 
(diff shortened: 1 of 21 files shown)


AmintorDusko (commented):
Did you test the wheels we build only when merging to master? (edit: I'm asking because maybe I'm missing something here) The [Wheel::Linux::ARM](https://github.com/PennyLaneAI/pennylane-lightning/blob/c4d43287d1347e5cb3e76170f69cb38f49ea9c5d/.github/workflows/wheel_linux_aarch64.yml) for example can sometimes take [more than 30 min](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/6240764078). I think we could only limit the Windows runners since they are the problematic ones. What do you think?

vincentmr:
> Did you test the wheels we build only when merging to master? (edit: I'm asking because maybe I'm missing something here) The [Wheel::Linux::ARM](https://github.com/PennyLaneAI/pennylane-lightning/blob/c4d43287d1347e5cb3e76170f69cb38f49ea9c5d/.github/workflows/wheel_linux_aarch64.yml) for example can sometimes take [more than 30 min](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/6240764078). I think we could only limit the Windows runners since they are the problematic ones. What do you think?

I think it good to have time limit on each job, to set expectations but also avoid wasting time in some cases. While we have an idea how long each job should take, new hires, contributors from other teams and external contributors don't. So I'd like to introduce timeouts and revise them as required. For `wheel_linux_aarch64`, I think you may be right, so I would just increase it to 45 minutes.

vincentmr:
I should note too that there are 30 minutes for building Kokkos and 30 minutes for the wheels, but I can still increase it a bit just in case.

mlxd:
Let's set these as 45 mins -- enough time to have a safety margin, and short enough to kill a task that runs over ahead of the 6 hr window.

AmintorDusko (approved):
Hey @vincentmr, thank you for that! Please triple-check the CIs building wheels only when merging.

mlxd (approved):
Thanks @vincentmr

PR #534, opened 2023-10-27
Add block-list for auditwheel builds with L-GPU

mlxd:
### Before submitting

**Context:** Prevents licensed libs being included during wheel-builds

**Description of the Change:** Adheres to original design from Lightning GPU repository.

**Benefits:** Prevents licensed libraries from being packaged in wheel-builds

**Possible Drawbacks:**

**Related GitHub Issues:**


AmintorDusko (approved):
Makes sense. Thank you for that!

bin/auditwheel
```diff
@@ -0,0 +1,25 @@
...
+from auditwheel.main import main
+from auditwheel.policy import _POLICIES as POLICIES
+
+# Do not include licensed dynamic libraries
+libs = [
+    "libcudart.so.11.0",
+    "libcublasLt.so.11",
+    "libcublas.so.11",
+    "libcusparse.so.11",
+    "libcustatevec.so.1",
```
vincentmr, line 16: I think this was asked before, but don't we want wildcards on the lib versions?

mlxd: As I recall, I wasn't able to validate this worked in the past. In the interest of time, and since we pin explicitly to these lib version (their names have remained invariant through symlinks) we should be fine with this until we move to CUDA 12.

multiphaseCFD: I'm curious about it too.

vincentmr (approved):
Looks good to me.

PR #565, opened 2023-11-24
Closes #557, in-place implementation of expval for LightningQubit

### LINKED ISSUE #557: Expval kernels for LightningQubit
### Important Note

⚠️ This issue is part of an internal assignment and not meant for external contributors.

### Details

The expectation value of a quantum gate $U$ for a circuit in the state $| \psi \rangle$ is $\langle \psi | U | \psi \rangle$. In PennyLane-Lightning this is calculated by first applying $U$ to $| \psi \rangle$ to get $| \psi\' \rangle$ and then take the inner product with $| \psi \rangle$. That is not optimal since one needs to allocate a copy of $| \psi \rangle$.

### Implementation

Modify the `expval` method in [MeasurementsLQubit.hpp](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp) (only the one with a `const std::string &operation` argument) to mimic the Lightning-Kokkos [implementation](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp), i.e. write specialized kernels that do not rely on `applyMatrix/applyOperation+innerProdC` but perform all operations at once. That will be done in a loop over the state vector entries (have a look at the PauliX functor in [ExpValFunctors.hpp](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/ExpValFunctors.hpp#L65) for example). That loop is embarrassingly parallelizable and should be parallelized as such.

### Requirements

A PR must satisfy the following:

- Write `expval` kernels for the `Identity`,  `PauliX`,  `PauliY`,  `PauliZ`,  `Hadamard` observables. 
- Parallelize the kernels with C++ threads or OpenMP.
- Validate the implementation with C++ & Python tests (you may reuse existing tests, but add some if coverage requires it).
- Pass clang-tidy, black and pylint checks.
- Include a changelog entry under "Improvements".
- Gather benchmark data using the Python interface of PennyLane (old versus new implementation & time versus number of threads for different qubits count) and discuss the performance trends. Put the benchmarking scripts in a Github gist and share the gist in the PR comments.

sbohloul:
**Context:**
This PR modifies the expval method for named operations using an in-place. OpenMP parallel, implementation to avoid creating an extra intemediate statevector.

The implementation closely follows that of [MeasurementsKokkos.hpp)](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp). It adds the corresponding functors in `ExpValFunctorsLQubit.hpp` and modifies `MeasurementsLQubit.hpp` accordingly. It uses the existing tests of `testNamedObsExpval` in `Test_MeasurementsBase.cpp` to validate the changes. 

**Benefits:**
- Avoiding creating an extra intermediate statevector
- An order of magnitude speedup in calculating the expected value of `PauliX`, `PauliY`, `PauliZ`, `Hadamard`, and `Identity` operators

**Possible Drawbacks:**

**Related GitHub Issues:**
Closes #557

### DIFF
--- a/pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
@@ -14,22 +14,22 @@
 #pragma once
 
 #include "BitUtil.hpp"
+#include "Util.hpp"
 
 /// @cond DEV
 namespace {
 using namespace Pennylane::Util;
-using Pennylane::Util::INVSQRT2;
 } // namespace
 /// @endcond
 
 namespace Pennylane::LightningQubit::Functors {
 
 template <class PrecisionT> struct getExpectationValueIdentityFunctor {
-    std::complex<PrecisionT> *arr;
+    const std::complex<PrecisionT> *arr;
     size_t num_qubits;
 
     getExpectationValueIdentityFunctor(
-        std::complex<PrecisionT> *arr_,
+        const std::complex<PrecisionT> *arr_,
         [[maybe_unused]] std::size_t num_qubits_,
         [[maybe_unused]] const std::vector<size_t> &wires) {
         arr = arr_;
@@ -49,7 +49,7 @@ template <class PrecisionT> struct getExpectationValueIdentityFunctor {
 };
 
 template <class PrecisionT> struct getExpectationValuePauliXFunctor {
-    std::complex<PrecisionT> *arr;
+    const std::complex<PrecisionT> *arr;
     size_t num_qubits;
 
     size_t rev_wire;
@@ -57,7 +57,7 @@ template <class PrecisionT> struct getExpectationValuePauliXFunctor {
     size_t wire_parity;
     size_t wire_parity_inv;
 
-    getExpectationValuePauliXFunctor(std::complex<PrecisionT> *arr_,
+    getExpectationValuePauliXFunctor(const std::complex<PrecisionT> *arr_,
                                      std::size_t num_qubits_,
                                      const std::vector<size_t> &wires) {
         arr = arr_;
@@ -69,7 +69,9 @@ template <class PrecisionT> struct getExpectationValuePauliXFunctor {
     }
 
     inline void operator()(PrecisionT &expval) const {
-        size_t k, i0, i1;
+        size_t k;
+        size_t i0;
+        size_t i1;
 #if defined(_OPENMP)
 #pragma omp parallel for default(none)                                         \
     shared(num_qubits, wire_parity_inv, wire_parity, rev_wire_shift,           \
@@ -86,7 +88,7 @@ template <class PrecisionT> struct getExpectationValuePauliXFunctor {
 };
 
 template <class PrecisionT> struct getExpectationValuePauliYFunctor {
-    std::complex<PrecisionT> *arr;
+    const std::complex<PrecisionT> *arr;
     size_t num_qubits;
 
     size_t rev_wire;
@@ -94,7 +96,7 @@ template <class PrecisionT> struct getExpectationValuePauliYFunctor {
     size_t wire_parity;
     size_t wire_parity_inv;
 
-    getExpectationValuePauliYFunctor(std::complex<PrecisionT> *arr_,
+    getExpectationValuePauliYFunctor(const std::complex<PrecisionT> *arr_,
                                      std::size_t num_qubits_,
                                      const std::vector<size_t> &wires) {
         arr = arr_;
@@ -106,8 +108,11 @@ template <class PrecisionT> struct getExpectationValuePauliYFunctor {
     }
 
     inline void operator()(PrecisionT &expval) const {
-        size_t k, i0, i1;
-        std::complex<PrecisionT> v0, v1;
+        size_t k;
+        size_t i0;
+        size_t i1;
+        std::complex<PrecisionT> v0;
+        std::complex<PrecisionT> v1;
 #if defined(_OPENMP)
 #pragma omp parallel for default(none)                                         \
     shared(num_qubits, wire_parity_inv, wire_parity, rev_wire_shift,           \
@@ -128,7 +133,7 @@ template <class PrecisionT> struct getExpectationValuePauliYFunctor {
 };
 
 template <class PrecisionT> struct getExpectationValuePauliZFunctor {
-    std::complex<PrecisionT> *arr;
+    const std::complex<PrecisionT> *arr;
     size_t num_qubits;
 
     size_t rev_wire;
@@ -136,7 +141,7 @@ template <class PrecisionT> struct getExpectationValuePauliZFunctor {
     size_t wire_parity;
     size_t wire_parity_inv;
 
-    getExpectationValuePauliZFunctor(std::complex<PrecisionT> *arr_,
+    getExpectationValuePauliZFunctor(const std::complex<PrecisionT> *arr_,
                                      std::size_t num_qubits_,
                                      const std::vector<size_t> &wires) {
         arr = arr_;
@@ -148,7 +153,9 @@ template <class PrecisionT> struct getExpectationValuePauliZFunctor {
     }
 
     inline void operator()(PrecisionT &expval) const {
-        size_t k, i0, i1;
+        size_t k;
+        size_t i0;
+        size_t i1;
 #if defined(_OPENMP)
 #pragma omp parallel for default(none)                                         \
     shared(num_qubits, wire_parity_inv, wire_parity, rev_wire_shift,           \
@@ -166,7 +173,7 @@ template <class PrecisionT> struct getExpectationValuePauliZFunctor {
 };
 
 template <class PrecisionT> struct getExpectationValueHadamardFunctor {
-    std::complex<PrecisionT> *arr;
+    const std::complex<PrecisionT> *arr;
     size_t num_qubits;
 
     size_t rev_wire;
@@ -174,9 +181,9 @@ template <class PrecisionT> struct getExpectationValueHadamardFunctor {
     size_t wire_parity;
     size_t wire_parity_inv;
 
-    constexpr static auto isqrt2 = INVSQRT2<PrecisionT>();
+    PrecisionT isqrt2 = INVSQRT2<PrecisionT>();
 
-    getExpectationValueHadamardFunctor(std::complex<PrecisionT> *arr_,
+    getExpectationValueHadamardFunctor(const std::complex<PrecisionT> *arr_,
                                        std::size_t num_qubits_,
                                        const std::vector<size_t> &wires) {
         arr = arr_;
@@ -188,8 +195,11 @@ template <class PrecisionT> struct getExpectationValueHadamardFunctor {
     }
 
     inline void operator()(PrecisionT &expval) const {
-        size_t k, i0, i1;
-        std::complex<PrecisionT> v0, v1;
+        size_t k;
+        size_t i0;
+        size_t i1;
+        std::complex<PrecisionT> v0;
+        std::complex<PrecisionT> v1;
 #if defined(_OPENMP)
 #pragma omp parallel for default(none)                                         \
     shared(num_qubits, wire_parity_inv, wire_parity, rev_wire_shift,           \
--- a/pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
@@ -1,0 +1,210 @@
+// Copyright 2018-2023 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the License);
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an AS IS BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+#include "BitUtil.hpp"
+
+/// @cond DEV
+namespace {
+using namespace Pennylane::Util;
+using Pennylane::Util::INVSQRT2;
+} // namespace
+/// @endcond
+
+namespace Pennylane::LightningQubit::Functors {
+
+template <class PrecisionT> struct getExpectationValueIdentityFunctor {
+    std::complex<PrecisionT> *arr;
+    size_t num_qubits;
+
+    getExpectationValueIdentityFunctor(
+        std::complex<PrecisionT> *arr_,
+        [[maybe_unused]] std::size_t num_qubits_,
+        [[maybe_unused]] const std::vector<size_t> &wires) {
+        arr = arr_;
+        num_qubits = num_qubits_;
+    }
+
+    inline void operator()(PrecisionT &expval) const {
+        size_t k;
+#if defined(_OPENMP)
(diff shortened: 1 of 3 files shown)


sbohloul:
@Alex-Preciado Please see [here](https://gist.github.com/sbohloul/332f57e68ccf118f1b854d17b2c07cd3) for the analysis script and artifacts.

Alex-Preciado:
Thank you so much for preparing this PR, @sbohloul ! The team will have a look next week, and we will come back to you with feedback. Let me know if you have any other questions in the meantime.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+/// @cond DEV
+namespace {
+using namespace Pennylane::Util;
+} // namespace
+/// @endcond
+
+namespace Pennylane::LightningQubit::Functors {
+
+template <class PrecisionT> struct getExpectationValueIdentityFunctor {
```
mlxd, line 27: What do you think would be a good docstring for this functor that matches the style of our other classes?

mlxd, line 27: Is there any good reason for this to be a struct or a class?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+    const std::complex<PrecisionT> *arr;
+    size_t num_qubits;
```
mlxd, line 29: What happens if we decide to modify these members directly with the created object? What would the potential issues be? Is there any way we can be more safe with these?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    getExpectationValueIdentityFunctor(
+        const std::complex<PrecisionT> *arr_,
+        [[maybe_unused]] std::size_t num_qubits_,
+        [[maybe_unused]] const std::vector<size_t> &wires) {
+        arr = arr_;
+        num_qubits = num_qubits_;
```
mlxd, line 36: Should we initialize these items in the constructor body? Are there any other ways to initialize these? What are the advantages of other ways?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    inline void operator()(PrecisionT &expval) const {
```
mlxd, line 39: Should we generate the `expval` result here by modification through reference? Is there another way we could define the functor operator that doesn't use this approach? What are the advantages for either way.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+        size_t k;
+#if defined(_OPENMP)
+#pragma omp parallel for default(none) shared(num_qubits, arr) private(k)      \
```
mlxd, line 42: What will happen with these OpenMP threads if we have multiple concurrent executions happening within the same process?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,219 @@
...
+template <class PrecisionT> struct getExpectationValueHadamardFunctor {
+    const std::complex<PrecisionT> *arr;
+    size_t num_qubits;
+
+    size_t rev_wire;
+    size_t rev_wire_shift;
+    size_t wire_parity;
+    size_t wire_parity_inv;
+
+    PrecisionT isqrt2 = INVSQRT2<PrecisionT>();
```
mlxd, line 183: since this value is invariant across all uses, is there any way we can make use of this knowledge to improve the safety of how we define it?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -661,5 +714,21 @@ class Measurements final
         }
         return init_idx;
     }
+
+    std::unordered_map<std::string, ExpValFunc> expval_funcs_;
```
mlxd, line 718: Is this the best place to define this class member?

Any thoughts on how we could make this `const`, but also populate it with the entries of `init_expval_funcs_` ?

mlxd (commented):
Thanks for the submission @sbohloul 

As suggested by @AmintorDusko it would be great to see some additional test coverage as reported by the CI.

Also, I have some questions based on the implementations:

sbohloul:
> Hi @sbohloul, thank you for your contribution. The Codecov report shows that some areas of your code are not been touched by the test suite. Could you please expand tests, so you can improve code coverage?

@AmintorDusko Thank you for your feedback. I expanded the coverage as suggested.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: Since `struct` default to public access privilege, both `*arr` and `num_qubits` can be modified through instantiated object (although values that `*arr` points to are protected via the `const` keyword). This could lead to unwanted modification of these variables so that the functor returns a wrong output. One way to improve this, is to explicitly declare these variables as `private` within the definition of the functor. I have modified the code accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: These items can also be initialized using constructor initializer list. Some of the advantages of using initializer list are:

- Member variables are initialized in the order they are declared in the `struct`, not the order they appear in the initializer list. This helps avoid potential issues related to the order of initialization.
- With initializer list the variables are only initialized once, directly with their specified values. In contrast, with assigning values in the constructor body, the members are first default-initialized and then assigned, leading to potentially unnecessary work.
- It is the only way to initialize the constant data members (not applicable to this case)

I have modified the code accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: We could modify the `operator` to generate and return `expval` by value. Since the operator returns a single value that is limited to its scope, return by value could be more inline with C++ coding standards. Here are some of pros and cons of each approach

*Modify through reference*

Pros:
- Directly modifies the variable, no need to create a copy

Cons:
- Not safe in terms of unintended modifications
- Not transparent regarding the initialization of the variable (expval could unintentionally be initialized with nonzero value)

*Return by value*

Pros:
- Explicitly returns the computed result, making it clear what the function does.
- Safer in terms of avoiding unintended modifications of external variables

Cons:
- Requires declaring a separate variable to store the result.

I have modified the code accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: I assume your point is about nested parallelism? In that case, If this region is an outer parallel region, the region is executed by creating `OMP_NUM_THREADS` threads. If this is a parallel region within an outer parallel region, with nested OpenMP enabled each thread launches `OMP_NUM_THREADS` to execute the loop, otherwise the loop is executed serially by one of the threads. This behavior can be controlled and made efficient with respect to available hardware resources by using OpenMP functions such as `omp_set_nested()` and `omp_get_max_threads()`, and explicitly setting `num_threads` for the parallel region. Please advise if this is what you mean, so I could modify the implmentaiton accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: We can consider making it a `constexpr static` variable which allows the value to be computed only once at the compile time, which also provides some benefits such as safety and performance. I have modified the code accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
sbohloul: With respect to location of this member within this class, perhaps it would be better to define it before the (private) member functions. Unless your point is about moving its definition to its parent class `MeasurementBase`? `const` keyword can be used to make this member constant, which then can only be initialized using constructor initializer list. In this case, `init_expval_funcs_` can be removed, and its body can be used directly in the initializer list. I have modified the code accordingly.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -581,6 +633,8 @@ class Measurements final
     }
 
   private:
+    const std::unordered_map<std::string, ExpValFunc> expval_funcs_;
```
AmintorDusko, line 636: Why not have this variable definition together with the others? Or even better, what do you think about merging the two private sections?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/tests/Test_MeasurementsLQubit.cpp
```diff
@@ -114,6 +119,11 @@ TEMPLATE_PRODUCT_TEST_CASE("Expected Values", "[Measurements]",
         exp_values = Measurer.expval(operations_list, wires_list);
         exp_values_ref = {0.58498357, 0.77015115, 0.91266780};
         REQUIRE_THAT(exp_values, Catch::Approx(exp_values_ref).margin(1e-6));
+
+        operations_list = {"Hadamard", "Hadamard", "Hadamard"};
+        exp_values = Measurer.expval(operations_list, wires_list);
+        exp_values_ref = {0.7620549436, 0.8420840225, 0.8449848566};
+        REQUIRE_THAT(exp_values, Catch::Approx(exp_values_ref).margin(1e-6));
```
AmintorDusko, line 126: How did you calculate these expected values?

AmintorDusko (commented):
Thank you @sbohloul, for the improvement in coverage. I have a few more questions/suggestions.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
```diff
@@ -0,0 +1,315 @@
...
+namespace Pennylane::LightningQubit::Functors {
+/**
+ * @brief Functor for computing expectation value of Identity operator.
+ *
+ * This functor provides an in-place implementation for computing the
+ * expectation value of Identity operator for a given statevector.
+ *
+ * @tparam PrecisionT precision data type of the statevector.
+ */
+template <class PrecisionT> struct getExpectationValueIdentityFunctor {
```
mlxd, line 41:
```suggestion
template <class PrecisionT> struct getExpectationValueIdentityFunctor {
```

Since we are now defining private members, we should make this a class :)

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
mlxd: Thanks @sbohloul No need to modify, just curious about your thoughts here.

mlxd (commented):
Thanks a lot @sbohloul 
Some final comments from me on this. Great work on the implementation.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/ExpValFunctorsLQubit.hpp
sbohloul: I just changed the `struct` to `class`.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
sbohloul: I moved the variable definition to the first private section. Regarding having two private sections, my intention was to minimize the modifications to the existing code. I found other examples with more than one `public` or `privat` sections in the codebase, such as [StateVectorLQubitManaged.hpp](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp), so I thought this might be part of coding style.

sbohloul:
@Alex-Preciado I was wondering if there are any other requirements to be fulfilled regarding this PR.

mlxd:
> @Alex-Preciado I was wondering if there are any other requirements to be fulfilled regarding this PR.

Thanks for the submission @sbohloul :) Happy to say we can come to this PR at a later date --- nothing left to do here for now. The team will reach out to you shortly about the next steps.

sbohloul:
@mlxd @AmintorDusko Thank you again for your helpful comment to navigate this PR. I would be happy to finalize the work here since we already spent some time on it. Please let me know what are your toughts about this.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -165,17 +206,30 @@ class Measurements final
...
-        ComplexT expected_value = innerProdC(this->_statevector.getData(),
-                                             operator_statevector.getData(),
-                                             this->_statevector.getLength());
-        return std::real(expected_value);
+        // In-place calculation of expval without creating duplicate of the
+        // statevector.
+        switch (expval_funcs_.at(operation)) {
+        case ExpValFunc::Identity:
+            return applyExpValNamedFunctor<getExpectationValueIdentityFunctor,
+                                           0>(wires);
```
vincentmr, line 214: Hi @sbohloul , nice work there. If you have time, I'd like us to update and merge this PR. We should first merge `master` in this branch, update the CHANGELOG and fix conflicts accordingly. Next, I would like us to change the implementation to avoid the use of functors. This is too much boilerplate for host-only code (sometimes required for device-generic code as in Kokkos however). Could you have a look at the [LM kernels](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp#L577)'s `applyNC1`? Could we write a lambda-templated method that accumulates the expval of a generic gate? We could then simply write something like 
```
switch (expval_funcs_.at(operation)) {
[...]
case ExpValFunc::PauliX:
auto core_function = [](std::complex<PrecisionT> *arr,
                                const std::size_t i0, const std::size_t i1) {
            std::swap(arr[i0], arr[i1]);
        };
[...]
}
applyExpVal1<PrecisionT, decltype(core_function)>(wires, core_function);
```

PR #570, opened 2023-12-06
Refactor obs base and safer LQRaw call for  measurement shots

multiphaseCFD:
### Before submitting

**Context:**

Shot-noise related methods now accommodate observable objects with arbitrary eigenvalues. Add a Kronecker product method for two diagonal matrices. 

Refactor shot-noise related methods of `MeasurementsBase` class in the C++ layer and eigenvalues are not limited to `1` and `-1`. Add `getObs()` method to Observables class. Refactor `applyInPlaceShots` to allow users to get eigenvalues of Observables object. Deprecated `_preprocess_state` method in `MeasurementsBase` class for safer use of the `LightningQubitRaw` backend.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
+++ b/pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
@@ -1253,7 +1253,7 @@ template <typename TypeList> void testHamiltonianObsExpvalShot() {
                 "Hadamard", std::vector<size_t>{1});
             auto obs1 = TensorProdObs<StateVectorT>::create({Y0, H1});
 
-            auto obs = Hamiltonian<StateVectorT>::create({1, 1}, {obs0, obs1});
+            auto obs = Hamiltonian<StateVectorT>::create({0.1, 0.3}, {obs0, obs1});
 
             Measurements<StateVectorT> Measurer_analytic(sv);
             auto expected = Measurer_analytic.expval(*obs);
--- a/pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
+++ b/pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
@@ -217,10 +217,12 @@ template <typename TypeList> void testProbabilitiesObs() {
 
         // Defining the Statevector that will be measured.
         auto statevector_data = createNonTrivialState<StateVectorT>();
+        auto sv_data = createNonTrivialState<StateVectorT>();
+
         StateVectorT statevector(statevector_data.data(),
                                  statevector_data.size());
 
-        StateVectorT sv(statevector_data.data(), statevector_data.size());
+        StateVectorT sv(sv_data.data(), sv_data.size());
 
         DYNAMIC_SECTION("Test PauliX"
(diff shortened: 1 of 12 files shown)


pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
```diff
@@ -217,10 +217,12 @@ template <typename TypeList> void testProbabilitiesObs() {
 
         // Defining the Statevector that will be measured.
         auto statevector_data = createNonTrivialState<StateVectorT>();
+        auto sv_data = createNonTrivialState<StateVectorT>();
```
vincentmr, line 220: If the goal is to create the same state, set to `statevector_data` so if the last changes this one will follow along.
```suggestion
        auto sv_data = statevector_data;
```

pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -509,4 +509,27 @@ auto transpose_state_tensor(const std::vector<T> &tensor,
     }
     return transposed_tensor;
 }
+
+/**
+ * @brief Kronecker product of two diagonal matrices. Only diagonal elements are
+ * stored.
+ *
+ * @tparam T Data type.
+ * @param A A diagonal matrix with only diagonal elements stored.
```
vincentmr, line 518:
```suggestion
 * @param diagA A vector containing the values of a diagonal matrix.
```

pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -509,4 +509,27 @@ auto transpose_state_tensor(const std::vector<T> &tensor,
...
+ * @param B A diagonal matrix with only diagonal elements stored.
+ * @return result Result matrix with only diagonal elements stored.
```
vincentmr, line 520:
```suggestion
 * @return kronAB A vector containing the diagonal values of the Kronecker product.
```

vincentmr (approved):
I left a few nitpicks, but I don't have much to add. Nice code generalization and reduction. One last thing, could you fill out the PR template for more context please?

pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
multiphaseCFD: Thanks @vincentmr ! The suggested change will work for all backends except for `LQRaw`. With `LQRaw`, both measurement objects will operate on the same data, `statevector_data` , if we set `sv_data = statevector_data`. In this case, I will keep current implementation.

maliasadi (approved):
Thanks Shuli :tada:

pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
vincentmr: It's a nitpick so I'm fine either way. But `createNonTrivialState` is returning a `vector` so `sv_data = statevector_data` should create a copy which is compatible with `LQRaw`, unlike passing `statevector_data.data()` into both SV constructors.

PR #572, opened 2023-12-07
Prevent reallocation of aligned memory

mlxd:
### Before submitting

**Context:** To ensure the memory allocated by Lightning is always aligned, and usable by the dispatcher AVX kernels, we must avoid any potential numpy-layer reallocations. This PR extends the aligned allocator functions bound through Pybind11 to allow data to be initialized and set for use. This avoids us needing to perform potential numpy casts, which could potentially reallocate the data. As a result, this removes overheads observed with lightning.qubit workloads depending on the need to handle multiple resets/allocations.

**Benefits:** Removes unneeded overheads from lightning.qubit, and redundant imports from LG/LK

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane_lightning/core/src/utils/Memory.hpp
```diff
@@ -35,23 +36,28 @@ namespace Pennylane::Util {
...
-    return _aligned_malloc(bytes, alignment);
+    p = _aligned_malloc(bytes, alignment);
 #else
-    return std::aligned_alloc(alignment, bytes);
+    p = std::aligned_alloc(alignment, bytes);
 #endif
+    if (zero_init) {
+        std::memset(p, 0, bytes);
+    }
+    return p;
```
maliasadi, line 60: Nice addition, but I cannot find any actual use case of  `zero_init = True` in the codebase? Is there a followup PR?

mlxd: And this is why code-reviews are so important: The following chunk is missing:
```
        def _create_basis_state(self, index):
            """Return a computational basis state over all wires.
            Args:
                index (int): integer representing the computational basis state
            Returns:
                array[complex]: complex array of shape ``[2]*self.num_wires``
                representing the statevector of the basis state
            Note: This function does not support broadcasted inputs yet.
            """
            state = allocate_aligned_array(2**self.num_wires, np.dtype(self.C_DTYPE), True)
            state[index] = 1
            return self._reshape(state, [2] * self.num_wires)
```

I'll add that in.

mlxd: And done.

pennylane_lightning/core/src/bindings/Bindings.hpp
```diff
@@ -172,20 +173,23 @@ auto alignedNumpyArray(CPUMemoryModel memory_model, size_t size) -> py::array {
  * @param size Size of the array to create
  * @param dt Pybind11's datatype object
  */
-auto allocateAlignedArray(size_t size, const py::dtype &dt) -> py::array {
+auto allocateAlignedArray(size_t size, const py::dtype &dt,
```
maliasadi, line 176: What do you think of adding `std::memset` to this method to remove `zeroInit` argument with some default value from the rest of C++ codebase? I assume `zeroInit` is mainly useful in this method and removing it from other places would reduce the number of code changes while keeping the signature of those methods simple.

pennylane_lightning/core/_version.py
```diff
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
```
maliasadi, line 16: Would you update the changelog :)

pennylane_lightning/core/src/bindings/Bindings.hpp
mlxd: Happy to in a follow up PR.

pennylane_lightning/core/_version.py
mlxd: Will do once approved

maliasadi (approved):
Thanks @mlxd! Looks good to go :t-rex:

pennylane_lightning/core/src/bindings/Bindings.hpp
mlxd: TODO added.

requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-pip==23.0
+pip~=22.0
```
multiphaseCFD, line 1: Want to know the reason why we use ~=22.0 here instead of 23.0

multiphaseCFD (approved):
Looks good to me! Thanks @mlxd !

requirements-dev.txt
mlxd: The latest version of the 23.x branch causes issues with pip-installed binaries being available and usable on path. I downgraded the previous major release for now, until we can identify the best path around this.

PR #516, opened 2023-10-11
Support arbitrary controlled operations in lightning.qubit

vincentmr:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**
Applying a gate such as `qml.ControlledQubitUnitary(U, control_wires=range(5), wires=[5])` should be quite fast compared with `qml.QubitUnitary(U, wires=range(6))` because the first requires applying a `2x2` matrix to 1/32th of all state vector entries and the second applying a `64x64` matrix to all entries. This also provides less control over the parametrization of `U`, which can be handy in adjoint-diff implementations.

**Description of the Change:**
Introduce three new kind of gates: controlled gates, controlled generators and controlled matrix ops. The new gates are analogous to the current gates, generators and matrix ops. The new gates accept arbitrary controls, for example `LM::NCMultiQubitOp` takes in `std::vector<std::size_t> &controlled_wires` in addition to the usual input of the other kernels. This extra input required modifying the dispatcher and all related definitions. The adjoint-diff classes also need to carry `controlled_wires` around. All new gate kernels are implemented in `GateImplementationsLM.hpp`. The Python layer also differentiate between `ControlledQubitUnitary` operations and others for now. 

**Benefits:**
- Easier to add gate kernels in LM.
- More gate supported in forward/backward mode.
- Improved multi-controlled ops performance.

**Possible Drawbacks:**
Some gates might be slower, but [benchmarks](https://docs.google.com/presentation/d/174ybszwNyI-YEUS9S6F3f4G5Qc-hMUj2Ww835tlvHA8/edit?usp=sharing) show that this is likely limited. For example, all non-controlled 1-qubit gates have the same performance.  

**Related GitHub Issues:**

PR #579, opened 2023-12-13
Add  GlobalPhase to all devices

vincentmr:
### Before submitting

**Context:**
Prior to #516 , the Lightning devices could not properly run Grover's algorithm (among others) because `GlobalPhase` and `C(GlobalPhase)` are not supported. This would also be true of any circuit which ends up with `GlobalPhase` or `C(GlobalPhase)` after decomposition. In addition, L-GPU's custom gate cache was not passed a parameter to distinguish between gates. This meant, for example, that `MultiControlledX([0, 1, 2], "111")` and `MultiControlledX([0, 2], "00")` were applied as the same operation. This likely affected any circuit with more than one version of 
```
"QubitUnitary",
"ControlledQubitUnitary",
"MultiControlledX",
"DiagonalQubitUnitary",
"PSWAP",
"OrbitalRotation",
```

Add `GlobalPhase` and `C(GlobalPhase)` support in L-GPU and L-Kokkos. For the sake of time, `C(GlobalPhase)` is implemented by generating the diagonal entries with `global_phase_diagonal` at the Python layer, and then passing it through the gate matrix argument down to the C++ layer and adding a special branch for it in `applyOperation`. To fix the gate cache, we pass the gate hash as a parameter when applying any of the above gates to make sure the gate cache adds a new matrix. 

**Benefits:**
Bug fix, can now run Grover with L-GPU and L-Kokkos.

**Possible Drawbacks:**
Higher memory demand in the gate cache.

**Related GitHub Issues:**


vincentmr:
Grover algorithm returns wrong results with [sc-51671]

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -95,6 +106,91 @@ void setStateVector_CUDA_call(GPUDataT *sv, index_type &num_indices,
...
+ * @param num_sv Number of state vector elements.
+ * @param phase Complex data pointer of input values (on device).
+ */
+template <class GPUDataT, class index_type>
+__global__ void globalPhaseStateVectorkernel(GPUDataT *sv, index_type num_sv,
+                                             GPUDataT phase) {
+    const unsigned int i = blockIdx.x * blockDim.x + threadIdx.x;
+    if (i < num_sv) {
+        sv[i] = ComplexMul(sv[i], phase);
+    }
```
mlxd, line 124: This kind of branching may be inefficient for CUDA kernels. Are we sure the `num_sv` value is at least as wide as a CUDA warp so that each SM is run at full capacity?

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -95,6 +106,91 @@ void setStateVector_CUDA_call(GPUDataT *sv, index_type &num_indices,
...
+ */
+template <class GPUDataT, class index_type>
+void globalPhaseStateVector_CUDA_call(GPUDataT *sv, index_type num_sv,
+                                      GPUDataT phase, size_t thread_per_block,
+                                      cudaStream_t stream_id) {
+    auto dv = std::div(static_cast<long>(num_sv), thread_per_block);
+    size_t num_blocks = dv.quot + (dv.rem == 0 ? 0 : 1);
+    const size_t block_per_grid = (num_blocks == 0 ? 1 : num_blocks);
+    dim3 blockSize(thread_per_block, 1, 1);
+    dim3 gridSize(block_per_grid, 1);
```
mlxd, line 144: Since these values are usually inferred from physical limits (as in, the total number of runnable threads, blocks, etc), are we sure the chosen numbers here, and arguments to the functions are sufficient to fit everything for all use-cases using this 1D indexing? Will we need to consider 2/3D blocks if things get larger?

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -95,6 +106,91 @@ void setStateVector_CUDA_call(GPUDataT *sv, index_type &num_indices,
...
+ * @param phase Complex data pointer of controlled global phase values (on
+ * device).
+ */
+template <class GPUDataT, class index_type>
+__global__ void cGlobalPhaseStateVectorkernel(GPUDataT *sv, index_type num_sv,
+                                              GPUDataT *phase) {
+    const unsigned int i = blockIdx.x * blockDim.x + threadIdx.x;
+    if (i < num_sv) {
+        sv[i] = ComplexMul(sv[i], phase[i]);
+    }
```
mlxd, line 166: Same here as earlier: are we sure branching won't happen in the execution of the kernel?

mlxd (commented):
Thanks @vincentmr 
I think we can come back and discuss this in a few weeks

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
vincentmr: Given the perceived urgency at the time, I simply mimicked the other L-GPU kernels without giving it much thought. That said, this is a common pattern in the CUDA [doc](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization). I think it is there to prevent out of bound access when the array is too small or when the array length is not divisible by the number of threads per block. So toward the end of the array you can't fill the SMs if the array is not commensurate with the number of SMs, but otherwise the branch should not prevent using the SMs fully.

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
vincentmr: I think that since the data (and access pattern) is intrinsically 1D, CUDA will in effect linearize any 2D/3D mapping. We could even avoid using `dim3` here (copy-pasting) and elsewhere. The `<<<numBlocks, threadsPerBlock>>>` syntax should be sufficient, but maybe we'd lose the possibility to call this asynchronously with streams?

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
```diff
@@ -35,6 +36,7 @@
 namespace {
 using namespace Pennylane;
 using namespace Pennylane::LightningGPU;
+std::mt19937_64 re{1337};
```
mlxd, line 39: Do we want this defined in the translation unit, or local to the tests it is used in?

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -14,6 +14,23 @@
 #include "cuError.hpp"
 #include <cuComplex.h>
 
+namespace {
+template <typename ComplexT>
+__device__ inline ComplexT ComplexConj(ComplexT a) {
```
mlxd, line 19: Isn't there a CUDA function for this already defined in their API ?

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -14,6 +14,23 @@
...
+    ComplexT b;
+    b.x = a.x;
+    b.y = -a.y;
+    return b;
+}
+template <typename ComplexT>
+__device__ inline ComplexT ComplexMul(ComplexT a, ComplexT b) {
```
mlxd, line 26: Same q for here?

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -140,4 +247,41 @@ void setBasisState_CUDA(cuDoubleComplex *sv, cuDoubleComplex &value,
...
+                                  cudaStream_t stream_id) {
+    if (adjoint) {
+        cGlobalPhaseStateVector_CUDA_call<cuDoubleComplex, size_t, true>(
+            sv, num_sv, phase, thread_per_block, stream_id);
+    } else {
+        cGlobalPhaseStateVector_CUDA_call<cuDoubleComplex, size_t, false>(
+            sv, num_sv, phase, thread_per_block, stream_id);
+    }
+}
+
```
mlxd, line 286: Not sure this file is the best overall for the location of these. Maybe we make an additional story to formalise the CUDA kernels when exploring the additions for the CPU-GPU box.

tests/test_gates.py
```diff
@@ -52,6 +53,7 @@ def op(op_name):
         "ControlledPhaseShift": [qml.ControlledPhaseShift, [], {"phi": 1.777, "wires": [0, 1]}],
         "CPhase": [qml.CPhase, [], {"phi": 1.777, "wires": [0, 1]}],
         "MultiRZ": [qml.MultiRZ, [], {"theta": 0.112, "wires": [0, 1, 2]}],
+        "GlobalPhase": [qml.GlobalPhase, [], {"phi": 0.112, "wires": [0, 1, 2]}],
```
mlxd, line 56: Should we add a controlled version test here too?

mlxd (commented):
Thanks @vincentmr 
Just a few suggestions, and things to follow up on but overall happy to approve for now when the comments are discussed

pennylane_lightning/core/src/simulators/lightning_gpu/utils/cuda_helpers.hpp
```diff
@@ -73,9 +73,10 @@ inline static auto Div(const CFP_t_T &a, const CFP_t_U &b) -> CFP_t_T {
  * @tparam CFP_t Complex data type. Supports std::complex<float>,
  * std::complex<double>, cuFloatComplex, cuDoubleComplex
  * @param a The given complex number
- * @return CFP_t The conjuagted complex number
+ * @return CFP_t The conjugated complex number
  */
-template <class CFP_t> inline static constexpr auto Conj(CFP_t a) -> CFP_t {
+template <class CFP_t>
+__device__ inline static constexpr auto Conj(CFP_t a) -> CFP_t {
```
mlxd, line 79:
```suggestion
__host__ __device__ inline static constexpr auto Conj(CFP_t a) -> CFP_t {
```

Should we also make this host compatible?

pennylane_lightning/core/src/simulators/lightning_gpu/utils/cuda_helpers.hpp
```diff
@@ -84,6 +85,25 @@ template <class CFP_t> inline static constexpr auto Conj(CFP_t a) -> CFP_t {
...
+ * @brief Multiplies two numbers for CXX & CUDA complex types
+ *
+ * @tparam CFP_t Complex data type. Supports std::complex<float>,
+ * std::complex<double>, cuFloatComplex, cuDoubleComplex
+ * @param a Complex number
+ * @param b Complex number
+ * @return CFP_t The multiplication result
+ */
+template <class CFP_t>
+__device__ inline static constexpr auto Cmul(CFP_t a, CFP_t b) -> CFP_t {
```
mlxd, line 98:
```suggestion
__host__ __device__ inline static constexpr auto Cmul(CFP_t a, CFP_t b) -> CFP_t {
```

Same q as above?

mlxd (commented):
Thanks @vincentmr 
Last 2 comments to add. Once the given are addressed, I can approve this. We can backlog the extension for adjoint support

pennylane_lightning/core/src/simulators/lightning_gpu/utils/cuda_helpers.hpp
vincentmr: It seems it is not required at present, but we might as well.

tests/test_gates.py
mlxd: Sounds good to me then. Thanks

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
mlxd: Yes, that exactly. I know this has caused problems in the Python layer in the past.

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaManaged.hpp
```diff
@@ -195,6 +213,44 @@ class StateVectorCudaManaged
                             thread_per_block, stream_id);
     }
 
+    /**
+     * @brief Multiplies the state-vector by a global phase.
+     *
+     * @param adjoint Indicates whether to use adjoint of gate.
+     * @param param Complex phase generator.
+     */
+    template <size_t thread_per_block = 256>
```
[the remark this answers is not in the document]
mlxd: Since this is currently working, we can accept as is, and aim to validate later with different tuning.

vincentmr:
Lightning adjoint diff supports GlobalPhase [sc-56497]

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
vincentmr: Ok, I'll fix this except for GPU-MPI which I'm not testing in this PR.

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
multiphaseCFD: Thanks @mlxd! Good question. Threads number per block is default as `256`, which is always less than the number of allowed threads per block as of `Ampere` card. `256` threads per block was chosen (which is mentioned in the following comment) is because this is the minimal number of threads per block (8 warps) required to hide maximal latency.

pennylane_lightning/core/_serialize.py
```diff
@@ -394,3 +394,18 @@ def get_wires(operation, single_op):
...
+    controls = np.array(controls)
+    control_values = np.array(control_values)
+    ind = np.argsort(controls)
+    controls = controls[ind[-1::-1]]
+    control_values = control_values[ind[-1::-1]]
+    idx = np.arange(2 ** len(wires), dtype=np.int64).reshape([2 for _ in wires])
+    for c, w in zip(control_values, controls):
+        idx = np.take(idx, np.array(int(c)), w)
+    diag[idx.ravel()] = np.exp(-1j * par)
+    return diag
```
AmintorDusko, line 411: This doesn't look like a serialize-related functionality. Let's look for a better place since this file is designed to host the `QuantumScriptSerializer` class.

.readthedocs.yml
```diff
@@ -21,11 +21,13 @@ build:
...
-    - nvidia-cuda-toolkit
+    - wget
   jobs:
     pre_install:
+      - wget https://developer.download.nvidia.com/compute/cuda/12.3.2/local_installers/cuda_12.3.2_545.23.08_linux.run
+      - sh cuda_12.3.2_545.23.08_linux.run --silent --toolkit --toolkitpath=${READTHEDOCS_VIRTUALENV_PATH}/cuda-12.3 || cat /tmp/cuda-installer.log
       - echo "setuptools~=66.0\npip~=22.0" >> ci_build_requirements.txt
-    post_install:
+    post_install: 
+      - rm -rf ./build && export PATH=${READTHEDOCS_VIRTUALENV_PATH}/cuda-12.3/bin${PATH:+:${PATH}} && export LD_LIBRARY_PATH=${READTHEDOCS_VIRTUALENV_PATH}/cuda-12.3/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}} && PL_BACKEND="lightning_gpu" python setup.py build_ext --define="PL_DISABLE_CUDA_SAFETY=1" && PL_BACKEND="lightning_gpu" python setup.py bdist_wheel
```
mlxd, line 31: Nice work on this.

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
```diff
@@ -1578,4 +1579,52 @@ TEMPLATE_TEST_CASE("LightningGPU::applyOperation multiple wires",
...
+
+TEMPLATE_TEST_CASE("StateVectorCudaManaged::applyControlledGlobalPhase",
+                   "[StateVectorCudaManaged_Param]", double) {
+    using ComplexT = StateVectorCudaManaged<TestType>::ComplexT;
+    std::mt19937_64 re{1337};
+    const size_t num_qubits = 3;
+    const bool inverse = GENERATE(false, true);
+    const size_t index = GENERATE(0, 1, 2);
+    // global_phase_diagonal(-np.pi/2, wires=[0, 1, 2], controls=[0, 1],
+    // control_values=[0, 1])
```
AmintorDusko, line 1614:
```suggestion
```

pennylane_lightning/core/src/simulators/lightning_qubit/gates/tests/Test_GateImplementations_Param.cpp
```diff
@@ -2662,4 +2663,54 @@ TEMPLATE_TEST_CASE(
...
+TEMPLATE_TEST_CASE("StateVectorLQubitManaged::applyControlledGlobalPhase",
+                   "[StateVectorLQubitManaged_Param]", double) {
+    using ComplexT = StateVectorLQubitManaged<TestType>::ComplexT;
+    std::mt19937_64 re{1337};
+
+    const TestType pi2 = 1.5707963267948966;
+    const size_t num_qubits = 3;
+    const bool inverse = GENERATE(false, true);
+    // global_phase_diagonal(-np.pi/2, wires=[0, 1, 2], controls=[0, 1],
+    // control_values=[0, 1])
```
AmintorDusko, line 2700:
```suggestion
```

pennylane_lightning/core/src/gates/Gates.hpp
```diff
@@ -120,8 +120,8 @@ static constexpr auto getS() -> std::vector<ComplexT<T>> {
 template <template <typename...> class ComplexT, typename T>
 static constexpr auto getT() -> std::vector<ComplexT<T>> {
     return {ONE<ComplexT, T>(), ZERO<ComplexT, T>(), ZERO<ComplexT, T>(),
-            ConstMult(SQRT2<decltype(ONE<ComplexT, T>().x)>() / 2,
-                      ConstSum(ONE<ComplexT, T>(), -IMAG<ComplexT, T>()))};
+            ConstMult(INVSQRT2<ComplexT, T>(),
+                      ConstSum(ONE<ComplexT, T>(), IMAG<ComplexT, T>()))};
```
mlxd, line 124: Wait, was this incorrect in the previous definition?

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
mlxd, line 1614:
```suggestion
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -522,21 +533,28 @@ def apply_lightning(self, operations):
                 method = getattr(self._gpu_state, name, None)
                 wires = self.wires.indices(ops.wires)
 
-                if method is None:
+                if ops.name == "C(GlobalPhase)":
+                    controls = ops.control_wires
+                    control_values = ops.control_values
+                    param = ops.base.parameters[0]
+                    matrix = global_phase_diagonal(param, self.wires, controls, control_values)
+                    self._gpu_state.apply(name, wires, False, [], matrix)
```
AmintorDusko, line 541: The `global_phase_diagonal` says that it returns only the diagonal of the operator.
Are we only sending the diagonal to be applied? I got confused here.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
```diff
@@ -248,7 +247,7 @@ TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applyRZ",
 
 TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applyPhaseShift",
                    "[StateVectorKokkosManaged_Param]", double) {
-    const bool inverse = GENERATE(true, false);
+    const bool inverse = GENERATE(false, true);
```
mlxd, line 250: Any reason for the order swap?

AmintorDusko (commented):
Hi @vincentmr, I left a few comments.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp
```diff
@@ -1832,6 +1836,51 @@ class GateImplementationsLM : public PauliGenerator<GateImplementationsLM> {
...
+                                      const std::size_t i0,
+                                      const std::size_t i1) {
+            arr[i0] *= phase;
+            arr[i1] *= phase;
+        };
+        std::size_t target{0};
+        if (!controlled_wires.empty()) {
+            for (std::size_t i = 0; i < num_qubits; i++) {
+                if (std::find(controlled_wires.begin(), controlled_wires.end(),
+                              i) == controlled_wires.end()) {
```
mlxd, line 1867: Is calling find in a loop like this ok for the usage we expect?

pennylane_lightning/core/src/utils/TestHelpersWires.hpp
```diff
@@ -53,6 +53,30 @@ inline auto createWires(Pennylane::Gates::GateOperation op, size_t num_qubits)
...
 
+inline auto createWires(Pennylane::Gates::ControlledGateOperation op,
+                        [[maybe_unused]] size_t num_qubits)
+    -> std::vector<size_t> {
+    // TODO: needed when introducing C(MultiRZ)
+    // if (array_has_elem(Pennylane::Gates::Constant::multi_qubit_gates, op)) {
+    //     std::vector<size_t> wires(num_qubits);
+    //     std::iota(wires.begin(), wires.end(), 0);
+    //     return wires;
+    // }
```
mlxd, line 64: Is this block needed?

requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 pip~=22.0
 git+https://github.com/PennyLaneAI/pennylane.git@master
+datetime
```
mlxd, line 3: Is this not a built in library?

tests/test_gates.py
mlxd, line 56: Should we add an out of order wire test too?

mlxd (commented):
Thanks @vincentmr 
Just a few final questions, and will be happy to approve once they are commented on.

pennylane_lightning/core/src/simulators/lightning_gpu/gates/tests/Test_StateVectorCudaManaged_Param.cpp
vincentmr: This is a comment saying what the `phase` diagonal is and how is it generate. I would like to keep it in case we want to change it in the future.

pennylane_lightning/core/_serialize.py
vincentmr: This functions takes in a [pennylane.operation.Operation](https://docs.pennylane.ai/en/stable/code/api/pennylane.operation.Operation.html) and spits out a NumPy array to be consumed by the C++ layer. It is a serialization util in that sense. Where do you see this go otherwise?

pennylane_lightning/core/src/simulators/lightning_qubit/gates/tests/Test_GateImplementations_Param.cpp
vincentmr: [Same](https://github.com/PennyLaneAI/pennylane-lightning/pull/579#discussion_r1491305967).

pennylane_lightning/lightning_gpu/lightning_gpu.py
vincentmr: Yes, the diagonal is passed via the `matrix` argument down to `applyOperation` which uses a special routine when the name of the operation is `C(GlobalPhase)`. I used that hack to get something working quick in L-GPU/Kokkos. The proper way to do things will be to add controlled-kernels as in L-Qubit, after which we can delete `global_phase_diagonal` and the associated logic.

pennylane_lightning/core/src/gates/Gates.hpp
vincentmr: `SQRT2<decltype(ONE<ComplexT, T>().x)>() / 2` is correct but a roundabout way to get `INVSQRT2<ComplexT, T>()`. 
`-IMAG<ComplexT, T>()` was a mistake. During the controlled-gate implementation I found many other such [typos] in that module which isn't tested much it seems.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
vincentmr: Just seemed more natural to iterate from 0 to 1 (casting in my head), or start with the default inverse value.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp
vincentmr: I didn't like that piece of code either, but it should not impact performance in any case. The problem is the following: 

- `controlled_wires` is a small array with length at most `num_qubits - 1`.
- we want a wire, any wire, that's not in `controlled_wires` and set the wire for the kernel to that.

Can you think of a nice 1-/2-liner to get this on top of your head?

pennylane_lightning/core/src/utils/TestHelpersWires.hpp
vincentmr: This is preparation for #614 . I would like to uncomment it once that gets merged.

requirements-dev.txt
vincentmr: Yes, but I was getting an error related to `datetime` in the CI so I tried upgrading it. Probably one of those CI flukes, let me try to remove it now.

tests/test_gates.py
vincentmr: I could, but then should we do it for all the other gates? This is probably the least sensitive gate since the `wires` argument is actually unused in the C++ kernels.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp
mlxd: Based on that, I think for now I think this is fine. If the order was sorted, it'd be faster to do bin-search, but while this isn't a bottleneck let's go for it.

requirements-dev.txt
mlxd: Weird. Fine to keep if needed for the CI.

tests/test_gates.py
mlxd: Let's look at this in a follow-up improvement of the test suite.

mlxd (approved):

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
```diff
@@ -140,4 +236,41 @@ void setBasisState_CUDA(cuDoubleComplex *sv, cuDoubleComplex &value,
...
+                                  size_t thread_per_block,
+                                  cudaStream_t stream_id) {
+    if (adjoint) {
+        cGlobalPhaseStateVector_CUDA_call<cuDoubleComplex, size_t, true>(
+            sv, num_sv, phase, thread_per_block, stream_id);
+    } else {
+        cGlobalPhaseStateVector_CUDA_call<cuDoubleComplex, size_t, false>(
+            sv, num_sv, phase, thread_per_block, stream_id);
+    }
+}
```
AmintorDusko, line 274: Can we do that?
```suggestion
void cGlobalPhaseStateVector_CUDA(cuComplex *sv, size_t num_sv, const bool adjoint,
                                  cuComplex *phase, size_t thread_per_block,
                                  cudaStream_t stream_id) {
    cGlobalPhaseStateVector_CUDA_call<cuComplex, size_t, adjoint>(
            sv, num_sv, phase, thread_per_block, stream_id);
}
void cGlobalPhaseStateVector_CUDA(cuDoubleComplex *sv, size_t num_sv,
                                  const bool adjoint, cuDoubleComplex *phase,
                                  size_t thread_per_block,
                                  cudaStream_t stream_id) {
    cGlobalPhaseStateVector_CUDA_call<cuDoubleComplex, size_t, adjoint>(
            sv, num_sv, phase, thread_per_block, stream_id);
}
```

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -680,6 +705,31 @@ class StateVectorKokkos final
...
+                Kokkos::RangePolicy<KokkosExecSpace>(0, exp2(num_qubits)),
+                globalPhaseFunctor<fp_t, false>(*data_, num_qubits, wires,
+                                                params));
+        } else {
+            Kokkos::parallel_for(
+                Kokkos::RangePolicy<KokkosExecSpace>(0, exp2(num_qubits)),
+                globalPhaseFunctor<fp_t, true>(*data_, num_qubits, wires,
+                                               params));
+        }
+    }
```
AmintorDusko, line 731:
```suggestion
    void applyGlobalPhase(const std::vector<size_t> &wires,
                          const bool inverse = false,
                          const std::vector<fp_t> &params = {}) {
        auto &&num_qubits = this->getNumQubits();

       Kokkos::parallel_for(
                Kokkos::RangePolicy<KokkosExecSpace>(0, exp2(num_qubits)),
                globalPhaseFunctor<fp_t, inverse>(*data_, num_qubits, wires,
                                               params));
    }
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
AmintorDusko: OK

pennylane_lightning/core/src/simulators/lightning_gpu/initSV.cu
vincentmr: No because `adjoint` needs to be a comptime constant.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
vincentmr: Same, we can't use a variable (even `const`) in a template parameter.

pennylane_lightning/core/_serialize.py
vincentmr: I would leave it here for the time being, but I don't mind moving it to `lightning_base.py` if you prefer. But I don't think we should create a module solely for that.

AmintorDusko: OK. We can come back to this in the future.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
AmintorDusko: Are you sure? Because I tested it with a std::array and it worked. :laughing:

AmintorDusko: Anyway, you don't need to spend time on that, we can leave it like this for now.

vincentmr: This does not work on my side, but you're right std::array templates are somehow more flexible.

PR #581, opened 2023-12-18
Update project URL in setup.py

dime10:
Should the project URL (shown by PyPI for example) point to the PennyLaneAI repo ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -194,7 +194,7 @@ info = {
     "version": version,
     "maintainer": "Xanadu Inc.",
     "maintainer_email": "software@xanadu.ai",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -194,7 +194,7 @@ def build_extension(self, ext: CMakeExtension):
     "version": version,
     "maintainer": "Xanadu Inc.",
     "maintainer_email": "software@xanadu.ai",
-    "url": "https://github.com/XanaduAI/pennylane-lightning",
+    "url": "https://github.com/PennyLaneAI/pennylane-lightning",
```
mlxd, line 197: Looks like this wasn't updated for a loooong time. Good catch

mlxd (approved):

PR #582, opened 2023-12-20
Fix the issue with assigning kernels to ops before registering kernels on macOS

maliasadi:
The addition of arbitrary controlled operations in `lightning.qubit` (PR #516) breaks the use of the `lightning.qubit` C++ API in Catalyst on macOS. This PR fixes the following issue:

``` console
libc++abi: terminating due to uncaught exception of type Pennylane::Util::LightningException: [./catalyst/runtime/build/_deps/pennylane_lightning-src/pennylane_lightning/core/src/simulators/lightning_qubit/gates/KernelMap.hpp][Line:326][Method:assignKernelForOp]: Error in PennyLane Lightning: The given kernel is not registered.
Fatal Python error: Aborted
```

Fixes https://github.com/PennyLaneAI/pennylane-lightning/issues/583

[sc-52500]

### DIFF
--- a/pennylane_lightning/core/src/simulators/lightning_qubit/gates/DynamicDispatcher.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_qubit/gates/DynamicDispatcher.hpp
@@ -105,11 +105,8 @@ template <typename PrecisionT> class DynamicDispatcher {
 
   private:
     std::unordered_map<std::string, GateOperation> str_to_gates_{};
-    std::unordered_map<std::string, ControlledGateOperation>
-        str_to_controlled_gates_{};
+
     std::unordered_map<std::string, GeneratorOperation> str_to_gntrs_{};
-    std::unordered_map<std::string, ControlledGeneratorOperation>
-        str_to_controlled_gntrs_{};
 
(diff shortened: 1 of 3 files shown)


pennylane_lightning/core/src/simulators/lightning_qubit/gates/DynamicDispatcher.hpp
```diff
@@ -122,6 +119,16 @@ template <typename PrecisionT> class DynamicDispatcher {
                        PairHash>
         matrix_kernels_{};
 
+    std::unordered_map<KernelType, std::string> kernel_names_{};
+
+    // Defining controlled gates before `kernel_names_` triggers
+    // `assignKernelForOp` before registering kernels at run time.
```
mlxd, line 125: Good find on this. I'd have anticipated clang-format to have notified us on this, but I guess since it is runtime that may  have been hard to reason.

Also, we can avoid keeping these comments here, as they'll be attached to the commit message and PR, so feel free to remove them.

mlxd (approved):
Thanks for the find @maliasadi

multiphaseCFD (approved):
Looks good to me! Thanks @maliasadi !

PR #599, opened 2024-01-23
Add support for `BlockEncode` gate to lightning devices

dime10:
A simple change to add support for BlockEncode to lightning devices via matrix application.

[sc-54971]


maliasadi (approved):
:tada:

AmintorDusko (approved):
Thank you for that.

PR #602, opened 2024-01-26
[RFC] Improve handling of nested control structures

dime10:
Looking for feedback on the following:

After merging #599, Lightning obtained support for the `BlockEncode` operator. However, even though Lightning in principle supports arbitrary control wires on all supported operators, it didn't automatically gain support for `C(BlockEncode)`. The reason is that this is treated as a separate gate altogether, and many operators are additionally added in their controlled form to the list of supported ops, which seems redundant. To make matters worse, operations of the form `C(C(...))` are still not supported.

The following patch is just one way we could handle controlled operations in a more generic fashion. `qml.simplify` is applied to any `Controlled` instance before processing it in order to flatten any nested control structures. Additionally, the `supports_operation` method is overridden to strip any `C(` prefixes from gate names to determine support status.

Note that nested control structures can easily arise when users write functions that call other functions while adding `qml.ctrl` into the mix.

Benefits:
- no redundant `C(...)` style specifications in the list of operations
- much faster processing of certain operations:
  For instance, `C(C(SWAP))` would previously need to be decomposed into many gates in order to simulate it, when it could just be applied in a single step. See the following benchmark for how this affects simulation time:
  Screenshot 2024-01-25 at 7 09 38 PM
  -> more than 10x improvement
- supported gates that cannot be decomposed (such as `BlockEncode`) are automatically also supported in their controlled version

[sc-55549]

### DIFF
--- a/pennylane_lightning/lightning_qubit/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit/lightning_qubit.py
@@ -117,27 +117,6 @@ if LQ_CPP_BINARY_AVAILABLE:
         "CRX",
         "CRY",
         "CRZ",
-        "C(PauliX)",
-        "C(PauliY)",
-        "C(PauliZ)",
-        "C(Hadamard)",
-        "C(S)",
-        "C(T)",
-        "C(PhaseShift)",
-        "C(RX)",
-        "C(RY)",
-        "C(RZ)",
-        "C(SWAP)",
-        "C(IsingXX)",
-        "C(IsingXY)",
-        "C(IsingYY)",
-        "C(IsingZZ)",
-        "C(SingleExcitation)",
-        "C(SingleExcitationMinus)",
-        "C(SingleExcitationPlus)",
-        "C(DoubleExcitation)",
-        "C(DoubleExcitationMinus)",
-        "C(DoubleExcitationPlus)",
         "CRot",
         "IsingXX",
         "IsingYY",
@@ -367,20 +346,36 @@ if LQ_CPP_BINARY_AVAILABLE:
             Returns:
                 array[complex]: the output state tensor
             """
-            basename = "PauliX" if operation.name == "MultiControlledX" else operation.base.name
-            if basename == "Identity":
-                return
-            method = getattr(sim, f"{basename}", None)
             control_wires = self.wires.indices(operation.control_wires)
-            control_values = (
-                [bool(int(i)) for i in operation.hyperparameters["control_values"]]
-                if operation.name == "MultiControlledX"
-                else operation.control_values
-            )
+
             if operation.name == "MultiControlledX":
+                basename = "PauliX"
+                control_values = [bool(int(i)) for i in operation.hyperparameters["control_values"]]
                 target_wires = list(set(self.wires.indices(operation.wires)) - set(control_wires))
             else:
+                basename = operation.base.name
+                control_values = operation.control_values
                 target_wires = self.wires.indices(operation.target_wires)
+
+            if basename == "Identity":
+                return
+            # This special handling is required because PennyLane doesn't canonicalize
+            # C(CNOT/Toffoli) in qml.simplify or qml.ctrl/Controlled, and we now allow `C(...)`
+            # instances through for any supported base gate. However, the C++ simulator `method`
+            # for CNOT/Toffoli does not accept additional control wires.
+            elif basename == "CNOT":
+                basename = "PauliX"
+                control_values += [True]
+                control_wires += operation.base.wires[0]
+                target_wires = [operation.base.wires[1]]
+            elif basename == "Toffoli":
+                basename = "PauliX"
+                control_values += [True, True]
+                control_wires += operation.base.wires[0:2]
+                target_wires = [operation.base.wires[2]]
+
+            method = getattr(sim, f"{basename}", None)
+
             if method is not None:  # apply n-controlled specialized gate
                 inv = False
                 param = operation.parameters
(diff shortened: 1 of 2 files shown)


vincentmr:
I really like the direction this is taking :) Also, wouldn't it be nice if, for example, CNOT, Toffoli, MultiControlledX were just aliases for C(PauliX)? This way we would not need special treatment for all these gates.

dime10:
> Also, wouldn't it be nice if, for example, CNOT, Toffoli, MultiControlledX were just aliases for C(PauliX)? This way we would not need special treatment for all these gates.

100%, luckily I think @astralcai is already working on ensuring all controlled-x type operations have a canonical representation.

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -402,6 +397,15 @@ def _apply_lightning_controlled(self, sim, operation):
...
                     )
 
+        def supports_operation(self, operation):
+            """Overwrite base class implementation to allow arbitrarily nested Controlled instances
+            to be applied by Lightning."""
+
+            while operation[0:2] == "C(":
+                operation = operation[2:-1]
+
+            return super().supports_operation(operation)
```
grwlf, line 407: @dime10 could you please clarify how e.g. `C(CX)` gate would be handled? From my understanding, Lighting does support `CX` but does not natively support the controlled version of this gate.

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -117,27 +117,6 @@
         "CRX",
         "CRY",
         "CRZ",
-        "C(PauliX)",
```
grwlf, line 120: (This is more a question, I am not asking for immediate changes in this PR)

The set of supported gates (in Lighthing at least) is defined [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/00aaa927d73d6eab6813c3cb12ad1755ac7641fa/pennylane_lightning/core/src/gates/GateOperation.hpp#L27) and in C++ (and I would say that the complete spec would also include threading model and the number of qubits). Is the Python list below - a duplicate of these C++ enums? 
Could we use the device API to read the supported gates somehow? Maybe, we can autogenerate Python sources from the C++ ones if we want to have this information in compile time without running C funcitons?

pennylane_lightning/lightning_qubit/lightning_qubit.py
vincentmr: It seems PennyLane will spit out `C(CX)` as a `Toffoli`, if it has a single control, and `MultiControlledX` otherwise. This is handled by the Python module and the C++ layer. The latter treats CNOT and Toffoli gates separately, but all `MultiControlledX` are understood as `C(PauliX)`, which are natively supported.

pennylane_lightning/lightning_qubit/lightning_qubit.py
vincentmr: The problem is we do not have a tight and robust mechanism to communicate natively supported gates. The `enum`s are just that, `enum`s. So they are not lists of the supported gates, just gates that might be supported on some device. What dictates which gates are decomposed are the `allowed_operations` dictionaries, but these are manually maintained and might not include gates that are natively supported (usually goes uncaught), and vice versa (usually fails some test). Some gates like `OrbitalRotation` are listed in `allowed_operations` but are not natively supported and do not crash any test. This is because gates that do not have a specialized implementation are applied via matrices. In the case of `OrbitalRotation`, this is fine because the number of wire is fixed at 4, a reasonable number. For other gates however, like `QFT`, this can cause issues since the number of wire is unlimited, possibly requiring the generation of a large dense matrix at the Python layer (filling out the memory).

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -367,20 +346,36 @@ def _apply_lightning_controlled(self, sim, operation):
...
-            control_values = (
-                [bool(int(i)) for i in operation.hyperparameters["control_values"]]
-                if operation.name == "MultiControlledX"
-                else operation.control_values
-            )
+
             if operation.name == "MultiControlledX":
+                basename = "PauliX"
+                control_values = [bool(int(i)) for i in operation.hyperparameters["control_values"]]
                 target_wires = list(set(self.wires.indices(operation.wires)) - set(control_wires))
```
astralcai, line 354: This special case is not necessary anymore. The `MultiControlledX` now inherits from `Controlled`, you can access its base like `operation.base`, its control values like `operation.control_values` just like you would with any other operator.

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -367,20 +346,36 @@ def _apply_lightning_controlled(self, sim, operation):
...
+            elif basename == "CNOT":
+                basename = "PauliX"
+                control_values += [True]
+                control_wires += operation.base.wires[0]
+                target_wires = [operation.base.wires[1]]
+            elif basename == "Toffoli":
+                basename = "PauliX"
+                control_values += [True, True]
+                control_wires += operation.base.wires[0:2]
+                target_wires = [operation.base.wires[2]]
```
astralcai, line 375: I believe after my PR, if you call `operation.simplify()`, it would flatten any nested controlled structures to a multi-controlled structure. You can test it out and let me know if anything doesn't work. This would allow you to get rid of these special treatment.

pennylane_lightning/lightning_qubit/lightning_qubit.py
astralcai: A slight caveat here: a `qml.ctrl(op, wire)` where `op` is a `CNOT` will spit out `Toffoli`, but if the operation is instantiated using the constructor of `Controlled` like `Controlled(op, wire)`, then it doesn't magically becomes a `Toffoli`, in which case you will need to call `op.simplify()` to flatten the nested control structure.

pennylane_lightning/lightning_qubit/lightning_qubit.py
astralcai:
```
>>> import pennylane as qml
>>> from pennylane.ops import Controlled
>>> op = Controlled(qml.CNOT([0,1],2))
>>> op.simplify()
Toffoli(wires=[2, 0, 1])
```

PR #601, opened 2024-01-25
Update lightning qubit memory management

AmintorDusko:
**Context:** Update Lightning qubit memory management to LightningQubitManaged.

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-53796]

PR #605, opened 2024-01-31
do not use measurementprocess.name

timmysilv:
**Context:**
`MeasurementProcess.name` has been deprecated, and we should avoid using it. Unfortunately, sometimes they disguise as observables. This is one of those times, and it [makes PennyLane CI fail](https://github.com/PennyLaneAI/pennylane/actions/runs/7729318298/job/21072294868?pr=5122) when we enable deprecation-warnings-as-errors

Just use `isinstance` to not check the name in case it's not an observable.

**Benefits:**
- No more deprecation warnings!
- my PR is unblocked 😄 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
PennyLaneAI/pennylane#5122, recently introduced in #601

### DIFF
--- a/pennylane_lightning/lightning_qubit/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit/lightning_qubit.py
@@ -625,7 +625,7 @@ if LQ_CPP_BINARY_AVAILABLE:
         # pylint: disable=attribute-defined-outside-init
         def sample(self, observable, shot_range=None, bin_size=None, counts=False):
             """Return samples of an observable."""
-            if isinstance(observable, qml.PauliZ):
+            if not isinstance(observable, qml.PauliZ):
                 self.apply_lightning(observable.diagonalizing_gates())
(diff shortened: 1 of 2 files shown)


pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -625,7 +625,7 @@ def probability_lightning(self, wires):
         # pylint: disable=attribute-defined-outside-init
         def sample(self, observable, shot_range=None, bin_size=None, counts=False):
             """Return samples of an observable."""
-            if observable.name != "PauliZ":
+            if isinstance(observable, qml.PauliZ):
```
timmysilv, line 628:
```suggestion
            if not isinstance(observable, qml.PauliZ):
```

vincentmr (approved):
Looks good to me, happy to approve & merge once the CI goes green!

PR #606, opened 2024-02-02
Add build and test support for CUDA 12

mlxd:
### Before submitting

**Context:** This PR adds build and test support for CUDA 12 with `lightning.gpu`.

**Description of the Change:** Updates the build tooling to use CUDA 12

**Benefits:** Natively supports CUDA 12

**Possible Drawbacks:** Additional build overhead while CUDA 11 is maintained.

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/tests_linux_x86_mpi_gpu.yml
+++ b/.github/workflows/tests_linux_x86_mpi_gpu.yml
@@ -39,6 +39,7 @@ jobs:
       max-parallel: 1
       matrix:
         mpilib: ["mpich", "openmpi"]
+        cuda_version: ["11", "12"]
     timeout-minutes: 30
 
     steps:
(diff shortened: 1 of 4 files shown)


AmintorDusko (approved):
Short and sweet! Thank you for that!

.github/workflows/tests_linux_x86_mpi_gpu.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       max-parallel: 1
       matrix:
         mpilib: ["mpich", "openmpi"]
+        cuda_version_maj: ["12"]
```
vincentmr, line 42: This will likely not be used in a `matrix`, i.e. if we add `"11"` then `cuda_version_min` should not take value `"2"`. So we could maybe just have a single version parameter `cuda_version_maj_min`.

.github/workflows/wheel_linux_x86_64_cuda.yml
```diff
@@ -67,24 +65,25 @@ jobs:
...
             cat /etc/yum.conf | sed "s/\[main\]/\[main\]\ntimeout=5/g" > /etc/yum.conf
-            python -m pip install ninja cmake~=3.24.3 auditwheel custatevec-cu11
+            python -m pip install ninja cmake~=3.24.3 auditwheel custatevec-cu${{ matrix.cuda_version }}
             yum clean all -y
             yum install centos-release-scl-rh -y
             yum install devtoolset-11-gcc-c++ -y
             source /opt/rh/devtoolset-11/enable -y
             yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel7/x86_64/cuda-rhel7.repo -y
-            yum -y install cuda-11-5 git openssh wget
+            yum -y install cuda-${{ matrix.cuda_version }}-0 git openssh wget
```
vincentmr, line 79: Why do we pick `12.0` here? Is `12.2` not available in `yum`?

vincentmr (commented):
Could you `grep -r cu11 .github/workflows` and fix the compat workflows and a few other string retaining `cu11`? Otherwise looks good, just a question.

.github/workflows/wheel_linux_x86_64_cuda.yml
mlxd: For building the wheels, if we choose to build against 12.2, users on a 12.0 system will not be able to use it without following the CUDA forward compat supports. If we build against 12.0 all newer version will work by default.

.github/workflows/tests_linux_x86_mpi_gpu.yml
mlxd: While a good idea, we often need separation of major and minor version (in this case between modules, loaded versions etc) that this is just convenient from a QOL perspective. We can migrate it to the JSON mapped style as a follow up, but I think this is fine for now.

vincentmr (approved):
Looks good to me now, happy to approve and feel free to merge once the CI goes green. Cheers!

mlxd:
OpenMPI 5.0 is causing failures with the distributed workload. We will need to rebuild the runner image using an aolder version and rerun before merging.

PR #613, opened 2024-02-13
Add Simulate and update Serialize for LightningQubit2 class

AmintorDusko:
**Context:** Add simulate logic (only expval) for use with the new device.

**Description of the Change:** This PR introduces two Lightning-specific Python classes `LightningStateVector` and `LightningMeasurements`. The `simulate` logic can be obtained with both classes.

**Benefits:** The class structure assembles all the Lightning specifics in classes allowing for multi-backend dispatching and further simplifying the Python class interfacing with PennyLane (`LightningQubit`).

**Possible Drawbacks:** Some updates will be necessary if we pursue multi-backend dispatching.

**Related GitHub Issues:**


pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -21,135 +21,184 @@
...
+
+
+def simulate(circuit: QuantumScript, dtype=np.complex128, debugger=None) -> Result:
+    """Simulate a single quantum script.
+
+    Args:
+        circuit (QuantumTape): The single circuit to simulate
+        dtype: Datatypes for state-vector representation. Must be one of
+            ``np.complex64`` or ``np.complex128``.
+        debugger (_Debugger): The debugger to use
```
albi3ro, line 60: I think we should probably leave `_Debugger` and snapshots out of scope for the project for now.  This could probably be added in later as part of a separate project and feature.

AmintorDusko: I think this makes sense, my idea here was to have it default to None, just to know where to act in the future.
Do you prefer if I remove it? Could you please clarify what you mean by snapshots in this context? Am I referring to it anywhere?

pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -34,24 +34,48 @@
...
+
+    Returns:
+        tuple(TensorLike): The results of the simulation
+
+    Note that this function can return measurements for non-commuting observables simultaneously.
+
+    """
+    state, is_state_batched = LightningStateVector(
```
albi3ro, line 68: Is it in the plan to support native broadcasted parameters in lightning? If not, we can just assume the circuit does not have any broadcasted parameters and that the state is never broadcasted.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -208,12 +232,13 @@ class LightningQubit2(Device):
...
 
     name = "lightning.qubit2"
+    _CPP_BINARY_AVAILABLE = LQ_CPP_BINARY_AVAILABLE
 
     _device_options = ["rng", "c_dtype", "batch_obs", "mcmc", "kernel_name", "num_burnin"]
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        wires=None,
+        wires=None,  # the number of wires are always needed for Lightning. Do we have a better way to get it
```
albi3ro: We can just make `wires` mandatory for lightning qubit.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+            circuit (QuantumScript): The single circuit to simulate
+            debugger (._Debugger): The debugger to use
+
+        Returns:
+            Tuple: A tuple containing the Lightning final state handler of the quantum script and
+                whether the state has a batch dimension.
+
+        """
+
+        circuit = circuit.map_to_standard_wires()
```
albi3ro, line 381: Now done in `LightningQubit2.execute`.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+            elif isinstance(operations[0], BasisState):
+                self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
+                operations = operations[1:]
+
+        for operation in operations:
+            if isinstance(operation, (StatePrep, BasisState)):
+                raise DeviceError(
+                    f"Operation {operation.name} cannot be used after other "
+                    f"Operations have already been applied on a {self.short_name} device."
+                )
```
albi3ro, line 361: Should be handled by preprocessing.  Probably fine to leave in, but we could also remove it. If someone skipped calling preprocessing, they are asking for things to fall over in confusing ways.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+                continue
+            method = getattr(state, name, None)
+            wires = self.wires.indices(operation.wires)
+
+            if method is not None:  # apply specialized gate
+                inv = False
+                param = operation.parameters
+                method(wires, inv, param)
+            elif (
+                name[0:2] == "C(" or name == "ControlledQubitUnitary" or name == "MultiControlledX"
```
albi3ro, line 331:
```suggestion
            elif isinstance(operation, qml.ops.Controlled):
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+        state = self.state_vector
+
+        # Skip over identity operations instead of performing
+        # matrix multiplication with it.
+        for operation in operations:
+            name = operation.name
+            if name == "Identity":
+                continue
+            method = getattr(state, name, None)
+            wires = self.wires.indices(operation.wires)
```
albi3ro, line 324: Given the wires are now just [0,1,2,3,4, ...], I think we can just do `wires = list(operation.wires)`.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+
+        Args:
+            operation (~pennylane.operation.Operation): operation to apply
+
+        Returns:
+            array[complex]: the output state tensor
+        """
+        state = self.state_vector
+
+        basename = "PauliX" if operation.name == "MultiControlledX" else operation.base.name
```
albi3ro, line 273: `MultiControlledX` now inherits from `Controlled`, so you can now just do `operation.base.name` in all cases.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+        if basename == "Identity":
+            return
+        method = getattr(state, f"{basename}", None)
+        control_wires = self.wires.indices(operation.control_wires)
```
albi3ro, line 277: Like below, we can now just do `list(operation.control_wires)`.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+        control_values = (
+            [bool(int(i)) for i in operation.hyperparameters["control_values"]]
+            if operation.name == "MultiControlledX"
+            else operation.control_values
+        )
```
albi3ro, line 282:
```suggestion
        control_values = operation.control_values
```
MultiControlledX now has the control values property.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+        if operation.name == "MultiControlledX":
+            target_wires = list(set(self.wires.indices(operation.wires)) - set(control_wires))
+        else:
+            target_wires = self.wires.indices(operation.target_wires)
```
albi3ro, line 286:
```suggestion
        target_wires = operation.target_wires
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,386 @@
...
+
+        if isinstance(state, self._qubit_state.__class__):
+            state_data = allocate_aligned_array(state.size, np.dtype(self.dtype), True)
+            self._qubit_state.getState(state_data)
+            state = state_data
+
+        ravelled_indices, state = self._preprocess_state_vector(state, device_wires)
+
+        # translate to wire labels used by device
+        device_wires = self.map_wires(device_wires)
```
albi3ro, line 237:
```suggestion
```

probably not needed anymore.

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,182 @@
...
+        Args:
+            circuit (QuantumScript): The single circuit to simulate
+            is_state_batched (bool): Whether the state has a batch dimension or not.
+
+        Returns:
+            Tuple[TensorLike]: The measurement results
+        """
+        if set(circuit.wires) != set(range(circuit.num_wires)):
+            wire_map = {w: i for i, w in enumerate(circuit.wires)}
+            circuit = qml.map_wires(circuit, wire_map)
```
albi3ro, line 169: Also handled in `LightningQubit2.execute`, so can be ignored here.

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,182 @@
...
+            if len(circuit.measurements) == 1:
+                return self.measurement(circuit.measurements[0])
+
+            results = []
+            for mp in circuit.measurements:
+                measure_val = self.measurement(mp)
+                results += [
+                    measure_val,
+                ]
+            return tuple(results)
```
albi3ro, line 182:
```suggestion
            return tuple(self.measurement(mp) for mp in circuit.measurements)
```

pennylane_lightning/lightning_qubit/lightning_qubit2.py
AmintorDusko: Not in the foreseeable future, I will remove it.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
AmintorDusko: I will remove the debugger references.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
AmintorDusko: I will make it mandatory.

pennylane_lightning/lightning_qubit/_state_vector.py
AmintorDusko: I will remove it.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
AmintorDusko: I'm rethinking that. I think it is better to query it from the tape. As we have now.

pennylane_lightning/lightning_qubit/_measurements.py
AmintorDusko: Beautiful.

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,172 @@
...
+        MeasurementsC64,
+        MeasurementsC128,
+    )
+except ImportError:
+    pass
+
+import numpy as np
+from typing import Callable, List
+
+import pennylane as qml
```
albi3ro, line 29:
```suggestion
```

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,172 @@
...
+        Returns:
+            Tuple[TensorLike]: The measurement results
+        """
+
+        if not circuit.shots:
+            # analytic case
+            if len(circuit.measurements) == 1:
+                return self.measurement(circuit.measurements[0])
+
+            return tuple(self.measurement(mp) for mp in circuit.measurements)
```
albi3ro, line 172:
```suggestion

        if circuit.shots:
            raise NotImplementedError
        # analytic case
        if len(circuit.measurements) == 1:
            return self.measurement(circuit.measurements[0])

        return tuple(self.measurement(mp) for mp in circuit.measurements)
```

vincentmr (commented):
Looks good to me so far. I have a few questions: 

1. I thought the goal here was to align the API with `DefaultQubit`. If that is the case, don't we want to start by implementing the following methods: `__init__`, `preprocess`, `execute`, `compute_derivatives`, `compute_jvp`, `compute_vjp`? 

1. Do we want to mimic the `devices.qubit` module hierarchy, defining Lightning's Python modules as follows 
```
.
├── adjoint_jacobian.md
├── adjoint_jacobian.py
├── apply_operation.py
├── initialize_state.py
├── __init__.py
├── measure.py
├── sampling.py
└── simulate.py
``` 

3. Related to the previous question, we probably want to mutate things in-place which makes the definition of a state-vector class natural, with methods to apply gates. To follow `DefaultQubit`'s coding style, could we consider moving the apply methods to an `apply_operation` module where a state-vector input is modified in-place.

3. Same with `simulate`.

pennylane_lightning/lightning_qubit/_serialize.py
albi3ro, line 1: How does this differ from `pennylane/core/_serialize.py`?  Should we extend or modify that class instead of duplicating it?

tomlqc:
Just a note: PR is related to [sc-56037].

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,172 @@
...
+            TensorLike: the result of the measurement
+        """
+        self._qubit_state.apply_operations(measurementprocess.diagonalizing_gates())
+
+        state_array = self._qubit_state.state
+        total_wires = int(np.log2(state_array.size))
+        wires = Wires(range(total_wires))
+        return measurementprocess.process_state(state_array, wires)
+
+    def expval(self, measurementprocess: MeasurementProcess):
```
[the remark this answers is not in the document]
AmintorDusko: They are to be added in the future. I will update the PR description.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -0,0 +1,69 @@
...
+        dtype: Datatypes for state-vector representation. Must be one of
+            ``np.complex64`` or ``np.complex128``.
+
+    Returns:
+        tuple(TensorLike): The results of the simulation
+
+    Note that this function can return measurements for non-commuting observables simultaneously.
+
+    """
+    final_state = state.reset_state()
```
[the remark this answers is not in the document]
AmintorDusko: You are right. Sorry about that.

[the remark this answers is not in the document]
AmintorDusko: Fixed

AmintorDusko:
> Looks good to me so far. I have a few questions:
> 
> 1. I thought the goal here was to align the API with `DefaultQubit`. If that is the case, don't we want to start by implementing the following methods: `__init__`, `preprocess`, `execute`, `compute_derivatives`, `compute_jvp`, `compute_vjp`?
> 2. Do we want to mimic the `devices.qubit` module hierarchy, defining Lightning's Python modules as follows
> 
> ```
> .
> ├── adjoint_jacobian.md
> ├── adjoint_jacobian.py
> ├── apply_operation.py
> ├── initialize_state.py
> ├── __init__.py
> ├── measure.py
> ├── sampling.py
> └── simulate.py
> ```
> 
> 3. Related to the previous question, we probably want to mutate things in-place which makes the definition of a state-vector class natural, with methods to apply gates. To follow `DefaultQubit`'s coding style, could we consider moving the apply methods to an `apply_operation` module where a state-vector input is modified in-place.
> 4. Same with `simulate`.

Hi @vincentmr, thank you for your suggestions.
Lightning has a different approach to how we implement some functionalities, so it is understandable if we don't follow the same file structure as default, and in fact, this is not a requirement. 
Regarding the changes in architecture, was decided we would proceed with this architecture for now, and come back later for updates when pushing the migration to our other backends.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,363 @@
...
+            device_wires (Wires): wires that get initialized in the state
+
+        Returns:
+            array[complex]: normalized input state of length ``2**len(wires)``
+                or broadcasted state of shape ``(batch_size, 2**len(wires))``
+            array[int]: indices for which the state is changed to input state vector elements
+        """
+
+        # translate to wire labels used by device
+        device_wires = self.map_wires(device_wires)
```
albi3ro, line 172:
```suggestion
```
No `map_wires` method as no longer necessary.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,363 @@
...
+        Args:
+            state (array[int]): computational basis state of shape ``(wires,)``
+                consisting of 0s and 1s
+            wires (Wires): wires that the provided computational state should be initialized on
+
+        Returns:
+            int: basis state index
+        """
+        # translate to wire labels used by device
+        device_wires = self.map_wires(wires)
```
albi3ro, line 205:
```suggestion
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,363 @@
...
+            array[int]: indices for which the state is changed to input state vector elements
+        """
+
+        # translate to wire labels used by device
+        device_wires = self.map_wires(device_wires)
+
+        # special case for integral types
+        if state.dtype.kind == "i":
+            state = qml.numpy.array(state, dtype=self.C_DTYPE)
+        state = self._asarray(state, dtype=self.C_DTYPE)
```
albi3ro, line 177:
```suggestion
            state = qml.numpy.array(state, dtype=self.dtype)
        state = self._asarray(state, dtype=self.dtype)
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,363 @@
...
+            state = state_data
+
+        ravelled_indices, state = self._preprocess_state_vector(state, device_wires)
+
+        # translate to wire labels used by device
+        output_shape = [2] * self.num_wires
+
+        if len(device_wires) == self.num_wires and Wires(sorted(device_wires)) == device_wires:
+            # Initialize the entire device state with the input state
+            state = self._reshape(state, output_shape).ravel(order="C")
```
albi3ro, line 241:
```
E           AttributeError: 'LightningStateVector' object has no attribute '_reshape'
```

pennylane_lightning/lightning_qubit/_serialize.py
albi3ro: Could we make a change to the existing one where we only map the wires if a wires map is provided?

Aka:
```
        wires = [wires_map[w] for w in observable.wires] if wires_map else observable.wires
```
and setting `wires_map=None` by default?

This would give us substantially less to update and test.

AmintorDusko: The classes were merged. Implementation and tests were updated.

albi3ro: Thanks!

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,173 @@
...
+
+from pennylane.measurements import StateMeasurement, MeasurementProcess, ExpectationMP
+from pennylane.typing import TensorLike, Result
+from pennylane.tape import QuantumScript
+from pennylane.wires import Wires
+
+from typing import List
+
+from pennylane_lightning.core._serialize import QuantumScriptSerializer
+from ._serialize import QuantumScriptSerializer
```
albi3ro, line 37:
```suggestion
```

pennylane_lightning/lightning_qubit/_state_vector.py
albi3ro:
```suggestion
            state = np.reshape(state, output_shape).ravel(order="C")
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,361 @@
...
+        """
+        # translate to wire labels used by device
+
+        # length of basis state parameter
+        n_basis_state = len(state)
+
+        if not set(state.tolist()).issubset({0, 1}):
+            raise ValueError("BasisState parameter must consist of 0 or 1 integers.")
+
+        if n_basis_state != len(device_wires):
```
albi3ro, line 211:
```suggestion
        if n_basis_state != len(wires):
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -202,19 +199,18 @@ def _get_basis_state_index(self, state, wires):
             int: basis state index
         """
         # translate to wire labels used by device
```
vincentmr, line 201:
```suggestion
```

pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -30,7 +25,7 @@
 
 
 try:
-    # pylint: disable=import-error, no-name-in-module
+    # pylint: disable=import-error, no-name-in-module, unused-import
     from pennylane_lightning.lightning_qubit_ops import (
```
vincentmr, line 29: Could we save a few lines here?
```suggestion
    import pennylane_lightning.lightning_qubit_ops
```

vincentmr (commented):
Hi @AmintorDusko , I don't have much to add at the moment except this is taking a really nice shape. Are you planning to reach full (line) coverage of the patch in the PR, or will this be difficult given the partial nature of this port?

pennylane_lightning/lightning_qubit/lightning_qubit2.py
AmintorDusko: Do you mean importing only the full binaries for checking, and not the Classes?

AmintorDusko:
> Hi @AmintorDusko , I don't have much to add at the moment except this is taking a really nice shape. Are you planning to reach full (line) coverage of the patch in the PR, or will this be difficult given the partial nature of this port?

Hey @vincentmr, I'm aiming for full coverage, but it may take a while to do that.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,361 @@
...
+            device_wires (Wires): wires that get initialized in the state
+
+        Returns:
+            array[complex]: normalized input state of length ``2**len(wires)``
+                or broadcasted state of shape ``(batch_size, 2**len(wires))``
+            array[int]: indices for which the state is changed to input state vector elements
+        """
+        # special case for integral types
+        if state.dtype.kind == "i":
+            state = qml.numpy.array(state, dtype=self.dtype)
```
albi3ro, line 178:
```suggestion
            state = np.array(state, dtype=self.dtype)
```

Why `qml.numpy` here? Shouldn't we be sticking to numpy at this point?

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,361 @@
...
+                method(
+                    qml.matrix(operation.base),
+                    control_wires,
+                    control_values,
+                    target_wires,
+                    False,
+                )
+            except AttributeError:  # pragma: no cover
+                # To support older versions of PL
+                method(operation.base.matrix, control_wires, control_values, target_wires, False)
```
albi3ro, line 295: I don't think we need to try and support versions of pennylane that old. Especially because our pl version support is now blocked by other things.

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,173 @@
...
+        """Apply a measurement to state when the measurement process has an observable with diagonalizing gates.
+            This method will is bypassing the measurement process to default.qubit implementation.
+
+        Args:
+            measurementprocess (StateMeasurement): measurement to apply to the state
+
+        Returns:
+            TensorLike: the result of the measurement
+        """
+        self._qubit_state.apply_operations(measurementprocess.diagonalizing_gates())
```
albi3ro, line 75: We may potentially want to forbid measurements to have diagonalizing gates at the point the pipeline.  We don't want to end up in a situation where we have the measurements `[qml.probs(op=qml.X(0)), qml.probs(op=qml.Y(0))]`, and then the second one ends up being wrong.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,361 @@
...
+
+        Args:
+            dtype (complex): Data complex type
+
+        Returns: the state vector class
+        """
+        if self.dtype not in [np.complex128, np.complex64]:  # pragma: no cover
+            raise ValueError(
+                f"Data type is not supported for state-vector computation: {self.dtype}"
+            )
```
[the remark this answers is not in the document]
AmintorDusko: I agree.

pennylane_lightning/lightning_qubit/_measurements.py
AmintorDusko: How do you propose doing that?

albi3ro:
```suggestion
       if measurementprocess.diagonalizing_gates():
           raise NotImplementedError("Measurement process {measuremnetprocess} should be diagonalized before submission to `dev.execute`")
```

I'll write up a transform to add to preprocessing for this case.

AmintorDusko: Sounds good!

pennylane_lightning/lightning_qubit/_state_vector.py
AmintorDusko: I decided to re-add it because with this we have full functionality for the class.
I also, think this is a small cost for a nice extra safety layer.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,356 @@
...
+        StateVectorC64,
+        StateVectorC128,
+    )
+except ImportError:
+    pass
+
+from itertools import product
+import numpy as np
+
+import pennylane as qml
```
vincentmr, line 32: Not a blocker at all, but I'm wondering whether we should introduce `isort` in addition to `black` to take care of imports. This would automatically fix several `pylint` warnings and issues.

tests/lightning_qubit/test_state_vector_class.py
```diff
@@ -0,0 +1,470 @@
...
+from pennylane_lightning.lightning_qubit import LightningQubit
+
+if not LightningQubit._CPP_BINARY_AVAILABLE:
+    pytest.skip("No binary module found. Skipping.", allow_module_level=True)
+
+if LightningDevice != LightningQubit:
+    pytest.skip("Exclusive tests for lightning.qubit. Skipping.", allow_module_level=True)
+
+
+@pytest.mark.parametrize("num_wires", [3, 10])
```
vincentmr, line 37: To specify what to do with `0` and test the frequent cases `1` and `2`. 
```suggestion
@pytest.mark.parametrize("num_wires", range(4))
```

tests/lightning_qubit/test_state_vector_class.py
```diff
@@ -0,0 +1,470 @@
...
+    state_vector = LightningStateVector(num_wires, dtype=dtype, device_name=device_name)
+    assert state_vector.dtype == dtype
+    assert state_vector.device_name == device_name
+    assert state_vector.wires == Wires(range(num_wires))
+
+
+def test_wrong_device_name():
+    """Test an invalid device name"""
+    with pytest.raises(qml.DeviceError, match="The device name"):
+        LightningStateVector(3, device_name="thunder.qubit")
```
vincentmr, line 51: I like that `thunder.qubit` :)

vincentmr (commented):
Looks great @AmintorDusko, I'm leaving a few questions and suggestions, but this is pretty much ready to go IMO. I'll come back shortly for an approval.

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,187 @@
...
+
+        state_array = self._qubit_state.state
+        total_wires = int(np.log2(state_array.size))
```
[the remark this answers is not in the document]
AmintorDusko: Updated!

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,356 @@
...
+            return arr
+
+        if not dtype:
+            dtype = arr.dtype
+
+        # We allocate a new aligned memory and copy data to there if alignment or dtype
+        # mismatches
+        # Note that get_alignment does not necessarily return CPUMemoryModel(Unaligned)
+        # numpy allocated memory as the memory location happens to be aligned.
+        if int(get_alignment(arr)) < int(best_alignment()) or arr.dtype != dtype:
```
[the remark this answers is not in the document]
AmintorDusko: I didn't have time to research if this would be needed at some point.
I say we keep it for now and come back to clean it when time allows.

tests/lightning_qubit/test_state_vector_class.py
AmintorDusko: Thank you for that!

tests/lightning_qubit/test_state_vector_class.py
```diff
@@ -0,0 +1,470 @@
...
+    (qml.CSWAP, [0, 0, 0, 0, 0, 0, 1, 0], [0, 0, 0, 0, 0, 1, 0, 0]),
+    (qml.Toffoli, [1, 0, 0, 0, 0, 0, 0, 0], [1, 0, 0, 0, 0, 0, 0, 0]),
+    (qml.Toffoli, [0, 1, 0, 0, 0, 0, 0, 0], [0, 1, 0, 0, 0, 0, 0, 0]),
+    (qml.Toffoli, [0, 0, 0, 0, 0, 0, 1, 0], [0, 0, 0, 0, 0, 0, 0, 1]),
+    (qml.Toffoli, [0, 0, 0, 0, 0, 0, 0, 1], [0, 0, 0, 0, 0, 0, 1, 0]),
+]
+
+
+@pytest.mark.parametrize("operation,input,expected_output", test_data_three_wires_no_parameters)
+def test_apply_operation_three_wires_no_parameters(tol, operation, input, expected_output):
```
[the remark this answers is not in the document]
AmintorDusko: Could you please point me to the tests?

AmintorDusko:
Note that there is an apparent repetition of tests but this is by design.
Some tapes/(gates calls) present in integration are present in the new unit tests.
The intention here is to, in the presence of a bug, more easily pinpoint if this is an integration problem.

tests/lightning_qubit/test_state_vector_class.py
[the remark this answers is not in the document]
vincentmr: This [test](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/tests/test_gates.py#L87). It is ok to have `test_apply_operation_three_wires_no_parameters` too, but if this becomes a maintenance burden we might remove it.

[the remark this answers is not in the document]
AmintorDusko: Thank you for clarifying. Yes, I'm aware that we have similar tests, but the intention here is to have unit tests for this class. It is OK to have similar tests if they are aiming at different parts of the stack. This apparent redundancy (not really a full redundancy) will only help the debugging process. We are talking about very lightweight tests here.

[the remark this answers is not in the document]
vincentmr: Agreed, this is a nitpick, no need to do anything here.

vincentmr (approved):
Thanks @AmintorDusko , looks good to me and happy to see this comes through.

tests/lightning_qubit/test_measurements_class.py
```diff
@@ -0,0 +1,643 @@
...
+        phi = -0.6
+        statevector = lightning_sv(num_wires=1)
+        statevector.apply_operations([qml.RX(phi, 0)])
+        m = LightningMeasurements(statevector)
+        result = getattr(m, method_name)(qml.expval(qml.Projector([0, 1], wires=0)))
+        assert qml.math.allclose(result, np.sin(phi / 2) ** 2)
+
+
+THETA = np.linspace(0.11, 1, 3)
+PHI = np.linspace(0.32, 1, 3)
```
maliasadi, line 194: Would you add these constants to the top?

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+
+    @property
+    def wires(self):
+        """All wires that can be addressed on this device"""
+        return self._wires
+
+    @property
+    def num_wires(self):
+        """Number of wires addressed on this device"""
+        return self._num_wires
```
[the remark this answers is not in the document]
AmintorDusko: In `LightningMeasurements` and `LightningAdjointJacobian` (not yet available for review) implementations we have several occurrences of querying the statevector for these properties. In the past, we just referred to the internal variables, I think it is a good call to have formal getters or properties. I can maybe with some work remove num_wires, and obtain such information from wires. If you think it is important. Let me know your thoughts.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def _state_dtype(self):
+        """Binding to Lightning Managed state vector C++ class.
+
+        Returns: the state vector class
+        """
+        return StateVectorC128 if self.dtype == np.complex128 else StateVectorC64
+
+    @staticmethod
+    def _asarray(arr, dtype=None):
```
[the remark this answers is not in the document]
AmintorDusko: I was afraid the functionality could be needed in the future.
But if you think it is a good idea to remove it, I will do it.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+
+        **Example**
+
+        >>> dev = qml.device('lightning.qubit', wires=1)
+        >>> dev.apply([qml.PauliX(wires=[0])])
+        >>> print(dev.state)
+        [0.+0.j 1.+0.j]
+        """
+        state = np.zeros(2**self._num_wires, dtype=self.dtype)
+        state = self._asarray(state, dtype=self.dtype)
```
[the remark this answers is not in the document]
AmintorDusko: Done!

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+        self._qubit_state.getState(state)
+        return state
+
+    @property
+    def state_vector(self):
+        """Returns a handle to the state vector."""
+        return self._qubit_state
```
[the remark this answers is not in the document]
AmintorDusko: yes

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -0,0 +1,184 @@
...
+    def dtype(self):
+        """Returns the simulation data type."""
+        return self._dtype
+
+    def _measurement_dtype(self):
+        """Binding to Lightning Measurements C++ class.
+
+        Returns: the Measurements class
+        """
+        return MeasurementsC64 if self.dtype == np.complex64 else MeasurementsC128
```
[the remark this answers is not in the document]
AmintorDusko: Although I agree that a base class would be good to have, I think we can come back to this in the future when we have all classes in place and time allows. Maybe when we start the work to bring the other devices to the new API.

tests/lightning_qubit/test_measurements_class.py
```diff
@@ -0,0 +1,643 @@
...
+from conftest import LightningDevice  # tested device
+from pennylane.devices import DefaultQubit
+
+try:
+    from pennylane_lightning.lightning_qubit_ops import (
+        MeasurementsC64,
+        MeasurementsC128,
+    )
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
AmintorDusko: Yes! Thank you for that!

tests/lightning_qubit/test_measurements_class.py
AmintorDusko: sure!

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+        Returns: the state vector class
+        """
+        return StateVectorC128 if self.dtype == np.complex128 else StateVectorC64
+
+    @staticmethod
+    def _asarray(arr, dtype=None):
+        """Verify data alignment and allocate aligned memory if needed.
+
+        Args:
+            arr (numpy.array): data array
```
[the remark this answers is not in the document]
AmintorDusko: The method was removed.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -0,0 +1,349 @@
...
+            dtype (dtype, optional): if provided will convert the array data type.
+
+        Returns:
+            np.array: numpy array with aligned data.
+        """
+        arr = np.asarray(arr)  # arr is not copied
```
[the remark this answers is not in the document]
AmintorDusko: The method was removed.

pennylane_lightning/lightning_qubit/_state_vector.py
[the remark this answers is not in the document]
maliasadi: Yep, please! if this is not required for the current support. let's remove it so that we always add tested features and helper methods to the codebase.

pennylane_lightning/lightning_qubit/_state_vector.py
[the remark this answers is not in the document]
maliasadi: Thanks Amintor for the clarification. I also think it is a good call and better practice if they have some use cases in other APIs :+1:

pennylane_lightning/lightning_qubit/_measurements.py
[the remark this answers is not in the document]
maliasadi: Then, it requires removals of more duplicated code! This is easy to maintain now than later. so, I may suggest reconsidering the addition of such base class now  to simplify the implementation of these classes instead of accumulating this technical dept for later ;)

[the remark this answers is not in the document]
maliasadi: Then, it requires removals of more duplicated code! This is easy to maintain now than later, but if details of this base class is not yet clear, we can get back to this later in followup PRs :)

pennylane_lightning/lightning_qubit/_measurements.py
mudit2812: @vincentmr tackled this in #627 by [applying the adjoint](https://github.com/PennyLaneAI/pennylane-lightning/blob/6ea2ee82070d94560d5095a5c160f8b30e0836b0/pennylane_lightning/lightning_qubit/_measurements.py#L151) of the diagonalizing gates after measuring. That is also a good approach to avoid raising an error.

maliasadi (approved):
:parachute:

pennylane_lightning/lightning_qubit/_measurements.py
AmintorDusko: Nice suggestion @mudit2812 and @vincentmr. I just pushed a change that implements that. @mudit2812, could you please double-check if it makes sense and mark this conversation as resolved if sensible?

pennylane_lightning/lightning_qubit/_state_vector.py
maliasadi: Let me add this to lightning :)

PR #621, opened 2024-02-28
Add MidMeasureMP C++ implementation

tomlqc:
### Before submitting

**Context:**
Native-execution of circuits with mid-measurements is enabled via the dynamic-one-shot transform. To support the transform, a device must implement methods to apply MidMeasureMP.

MidMeasureMP is a special kind of operation where a sample is drawn from the state vector probability distribution and a projector is applied to the state vector accordingly. It may also support options like reset and post-selection.

The most straightforward way to deal with the requirements is to enhance the state vector class StateVectorLQubit.

- [x] Add method to obtain the probability in the z-basis on a given wire.
- [x] Add RNG attribute.
- [x] Add method to init the RNG attribute with a given seed.
- [x] Add method to sample from a binomial distribution.
- [x] Add method to collapse the state vector onto a given branch.
- [x] Add method which implements the MidMeasureMP workflow.
- [x] Add bindings to the MidMeasureMP method.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

**Related Shortcut Story:**
[sc-57704]


pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+     * @brief Apply a mid-circuit measurement.
+     *
+     * @param wires Wires to sample from.
+     */
+    inline auto applyMidMeasureMP(const std::vector<size_t> &wires,
+                                  const std::vector<size_t> &postselect = {},
+                                  bool reset = false) -> int {
+        PL_ABORT_IF_NOT(wires.size() == 1,
+                        "MidMeasureMP should have a single wire.")
+        PL_ABORT_IF_NOT(postselect.size() > 1,
```
vincentmr, line 657:
```suggestion
        PL_ABORT_IF(postselect.size() > 1,
```

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+     */
+    const auto measure(const std::size_t wire, const int postselect = -1,
+                       const bool reset = false) {
+
+        std::vector<PrecisionT> probs_ = probs(wire);
+        auto sample = random_sample(probs_[0]);
+        if (postselect != -1 && postselect != sample) {
+            setBasisState(0);
+            this->getData()[0] = 0.;
+            sample = -1;
```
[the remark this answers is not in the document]
tomlqc: I thought we need to reset the state. But you're right, if the circuit is stopped at that point, then no need to do anything.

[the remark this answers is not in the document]
tomlqc: @vincentmr are you sure the python layer doesn't do anything later with the statevector?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        return distrib(rng);
+    }
+
+    /**
+     * @brief Calculate probabilities of measuring 0 or 1 on a specific wire
+     *
+     * @param wire Wire
+     * @return std::vector<PrecisionT> Both probabilities of getting 0 or 1
+     */
+    auto probs(const std::size_t wire) -> std::vector<PrecisionT> {
```
vincentmr, line 687: Could we remove this in favor of using `probs` from `core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp`?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+    }
+
+    /**
+     * @brief Collapse the state vector as after having measured one of the
+     * qubits
+     *
+     * @param wire Wire to collapse.
+     * @param branch Branch 0 or 1.
+     */
+    void collapse(const std::size_t wire, const bool branch) {
```
vincentmr, line 718: We'll need to add bindings for this in order to use it during tree-traversal. No need to do it in this PR however.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        }
+
+        PrecisionT norm = std::sqrt(squaredNorm(arr, this->getLength()));
+
+        if (norm > std::numeric_limits<PrecisionT>::epsilon() * 1e2) {
+            std::complex<PrecisionT> inv_norm = 1. / norm;
+            for (size_t k = 0; k < this->getLength(); k++) {
+                arr[k] *= inv_norm;
+            }
+        }
```
vincentmr, line 745: It might be useful to have this in a separate method and add bindings.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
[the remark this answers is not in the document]
vincentmr: > I thought we need to reset the state.

This is what we do in Python to signal the shot is lost, but here we use `sample = -1` so I think there is no need to touch the state. If things changes we could revise that.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitRaw.hpp
```diff
@@ -116,6 +116,18 @@ class StateVectorLQubitRaw final
...
+    /**
+     * @brief Prepares a single computational basis state.
+     *
+     * @param index Index of the target element.
+     */
+    void setBasisState(const std::size_t index) {
+        for (size_t i = 0; i < length_; i++) {
+            data_[i] = {0, 0};
+        }
+        data_[index] = {1, 0};
```
vincentmr, line 128: Nitpick warning:
```suggestion
            data_[i] = {0.0, 0.0};
        }
        data_[index] = {1.0, 0.0};
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
+# Copyright 2023 Xanadu Quantum Technologies Inc.
```
vincentmr, line 1:
```suggestion
# Copyright 2024 Xanadu Quantum Technologies Inc.
```

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +643,133 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+    /**
+     * @brief Measure one of the qubits and collapse the state accordingly
+     *
+     * @param wire Wire to measure.
+     * @param postselect Wire sample value to postselect (0 or 1) or -1 to
+     * ignore
+     * @param reset Flag to reset wire to 0
+     * @param branch Branch 0 or 1.
+     */
+    const auto measure(const std::size_t wire, const int postselect = -1,
```
vincentmr, line 757: When we're done we could move this up in `applyMidMeasureMP` and delete `measure`. I think the Catalyst folks may then redefine the interface at their convenience.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
tomlqc: This introduces a circular dependency between `MeasurementsLQubit.hpp` and `StateVectorLQubit.hpp`. Let's keep it as it is right now, and fix it in another story.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+* `lightning.qubit` supports mid-circuit measurements.
```
tomlqc, line 5: Actually that's not Release 0.35 any more. Should we create a new Release 0.36 section right now?

vincentmr: No, it is generally done in a separate PR #626 .

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+     * @param seed Seed
+     */
+    void seed(const size_t seed) { rng.seed(seed); }
+
+    /**
+     * @brief Sample 0 or 1 for given probabilities
+     *
+     * @param probs Probabilities of 0.
+     */
+    auto random_sample(const PrecisionT prob_0) {
```
vincentmr, line 682: We usually employ lower camel case for methods.
```suggestion
    auto getRandomSample(const PrecisionT prob_0) {
```

vincentmr (commented):
Just two minors suggestions. I'll wait to read others' comments before approving, but this looks good to me.

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -63,19 +63,19 @@
     from pennylane.operation import Tensor
     from pennylane.ops.op_math import Adjoint
     from pennylane.wires import Wires
+    from pennylane_lightning.lightning_kokkos_ops.algorithms import (
```
tomlqc, line 66: Fun fact: that's `make format` that made this change. It doesn't seem to consider `pylint: ungrouped-imports` below.

vincentmr: Yes, we should put a `pylint: disable=` right on top if `from pennylane_lightning.lightning_kokkos_ops.algorithms import` to fix this.

tomlqc: I already tried this locally. The `make format` still moves the code snippet about `algorithms` to the top. But I will commit now, may be it resolves the CodeFactor issue.

tomlqc: CodeFactor is happy

vincentmr (approved):
Looks good to me, thanks @tomlqc .

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        PL_ABORT_IF_NOT(wires.size() == 1,
+                        "MidMeasureMP should have a single wire.")
+        PL_ABORT_IF(postselect.size() > 1,
+                    "MidMeasureMP accepts at most one postselect value.")
+        const int ps = postselect.empty() ? -1 : postselect[0];
+
+        std::vector<PrecisionT> probs_ = probs(wires[0]);
+        auto sample = randomSample(probs_[0]);
+        if (ps != -1 && ps != sample) {
+            sample = -1;
```
dime10, line 660: Does this mean the operation can return a measurement result of -1?

dime10 (commented):
Are we extending the support to kokkos and cuquantum as well?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
vincentmr: How do you deal with filtering out a shot in Catalyst?

vincentmr:
> Are we extending the support to kokkos and cuquantum as well?

That's the plan ;)

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
dime10: The short answer is: we don't 😅 

QNode execution in Catalyst is currently exclusively managed by the user program (user program here means "all binary code staged out from python"). The Catalyst runtime environment merely facilitates communication between the user program and the quantum device, but does not manage execution.

I've provided some more context offline so we might need to come a decision on that discussion first before determining whether the proposed implementation here is compatible with Catalyst (at the moment it isn't).

.github/workflows/tests_linux.yml
```diff
@@ -590,7 +591,7 @@ jobs:
         run: |
           cd main/
           DEVICENAME=`echo ${{ matrix.pl_backend }} | sed "s/_/./g"`
-          PL_DEVICE=${DEVICENAME} python -m pytest tests/ $COVERAGE_FLAGS
+          PL_DEVICE=${DEVICENAME} python -m pytest tests/ -k "not test_native_mcm" $COVERAGE_FLAGS
```
AmintorDusko, line 594: Do we need to do that, since we have a [device name control ](https://github.com/PennyLaneAI/pennylane-lightning/pull/621/files#diff-cd218047f28fcee04400b9e924bdc451f050d553f4867212944603325d909d3fR28-R29)already in place?

AmintorDusko (commented):
I left some comments, but mostly questions.
It would be great if this new measurement process could be incorporated to the Measurement class. Such implementation of a measurement as part of the state vector is against the design we have in place now.

.github/workflows/tests_linux.yml
vincentmr: This is not required, just to be safe.

.github/workflows/tests_without_binary.yml
```diff
@@ -106,7 +106,7 @@ jobs:
         run: |
           cd main/
           DEVICENAME=`echo ${{ matrix.pl_backend }} | sed "s/_/./g"`
-          PL_DEVICE=${DEVICENAME} python -m pytest tests/ $COVERAGE_FLAGS
+          PL_DEVICE=${DEVICENAME} python -m pytest tests/ -k "not test_native_mcm" $COVERAGE_FLAGS
```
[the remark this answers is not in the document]
vincentmr: This is not required, just to be safe.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+
+    /**
+     * @brief Apply a mid-circuit measurement.
+     *
+     * @param wires Wires to sample from.
+     * @param postselect Value to postselect (0 or 1) or empty to ignore
+     * @param reset Flag to reset wire to 0
+     */
+    inline auto
+    applyMidMeasureMP(const std::vector<std::size_t> &wires,
```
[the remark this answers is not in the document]
vincentmr: Yes.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
vincentmr: Yes, it is in the Python layer [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/621/files/7fb6faaba1becce5a38cdd244f745d2e575536d6#diff-068e2aa5237a375b6af67a124f81018aa46c2d335fecf3035346f0627e6ef7a4R454) and then in PennyLane.

.github/workflows/tests_linux.yml
```diff
@@ -184,7 +184,8 @@ jobs:
...
+          OMP_NUM_THREADS=1 PL_DEVICE=${DEVICENAME} python -m pytest -n auto tests/test_native_mcm.py $COVERAGE_FLAGS
```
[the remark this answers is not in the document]
AmintorDusko: But you are making ` OMP_NUM_THREADS=1`, isn't the number of parallel cores in pytest controlled with the OMP_NUM_THREADS variable?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
AmintorDusko: Well if this is used in PennyLane the incompatibility goes a little bit deeper. We need to think about that.

.github/workflows/tests_linux.yml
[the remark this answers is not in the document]
vincentmr: `OMP_NUM_THREADS=1` sets the number of OpenMP threads per pytest worker to `1`, then `-n` sets the number of pytest workers. So `OMP_NUM_THREADS=1` is to avoid oversubscription.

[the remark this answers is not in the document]
AmintorDusko: Hmm. Nice to know. Thank you for clarifying.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
mudit2812: We need the marginal probability distribution of the measured wire to get a sample which will determine which state the state vector collapses to

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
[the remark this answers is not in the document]
AmintorDusko: Thanks!

tomlqc:
> I left some comments, but mostly questions. It would be great if this new measurement process could be incorporated to the Measurement class. Such implementation of a measurement as part of the state vector is against the design we have in place now.

@vincentmr I was just wondering if moving the new methods to `Measurement` class would be also beneficial as then we could use its `probs()` method (it isn't straight forward right now because of the circular dependencies). Are there other design aspects that favors the implementation in `StateVector`?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        for (size_t idx = 0; idx < half_section_size; idx++) {
+            const size_t offset = stride * 2 * idx;
+            for (size_t ids = 0; ids < stride; ids++) {
+                prob_0 += std::norm(arr[offset + ids]);
+            }
+        }
+
+        std::vector<PrecisionT> probs{prob_0, PrecisionT(1.) - prob_0};
+        return probs;
+    }
```
AmintorDusko, line 716: As previously discussed this part of the logic can be moved to the measurements class.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+    void normalize() {
+        auto *arr = this->getData();
+        PrecisionT norm = std::sqrt(squaredNorm(arr, this->getLength()));
+
+        if (norm > std::numeric_limits<PrecisionT>::epsilon() * 1e2) {
+            std::complex<PrecisionT> inv_norm = 1. / norm;
+            for (size_t k = 0; k < this->getLength(); k++) {
+                arr[k] *= inv_norm;
+            }
+        }
```
AmintorDusko, line 760: It makes sense to have this method in the state vector class. Please verify if we can have this added to the [lightning base](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/simulators/base/StateVectorBase.hpp). If not, this is a nice place because both raw and managed will have access to this functionality.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        const size_t k = branch ? 0 : 1;
+        for (size_t idx = 0; idx < half_section_size; idx++) {
+            const size_t offset = stride * (k + 2 * idx);
+            for (size_t ids = 0; ids < stride; ids++) {
+                arr[offset + ids] = {0., 0.};
+            }
+        }
+
+        normalize();
+    }
```
AmintorDusko, line 746: We can keep it here since this is strongly state vector oriented. But please expand the docstring to provide more context for the branch variable.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+        if (ps != -1 && ps != sample) {
+            sample = -1;
+        } else {
+            collapse(wires[0], sample);
+            if (reset && sample == 1) {
+                applyOperation("PauliX", wires);
+            }
+        }
+        return sample;
+    }
```
AmintorDusko, line 668: This class is realizing a measurement. It would be nice to have it in the measurement class. Without looking too deep into that, it looks like this will integrate nicely with the class design we have in place for LightningQubit with the new device API.

AmintorDusko (commented):
Some more thoughts. Let's see if overall we can move some code reused by LKokkos to the respective base classes.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +636,128 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+
+    /**
+     * @brief Normalize vector (to have norm 1).
+     */
+    void normalize() {
+        auto *arr = this->getData();
+        PrecisionT norm = std::sqrt(squaredNorm(arr, this->getLength()));
+
+        if (norm > std::numeric_limits<PrecisionT>::epsilon() * 1e2) {
```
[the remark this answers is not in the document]
tomlqc: In this `normalize` function, I just want to make sure we don't divide by zero.

- If the norm is zero in our mid-measure scenario, the collapse was done by trying to keep elements of the vector that already had zero amplitude. 
- In the dynamic shots approach, as we sample from the probabilities, this case will not occur.
- In the fixed shots approach (catalyst), this case may occur. But we can avoid this if catalyst checks the probabilities first and stops in case probability of getting `branch` is zero.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
tomlqc: ok, I'll move `applyMidMeasureMP` to Measurement

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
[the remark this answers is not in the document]
AmintorDusko: Please check with the catalyst team the best course of action here, but if this is not needed I'm in favor of removing it.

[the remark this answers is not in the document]
vincentmr: The analogy is usually give is this is that projecting the state on a zero-probability branch is the quantum programming equivalent of a division by zero. It thus requires special handling at either compile time or runtime as in classical programming.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
vincentmr: Yes, it is in fact already in the measurement class, but we could not use it without tricking CMake to use GCC-specific flags. We'll find a way as per our discussion.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
vincentmr: It will be difficult to have an implementation in `base` because L-Kokkos will need a Kokkos kernel and L-GPU a CUDA kernel or a call to cuQuantum for that.

AmintorDusko: No worries.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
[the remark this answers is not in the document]
dime10: It looks like right now if the probabilities are very small, the normalization is just skipped. Is that better than having a statevector with infinities? 🤔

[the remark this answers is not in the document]
vincentmr: This should not happen with the revised workflow, so @tomlqc how about simply throwing an error with `PL_ABORT_IF`?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
tomlqc: As agreed upon, `applyMidMeasureMP` has been moved to the python layer

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
tomlqc: As `applyMidMeasure` has been moved to python layer, currently not need for `probs()`

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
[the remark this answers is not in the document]
tomlqc: I'm just unsure about `std::numeric_limits<PrecisionT>::epsilon() * 1e2`. The `1e2` is arbitrary (I've searched for similar usage in lightning and found some 1e2 or 1e4 at other places). 
Does everyone agree with `1e2` and `PL_ABORT_IF` ?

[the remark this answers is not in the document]
AmintorDusko: Thank you for clarifying, we can keep it like that, but possibly throwing an error as suggested by @vincentmr.

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+        func(*params)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000, [5000, 5001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+def test_simple_mcm(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a single mid-circuit measurement and a
```
AmintorDusko, line 170:
```suggestion
    """Tests that LightningQubit handles a circuit with a single mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+        validate_measurements(measure_f, shots, results1, results2)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [1000, [1000, 1001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+def test_single_mcm_single_measure_mcm(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a single mid-circuit measurement and a
```
AmintorDusko, line 200:
```suggestion
    """Tests that LightningQubit handles a circuit with a single mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000, [5000, 5001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+@pytest.mark.parametrize("obs", [qml.PauliZ(0), qml.PauliY(1), qml.PauliZ(0) @ qml.PauliY(1)])
+def test_single_mcm_single_measure_obs(shots, postselect, reset, measure_f, obs):
+    """Tests that DefaultQubit handles a circuit with a single mid-circuit measurement and a
```
AmintorDusko, line 250:
```suggestion
    """Tests that LightningQubit handles a circuit with a single mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [3000, [3000, 3001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.probs, qml.sample])
+@pytest.mark.parametrize("wires", [[0], [0, 1]])
+def test_single_mcm_single_measure_wires(shots, postselect, reset, measure_f, wires):
+    """Tests that DefaultQubit handles a circuit with a single mid-circuit measurement and a
```
AmintorDusko, line 278:
```suggestion
    """Tests that LightningQubit handles a circuit with a single mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+        validate_measurements(measure_f, shots, results1, results2)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+def test_single_mcm_multiple_measurements(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a single mid-circuit measurement with reset
```
AmintorDusko, line 307:
```suggestion
    """Tests that LightningQubit handles a circuit with a single mid-circuit measurement with reset
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+            validate_measurements(measure_f, shots, r1, r2)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000, [5000, 5001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.sample, qml.var])
+def test_composite_mcm_measure_composite_mcm(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a composite mid-circuit measurement and a
```
AmintorDusko, line 337:
```suggestion
    """Tests that LightningQubit handles a circuit with a composite mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+        validate_measurements(measure_f, shots, results1, results2)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000, [5000, 5001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+def test_composite_mcm_single_measure_obs(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a composite mid-circuit measurement and a
```
AmintorDusko, line 370:
```suggestion
    """Tests that LightningQubit handles a circuit with a composite mid-circuit measurement and a
```

tests/test_native_mcm.py
```diff
@@ -0,0 +1,423 @@
...
+        validate_measurements(measure_f, shots, results1, results2)
+
+
+@flaky(max_runs=5)
+@pytest.mark.parametrize("shots", [5000, [5000, 5001]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("reset", [False, True])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.probs, qml.sample])
+def test_composite_mcm_measure_value_list(shots, postselect, reset, measure_f):
+    """Tests that DefaultQubit handles a circuit with a composite mid-circuit measurement and a
```
AmintorDusko, line 400:
```suggestion
    """Tests that LightningQubit handles a circuit with a composite mid-circuit measurement and a
```

AmintorDusko (approved):
Beautiful! You have my approval already, but please check my suggestions.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -65,6 +67,23 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         : _statevector{statevector} {};
 #endif
 
+    /**
+     * @brief Set the seed of the internal random generator
+     *
+     * @param seed Seed
+     */
+    void setSeed(const size_t seed) { rng.seed(seed); }
```
dime10, line 75: Nice! So the seed has to be set on each measurement object right?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -632,5 +633,53 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
...
+            }
+        }
+
+        normalize();
+    }
+
+    /**
+     * @brief Normalize vector (to have norm 1).
+     */
+    void normalize() {
```
dime10, line 672: Nice, I like the reusable pieces 👍

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -419,32 +432,56 @@ def _apply_lightning_controlled(self, operation):
...
+
+            Returns:
+                None
+            """
+            wires = self.wires.indices(operation.wires)
+            wire = list(wires)[0]
+            sample = qml.math.reshape(self.generate_samples(shots=1), (-1,))[wire]
+            if operation.postselect is not None and sample != operation.postselect:
+                mid_measurements[operation] = -1
+                return
```
dime10, line 449: Ok just confirming here, an invalid shot is recorded as a `-1` outcome and the statevector is left untouched right?

dime10 (approved):
Looks good from my side 💯  So on the Catalyst side we move to just calling the `sample` and `collapse` function ourselves [here](https://github.com/PennyLaneAI/catalyst/blob/ac3b90f9b69fefa0f24b15946ddfeacf6be36949/runtime/lib/backend/lightning/lightning_dynamic/LightningSimulator.cpp#L423) (similar to what lightningqubit.py is doing) ?

pennylane_lightning/lightning_qubit/lightning_qubit.py
tomlqc: Correct

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
tomlqc: That's the current way in LQubit or GPU implementations (set randomly).

PR #627, opened 2024-03-05
Lightning qubit2 add measurements

vincentmr:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

**Description of the Change:**
Add `probs` and `var` support in the new device measurement module.
Add parametrized tests, checkout single/double observable support.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #628, opened 2024-03-05
Lightning qubit2 upgrade api

vincentmr:
### Before submitting

**Context:**
This PR merges #607 and #627 to enable integration testing.

Since #607 enables QNode execution, several tests under `tests` (namely all except those explicitly testing the old device API) become available, in principle that is since features like adjoint at still missing. This PR therefore introduces (as a temporary measure and until all old device API backends become deprecated) a `_new_API` attribute which allows discriminating between backends. All tests with missing functionality (e.g. adjoint, shots, etc.) are marked `skip` or `xfail`. 

**Benefits:**
Leverage unit/integration tests that already exists.
Enable coexistence of old/new device API backends, likely needed if we cannot complete the port of L-GPU/Kokkos before next release.

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -194,7 +201,13 @@ class LightningQubit2(Device):
 
     _device_options = ("rng", "c_dtype", "batch_obs", "mcmc", "kernel_name", "num_burnin")
 
+    _new_API = True
     _CPP_BINARY_AVAILABLE = LQ_CPP_BINARY_AVAILABLE
+    short_name = "lightning.qubit2"
+    operations = _operations
+    observables = _observables
```
albi3ro, line 208: So we do have these as class properties below as well?

pennylane_lightning/lightning_qubit/lightning_qubit2.py
```diff
@@ -194,7 +201,13 @@ class LightningQubit2(Device):
 
     _device_options = ("rng", "c_dtype", "batch_obs", "mcmc", "kernel_name", "num_burnin")
 
+    _new_API = True
     _CPP_BINARY_AVAILABLE = LQ_CPP_BINARY_AVAILABLE
+    short_name = "lightning.qubit2"
```
albi3ro, line 206: Was the lack of a `short_name` breaking something in particular?

vincentmr: At first I could not properly initialize a device with it's name and I thought I should add this by analogy with `lightning.qubit`. But the issue was the missing symlink `lightning_qubit2`, I think. So we can remove or keep it.

pennylane_lightning/lightning_qubit/lightning_qubit2.py
vincentmr: I think the duplication was introduced during a merge. Let me keep only these two.

PR #635, opened 2024-03-08
Fix random `coverage xml` CI issues

maliasadi:
Fix random `coverage xml` CI non-deterministic failures with ignoring "No source for code  `__init__,py`". This shouldn't affect the coverage as `__init__` only includes the package version.  

[sc-58646]

### DIFF
--- a/.github/workflows/tests_linux.yml
+++ b/.github/workflows/tests_linux.yml
@@ -645,6 +645,7 @@ jobs:
           python -m pip install coverage
           python -m coverage combine .coverage-python*
           # Added cov xml -i to ignore "No source for code" random errors
+          # https://stackoverflow.com/questions/2386975/no-source-for-code-message-in-coverage-py
           python -m coverage xml -i -o coverage-${{ github.job }}.xml
 
       - name: Upload to Codecov
--- a/.github/workflows/tests_linux.yml
+++ b/.github/workflows/tests_linux.yml
@@ -644,7 +644,8 @@ jobs:
         run: |
           python -m pip install coverage
           python -m coverage combine .coverage-python*
-          python -m coverage xml -o coverage-${{ github.job }}.xml
+          # coverage -i is added to ignore "No source for code" unspecified errors
+          python -m coverage -i xml -o coverage-${{ github.job }}.xml
 
       - name: Upload to Codecov
         uses: codecov/codecov-action@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/tests_linux.yml
```diff
@@ -644,7 +644,8 @@ jobs:
         run: |
           python -m pip install coverage
           python -m coverage combine .coverage-python*
-          python -m coverage xml -o coverage-${{ github.job }}.xml
+          # coverage -i is added to ignore "No source for code" unspecified errors
+          python -m coverage -i xml -o coverage-${{ github.job }}.xml
```
maliasadi, line 648:
```suggestion
          # Added cov xml -i to ignore "No source for code" ramdom errors
          python -m coverage xml -i -o coverage-${{ github.job }}.xml
```

vincentmr (approved):
Looks good to me, thanks @maliasadi .

.github/workflows/tests_linux.yml
```diff
@@ -644,7 +644,9 @@ jobs:
         run: |
           python -m pip install coverage
           python -m coverage combine .coverage-python*
-          python -m coverage xml -o coverage-${{ github.job }}.xml
+          # Added cov xml -i to ignore "No source for code" random errors
+          # https://stackoverflow.com/questions/2386975/no-source-for-code-message-in-coverage-py
+          python -m coverage xml -i -o coverage-${{ github.job }}.xml
```
tomlqc, line 650: Hi guys, you're sure we won't miss important errors with:
`-i, --ignore-errors   Ignore errors while reading source files`

maliasadi: @tomlqc I'd argue that if there are any errors in uploading all coverage reports (due to missing some important errors at this step), then the codecov CI report will fail because of the coverage reduction in the code base. We can always revert this change if we find out this patch is defective.

maliasadi: @tomlqc I presume that in the event of any errors during the export of coverage reports, the codecov CI report may fail due to a reduction in code base coverage. I didn't see any issue with > 10-time triggering CIs that may be a good sign. If we discover that this patch is defective, we have the option to revert this change at anytime.

AmintorDusko (approved):
Thank you for that!

tomlqc (approved):

PR #569, opened 2023-12-05
add Hermitian support for shot-noise & LAPACK linking

multiphaseCFD:
### Before submitting

Please complete the following checklist when submitting a PR:

- [ ] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [ ] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

Add `Hermitian` observable support for stochastic measurement to C++ layer. `LAPACK` is required to solve `eigenvalues` and the `unitary` of a given `Hermitian`. The main challenge of this task is `LAPACK` linking for Windows system. To solve this issue, `vcpkg` is used to install `LAPACK` and the only thing Windows users need to do is to specify the path to set `CMAKE_TOOLCHAIN_FILE` with `-DCMAKE_TOOLCHAIN_FILE=<vcpkg-root>/scripts/buildsystems/vcpkg.cmake`.

**Description of the Change:**

C++ backends now can link against LAPACK.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #642, opened 2024-03-13
Update device toml specs to schema 2

grwlf:
**Context:** Catalyst supports quantum device specification in a new format in addition to the existing one, supporting per-gate parameters. 

**Description of the Change:** By this PR we update existing device specifications to the new format 

**Benefits:** Per-gate properties allow device authors to query PL for certain behavior on programs (decomposition, adding quantum control, validation) 

**Possible Drawbacks:**

**Related GitHub Issues:**
* https://github.com/PennyLaneAI/catalyst/pull/629
* https://github.com/PennyLaneAI/catalyst/pull/554
* https://github.com/PennyLaneAI/catalyst/pull/712
* https://github.com/PennyLaneAI/catalyst/pull/718
* https://github.com/PennyLaneAI/pennylane-lightning/pull/665


pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
-        "DoubleExcitation",
-        "DoubleExcitationPlus",
-        "DoubleExcitationMinus",
-        "MultiRZ",
-        "QubitUnitary",
-        "GlobalPhase",
-]
+[operators.gates.native]
+
+CNOT = { properties = [ "invertible", "differentiable" ] }
```
mlxd, line 7: Is there a formatter for TOML that we can use to make this a little easier to read? Maybe some newlines and indenting? If not, we can keep as is for now.

mlxd (commented):
Hey @grwlf 
Just a few quick questions

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
+QubitCarry = {}
+QubitSum = {}
+SISWAP = {}
+SQISW = {}
+SX = {}
+
+# Observables supported by the device
+[operators.observables]
+
+Exp = {}
```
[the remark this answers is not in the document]
grwlf: I believe it is the name of PennyLane operation. The same as we had in the current `schema 1` config [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/2c69c59fa5c2685a461fac6c33c9067f076f59c6/pennylane_lightning/lightning_kokkos/lightning_kokkos.toml#L20).

I would suggest to think about these configs as of the place where we specify what should PennyLane do to run a program on a particular device. By listing `Exp` here we essentially allow it to keep this operation on a quantum tape.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
+Hadamard = {}
+Hamiltonian = {}
+Hermitian = {}
+Identity = {}
+PauliX = {}
+PauliY = {}
+PauliZ = {}
+Prod = {}
```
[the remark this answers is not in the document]
grwlf: The current consequence of these lines - the mentioned operations will appear in the `device.observables` set. This would prevent PL from decomposing. Again, here we just change the format of the config. The old one listed the same data as list of string [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/2c69c59fa5c2685a461fac6c33c9067f076f59c6/pennylane_lightning/lightning_kokkos/lightning_kokkos.toml#L8).

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
-	"Var",
-	"Probs",
-	"Sample",
-	"Counts",
-]
+
+Expval = {}
+Var = {}
+Probs = {}
+State = { condition = [ "analytic" ] }
```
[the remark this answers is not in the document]
grwlf: I believe that Lightning does not have to process it in any way. It is PennyLane (Catalyst) who does. This condition tells it that if the user asks for `shots=None` then we must add `State` to the set of supported observables for this device. Catalyst will act accordingly by allowing State to appear in the emitted LLVM code for this program.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
+Hadamard = {}
+Hamiltonian = {}
+Hermitian = {}
+Identity = {}
+PauliX = {}
+PauliY = {}
+PauliZ = {}
+Prod = {}
+SparseHamiltonian = {}
+SProd = {}
```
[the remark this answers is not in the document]
grwlf: See the Exp discussion above.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,109 @@
...
+Sum = {}
```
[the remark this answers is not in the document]
grwlf: See the Exp discussion above.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Indeed it might be overly verbose, but do you think that newlines will help in getting the whole picture? What do you think about table-like formatting, like
``` toml
DoubleExcitationPlus = { properties = [ "invertible",                       "differentiable" ] }
DoubleExcitation        = { properties = [ "invertible",                       "differentiable" ] }
GlobalPhase              = { properties = [ "invertible", "controllable", "differentiable" ] }
```
?

mlxd: Yep, honestly that looks good to me too. Having a quick 'n' easy way to scan it will make checks here easy. Thanks @grwlf

grwlf: @mlxd Done!

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,110 @@
...
+CRX                    = { properties = [ "invertible",                 "differentiable" ] }
+CRY                    = { properties = [ "invertible",                 "differentiable" ] }
+CRZ                    = { properties = [ "invertible",                 "differentiable" ] }
+CSWAP                  = { properties = [ "invertible",                 "differentiable" ] }
+CY                     = { properties = [ "invertible",                 "differentiable" ] }
+CZ                     = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationMinus  = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationPlus   = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitation       = { properties = [ "invertible",                 "differentiable" ] }
+GlobalPhase            = { properties = [ "invertible", "controllable", "differentiable" ] }
```
vincentmr, line 20: `C(GlobalPhase)` isn't supported by the C++ device.
```suggestion
GlobalPhase            = { properties = [ "invertible", "differentiable" ] }
```

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,131 +1,114 @@
...
-        "Identity",
-        "Projector",
-        "SparseHamiltonian",
-        "Hamiltonian",
-        "Sum",
-        "SProd",
-        "Prod",
-        "Exp",
-]
+CNOT                   = { properties = [ "invertible",                 "differentiable" ] }
```
vincentmr, line 5: We could make this `controllable`, if that helps in any way.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,131 +1,114 @@
...
+IsingXX                = { properties = [ "invertible", "controllable", "differentiable" ] }
+IsingXY                = { properties = [ "invertible", "controllable", "differentiable" ] }
+IsingYY                = { properties = [ "invertible", "controllable", "differentiable" ] }
+IsingZZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
+MultiRZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliX                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliY                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliZ                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PhaseShift             = { properties = [ "invertible", "controllable", "differentiable" ] }
+QubitUnitary           = { properties = [ "invertible",                 "differentiable" ] }
```
vincentmr, line 30: This is technically controllable, but I guess this is covered by `ControlledQubitUnitary`?

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,131 +1,114 @@
...
 # strategy in PennyLane.
-decomp = [
-        "BasisState",
-        "QubitStateVector",
-        "StatePrep",
-        "QFT",
-        "MultiControlledX",
-]
+BasisState             = {}
+MultiControlledX       = {}
```
vincentmr, line 51: Should that be here given that `PauliX` above is `controllable`?

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,110 @@
...
-        "OrbitalRotation",
-        "QubitCarry",
-        "QubitSum",
-        "DiagonalQubitUnitary",
-        "BlockEncode",
-]
+[operators.gates.matrix]
+
+BlockEncode            = {}
+CPhase                 = {}
```
vincentmr, line 62: Is that not `ControlledPhaseShift`?

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,110 @@
...
+Exp                    = {}
+Hadamard               = {}
+Hamiltonian            = {}
+Hermitian              = {}
+Identity               = {}
+PauliX                 = {}
+PauliY                 = {}
+PauliZ                 = {}
+Prod                   = {}
+Projector              = {}
```
vincentmr, line 86: I don't think this is supported.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: Yes, exactly. The absence of `controllable` attribute means that Catalyst should instruct PennyLane to avoid creating `C(QubitUnitary)`. Currently it happens automatically since calling `qml.ctrl( QubitUnitary )` creates the `ControlledQubitUnitary`. So I see it like we just describe this fact here.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: Make sense, but my intention for now is to describe state of the art here.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: PennyLane has this alias [here](https://github.com/grwlf/pennylane/blob/58ad1dfdc661fb58034225743503167e3006d86f/pennylane/ops/op_math/controlled_ops.py#L2000) so technically it is a valid gate. The purpose of this config is to describe what to do with operations, and this line should instructs it to decompose `CPhase` into a matrix. But we should do the same with `ControlledPhaseShift`, while in the current version we mark it as native. This is strange, I think we need to move `CPhase` to the `native` set as well.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: The idea of this config is to describe what to do with various PennyLane operations in order to make a program to work on this device. PL does have some ambiguities but I suggest to treat them as a matter of fact.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Hmm, but I thing that the Kokkos device does report this op in it `device.operations` set. I need to double-check.

grwlf: @vincentmr could you please check the purpose of [this](https://github.com/PennyLaneAI/pennylane-lightning/blob/1dae57c2ebe38793be929aa4e5c73d3ab1243b0f/pennylane_lightning/lightning_kokkos/lightning_kokkos.py#L157) ?

vincentmr: I'll repost something from Slack here, for confirmation from the comp-team and to converge on a simple message for everyone who asks. 

> The TOML files are for Catalyst use, they mean there is a special gate implementation at the C++ level. The `operations` set means there is support at the Python layer (at least) and this is used by the QNode for controlling decomposition. For example `C(GlobalPhase)` is not in the L-Kokkos TOML, because there is no C++ implementation. It is in `allowed_operations` however because the device uses a mix of Python/C++ functions to apply it, and hence it is supported at the Python level.

This is outdated by this PR, but the idea remains the same: the TOML files described the C++ backend period; the `operations` and `observables` frozensets describe the capabilities of the Python device, which may surpass the C++ backend (e.g. when something is implemented in Python), or which may be surpassed by the C++ backend (e.g. when bindings do not exist and there is no corresponding Python-level functionality).

grwlf: Ah, I see. Indeed, there are two PL-L APIs: the Python and the C++ one, and they need not to be the same of cause.
The first conclusion from this: we are using our validation procedure incorrectly when it comes to observables, becuase we compare `device.observables` (Python) with `toml.observables` (C++) and treat mismatch as an error.
The second conclusion: the toml schema 2, as of now, does not describe the Python API at all, so we can not make the strict validation against Python's `device`. Ideally we would need `operators.observables.unsupported`.

I will raise this question in Slack, for now I am going to remove the unsupported observables and relax the validation procedure.

grwlf: Done:
* Ali removed non-native observables
* Sergei opened a Catalyst PR relaxing the observables validation

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,105 @@
...
-        "DoubleExcitation",
-        "DoubleExcitationPlus",
-        "DoubleExcitationMinus",
-        "MultiRZ",
-        "QubitUnitary",
-        "GlobalPhase",
-]
+[operators.gates.native]
+
+CNOT                   = { properties = [ "invertible",                 "differentiable" ] }
```
vincentmr, line 7: It may be just a matter of convention, but do we want to mark non-parametric gates differentiable? One issue which might come up is that if one reads that and tries to apply Lightning's adjoint on CNOT, it will fail because CNOT has no generator (instead of an all-zero generator).

dime10: I suppose it is differentiable in the sense that it is allowed in adjoint Jacobian circuit. The purpose is to identify gates that need to be decomposed because they are not allowed in an adjoint Jacobian circuit 🤔

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,131 +1,109 @@
...
+IsingXY                = { properties = [ "invertible", "controllable", "differentiable" ] }
+IsingYY                = { properties = [ "invertible", "controllable", "differentiable" ] }
+IsingZZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
+MultiRZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliX                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliY                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PauliZ                 = { properties = [ "invertible", "controllable", "differentiable" ] }
+PhaseShift             = { properties = [ "invertible", "controllable", "differentiable" ] }
+QubitUnitary           = { properties = [ "invertible",                 "differentiable" ] }
+Rot                    = { properties = [ "invertible", "controllable", "differentiable" ] }
```
dime10, line 31: Apparently multi-parameter gates are not differentiable by lightning?

vincentmr: Nope.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: @vincentmr if we follow the convention saying that this file describes device APIs (to which I agree), and, specifically, that the `gates.native` section describes the C++ API of the device, then I think we should not assume direct interpretation by a user.

dime10: I agree with Vincent that it is mainly a matter of convention, but one advantage I see with attaching the "differentiable" tag to non-parametrized gates is that we don't need any additional information on whether to decompose the gate inside an adjoint-Jacobian circuit. With the other scheme we would additionally need to query the gate for whether it has differentiable parameters, and only if it does should we check for the "differentiable" tag in the TOML. But I'm open to go with that solution as well if it offers other benefits.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
dime10: @grwlf The "differentiable" tag needs to removed then from all multi-parameter gates in the lightning gate set, like `Rot`.

vincentmr: I think that is only `Rot`, `QubitUnitary` and their controlled versions.

grwlf: Removed Rot/QubitUnitary differentiable flag, @vincentmr could you please check?

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
vincentmr: I don't mind either way, do as you see fit.

vincentmr (approved):
Looks go to me, cheers!

grwlf:
@dime10 could you please take another look at this pr?

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,105 @@
...
+CRX                    = { properties = [ "invertible",                 "differentiable" ] }
+CRY                    = { properties = [ "invertible",                 "differentiable" ] }
+CRZ                    = { properties = [ "invertible",                 "differentiable" ] }
+CSWAP                  = { properties = [ "invertible",                 "differentiable" ] }
+CY                     = { properties = [ "invertible",                 "differentiable" ] }
+CZ                     = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationMinus  = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationPlus   = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitation       = { properties = [ "invertible",                 "differentiable" ] }
+GlobalPhase            = { properties = [ "invertible", "differentiable" ] }
```
dime10, line 20: Minor, but since you have this nice formatting in place:
```suggestion
GlobalPhase            = { properties = [ "invertible",                 "differentiable" ] }
```

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
dime10: This update was forgotten, and I agree that CPhase and ControlledPhaseShift should not be in different operation types. However, the weird part is that `CPhase` is not even its own class, so you would never see one on tape. I think the correct course of action is remove it from the TOML as well as removing it from the device class:
https://github.com/PennyLaneAI/pennylane-lightning/blob/a254c95ad5140c94a51c95f7fafab1b344757c45/pennylane_lightning/lightning_kokkos/lightning_kokkos.py#L116

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,105 @@
...
+OrbitalRotation        = {}
+PSWAP                  = {}
+QubitCarry             = {}
+QubitSum               = {}
+SISWAP                 = {}
+SQISW                  = {}
+SX                     = {}
+
+# Observables supported by the device
+[operators.observables]
```
dime10, line 75: The case of `"Sum", "SProd", "Prod", "Exp"` is interesting. I see why they were removed (there is no native API for them), but Catalyst will convert them to the old style observables after the tape is processed, so we'll have to remember in our decomposition/verification functions that they are not flagged unsupported unless the device also doesn't support the equivalent old style (Hamiltonian, Tensor, etc) observable.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,105 @@
...
+
+Identity               = {}
+PauliX                 = {}
+PauliY                 = {}
+PauliZ                 = {}
+Hadamard               = {}
+Hermitian              = {}
+Hamiltonian            = {}
+SparseHamiltonian      = {}
```
dime10, line 84: No tensor observable support on the kokkos device?

pennylane_lightning/lightning_qubit/lightning_qubit.toml
dime10: Unless the CNOT C++ kernel accepts arbitrary _additional_ control wires it shouldn't be marked controllable, but if it does I don't see why we couldn't add it? @grwlf

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,131 +1,109 @@
...
+CRZ                    = { properties = [ "invertible",                 "differentiable" ] }
+CSWAP                  = { properties = [ "invertible",                 "differentiable" ] }
+CY                     = { properties = [ "invertible",                 "differentiable" ] }
+CZ                     = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationMinus  = { properties = [ "invertible", "controllable", "differentiable" ] }
+DoubleExcitationPlus   = { properties = [ "invertible", "controllable", "differentiable" ] }
+DoubleExcitation       = { properties = [ "invertible", "controllable", "differentiable" ] }
+GlobalPhase            = { properties = [ "invertible", "controllable", "differentiable" ] }
+Hadamard               = { properties = [ "invertible", "controllable", "differentiable" ] }
+Identity               = { properties = [ "invertible",                 "differentiable" ] }
```
dime10, line 20: Just curious, but Identity is actually supported as a no-op on the lightning devices? :) @vincentmr

pennylane_lightning/lightning_qubit/lightning_qubit.toml
dime10: Yep, to help @vincentmr understand the subtlety here: `PauliX(control_wires=...)` and `MultiControlledX` are two different classes and thus share no apparent relation to each other (without explicit equivalence rules). So unless the C++ API has a named gate called `MultiControlledX` we cannot just pass it down as if it were a natively supported gate.
The compiler however may chose to turn `MultiControlledX` into `PauliX(control_wires=...)` if it sees the latter is supported.

dime10 (changes requested):
Looks great, although I think I identified a few specs worth looking over again.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
dime10, line 84: @grwlf In lightning.qubit you marked the observables differentiable, does that not apply here?

rmoyard:
Please let's wait on this one to be merged https://github.com/PennyLaneAI/pennylane-lightning/pull/665 first.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: @dime10 CNOT is [absent](https://github.com/PennyLaneAI/pennylane-lightning/blob/37f19012eab034db981969c42e4c5705825fcb05/pennylane_lightning/core/src/gates/GateOperation.hpp#L73) from the Controlled enum. I believe it means it is not controllable.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: I think the Identity is [supported](https://github.com/PennyLaneAI/pennylane-lightning/blob/37f19012eab034db981969c42e4c5705825fcb05/pennylane_lightning/core/src/gates/GateOperation.hpp#L30), but it is not controllable. Vincent, please correct me if I am wrong.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,124 +1,105 @@
...
+Hadamard               = { properties = [ "invertible",                 "differentiable" ] }
+Identity               = { properties = [ "invertible",                 "differentiable" ] }
+IsingXX                = { properties = [ "invertible",                 "differentiable" ] }
+IsingXY                = { properties = [ "invertible",                 "differentiable" ] }
+IsingYY                = { properties = [ "invertible",                 "differentiable" ] }
+IsingZZ                = { properties = [ "invertible",                 "differentiable" ] }
+MultiRZ                = { properties = [ "invertible",                 "differentiable" ] }
+PauliX                 = { properties = [ "invertible",                 "differentiable" ] }
+PauliY                 = { properties = [ "invertible",                 "differentiable" ] }
+PauliZ                 = { properties = [ "invertible",                 "differentiable" ] }
```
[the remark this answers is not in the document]
grwlf: I marked them so because the original config file has global controllable flag set to False kokkos devices.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
vincentmr: There is, it is part of the [base](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/observables/Observables.hpp#L318) API.

grwlf: Indeed, they are differentiable (checks are in this [internal doc](https://github.com/XanaduAI/sergei-environment/blob/master/src/qcontrol/toml_diff.md)). Thanks, David!

grwlf: FYI kokkos device [contains](https://github.com/PennyLaneAI/pennylane-lightning/blob/50864f505071ffb8a67430cebc93a019eee0a076/pennylane_lightning/core/src/simulators/lightning_kokkos/observables/tests/Test_ObservablesKokkos.cpp#L94) tests mentioning tensor observables

dime10: Right so let's add it :)

grwlf:
Hi all, I have run Catalyst tests against the recent version of this PR and discovered a number of device/spec validation failures. I have updates the PR to pass validation and put FIXMEs near related lines. @vincentmr could you please check it?
Here are the list:
* Lightning.qubit: Python API of the device reports no support for `BasisState, QubitStateVector, StatePrep` operations in the Python API ([link](https://github.com/PennyLaneAI/pennylane-lightning/blob/3104634ce7d1a5bc8cedb7a72640c4e84cddf683/pennylane_lightning/lightning_qubit/lightning_qubit.toml#L52)). Should we mark them as supported in Python?
* Lightning.kokkos: Python API reports support for `C(GlobalPhase)` while in spec it is marked as non-controllable. Do you know what is the situation?
* A few tests fails even after my fixups. Catalyst is able to compile and run a program, involving finiteshots-Hermitian, while PennyLane rejects it. [link](https://github.com/PennyLaneAI/catalyst/blob/6c7b935d8afcebc7a5a58f82219a389bec204d1a/frontend/test/pytest/test_measurements_shots_results.py#L340). What is the PLL status of this mode?


CC @dime10

pennylane_lightning/lightning_gpu/lightning_gpu.toml
```diff
@@ -6,18 +6,14 @@ name = "lightning.gpu"
...
         "PauliZ",
         "Hadamard",
         "Hermitian",
-        "Identity",
-        "SparseHamiltonian",
         "Hamiltonian",
-        "Sum",
-        "SProd",
-        "Prod",
-        "Exp",
```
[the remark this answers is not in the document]
grwlf: OK I have done my best to update this file, but I realized that I can not check it immediately: I have no cuda device available locally. I'll request an access to some remote device and update later.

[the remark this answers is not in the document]
grwlf: The original toml config has `qjit_compatible=false`. Does it mean that there are currently no ways to validate the config? (I am only aware of the Catalyst-based validation procedure)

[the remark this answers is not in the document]
dime10: Thank you! Yes Catalyst doesn't support lightning.gpu yet so we can't test the device. However doing most of the legwork to keep it up to date with the toml format is still a good idea I think, simplifies the task going forward.

vincentmr:
> Hi all, I have run Catalyst tests against the recent version of this PR and discovered a number of device/spec validation failures. I have updated the PR to make it pass the validation and put FIXMEs near the related lines. @vincentmr could you please check them? Here are the list:
> 
> * Lightning.qubit: Python API of the device reports no support for `BasisState, QubitStateVector, StatePrep` operations, so I disabled them in the spec for now ([link](https://github.com/PennyLaneAI/pennylane-lightning/blob/3104634ce7d1a5bc8cedb7a72640c4e84cddf683/pennylane_lightning/lightning_qubit/lightning_qubit.toml#L52)). Should we mark them supported by updating PennyLane-lightnig and re-enable in the config?

I think @erick-xanadu is taking care of this in #698 .

> * Lightning.kokkos: Python API reports support for `C(GlobalPhase)` while in spec it was marked as non-controllable. If it is true, it would be the only controllable gate supported by the kokkos device. Do you know what is the situation here?

Supporting this requires calling [global_phase_diagonal](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/_serialize.py#L429) from Python, so it is not supported by the C++ backend at the moment.

> * A few tests fails even after my fixups. Catalyst is able to compile and run a program, involving finiteshots-Hermitian, while pure-PennyLane lightning.qubit device rejects it. [link](https://github.com/PennyLaneAI/catalyst/blob/6c7b935d8afcebc7a5a58f82219a389bec204d1a/frontend/test/pytest/test_measurements_shots_results.py#L340). What is the PLL status for this mode?

I think this was added in #569 but without Python bindings. So it's only for you guys ;)

> 
> CC @dime10

pennylane_lightning/lightning_gpu/lightning_gpu.toml
```diff
@@ -1,121 +1,103 @@
...
-	"Sample",
-	"Counts",
-]
+
+Expval                 = {}
+Var                    = {}
+Probs                  = {}
+State                  = { condition = [ "analytic" ] }
+Sample                 = { condition = [ "finiteshots" ] }
+Count                  = { condition = [ "finiteshots" ] }
```
rmoyard, line 94:
```suggestion
Counts                  = { condition = [ "finiteshots" ] }
```

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,132 +1,115 @@
...
-	"Sample",
-	"Counts",
-]
+
+Expval                 = {}
+Var                    = {}
+Probs                  = {}
+State                  = { condition = [ "analytic" ] }
+Sample                 = { condition = [ "finiteshots" ] }
+Count                  = { condition = [ "finiteshots" ] }
```
rmoyard, line 97:
```suggestion
Counts                  = { condition = [ "finiteshots" ] }
```

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Added [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/642/files#diff-45f8d82a726b88e990143dd273747c086a5dcfaec6a302400d07e0c25e6d1415R88)

grwlf:
I have run the lightning and kokkos cross-checks using the following components:
* Catalyst https://github.com/PennyLaneAI/catalyst/blob/1babc99e798278672b568028a644618420f0e9f1/README.md#L1
* PennyLane-lightning https://github.com/PennyLaneAI/pennylane-lightning/blob/d9507607c0d77500f20dc710fd00c118ad54d038/README.rst#L1

As far as I see, after a minor Catalyst fixup (the PR is linked above), all checks pass.

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
dime10: It is not a class name, it is just a local variable name. So this gate doesn't exist!

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,125 +1,111 @@
...
+CRY                    = { properties = [ "invertible",                 "differentiable" ] }
+CRZ                    = { properties = [ "invertible",                 "differentiable" ] }
+CSWAP                  = { properties = [ "invertible",                 "differentiable" ] }
+CY                     = { properties = [ "invertible",                 "differentiable" ] }
+CZ                     = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationMinus  = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitationPlus   = { properties = [ "invertible",                 "differentiable" ] }
+DoubleExcitation       = { properties = [ "invertible",                 "differentiable" ] }
+# FIXME: device reports it is controllable. Why is that?
+GlobalPhase            = { properties = [ "invertible", "controllable", "differentiable" ] }
```
dime10, line 22: Vincent commented this was not controllable, what was the resolution?

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -1,129 +1,112 @@
...
-        "Hamiltonian",
-        "Sum",
-        "SProd",
-        "Prod",
-        "Exp",
-        "LinearCombination",
-]
+CNOT                   = { properties = [ "invertible",                 "differentiable" ] }
+ControlledPhaseShift   = { properties = [ "invertible",                 "differentiable" ] }
+CPhase                 = { properties = [ "invertible",                 "differentiable" ] }
```
dime10, line 7: Ali is creating a PR that removes `CPhase` from the device list.
```suggestion
```

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
```diff
@@ -1,125 +1,111 @@
...
-        "DoubleExcitationMinus",
-        "MultiRZ",
-        "QubitUnitary",
-        "GlobalPhase",
-]
+[operators.gates.native]
+
+CNOT                   = { properties = [ "invertible",                 "differentiable" ] }
+ControlledPhaseShift   = { properties = [ "invertible",                 "differentiable" ] }
+CPhase                 = { properties = [ "invertible",                 "differentiable" ] }
```
dime10, line 9:
```suggestion
```

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Oh, right. Hmm, I think we need to hardcode a special case for schema 2 to pass validation in Catalyst then.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: @maliasadi can you link your pr here?

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
dime10: Hmm, yes the toml is about the low-level capabilities. The validation is mainly there so the toml doesn't go out of sync too easily. M)aybe it would be better to go back to the simple validation we had before? (that is do not compare Adjoint(Gate) or C(Gate)

grwlf: I would suggest to introduce a smaller relaxation since it is the only exception we have. See https://github.com/PennyLaneAI/catalyst/pull/718

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: OK, got it, we can remove it from toml schema. Still, there are minor problems: 1) CPhase exists in the toml schema 1 configs 2) CPhase does exists in PL's `device.operations` as a separate gate. You said in other discussion that Ali is addressing (2). Meanwhile I have addressed (1) in  https://github.com/PennyLaneAI/catalyst/pull/718

pennylane_lightning/lightning_qubit/lightning_qubit.toml
grwlf: CPhase is removed

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Done

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
grwlf: Meanwhile I have removed the controllable flag here.

dime10 (approved):
Good to go 👍

maliasadi (approved):
:raised_hands:

PR #644, opened 2024-03-15
`LightningQubit` VJP support

mudit2812:
**Context:**
This PR adds adjoint VJP computation support to `lightning.qubit` with the new API.

* Add `LightningQubit.compute_vjp`, `supports_vjp`, and `execute_and_compute_vjp` methods along with appropriate helper functions.
* Update `tests/test_vjp.py` to stop skipping new API devices.
* Add new tests to `tests/new_api/test_device.py`.

**Benefits:**
* QNodes using `LightningQubit` can now set the `device_vjp` argument to automatically use the VJP pipeline for differentiation.

**Possible Drawbacks:**

**Related GitHub Issues:**


.github/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Add `lightning.qubit2` device which uses the new device API.
   [(#607)](https://github.com/PennyLaneAI/pennylane-lightning/pull/607)
   [(#628)](https://github.com/PennyLaneAI/pennylane-lightning/pull/628)
+  [(#644)](https://github.com/PennyLaneAI/pennylane-lightning/pull/644)
```
AmintorDusko, line 17: I think we can have it in as a new entry.

mudit2812:
[sc-59045]

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -115,6 +116,55 @@ def simulate_and_jacobian(circuit: QuantumTape, state: LightningStateVector, bat
...
+        state (LightningStateVector): handle to Lightning state vector
+        batch_obs (bool): Determine whether we process observables in parallel when
+            computing the VJP. This value is only relevant when the lightning
+            qubit is built with OpenMP.
+    Returns:
+        TensorLike: The VJP of the quantum script
+    """
+    circuit = circuit.map_to_standard_wires()
+    state.reset_state()
+    final_state = state.get_final_state(circuit)
```
albi3ro, line 138: Given this could potentially be expensive, we could potentially keep track of the `tape.hash` that corresponds to the current state, and only reseting and recalculating the state when the hash is different.

mudit2812:
[Lots of test failures in Pennylane](https://github.com/PennyLaneAI/pennylane/actions/runs/8393761839/job/22989473218?pr=5434) due to lack of VJP support in Lightning. I am skipping such tests for now, but they should stop being skipped once this PR is merged. See the files changed in PL [#5434](https://github.com/PennyLaneAI/pennylane/pull/5434) to find which tests need to be updated.

pennylane_lightning/lightning_qubit/lightning_qubit.py
AmintorDusko: @albi3ro, you mean that calculating the final state is expensive, right?
I think it would be a smart move to hash in order to avoid re-calculating. However, wouldn't it make more sense to hash `tape.operations` instead? This is because we essentially use `tape.operations` to build the state vector, and even if the observables change, the state vector would still be the same. The same logic applies to both the simulate and jacobian parts, right?

mudit2812: I'm not sure how significant the benefit would be for caching using `tape.operations` hash vs `tape.hash`. However, if we consider doing this, same change will also benefit `default.qubit`.

mudit2812:
TODO in pennylane: Uncomment interface tests using `lightning.qubit` with `device_vjp=True`. These were commented out due to `LightningQubit` not being able to compute VJPs.

tests/new_api/test_device.py
```diff
@@ -729,3 +726,337 @@ def test_derivatives_tape_batch(self, phi, execute_and_derivatives, batch_obs):
         assert len(jacs[0]) == len(expected_jac[0])
         assert np.allclose(jacs[0], expected_jac[0])
         assert np.allclose(jacs[1], expected_jac[1])
+
+
+@pytest.mark.parametrize("batch_obs", [True, False])
+class TestVJP:
```
mudit2812, line 732: Most of these tests are copies of the `TestDerivatives` class. I _could_ have parametrized the tests in that class, but it would've made the tests a lot less readable so I decided to make a separate class with tests dedicated to VJPs

pennylane_lightning/lightning_qubit/lightning_qubit.py
AmintorDusko: I think we can let this for now, and come back in the future if needed.

AmintorDusko (approved):
Nice job! Thanks!

vincentmr (approved):
Looks good to me, nice work @mudit2812 !

PR #647, opened 2024-03-19
Fix isort config

vincentmr:
### Before submitting

**Context:**
`isort` produces different results across Python version and environments.

Fix `isort` config and use GHA for isort and black. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


.github/workflows/format.yml
```diff
@@ -14,23 +14,22 @@ jobs:
...
+          requirements-files: "requirements-dev.txt"
+          sort-paths: "./pennylane_lightning/ ./mpitests ./tests"
 
-      - name: Run Black
-        run: python -m black -l 100 pennylane_lightning/ tests/ --check --verbose
+      - uses: psf/black@stable
+        with:
+          options: "-l 100 --check --verbose"
+          src: "./pennylane_lightning/ ./mpitests ./tests"
+          version: "23.7.0"
```
AmintorDusko, line 32: We usually avoid third-party actions, but I guess we can rely on that.

Makefile
```diff
@@ -113,7 +113,7 @@ format-cpp:
 	./bin/format $(CHECK) ./pennylane_lightning
 
 format-python:
-	isort --profile black ./pennylane_lightning/ ./mpitests ./tests $(CHECK)
+	isort --py 311 --profile black -l 100 -p pennylane_lightning ./pennylane_lightning/ ./mpitests ./tests $(CHECK)
```
AmintorDusko, line 116: So with this, we can fix the default to follow python3.11 version, right?

AmintorDusko (approved):
Hey @vincentmr, I assume you tested it with Python 3.9 (supported by us) and observed no more changes coming from it, right? In this case, thank you very much for that. Now that we have a consistent development environment, I'm happy. I will approve, but please wait for #646 to be merged, before merging this PR.

.github/workflows/format.yml
vincentmr: Yes, but they are the official isort/black collab actions so I thought it would be ok.

Makefile
vincentmr: We needed all those options before things stabilized. I think the main issue was actually `-p pennylane_lightning` because sometimes it would be recognize as first party and others third party.

AmintorDusko: Interesting. Thanks!

.github/workflows/format.yml
AmintorDusko: No problems for me.

maliasadi (changes requested):
@vincentmr What about using these custom options in `pyproject.toml` so that we can reuse them in our GH action. Look at what we've done in Catalyst [for an example](https://github.com/PennyLaneAI/catalyst/blob/main/pyproject.toml).

Let's not use 3rd-part actions in our recipes where we can simply install these packages from PyPI ;)

vincentmr:
> @vincentmr What about using these custom options in `pyproject.toml` so that we can reuse them in our GH action. Look at what we've done in Catalyst [for an example](https://github.com/PennyLaneAI/catalyst/blob/main/pyproject.toml).

Since this is only for `isort` at this point, I think the CLI is still doing the job.

> Let's not use 3rd-part actions in our recipes where we can simply install these packages from PyPI ;)

I would agree with the exception when the action is from the package collab, as in the present case.

.github/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@
 
 ### Bug fixes
 
+* Specify `isort` `--py` (Python version) and `-l` (max line length) to stabilize `isort` across Python versions and environments.
```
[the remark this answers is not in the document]
vincentmr: I think it still is, I saw a case where `isort` will `black -l` is more opiniated than `black` itself.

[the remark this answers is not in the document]
vincentmr: I would therefore keep it for good measure.

.github/workflows/format.yml
```diff
@@ -14,23 +14,22 @@ jobs:
...
     steps:
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.11'
 
-      - name: Install dependencies
-        run:
-            python -m pip install click==8.0.4 black==23.7.0 isort==5.13.2
+      - uses: actions/checkout@v3
```
rashidnhm, line 18:
```suggestion
      - uses: actions/checkout@v4
```
Let's drop usage of `checkout@v3` across all our workflows as it uses a deprecated version of NodeJS

.github/workflows/format.yml
```diff
@@ -14,23 +14,22 @@ jobs:
...
 
-      - name: Checkout PennyLane-Lightning
-        uses: actions/checkout@v3
-
-      - name: Run isort
-        run: python -m isort --profile black ./pennylane_lightning/ ./mpitests ./tests --check --diff
+      - name: python-isort
+        uses: isort/isort-action@v1.1.0
```
rashidnhm, line 21: If the action is provided by a trustworthy party, then *it should be* safe for use. I did look through the source code of the action and it seems simple enough, and does what we were doing previously with cli call

.github/workflows/format.yml
```diff
@@ -14,23 +14,22 @@ jobs:
...
+        with:
+          configuration: --py 311 --profile black -l 100 -p pennylane_lightning --check --diff
+          isort-version: 5.13.2
+          requirements-files: "requirements-dev.txt"
+          sort-paths: "./pennylane_lightning/ ./mpitests ./tests"
```
rashidnhm, line 26:
```suggestion
          requirements-files: 'requirements-dev.txt'
          sort-paths: './pennylane_lightning/ ./mpitests ./tests'
```

rashidnhm (commented):
Would it make sense to use `make format-python` within CI instead of the actions? One benefit of doing that is it would ensure we're running identical code across both CI and local development environments.

vincentmr:
> Would it make sense to use `make format-python` within CI instead of the actions? One benefit of doing that is it would ensure we're running identical code across both CI and local development environments.

Do you mean `make format-python check=1`?

Makefile
```diff
@@ -113,8 +115,8 @@ format-cpp:
 	./bin/format $(CHECK) ./pennylane_lightning
 
 format-python:
-	isort --profile black ./pennylane_lightning/ ./mpitests ./tests $(CHECK)
-	black -l 100 ./pennylane_lightning/ ./mpitests ./tests $(CHECK)
+	isort --py 311 --profile black -l 100 -p pennylane_lightning ./pennylane_lightning/ ./mpitests ./tests $(ICHECK) $(VERBOSE)
```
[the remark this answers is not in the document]
maliasadi: chatted offline.. let's keep it for consistency

rashidnhm (approved):

PR #649, opened 2024-03-19
Add support for `LinearCombination`

mudit2812:
**Context:**
PennyLane is switching to the new Operator Arithmetic as the default. This PR updates Lightning to be compatible with this change.

* Update `QuantumScriptSerializer` to be smarter with new op math, and to accommodate `LinearCombination`.
* Update tests with new fixtures that run tests on both new and old op arithmetic
* Add support to all devices for detecting Projectors in `Sum`, `Prod`, and `SProd` instances when validating for adjoint diff. This function was the same between LKokkos and LGPU, so I moved this logic to `LightningBase`.

**Benefits:**
New op math as the default works with Lightning.

**Possible Drawbacks:**
Serialization with new op math has some weird nesting. Will be addressed in a follow up PR.

**Related GitHub Issues:**

Related work in PennyLane:
[#5335](https://github.com/PennyLaneAI/pennylane/pull/5335), [#5216](https://github.com/PennyLaneAI/pennylane/pull/5216), [#5269](https://github.com/PennyLaneAI/pennylane/pull/5269)


mudit2812:
[sc-58734]

pennylane_lightning/core/_serialize.py
```diff
@@ -261,16 +263,16 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
...
-            return self._hamiltonian(observable, wires_map)
-        if observable.name == "SparseHamiltonian":
-            return self._sparse_hamiltonian(observable, wires_map)
         if isinstance(observable, (PauliX, PauliY, PauliZ, Identity, Hadamard)):
             return self._named_obs(observable, wires_map)
         if observable._pauli_rep is not None:
             return self._pauli_sentence(observable._pauli_rep, wires_map)
+        if isinstance(observable, (Tensor, Prod)):
+            return self._tensor_ob(observable, wires_map)
+        if observable.name in ("Hamiltonian", "LinearCombination"):
```
trbromley, line 272: Could this just be an `isinstance` check?

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -199,6 +199,7 @@ def _mebibytesToBytes(mebibytes):
         "Hadamard",
         "SparseHamiltonian",
         "Hamiltonian",
+        "LinearCombination",
```
trbromley, line 202: So we need this because, rather than instance checks, we are relying on operator names? And `LinearCombination`'s name is not `Hamiltonian`?

tests/test_adjoint_jacobian.py
```diff
@@ -215,7 +215,7 @@ def test_proj_unsupported(self, dev):
 
         with qml.tape.QuantumTape() as tape:
             qml.CRX(0.1, wires=[0, 1])
-            qml.expval(qml.Projector([0], wires=[0]) @ qml.PauliZ(0))
+            qml.expval(qml.Projector([0], wires=[0]) @ qml.PauliZ(1))
```
mudit2812, line 218: Got [this failure](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8363827153/job/22897751458?pr=649) on this test, as well as the same test from `test_vjp.py`. Why was this test working to begin with?

mudit2812:
Some LGPU test failures [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8365545642/job/22903678794?pr=649). The failing test compares jacobian results against `default.qubit`, and `default.qubit` is the cause of the error, not LGPU. Not really sure what's happening as I can't recreate the error locally.

pennylane_lightning/lightning_gpu/lightning_gpu.py
mudit2812: Yes. Operator.name is `Operator.__class__.__name__` by default.

pennylane_lightning/core/_serialize.py
mudit2812: Definitely can be.

tests/test_adjoint_jacobian.py
mudit2812: Was happening because the adjoint validation wasn't catching `Projector`s inside `Prod`s, so the C++ layer raised the error in the failure. With `Tensor` the validation error would get raised way before we reached the C++ layer.

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 pip~=22.0
-git+https://github.com/PennyLaneAI/pennylane.git@master
+git+https://github.com/PennyLaneAI/pennylane.git@lightning-op-math
```
mudit2812, line 2: Leaving this here for later. Should be done AFTER two approvals
```suggestion
git+https://github.com/PennyLaneAI/pennylane.git@master
```

mudit2812:
> Some LGPU test failures [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8365545642/job/22903678794?pr=649). The failing test compares jacobian results against `default.qubit`, and `default.qubit` is the cause of the error, not LGPU. Not really sure what's happening as I can't recreate the error locally.

This is now fixed.

tests/test_vjp.py
```diff
@@ -98,8 +98,11 @@ def test_multiple_measurements(self, tol, dev):
         dy = np.array([1.0, 2.0, 3.0])
         tape1.trainable_params = {1, 2, 3}
 
+        ham = qml.Hamiltonian(dy, [qml.PauliX(0), qml.PauliY(1), qml.PauliY(1)])
+        print(ham)
+
         with qml.tape.QuantumTape() as tape2:
-            ham = qml.Hamiltonian(dy, [qml.PauliX(0), qml.PauliY(1), qml.PauliY(1)])
+            ham
```
vincentmr, line 105: Can we remove `print` and move `qml.Ham` back in the `QuantumTape` scope?

AmintorDusko: These sneak prints!!!!

AmintorDusko: done!

pennylane_lightning/core/_serialize.py
```diff
@@ -254,23 +257,31 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
...
             return self._sparse_hamiltonian(observable, wires_map)
         if isinstance(observable, (PauliX, PauliY, PauliZ, Identity, Hadamard)):
             return self._named_obs(observable, wires_map)
-        if observable._pauli_rep is not None:
-            return self._pauli_sentence(observable._pauli_rep, wires_map)
+        if observable.pauli_rep is not None:
+            return self._pauli_sentence(observable.pauli_rep, wires_map)
+        # if isinstance(observable, (Prod, Sum)):
+        #     return self._hamiltonian(observable, wires_map)
         return self._hermitian_ob(observable, wires_map)
```
maliasadi, line 285: what about raising a `QuantumFunctionError` error as the default behaviour for this method and conditionally patch to `_hermitian_ob`? :thinking: 

Something like,
```suggestion
       if isinstance(observable, (Hermitian, Projector, any_other_obs))
            return self._hermitian_ob(observable, wires_map)
       raise qml.QuantumFunctionError(f"cannot convert observable {observable}")
```

pennylane_lightning/core/_serialize.py
```diff
@@ -183,19 +184,21 @@ def _named_obs(self, observable, wires_map: dict = None):
 
     def _hermitian_ob(self, observable, wires_map: dict = None):
         """Serializes a Hermitian observable"""
-        assert not isinstance(observable, Tensor)
+        assert not isinstance(observable, (Tensor, Prod))
```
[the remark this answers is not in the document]
mudit2812: Yeah I don't see the point of having this assertion.

> What about other obs like Hamiltonian, SProd, and Sum?

If we want to keep the assertion it makes sense to add `Hamiltonian` to the list, but `Sum` and `SProd` are handled by this branch. I think we should update `QuantumScriptSerializer._ob` to dispatch to `_hamiltonian_ob` for `SProd` and `Sum` though. What are your thoughts? cc @vincentmr @AmintorDusko

pennylane_lightning/core/_serialize.py
mudit2812: Hmm. I personally think it's best not to raise an error as `_hermitian_ob` is a fallback for many cases (such as `Projector` as we saw earlier)

pennylane_lightning/core/_serialize.py
```diff
@@ -254,23 +257,31 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
         terms = [self._pauli_word(pw, wires_map) for pw in pwords]
         coeffs = np.array(coeffs).astype(self.rtype)
 
+        # TODO: Add this
+        # if len(terms) == 1 and coeffs[0] == 1.0:
+        #     return terms[0]
```
mudit2812, line 262: Optimization for later: Unwrap single term Hamiltonians.

pennylane_lightning/core/_serialize.py
```diff
@@ -254,23 +257,31 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
...
+
         if self.split_obs:
             return [self.hamiltonian_obs([c], [t]) for (c, t) in zip(coeffs, terms)]
         return self.hamiltonian_obs(coeffs, terms)
 
-    # pylint: disable=protected-access
+    # pylint: disable=protected-access, too-many-return-statements
     def _ob(self, observable, wires_map: dict = None):
         """Serialize a :class:`pennylane.operation.Observable` into an Observable."""
```
mudit2812, line 270: Optimization for later: Reordering priority as follows: named obs -> Hamiltonian/LinearCombination (Maybe also Sum and SProd) -> Pauli rep -> Tensor/Prod -> Hermitian

Might make sense to prioritize Pauli rep over Hamiltonian as well.

pennylane_lightning/core/_serialize.py
```diff
@@ -254,23 +257,31 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
...
         if observable.name == "SparseHamiltonian":
             return self._sparse_hamiltonian(observable, wires_map)
         if isinstance(observable, (PauliX, PauliY, PauliZ, Identity, Hadamard)):
             return self._named_obs(observable, wires_map)
-        if observable._pauli_rep is not None:
-            return self._pauli_sentence(observable._pauli_rep, wires_map)
+        if observable.pauli_rep is not None:
+            return self._pauli_sentence(observable.pauli_rep, wires_map)
+        # if isinstance(observable, (Prod, Sum)):
+        #     return self._hamiltonian(observable, wires_map)
```
mudit2812, line 284: Optimization for later

pennylane_lightning/core/_serialize.py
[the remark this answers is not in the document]
vincentmr: I agree, I think we can remove that assertion. 

Since it's almost Friday, I'd say if a change doesn't work off the bat, let's follow-up in Q2. For `SProd` and `Sum`, can't the underlying observables be anything, including Hermitian observables? In this case, we'd need more checks to dispatch to `_hamiltonian` no?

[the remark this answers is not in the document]
mudit2812: I don't think we should. The recursive structure of the serialization should accommodate for this. We might just end up with weird looking serializations, but they will still be correct.

pennylane_lightning/core/_serialize.py
mudit2812: This Catalyst function is a good reference for what order to use: https://github.com/PennyLaneAI/catalyst/blob/f016a31f69d1b8a84bc9612af1bc64f0575506e9/frontend/catalyst/jax_tracer.py#L479

mudit2812:
Looks like codecov is failing due to MPI tests being skipped. Is this new?

pennylane_lightning/core/_serialize.py
```diff
@@ -183,19 +184,21 @@ def _named_obs(self, observable, wires_map: dict = None):
...
+        assert not isinstance(observable, (Tensor, Prod))
 
         wires = [wires_map[w] for w in observable.wires] if wires_map else observable.wires.tolist()
         return self.hermitian_obs(matrix(observable).ravel().astype(self.ctype), wires)
 
     def _tensor_ob(self, observable, wires_map: dict = None):
         """Serialize a tensor observable"""
-        assert isinstance(observable, Tensor)
-        return self.tensor_obs([self._ob(obs, wires_map) for obs in observable.obs])
+        assert isinstance(observable, (Tensor, Prod))
```
vincentmr, line 194: Should we raise an error here or simply `assert`?

pennylane_lightning/core/_serialize.py
vincentmr: Good idea. Should we use `qml.math.allclose` when comparing floats (`coeffs[0]`) however?

pennylane_lightning/core/_serialize.py
vincentmr: > Might make sense to prioritize Pauli rep over Hamiltonian as well.

Yes, especially if specialized routines like [PauliSentence.dot](https://docs.pennylane.ai/en/stable/code/api/pennylane.pauli.PauliSentence.html#pennylane.pauli.PauliSentence.dot).

pennylane_lightning/core/_serialize.py
vincentmr: Let's do this when we stumble upon such cases.

pennylane_lightning/core/_serialize.py
vincentmr: Does it mean `Prod` won't be handled by `_tensor_ob`? Or not, because operators are allowed to act on overlapping wires?

pennylane_lightning/core/lightning_base.py
```diff
@@ -306,18 +307,69 @@ def _process_jacobian_tape(
             "obs_idx_offsets": obs_idx_offsets,
         }
 
+    @staticmethod
+    def _assert_adjdiff_no_projectors(observable):
```
vincentmr, line 311: Neat. This could be used to recursively check for other unsupported observables.

pennylane_lightning/core/lightning_base.py
```diff
@@ -306,18 +307,69 @@ def _process_jacobian_tape(
...
+
+        # The return_type of measurement processes must be expectation
+        if any(m.return_type is not Expectation for m in measurements):
+            raise qml.QuantumFunctionError(
+                "Adjoint differentiation method does not support expectation return type "
+                "mixed with other return types"
+            )
+
+        for measurement in measurements:
+            LightningBase._assert_adjdiff_no_projectors(measurement.obs)
```
vincentmr, line 371: Nothing to do here, but if this becomes somehow expensive, because of recursive depth for example, I wonder whether inputting `measurements` would help.

pennylane_lightning/core/src/bindings/Bindings.hpp
```diff
@@ -350,6 +350,8 @@ void registerBackendAgnosticObservables(py::module_ &m) {
         .def("__repr__", &HermitianObs<StateVectorT>::getObsName)
         .def("get_wires", &HermitianObs<StateVectorT>::getWires,
              "Get wires of observables")
+        .def("get_matrix", &HermitianObs<StateVectorT>::getMatrix,
```
vincentmr, line 353: I don't mind adding this, but where are these new bindings used?

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -235,6 +238,26 @@ def accepted_observables(obs: qml.operation.Operator) -> bool:
     return obs.name in _observables
 
 
+def adjoint_observables(obs: qml.operation.Operator) -> bool:
```
vincentmr, line 241: Is the function still necessary after the introduction of `_assert_adjdiff_no_projectors`?

tests/conftest.py
```diff
@@ -152,6 +152,34 @@ def _device(wires, shots=None):
...
+def use_new_opmath():
+    with qml.operation.enable_new_opmath_cm() as cm:
+        yield cm
+
+
+@pytest.fixture(
+    params=[qml.operation.disable_new_opmath_cm, qml.operation.enable_new_opmath_cm],
+    scope="function",
+)
+def use_legacy_and_new_opmath(request):
```
vincentmr, line 175: Nice.

tests/new_api/test_device.py
```diff
@@ -29,6 +29,7 @@
     _supports_adjoint,
     accepted_observables,
     adjoint_measurements,
+    adjoint_observables,
```
vincentmr, line 32: Should we test `_assert_adjdiff_no_projectors` instead?

tests/new_api/test_device.py
```diff
@@ -269,11 +292,11 @@ def test_preprocess(self, adjoint):
             qml.probs(op=qml.Z(2)),
             qml.expval(qml.Z(2)),
             qml.var(qml.X(2)),
-            qml.expval(qml.sum(qml.X(0), qml.Z(0))),
+            qml.expval(qml.X(0) + qml.Z(0)),
```
vincentmr, line 295: Nice.

tests/test_adjoint_jacobian.py
vincentmr: Can we add a case with overlapping wires if this is supposed to be supported? I guess it should as this simply means multiplying the matrices under the hood.

pennylane_lightning/core/_serialize.py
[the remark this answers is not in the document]
AmintorDusko: Sounds good to me! Let's focus on the Q1 deliverables.  But please create follow-up stories for Q2.

pennylane_lightning/core/_serialize.py
AmintorDusko: Sounds god, @mudit2812.

pennylane_lightning/core/_serialize.py
AmintorDusko: Nice!

tests/test_serialize.py
```diff
@@ -234,26 +245,44 @@ def test_hamiltonian_return(self, use_csingle, wires_map):
...
+                    tensor_prod_obs(
+                        [
+                            hermitian_obs(np.eye(4, dtype=c_dtype).ravel(), [0, 1]),
+                            named_obs("PauliY", [2]),
+                        ]
+                    ),
+                    tensor_prod_obs([named_obs("PauliX", [0]), named_obs("PauliY", [2])]),
+                    hermitian_obs(np.ones(64, dtype=c_dtype), [0, 1, 2]),
+                ],
+            )
```
vincentmr, line 278:
```suggestion
        s_expected = hamiltonian_obs(
            np.array([0.3, 0.5, 0.4], dtype=r_dtype),
            [
                tensor_prod_obs(
                    [
                        hermitian_obs(np.eye(4, dtype=c_dtype).ravel(), [0, 1]),
                        named_obs("PauliY", [2]),
                    ]
                ),
                hamiltonian_obs(
                    np.array([1], dtype=r_dtype),
                    [tensor_prod_obs([named_obs("PauliY", [2]), named_obs("PauliX", [0])])],
                )
                if qml.operation.active_new_opmath()
                else tensor_prod_obs([named_obs("PauliX", [0]), named_obs("PauliY", [2])]),
                hermitian_obs(np.ones(64, dtype=c_dtype), [0, 1, 2]),
            ],
        )
```

tests/test_serialize.py
```diff
@@ -333,31 +405,64 @@ def test_hamiltonian_mix_return(self, use_csingle, wires_map):
...
+            s_expected2 = hamiltonian_obs(
+                np.array([0.7, 0.3], dtype=r_dtype),
+                [
+                    tensor_prod_obs(
+                        [
+                            named_obs("PauliX", [0]),
+                            hermitian_obs(np.eye(4, dtype=c_dtype).ravel(), [1, 2]),
+                        ]
+                    ),
```
vincentmr, line 433:
```suggestion
        s_expected1 = hamiltonian_obs(
            np.array([0.3, 0.5, 0.4], dtype=r_dtype),
            [
                tensor_prod_obs(
                    [
                        hermitian_obs(np.eye(4, dtype=c_dtype).ravel(), [0, 1]),
                        named_obs("PauliY", [2]),
                    ]
                ),
                hamiltonian_obs(
                    np.array([1], dtype=r_dtype),
                    [tensor_prod_obs([named_obs("PauliY", [2]), named_obs("PauliX", [0])])],
                )
                if qml.operation.active_new_opmath()
                else tensor_prod_obs([named_obs("PauliX", [0]), named_obs("PauliY", [2])]),
                hermitian_obs(np.ones(64, dtype=c_dtype), [0, 1, 2]),
            ],
        )
        s_expected2 = hamiltonian_obs(
            np.array([0.7, 0.3], dtype=r_dtype),
            [
                tensor_prod_obs(
                    [
                        named_obs("PauliX", [0]),
                        hermitian_obs(np.eye(4, dtype=c_dtype).ravel(), [1, 2]),
                    ]
                ),
                hamiltonian_obs(
                    np.array([1], dtype=r_dtype),
                    [tensor_prod_obs([named_obs("PauliX", [2]), named_obs("PauliY", [0])])],
                )
                if qml.operation.active_new_opmath()
                else tensor_prod_obs([named_obs("PauliY", [0]), named_obs("PauliX", [2])]),
            ],
        )
```

vincentmr (commented):
Nice work @mudit2812 . I'm leaving a few suggestions but this is essentially ready to merge IMO.

pennylane_lightning/core/_serialize.py
AmintorDusko: Agree.

requirements-dev.txt
AmintorDusko: We need the PennyLane PR to be merged first, right? What is the status of that?

pennylane_lightning/core/_serialize.py
maliasadi: I don't have any problem with it as long as this is clearly noted in the serialization docs that Hermitian is our default fallback. I generally don't like these "default" fallback in the codebase as it adds more obscurity to the code. As discussed offline, I'd prefer a tuple for observables that we convert to Hermitian in the serialization module.

AmintorDusko (commented):
Nice work @mudit2812. Very close to ready to merge. While you work on the outstanding review comments, I will expand testing for measurements and measurements_sparse. We may be able to merge it this morning. Please consider any non-essential updates and suggestions to be developed next quarter as a follow-up PR. I will review it again after you work through all the suggestions.

AmintorDusko:
Also, I activated labels to run tests for multi-GPU and build wheels, but I will remove it for now it is better to run it at the end, as a final round of CIs.

pennylane_lightning/core/_serialize.py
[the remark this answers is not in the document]
maliasadi: If there is any point about keeping this assertion, I need to hear it as I am not yet convinced :)

pennylane_lightning/core/_serialize.py
mudit2812: In my opinion, the less changes we have to make to get new operator arithmetic to work, the better. The assertions were there before, and I don't really see a reason to change them.

pennylane_lightning/core/_serialize.py
mudit2812: I thought about this, but came to the conclusion that this should really only matter for pauli reps of non-Hamiltonian/LinearCombination objects, such as `Prod`, `Tensor`, or other named observables because their pauli rep is defined as `1.0 * corresponding_pauli_word`, in which case the coefficient can be compared exactly to `1.0`. Let me know your thoughts. This is not a blocker however, so we can also address this in a follow up.

AmintorDusko:
@mudit2812, please add a link to the pennylane PR to this PR description.

pennylane_lightning/core/_serialize.py
maliasadi: is there any problem with uncommenting the if-condition in this PR?

.github/workflows/format.yml
```diff
@@ -19,13 +19,13 @@ jobs:
...
-      - name: Install dependencies
-        run:
-            python -m pip install click==8.0.4 black==23.7.0 isort==5.13.2
-
       - name: Checkout PennyLane-Lightning
         uses: actions/checkout@v3
 
+      - name: Install dependencies
+        run:
+            python -m pip install -r requirements-dev.txt
```
maliasadi, line 27: This will result the installation of many unused python packages. let's update this as follows,
 
```suggestion
            python -m pip install click==8.0.4 black==23.7.0 isort==5.13.2
```

mpitests/test_apply.py
```diff
@@ -556,7 +556,7 @@ def circuit():
 
         cpu_qnode = qml.QNode(circuit, dev_cpu)
         expected_output_cpu = cpu_qnode()
-        comm.Bcast(expected_output_cpu, root=0)
+        comm.Bcast(np.array(expected_output_cpu), root=0)
```
maliasadi, line 559: is this casting necessary?

hmm.. will it be not working without casting to a numpy array? :thinking: 
 
```suggestion
        comm.Bcast(expected_output_cpu, root=0)
```

pennylane_lightning/core/_serialize.py
```diff
@@ -254,23 +260,31 @@ def _pauli_sentence(self, observable, wires_map: dict = None):
...
             return [self.hamiltonian_obs([c], [t]) for (c, t) in zip(coeffs, terms)]
         return self.hamiltonian_obs(coeffs, terms)
 
-    # pylint: disable=protected-access
+    # pylint: disable=protected-access, too-many-return-statements
     def _ob(self, observable, wires_map: dict = None):
         """Serialize a :class:`pennylane.operation.Observable` into an Observable."""
-        if isinstance(observable, Tensor):
+        if isinstance(observable, (Prod, Sum, SProd)) and observable.pauli_rep is not None:
+            return self._pauli_sentence(observable.pauli_rep, wires_map)
```
maliasadi, line 275: This seems to be redundent with another if-condition (line 284). this can be re-written in some format closer to Catalyst as we discussed.

pennylane_lightning/core/lightning_base.py
```diff
@@ -306,18 +307,69 @@ def _process_jacobian_tape(
             "obs_idx_offsets": obs_idx_offsets,
         }
 
+    @staticmethod
+    def _assert_adjdiff_no_projectors(observable):
+        """Helper function to validate that an observable is not or does not contain
+        Projectors
+
+        Args:
+            observable (.Operator): Observable to check
```
maliasadi, line 316:
```suggestion
            observable (~pennylane.operation.Operator): Observable to check
```

maliasadi (commented):
Nice work @mudit2812! :100: I've left a few comments but it should to be good to merge. we can discuss the additional improvements in your follow-up PR.

.github/workflows/format.yml
vincentmr: Indeed, but dependabot won't update `format.yml` so since this is one of the quickest jobs I would pay the price of automation here.

mpitests/test_apply.py
vincentmr: Not without fixing something else.

requirements-dev.txt
mudit2812: Not merged yet. I'm not sure if it will get merged today or not.

pennylane_lightning/core/_serialize.py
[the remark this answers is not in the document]
mudit2812: I'm convinced, I don't think we need the assertion. I will remove it.

pennylane_lightning/core/_serialize.py
mudit2812: Just that serialization tests would have to be updated again, as they expect all observables going through the pauli rep pipeline to be hamiltonians. I'd rather leave it for now and address it in bulk with everything else later.

pennylane_lightning/core/_serialize.py
mudit2812: `Prod` does allow overlapping wires, true. I hadn't considered that, so maybe we check if there are overlapping ops, and if that's the case then we dispatch to the `Hermitian` pipeline, otherwise `Tensor`. The reason I was testing out `Hamiltonian` was because `Prod` can have `Sum` ops inside it, but I think it's better to just send it to the `Tensor` pipeline.

pennylane_lightning/core/_serialize.py
mudit2812: As per above discussion, the assertions will be removed.

tests/test_adjoint_jacobian.py
mudit2812: In that case I will have to update the serialization to handle `Prod` with overlapping wires.

pennylane_lightning/core/src/bindings/Bindings.hpp
mudit2812: Lee added these so that we can debug serialization issues more easily. We were running into a serialization issue where the actual and expected serialization had the exact same repr, but the test asserting equivalence between the two was failing, and because this goes into the C++ layer it was really hard to debug.

pennylane_lightning/lightning_qubit/lightning_qubit.py
mudit2812: `_assert_adjdiff_no_projectors` is inside `LightningBase`. My intention was to leave that as is, and add this function here which can be reused for kokkos and GPU once those are ported over to use the new API.

pennylane_lightning/core/_serialize.py
mudit2812: The reason for this was because to minimize the number of tests that needed to be updated, only the new op math observables (except for `LinearCombination`) needed to use the pauli rep. Another thing that I'd like to leave as it for now and address in the next PR. The plan is to use a similar structure to Catalyst in that PR anyway, so this will become a non-issue.

tests/new_api/test_device.py
mudit2812: See above comment.

pennylane_lightning/core/_serialize.py
```diff
@@ -183,19 +184,24 @@ def _named_obs(self, observable, wires_map: dict = None):
...
+        assert not isinstance(observable, (Tensor, Prod))
 
         wires = [wires_map[w] for w in observable.wires] if wires_map else observable.wires.tolist()
         return self.hermitian_obs(matrix(observable).ravel().astype(self.ctype), wires)
 
     def _tensor_ob(self, observable, wires_map: dict = None):
         """Serialize a tensor observable"""
-        assert isinstance(observable, Tensor)
-        return self.tensor_obs([self._ob(obs, wires_map) for obs in observable.obs])
+        assert isinstance(observable, (Tensor, Prod))
```
mudit2812, line 194:
```suggestion
```

.github/workflows/format.yml
mlxd: Let's avoid splitting deps out of requirements-dev.txt. I'll resolve this

mpitests/test_apply.py
maliasadi: what is this "something else"? :)

.github/workflows/format.yml
maliasadi: absolutely! in that case, let's fix them in all places is one single PR, not partially in multiple ones.

mpitests/test_apply.py
mlxd: I assume this is due to MPI4PY having special buffers for numpy data. I suspect the output is Pythonic lists/tuples, which don't use this buffer, so the cast is needed

pennylane_lightning/core/src/bindings/Bindings.hpp
mlxd: This was required to debug changes in the tests for the serialisation process, and is a low-hanging fruit that should've been added previously. No harm keeping this in.

mlxd (approved):
Thanks a bunch @mudit2812 

No concerns from me. I'll mark as approved, and can override anything once https://github.com/PennyLaneAI/pennylane/pull/5434 merges

vincentmr (approved):
I'll approve too, this is ready to go as soon as fixes get merge on the PL side.

maliasadi (approved):
:parachute:

mpitests/test_apply.py
maliasadi: Thanks @mlxd! I'll continue this in pv to avoid blocking merge of this PR.

pennylane_lightning/core/lightning_base.py
mudit2812: I don't think recursive depth will become an issue here as the recursion never brings us back to this function, only to `_assert_adjdiff_no_projectors`.

mudit2812:
`cmake` had to be pinned due to issues with the latest release. See [these commits](https://github.com/PennyLaneAI/pennylane-lightning/pull/649/files/feca13c9fb9686b8cd42790aed761c6e11ae82ca..c54af4574b913b8ee30839bdaff4919cea329071) to see which files were changed and would need to be unpinned eventually.

.github/workflows/tests_windows.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - name: Install dependencies
         if: steps.kokkos-cache.outputs.cache-hit != 'true'
         run: |
-          python -m pip install cmake build
+          python -m pip install "cmake!=3.29.0" build
```
maliasadi, line 78: The default Python version in this action is > 3.9. We may don't need this change here,
```suggestion
          python -m pip install cmake build
```

.github/workflows/tests_windows.yml
```diff
@@ -120,7 +120,7 @@ jobs:
       - name: Install dependencies
         if: steps.vcpkg-cache.outputs.cache-hit != 'true'
         run: |
-          python -m pip install cmake build ninja
+          python -m pip install "cmake!=3.29.0" build ninja
```
maliasadi, line 123:
```suggestion
          python -m pip install cmake build ninja
```

requirements-dev.txt
```diff
@@ -14,6 +14,7 @@ clang-tidy~=16.0
 clang-format~=16.0
 isort==5.13.2
 click==8.0.4
-cmake==3.28.4
+# cmake 3.29.0 does not work with python <3.10
+cmake!=3.29.0
```
maliasadi, line 18: I would leave this untouched as all other deps in this file are pinned to specific versions.
```suggestion
cmake==3.28.4
```

.github/workflows/tests_gpu_cuda.yml
```diff
@@ -119,7 +119,7 @@ jobs:
 
       - name: Install required packages
         run: |
-          python -m pip install ninja cmake custatevec-cu${{ matrix.cuda_version }}
+          python -m pip install ninja "cmake!=3.29.0" custatevec-cu${{ matrix.cuda_version }}
```
maliasadi, line 122: Could you also add some comment about this change for future reference?

Maybe something like,
 ```suggestion
          # Omitting the installation of cmake v3.29.0 due to
          # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install ninja "cmake!=3.29.0" custatevec-cu${{ matrix.cuda_version }}
```

File: .github/workflows/tests_gpu_cuda.yml [hunk 30]
```diff
@@ -241,7 +241,7 @@ jobs:
         run: |
           cd main
           python -m pip install -r requirements-dev.txt
-          python -m pip install cmake custatevec-cu${{ matrix.cuda_version }} openfermionpyscf
+          python -m pip install "cmake!=3.29.0" custatevec-cu${{ matrix.cuda_version }} openfermionpyscf
```
mudit2812, line 244:
```suggestion
         # Omitting the installation of cmake v3.29.0 due to
         # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install "cmake!=3.29.0" custatevec-cu${{ matrix.cuda_version }} openfermionpyscf
```

.github/workflows/tests_linux_x86_mpi_gpu.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - name: Install required packages
         run: |
           python -m pip install -r requirements-dev.txt
-          python -m pip install cmake custatevec-cu12
+          python -m pip install "cmake!=3.29.0" custatevec-cu12
```
mudit2812, line 97:
```suggestion
         # Omitting the installation of cmake v3.29.0 due to
         # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install "cmake!=3.29.0" custatevec-cu12
```

.github/workflows/wheel_macos_x86_64.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         run: |
           mkdir -p ${{ github.workspace}}/Kokkos_install/${{ matrix.exec_model }}
           cd kokkos
-          python -m pip install cmake ninja
+          python -m pip install "cmake!=3.29.0" ninja
```
mudit2812, line 71:
```suggestion
         # Omitting the installation of cmake v3.29.0 due to
         # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install "cmake!=3.29.0" ninja
```

.github/workflows/wheel_noarch.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
       - name: Install CMake and ninja
         run: |
-          python -m pip install --upgrade cmake ninja
+          python -m pip install --upgrade "cmake!=3.29.0" ninja
```
mudit2812, line 47:
```suggestion
         # Omitting the installation of cmake v3.29.0 due to
         # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install --upgrade "cmake!=3.29.0" ninja
```

.github/workflows/wheel_win_x86_64.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Install dependencies
         if: steps.kokkos-cache.outputs.cache-hit != 'true'
         run: |
-          python -m pip install cmake build
+          python -m pip install "cmake!=3.29.0" build
```
mudit2812, line 67:
```suggestion
         # Omitting the installation of cmake v3.29.0 due to
         # https://github.com/scikit-build/cmake-python-distributions/pull/474
          python -m pip install "cmake!=3.29.0" build
```

maliasadi (approved):
Nice work @mudit2812!

PR #659, opened 2024-03-21
Remove non-native observables from TOML files

maliasadi:
This PR removes non-native observables (those obs which are not supported natively by the device) from TOML files.

### DIFF
--- a/pennylane_lightning/core/_version.py
+++ b/pennylane_lightning/core/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.36.0-dev16"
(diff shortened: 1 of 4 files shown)


vincentmr (approved):
Looks good to me. Just a tiny thing, we could move `Identity` up the list with the other Hermitian gates and use the canonical order: gates, `Hermitian`, `Hamiltonian`, `SparseHamiltonian`. I mean canonical in the Lightning sense, that's usually the order in which they are listed.

erick-xanadu (approved):
Thanks for bringing this to attention.

grwlf:
@Ali, just to double-check: are the new versions of toml configs compatible with the current Catalyst validation?

PR #668, opened 2024-04-01
Fix the stochastic test failure on Kokkos (ubuntu, CUDA)

maliasadi:


tests/lightning_qubit/test_measurements_class.py
```diff
@@ -474,7 +474,8 @@ def test_single_return_value(self, measurement, observable, lightning_sv, tol):
         result = m.measure_final_state(tape)
 
         # a few tests may fail in single precision, and hence we increase the tolerance
-        assert np.allclose(result, expected, max(tol, 1.0e-5))
+        # Update the tolerance to `1.0e-4` to fix stochastic failures with Kokkos (ubuntu, CUDA)
+        assert np.allclose(result, expected, max(tol, 1.0e-4))
```
maliasadi, line 478: @vincentmr I know you've looked at this in the past. I'm wondering what you think the bast approach is here? This is not the only test failing in this action and there are more actions that suffer from this stochastic issue. PL uses custom `tol` in their unit tests and we can adjust our `tol`s to reduce the likelihood of these random failures.

vincentmr: This one is a bit tricky as I think we should find a better match, but it's not wrong either. I'd say at this point we can simply increase the tolerance or the number of flaky runs, or both.

PR #673, opened 2024-04-08
Add the missing pytest-xdist to compat-check-stable-stable

maliasadi:
Fix failed GH jobs [here]( https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8532587255/job/23373981601)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ pybind11
 pytest~=8.0.0
 pytest-cov
 pytest-mock
+pytest-xdist
--- a/.github/workflows/tests_linux.yml
+++ b/.github/workflows/tests_linux.yml
@@ -182,6 +182,8 @@ jobs:
 
       - name: Run PennyLane-Lightning unit tests
         run: |
+          # TODO: Remove installing pytest-xdist with release v0.36.0
+          python -m pip install pytest-xdist
           cd main/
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,4 +4,5 @@ pennylane>=0.34
 pybind11
 pytest~=8.0.0
 pytest-cov
-pytest-mock
\ No newline at end of file
+pytest-mock
+pytest-xdist
```
vincentmr, line 8: We typically use `requirements-dev.txt` to install dev/ci deps, so I don't think this is required. Doesn't hurt either. I think `pytest` is only included here so that PennyLane's `pl-device-test` can run.

vincentmr (approved):
Looks good otherwise, I'm ready to approve. Cheers @maliasadi !

requirements.txt
maliasadi: yeah, I usually run `requirements.txt` in a fresh container before checking lightning unit tests and it may be good to also install `pytest-xdist` (along with `pytest`) to boost up `pl-device-test`ing even for developers.

AmintorDusko (approved):
Thank you for that!

maliasadi:
https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8603399125

PR #682, opened 2024-04-15
Fix the CodeCov issue with compat-check actions

maliasadi:
This PR fixes the CodeCov job failures in compat-check actions by
- Updating the version of codecov-action from v3 to v4
- Add the codecov token to be used when uploading coverage files

### DIFF
--- a/.github/workflows/tests_linux.yml
+++ b/.github/workflows/tests_linux.yml
@@ -1,6 +1,10 @@
 name: Testing (Linux)
 on:
   workflow_call:
+    secrets:
+      codecov_token:
+        description: The codecov token to use when uploading coverage files
+        required: true
     inputs:
       lightning-version:
         type: string
@@ -622,10 +626,10 @@ jobs:
           python -m coverage xml -i -o coverage-${{ github.job }}.xml
 
       - name: Upload to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           fail_ci_if_error: true
-          token: ${{ secrets.CODECOV_TOKEN }}
+          token: ${{ secrets.codecov_token }}
 
   upload-to-codecov-linux-cpp:
     needs: [cpptests, cpptestswithOpenBLAS, cpptestswithKokkos]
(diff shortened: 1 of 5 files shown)


.github/workflows/compat-check-latest-latest.yml
```diff
@@ -12,27 +12,32 @@ jobs:
     with:
       lightning-version: latest
       pennylane-version: latest
+    secrets: inherit # pass all secrets
```
mlxd, line 15: Any worries about using this? Will it allow secrets to leak to 3rd party workflows?

.github/workflows/tests_linux.yml
```diff
@@ -555,12 +555,12 @@ jobs:
       - name: Run PennyLane-Lightning unit tests
         if: ${{ matrix.pl_backend != 'all'}}
         run: |
-          # TODO: Remove installing pytest-xdist with release v0.36.0
-          python -m pip install pytest-xdist
+          # TODO: Remove installing pytest-xdist and pytest-custom_exit_code with release v0.36.0
+          python -m pip install pytest-xdist pytest-custom_exit_code
```
mlxd, line 559: I don't know if we should depend upon a 3rd party package that hasn't been updated in 5 years. Is there an alternative?

mlxd (commented):
Thanks @maliasadi 
Just 2 comments:

.github/workflows/compat-check-latest-latest.yml
maliasadi: As checked with @rashidnhm, this should be fine here as we own these workflows.

.github/workflows/tests_linux.yml
maliasadi: This package is the recommended way to suppress ExitCode 5 by [the official pytest docs](https://docs.pytest.org/en/stable/reference/exit-codes.html). The logic is fairly simple but I didn't want to spend anytime on it as this is just temporary and is going to be removed with release v0.36

mlxd: Hmm, I'm not really convinced we should use this. However, if temporary, then can we pin the installed version for now to `0.3.0` so it is guaranteed to be the one we can currently see on PyPI, and then ensure it is removed before the release goes out.

maliasadi: okay, we don't need this package as the problem is limited to the native mcm test file ;) I fixed this using if-cond and we will remove it with v0.36 release.

maliasadi: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/8699050876 :green_circle:

AmintorDusko (approved):
Looks good to me! Thank you, @maliasadi!

rashidnhm (approved):

PR #685, opened 2024-04-16
Refactor of dev prerelease auto-update-version workflow

rashidnhm:
**Context:** Changes to the update_dev_version.yml

This PR introduces the following changes to the `update_dev_version.yml` workflow:

- Refactor `dev_version_script.py` to use [python-semver](https://python-semver.readthedocs.io/en/3.0.2/)
    - **Why?** -> It is generally more robust to parse and bump versions. In the previous version, it was assumed that the version on the PR was the same as master, then the pr version got bumped by 1. Though that is true for 99% of the cases, it left out a chance for a bad version to get by. The check is a bit more strict now, where the version is taken from master, bumped up by 1, then checked against the pr version.

Example of this in action:
- Create a branch off of master
- As of writing, the version in master is `0.36.0-dev29`
- In your branch, update the version manually to `0.36.0-dev25` (Simulates someone accidentally drifting away from the master version by more than 1)

With the previous version of the workflow, the pr would get auto updated to: `0.36.0-dev26`. This is technically invalid as that version already exists. Though it can be caught during manual PR review it might get by as well.

With the new workflow, the pr would get auto updated to: `0.36.0-dev30`.

- Change `update_dev_version.yml` to run on `pull_request_target`
    - **Why?** -> This would let the action trigger from the master branch instead of the PR branch, which means it will have access to secrets (even if the PR is from a fork!) and would be able to push to PRs as ringo bot. Usage of the token also means that when a new commit is pushed, new CI will be kicked off.

Other notable changes:
- The commit pushed by the bot now includes the version it bumped from/to.
image
image

**Benefits:**
Noted in the section above.

**Possible Drawbacks:**
With CI triggering itself, there is always a possibility CI recursively calling itself infinitely. The base case here is if the version matches what is expected, then nothing is pushed to the PR.

**Related GitHub Issues:**
None. [sc-61382](https://app.shortcut.com/xanaduai/story/61382/pennylane-lightning-fix-auto-update-bot-to-use-pr-comment-suggestion)


.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
-    if pr_version == master_version:
-        if is_dev(pr_version):
-            print("Automatically update version string.")
-            update_dev_version(Path(args.pr), pr_version)
+    print("Got Package Version from 'master' ->", str(master_version))
+    print("Got Package Version from 'pr' ->", str(pr_version))
+
+    # Only attempt to bump the version if the pull_request is:
+    #  - A prerelease, has `X.Y.Z-prerelease` in _version.py
+    #  - The prerelease startswith `dev`. We do not want to auto bump for non-dev prerelease.
```
mlxd, line 96: We should likely want this to work for RC too, support both https://peps.python.org/pep-0440/#developmental-releases and https://peps.python.org/pep-0440/#pre-releases. Though, we can add RC in a follow-up if this solves the dev issue for now.

.github/workflows/update_dev_version.yml
```diff
@@ -14,31 +16,50 @@ jobs:
...
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          ref: ${{ github.head_ref }}
-          path: pr
+      - name: Install Python Dependencies
+        run: pip install semver
+
+      - name: Capture Current Version in Pull Request
+        id: old_version
+        run: echo "version=$(grep -o '"[^"]\+[0-9]\+"' pr/pennylane_lightning/core/_version.py | sed 's/"//g')" >> $GITHUB_OUTPUT
```
mlxd, line 40: Should we validate that a number was captured here?

.github/workflows/update_dev_version.yml
```diff
@@ -14,31 +16,50 @@ jobs:
...
           cd ./pr
           if [[ -n $(git status -s) ]]; then
-            git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-            git config --global user.name "Dev version update bot"
+            git config --global user.email 'github-ringo-but-quantum@xanadu.ai'
+            git config --global user.name "ringo-but-quantum"
             git add .
-            git commit -m 'Auto update version'
+            git commit -m "Auto update version from '$OLD_VERSION' to '$NEW_VERSION'"
             git push
```
mlxd, line 64: Any concerns about this over time?

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+except ImportError as exc:
+    raise ImportError("Unable to import semver. Install semver by running `pip install semver`") from exc
 
-VERSION_FILE_PATH = "pennylane_lightning/core/_version.py"
+DEV_PRERELEASE_TAG_PREFIX = "dev"
+DEV_PRERELEASE_TAG_START = "dev1"
+VERSION_FILE_PATH = Path("pennylane_lightning/core/_version.py")
 
-rgx_ver = re.compile('^__version__ = "(.*?)"$')
+rgx_ver = re.compile(pattern=r"^__version__ = \"(.*)\"$", flags=re.MULTILINE)
```
mlxd, line 28: Just checking that this won't cause any problems if we use "RC" or non pre-release versions with the string?

mlxd (commented):
Thanks @rashidnhm 
Just a few quick questions

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
 # limitations under the License.
+
+import re
 import argparse
 from pathlib import Path
-import importlib
 
-import re
+try:
+    import semver
```
maliasadi, line 20: Good call :+1:

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+except ImportError as exc:
+    raise ImportError("Unable to import semver. Install semver by running `pip install semver`") from exc
 
-VERSION_FILE_PATH = "pennylane_lightning/core/_version.py"
+DEV_PRERELEASE_TAG_PREFIX = "dev"
+DEV_PRERELEASE_TAG_START = "dev1"
```
maliasadi, line 25: The dev-version in previous cycles started with dev0.

```suggestion
DEV_PRERELEASE_TAG_START = "dev0"
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
-    master_version = extract_version(Path(args.master))
+    pr_version = extract_version(args.pr)
+    master_version = extract_version(args.master)
 
-    if pr_version == master_version:
-        if is_dev(pr_version):
-            print("Automatically update version string.")
-            update_dev_version(Path(args.pr), pr_version)
+    print("Got Package Version from 'master' ->", str(master_version))
+    print("Got Package Version from 'pr' ->", str(pr_version))
```
maliasadi, line 92:
```suggestion
    print("Got Package Version from 'PR' ->", str(pr_version))
```

.github/workflows/dev_version_script.py
maliasadi, line 20:
```suggestion
    from semver import Version
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+DEV_PRERELEASE_TAG_PREFIX = "dev"
+DEV_PRERELEASE_TAG_START = "dev1"
+VERSION_FILE_PATH = Path("pennylane_lightning/core/_version.py")
 
-rgx_ver = re.compile('^__version__ = "(.*?)"$')
+rgx_ver = re.compile(pattern=r"^__version__ = \"(.*)\"$", flags=re.MULTILINE)
 
-rgx_dev_ver = re.compile("^(\d*\.\d*\.\d*)-dev(\d*)$")
 
+def extract_version(repo_root_path: Path) -> semver.Version:
```
maliasadi, line 31:
```suggestion
def extract_version(repo_root_path: Path) -> Version:
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
-                line = line.strip()
-                m = rgx_ver.match(line)
-                return m.group(1)
+                if (m := rgx_ver.match(line.strip())) is not None:
+                    if not m.groups():
+                        raise ValueError(f"Unable to find valid semver for __version__. Got: '{line}'")
+                    parsed_semver = m.group(1)
+                    if not semver.Version.is_valid(parsed_semver):
+                        raise ValueError(f"Invalid semver for __version__. Got: '{parsed_semver}' from line '{line}'")
+                    return semver.Version.parse(parsed_semver)
```
maliasadi, line 52:
```suggestion
                    if not Version.is_valid(parsed_semver):
                        raise ValueError(f"Invalid semver for __version__. Got: '{parsed_semver}' from line '{line}'")
                    return Version.parse(parsed_semver)
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+                raise ValueError(f"Unable to find valid semver for __version__. Got: '{line}'")
     raise ValueError("Cannot parse version")
 
 
-def is_dev(version_str):
-    m = rgx_dev_ver.fullmatch(version_str)
-    return m is not None
+def update_prerelease_version(repo_root_path: Path, version: semver.Version):
```
maliasadi, line 57:
```suggestion
def update_prerelease_version(repo_root_path: Path, version: Version):
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+            print(f"Updating pr package version from -> '{pr_version}', to -> {new_version}")
+            update_prerelease_version(args.pr, new_version)
         else:
-            print(
-                "Even though version of this PR is different from the master, as the PR is not dev, we do nothing."
-            )
+            print(f"pr is on the expected version '{new_version}' ... Nothing to do!")
     else:
-        print("Version of this PR is already different from master. Do nothing.")
+        print("pr is not a dev prerelease ... Nothing to do!")
```
maliasadi, line 112:
```suggestion
        print("PR is not a dev prerelease ... Nothing to do!")
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+            next_prerelease_version = master_version.next_version("prerelease").prerelease
+        new_version = master_version.replace(prerelease=next_prerelease_version)
+        if pr_version != new_version:
+            print(f"Updating pr package version from -> '{pr_version}', to -> {new_version}")
+            update_prerelease_version(args.pr, new_version)
         else:
-            print(
-                "Even though version of this PR is different from the master, as the PR is not dev, we do nothing."
-            )
+            print(f"pr is on the expected version '{new_version}' ... Nothing to do!")
```
maliasadi, line 110:
```suggestion
            print(f"PR is on the expected version '{new_version}' ... Nothing to do!")
```

.github/workflows/dev_version_script.py
```diff
@@ -11,74 +11,102 @@
...
+        # If master branch does not have a prerelease (for any reason) OR does not have an ending number
+        # Then default to the starting tag
+        if not master_version.prerelease or master_version.prerelease == DEV_PRERELEASE_TAG_PREFIX:
+            next_prerelease_version = DEV_PRERELEASE_TAG_START
+        else:
+            # Generate the next prerelease version (eg: dev1 -> dev2). Sourcing from master version.
+            next_prerelease_version = master_version.next_version("prerelease").prerelease
+        new_version = master_version.replace(prerelease=next_prerelease_version)
+        if pr_version != new_version:
+            print(f"Updating pr package version from -> '{pr_version}', to -> {new_version}")
```
maliasadi, line 107:
```suggestion
            print(f"Updating PR package version from -> '{pr_version}', to -> {new_version}")
```

maliasadi (commented):
Nice work! :partying_face:

.github/workflows/dev_version_script.py
rashidnhm: Supporting RC would likely be a future feature to add. Even as it stands right now, this update script (what exists on master) branch does not support RC ([See Here](https://github.com/PennyLaneAI/pennylane-lightning/blob/288b08d1f491319fde56307a27d8283b8e8a846b/.github/workflows/dev_version_script.py#L37-L39)).

I retained that logic as part of the refactor, but we can change it in the future :+1:

.github/workflows/update_dev_version.yml
rashidnhm: Potentially, yes? Though I didn't think it was necessary as if there wasn't a valid semver in the _version.py file, then the subsequent step would error out and this value wouldn't matter anyways.

mlxd (approved):
Thanks @rashidnhm 
No further issues from my end

PR #664, opened 2024-03-27
Add compile-time support for AVX2/512 streaming operations in LQ

mlxd:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:** This PR enable compile-time support for AVX2/512 streaming operations, as validated and used for the Lightning paper workloads listed in https://github.com/PennyLaneAI/lightning-on-hpc and was promised for release in 0.36

**Description of the Change:** Adds compile-time support for streaming intrinsics to bypass cache-level updates.

**Benefits:** Improves performance for multithreading intrinsic-enabled kernels using high thread-counts.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-62009]

PR #695, opened 2024-04-24
Run CI conditionally with paths

vincentmr:
### Before submitting

**Context:**
The CI is slow enough to be a real nuisance and productivity killer. This is making devs nervous, especially with freeze week coming.

Split tests into CPP and Python tests. CPP tests are run only when changes are made to the CPP source files. Python tests always run.

Test skipping can be seen in action in this [PR](https://github.com/PennyLaneAI/pennylane-lightning/pull/696), targeting the current one.

**Benefits:**
Run fewer tests which cannot be impacted by changes.
Faster CI runs in Lightning.
Increase GHA runners availability across the organization.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-61920]


.github/workflows/tests_lgpu_cpp.yml
```diff
@@ -0,0 +1,197 @@
...
+      - .github/workflows/tests_lgpu_cpp.yml
+      - pennylane_lightning/core/src/*
+      - pennylane_lightning/core/src/algorithms/**
+      - pennylane_lightning/core/src/bindings/**
+      - pennylane_lightning/core/src/gates/**
+      - pennylane_lightning/core/src/measurements/**
+      - pennylane_lightning/core/src/observables/**
+      - pennylane_lightning/core/src/simulators/base/**
+      - pennylane_lightning/core/src/simulators/lightning_gpu/**
+      - pennylane_lightning/core/src/utils
```
mlxd, line 25: Fine as is, but would this be tidier with `pennylane_lightning/core/src/**` in paths, and use a path exclude to remove the non-testing simulators? No preference for me, if opt in is easier than opt out, all good here.

.github/workflows/tests_lkcuda_cpp.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Cache installation directories
+        id: kokkos-cache
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ matrix.os }}-kokkos${{ matrix.kokkos_version }}-${{ matrix.exec_model }}-${{ github.ref }}-${{ github.sha }}
+          path: ${{ github.workspace }}/Kokkos_install/${{ matrix.exec_model }}
+
+  cpptestswithKokkos:
+    if: ${{ github.sha == 'xxxxxxxx' }}
```
mlxd, line 110: Do we need to set something for this?

mlxd (commented):
Thanks folks. No major concerns from me, but a few questions:
* Will this require us to make the CPP tests as optional, rather than explicit for the CI checks?
* How much better are the CI runs after this? From my experience, the CPP tests were never the bottleneck, but I could be wrong. 
* Do we have any any issues with coverage checks, since now we are cutting coverage on the overall ecosystem on a per-PR basis?
* To follow no form the above, do we have issues uploading mutliple coverage files to codecov, or is the number the same as before?

.github/workflows/tests_lgpu_cpp.yml
vincentmr: Good idea, let me tidy this up.

.github/workflows/tests_lkcuda_cpp.yml
vincentmr: This is still a todo. At the time LK was migrated, the CUDA-12 `nvcc` would run out of memory, even on a Perlmutter node with 256GB of RAM, and hence this is always skipped for now.

vincentmr:
I can confirm CodeCov plays nice in [3c40f07](https://github.com/PennyLaneAI/pennylane-lightning/pull/695/commits/3c40f0799c3370f536634a02ceb5296dd9f92171).

.github/workflows/tests_lkcuda_cpp.yml
mlxd: Maybe we upgrade the CUDA SDK version as a follow-on to this, in the hope it will be ok.

.github/workflows/tests_without_binary.yml
```diff
@@ -11,11 +11,17 @@ on:
         required: true
         description: The version of PennyLane to use. Valid values are either 'release' (most recent release candidate), 'stable' (most recent git-tag) or 'latest' (most recent commit from master)
   pull_request:
+    paths:
+      - .github/workflows/tests_without_binary.yml
+      - pennylane_lightning/core/*
```
mlxd, line 16: Do we need `core` and `pennylane_lightning/lightning_qubit/` for the no-bin wheel to be testable?

mlxd (approved):
No major blocker from me, thanks. I've a few quick comments left, but you can ignore these if no longer relevant.

.github/workflows/tests_without_binary.yml
vincentmr: `core` contains a few Python base modules so the files therein must be included. I think I'll flip this upside-down and change for 
```
    paths-ignore:
      - pennylane_lightning/core/src/**
```
so run if there are changes other than the C++ source files. This is better for Python tests which could also be affected to changes in `setup.py` for example.

mlxd:
Since https://github.com/PennyLaneAI/pennylane-lightning/pull/664 will likely need an update or two to match this CI change, can that PR go in first?

vincentmr:
> Since #664 will likely need an update or two to match this CI change, can that PR go in first?

Yes, it will be a good test. This might require a few adjustments.

.github/workflows/tests_lgpu_cpp.yml
```diff
@@ -0,0 +1,197 @@
...
```
maliasadi, line 25: Could we have these paths store in a file such as `./.cpp-paths` or a different action and use it in all of these actions recipes?

vincentmr: I don't think this is possible, but looking again at the doc I found something to simplify this :)

maliasadi: I like the fix! :100:

mlxd:
I resolved the merge conflicts locally with the streaming ops. Everything now should adhere to this PR's design

PR #705, opened 2024-05-02
Use wire order specified on instantiation

albi3ro:
**Context:**

We recently discovered an issue with wire ordering:
```
ops = [qml.RZ(0.64, 2), qml.Z(2), qml.S(2), qml.Hadamard(2)]

tape = qml.tape.QuantumScript(ops, [qml.probs(wires=[0,1,2])])

dev1 = qml.device("lightning.qubit", wires=3)
dev2 = qml.device("default.qubit", wires=3)
res1 = dev1.execute(tape)
res2 = dev2.execute(tape)
print(res1, res2)
```

Basically, we were using `QuantumScript.map_to_standard_wires`, which changed the wire order on the probability to `probs(wires=[1, 2, 0])`.  But becuase lightning cannot handle non-ascending wire orders, the wrong result came back without error.

Another PR wil fix the fact we can't handle `probs` in a non-ascending wire order, but this PR is eliminating the use of `map_to_standard_wires`. For lightning, we want to make sure to use the wire order specified on instantiation to give users more control over the simulation indices that will be used.

Switch use of `map_to_standard_wires` with `qml.map_wires` with a device wire map.

**Benefits:**

Better control over simulation indices.

**Possible Drawbacks:**

Unsure how to test this. The correct answer should be outputted regardless of the simulation indices used. The only real difference is how a user can control the performance of the simulation.  We may just leave this untested, as it continues to give correct results.

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -106,7 +106,7 @@
 ### Bug fixes
 
 * Lightning Qubit once again respects the wire order specified on device instantiation.
-  [(#705)](https://github.com/PennyLaneAI/pennylane-lightning/pull/705/)
+  [(#705)](https://github.com/PennyLaneAI/pennylane-lightning/pull/705)
 
 * `dynamic_one_shot` was refactored to use `SampleMP` measurements as a way to return the mid-circuit measurement samples. `LightningQubit`'s `simulate` is modified accordingly.
   [(#694)](https://github.com/PennyLaneAI/pennylane/pull/694)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -105,6 +105,8 @@
 
 ### Bug fixes
 
+* Lightning Qubit once again respects the wire order specified on device instantiation.
+
 * `dynamic_one_shot` was refactored to use `SampleMP` measurements as a way to return the mid-circuit measurement samples. `LightningQubit`'s `simulate` is modified accordingly.
   [(#694)](https://github.com/PennyLaneAI/pennylane/pull/694)
 
--- a/pennylane_lightning/lightning_qubit/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit/lightning_qubit.py
@@ -108,7 +108,9 @@ def jacobian(circuit: QuantumTape, state: LightningStateVector, batch_obs=False,
     return LightningAdjointJacobian(final_state, batch_obs=batch_obs).calculate_jacobian(circuit)
 
 
-def simulate_and_jacobian(circuit: QuantumTape, state: LightningStateVector, batch_obs=False, wire_map=None):
+def simulate_and_jacobian(
+    circuit: QuantumTape, state: LightningStateVector, batch_obs=False, wire_map=None
+):
     """Simulate a single quantum script and compute its Jacobian.
 
     Args:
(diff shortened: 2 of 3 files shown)


.github/CHANGELOG.md
```diff
@@ -105,6 +105,8 @@
 
 ### Bug fixes
 
+* Lightning Qubit once again respects the wire order specified on device instantiation.
```
albi3ro, line 108:
```suggestion
* Lightning Qubit once again respects the wire order specified on device instantiation.
  [(#705)](https://github.com/PennyLaneAI/pennylane-lightning/pull/705/)
```

.github/CHANGELOG.md
```diff
@@ -105,6 +105,9 @@
...
+  [(#705)](https://github.com/PennyLaneAI/pennylane-lightning/pull/705/)
```
maliasadi, line 109:
```suggestion
  [(#705)](https://github.com/PennyLaneAI/pennylane-lightning/pull/705)
```

AmintorDusko:
@albi3ro, please elaborate on the Possible Drawbacks you added to the PR's description.
Wouldn't the example you gave in the description work as a test?

vincentmr:
Consider 
```
for wires in itertools.permutations((0,1,2), 3):
    tape = qml.tape.QuantumScript(ops, [qml.probs(wires=wires)])

    dev1 = qml.device("lightning.qubit", wires=3)
    dev2 = qml.device("default.qubit", wires=3)
    try:
        res1 = dev1.execute(tape)
        res2 = dev2.execute(tape)
        print(np.allclose(res1, res2), wires, res1, res2)
    except:
        pass
```
L-GPU and L-Kokkos will raise in all but the first `wires`. L-Qubit manages to get [0, 1, 2, 5] correct, but not [3, 4]. Should we simply raise as in L-GPU/Kokkos?

albi3ro:
> Consider
> 
> ```
> for wires in itertools.permutations((0,1,2), 3):
>     tape = qml.tape.QuantumScript(ops, [qml.probs(wires=wires)])
> 
>     dev1 = qml.device("lightning.qubit", wires=3)
>     dev2 = qml.device("default.qubit", wires=3)
>     try:
>         res1 = dev1.execute(tape)
>         res2 = dev2.execute(tape)
>         print(np.allclose(res1, res2), wires, res1, res2)
>     except:
>         pass
> ```
> 
> L-GPU and L-Kokkos will raise in all but the first `wires`. L-Qubit manages to get [0, 1, 2, 5] correct, but not [3, 4]. Should we simply raise as in L-GPU/Kokkos?

I believe @mlxd is currently working on adding an ability to wires in different orders.  

Option 1 was to add in an error if the wires were not in ascending order. But since that validation would be in the python layer, catalyst would still be wrong. So we're trying to add support for the other wire orders in the c++ layer.

vincentmr:
> I'm a little unsure about the best way to add tests for this change.

Why don't we add the example in the description as a test?

maliasadi:
> @maliasadi I'm a little unsure about the best way to add tests for this change.

@albi3ro could we use the failed test now to check the fix?

``` py
ops = [qml.RZ(0.64, 2), qml.Z(2), qml.S(2), qml.Hadamard(2)]

tape = qml.tape.QuantumScript(ops, [qml.probs(wires=[0, 1, 2])])

dev1 = qml.device("lightning.qubit", wires=3)
dev2 = qml.device("default.qubit", wires=3)
res1 = dev1.execute(tape)
res2 = dev2.execute(tape)
print(res1, res2)
```
 
> Once we are able to handle probs with different wire orders, the results will be the same regardless of the wire order the simulation uses. This is more about giving users the ability to control the simulation indices more precisely.

I'd suggest adding test cases with different wire orders to verify the correctness of the probs support would be sufficient. Unless you think otherwise. This is essential for handling probs with arbitrary wire orders. Currently, the results are not valid, so it's crucial to have tests that check the support for each patch fix.

AmintorDusko (approved):
Thank you, @albi3ro!

vincentmr (approved):
Looks good to me, thanks for the fix @albi3ro.

PR #722, opened 2024-05-10
Improve Dockerfile robustness

vincentmr:
### Before submitting

**Context:**
When PennyLane lags Lightning, like around release dates, there is a possibility that Docker images for Lightning will be built before PennyLane is release and hence the versions mismatch. 

Add PENNYLANE_VERSION build arg.
Remove dependency on requirement files and install only deps that are needed.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-63094]


docker/Dockerfile
```diff
@@ -85,7 +85,7 @@ RUN python setup.py bdist_wheel
 FROM base-runtime AS wheel-lightning-qubit
 COPY --from=build-wheel-lightning-qubit /opt/pennylane-lightning/dist/ /
 RUN pip install --force-reinstall --no-cache-dir PennyLane_Lightning*.whl && rm PennyLane_Lightning*.whl
-RUN pip install --no-cache-dir -U pennylane
+RUN pip install --no-cache-dir git+https://github.com/PennyLaneAI/pennylane.git@${PENNYLANE_VERSION}
```
vincentmr, line 88: Enable installing against any pennylane version and new PennyLane releases when wheels have not been updated to PyPI yet. This is convenient around release dates.

docker/Dockerfile
```diff
@@ -19,14 +19,17 @@ FROM ubuntu:jammy AS base-runtime
...
 ARG KOKKOS_VERSION=4.2.00
-ARG LIGHTNING_VERSION=v0.34.0
+ARG PENNYLANE_VERSION=v0.36.0
+ARG LIGHTNING_VERSION=v0.36.0
 ARG ROCM_INSTALLER=https://repo.radeon.com/amdgpu-install/5.7/ubuntu/jammy/amdgpu-install_5.7.50700-1_all.deb
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
     apt-utils \
     ca-certificates \
+    git \
```
vincentmr, line 32: `git` now required to install generic PennyLane versions.

docker/Dockerfile
```diff
@@ -19,14 +19,17 @@ FROM ubuntu:jammy AS base-runtime
 ARG AMD_ARCH=AMD_GFX90A
 ARG CUDA_ARCH=AMPERE80
 ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.2.2/local_installers/cuda_12.2.2_535.104.05_linux.run
+ARG DEBIAN_FRONTEND=noninteractive
```
vincentmr, line 22: Required to install `git` non-interactively (without choosing geographic regions, etc.)

docker/Dockerfile
```diff
@@ -70,9 +72,7 @@ RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN rm -rf tmp && git clone --depth 1 --branch ${LIGHTNING_VERSION} https://github.com/PennyLaneAI/pennylane-lightning.git tmp\
     && mv tmp/* /opt/pennylane-lightning && rm -rf tmp
-RUN pip install --no-cache-dir wheel
-RUN pip install --no-cache-dir -r requirements-dev.txt
-RUN pip install --no-cache-dir -r doc/requirements.txt
+RUN pip install --no-cache-dir cmake ninja pybind11 wheel
```
vincentmr, line 75: That's all we need from `requirements-dev.txt`.

.github/workflows/docker_linux_x86_64.yml
```diff
@@ -12,7 +12,7 @@ on:
       lightning-version:
         type: string
         required: true
-        description: The version of Lightning to use. This should be a valid semantic version, e.g. 0.34.0.
+        description: The version of Lightning to use. This should be a valid git reference, e.g. v0.36.0.
```
jay-selby, line 15:
```suggestion
        description: The version of Lightning to use. This should be a valid git tag, e.g. v0.36.0.
```

.github/workflows/docker_linux_x86_64.yml
```diff
@@ -22,7 +22,7 @@ on:
...
```
jay-selby, line 25:
```suggestion
        description: The version of Lightning to use. This should be a valid git tag, e.g. v0.36.0.
```

docker/Dockerfile
```diff
@@ -19,14 +19,17 @@ FROM ubuntu:jammy AS base-runtime
 ARG AMD_ARCH=AMD_GFX90A
 ARG CUDA_ARCH=AMPERE80
 ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.2.2/local_installers/cuda_12.2.2_535.104.05_linux.run
+ARG DEBIAN_FRONTEND=noninteractive
 ARG GCC_VERSION=11
 ARG KOKKOS_VERSION=4.2.00
-ARG LIGHTNING_VERSION=v0.34.0
+ARG PENNYLANE_VERSION=v0.36.0
+ARG LIGHTNING_VERSION=v0.36.0
```
jay-selby, line 26: Is it better to fetch these from the environment?  Would anyone every want to build a very specific container by setting a PENNYLANE_VERSION env var and then running make?

jay-selby (changes requested):

docker/Dockerfile
maliasadi: Why is this required now?

docker/Dockerfile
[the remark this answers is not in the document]
vincentmr: > PL and LQ versions should be always the same.

This is only true for official releases, but one of the main reason why we have these Docker images is to easily deploy on HPC systems (and for example test new features between releases, like MPI parallelisation of L-Kokkos for QChem). It is thus designed to be able to build an image for any Lightning branch/tag. PennyLane can also be a limitation so it is natural to allow building an image with any branch/tag there too.

docker/Dockerfile
vincentmr: Seems like moving git to that stage changed the behaviour of `apt`.

docker/Dockerfile
vincentmr: Some packages like `clang-tidy` have been a source of failures on some platform as reported by @jay-selby previously, and hence I'm removing it to favor a minimal set of dependencies. `requirements-dev.txt` has many unnecessary packages and Docker style guides usually advocates minimal stages and images. `quimb` is not really integrated in the Lightning env yet, but we can add it if/when it is.

Makefile
```diff
@@ -150,7 +150,7 @@ endif
 ifdef version
     VERSION := $(version)
 else
-    VERSION := v0.35.0
+    VERSION := v0.36.0
```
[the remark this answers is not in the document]
vincentmr: We seldom use that `Makefile`, it is more like a cheat sheet for invoking `docker build` for me. But if you think it is useful, I can look into it.

docker/Dockerfile
vincentmr: I'll repost my comment below

> We seldom use that Makefile, it is more like a cheat sheet for invoking docker build for me.

So I'd say, no need for an env var, let's just pass extra options with `--build-arg`.

docker/Dockerfile
maliasadi: Sounds good! :+1:

docker/Dockerfile
```diff
@@ -85,44 +84,22 @@ RUN python setup.py bdist_wheel
...
 # Download Lightning release and build lightning-kokkos backend with Kokkos-OpenMP
 FROM base-build-python AS build-wheel-lightning-kokkos-openmp
 WORKDIR /opt/pennylane-lightning
-COPY --from=build-kokkos-openmp /opt/kokkos /opt/kokkos
-ENV CMAKE_PREFIX_PATH=/opt/kokkos:$CMAKE_PREFIX_PATH
+ENV PL_BACKEND=lightning_kokkos
 RUN pip uninstall -y pennylane-lightning
-RUN PL_BACKEND=lightning_kokkos python setup.py build_ext
-RUN PL_BACKEND=lightning_kokkos python setup.py bdist_wheel
+RUN python setup.py build_ext -i --define="PL_BACKEND=lightning_kokkos;Kokkos_ENABLE_SERIAL:BOOL=ON;Kokkos_ENABLE_OPENMP:BOOL=ON"
```
maliasadi, line 94: This seems not be required as `PL_BACKEND` is already set to LK (line 92) 
```suggestion
RUN python setup.py build_ext -i --define="Kokkos_ENABLE_SERIAL:BOOL=ON;Kokkos_ENABLE_OPENMP:BOOL=ON"
```

docker/Dockerfile
```diff
@@ -133,45 +110,23 @@ RUN ./cuda-install.run --silent --toolkit --toolkitpath=/usr/local/cuda-$(echo $
...
 FROM base-build-cuda AS build-wheel-lightning-kokkos-cuda
 WORKDIR /opt/pennylane-lightning
-COPY --from=build-kokkos-cuda /opt/kokkos /opt/kokkos
-ENV CMAKE_PREFIX_PATH=/opt/kokkos:$CMAKE_PREFIX_PATH
+ENV PL_BACKEND=lightning_kokkos
 RUN pip uninstall -y pennylane-lightning
 RUN echo >> cmake/support_kokkos.cmake && echo "find_package(CUDAToolkit REQUIRED)" >> cmake/support_kokkos.cmake
-RUN PL_BACKEND=lightning_kokkos python setup.py build_ext
-RUN PL_BACKEND=lightning_kokkos python setup.py bdist_wheel
+RUN python setup.py build_ext -i --define="PL_BACKEND=lightning_kokkos;Kokkos_ENABLE_SERIAL:BOOL=ON;Kokkos_ENABLE_OPENMP:BOOL=ON;Kokkos_ENABLE_CUDA:BOOL=ON;Kokkos_ARCH_${CUDA_ARCH}=ON"
```
maliasadi, line 119: Same here,
```suggestion
RUN python setup.py build_ext -i --define="Kokkos_ENABLE_SERIAL:BOOL=ON;Kokkos_ENABLE_OPENMP:BOOL=ON;Kokkos_ENABLE_CUDA:BOOL=ON;Kokkos_ARCH_${CUDA_ARCH}=ON"
```

maliasadi (approved):
Nice work @vincentmr! Don't forget to update the changelog.

jay-selby (approved):

docker/Dockerfile
vincentmr: > Maybe prioritize robustness here since that is the goal of the story?

Agreed, but in this case I would argue adding extra requirements was decreasing robustness since some unrequired dependencies in `requirements-dev.txt` were failing the build. So I would like to keep it as is.

PR #726, opened 2024-05-13
Pin autoray < 0.6.10

maliasadi:
CIs currently are failing due to https://github.com/jcmgray/autoray/pull/23, we pin to the previous version to pass CIs.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -19,4 +19,5 @@ cmake
 custatevec-cu12
 pylint
 scipy~=1.12.0
+autoray<0.6.10
 git+https://github.com/jcmgray/quimb.git
--- a/pennylane_lightning/core/_version.py
+++ b/pennylane_lightning/core/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.37.0-dev8"
+__version__ = "0.37.0-dev9"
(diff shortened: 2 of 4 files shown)


maliasadi:
The merge of this PR is currently blocked by https://github.com/PennyLaneAI/pennylane/pull/5595

AmintorDusko (approved):
Thank you, @maliasadi!

requirements-dev.txt
```diff
@@ -19,4 +19,5 @@ cmake
 custatevec-cu12
 pylint
 scipy~=1.12.0
+autoray<0.6.10
```
vincentmr, line 22: In PL we have a req `autoray>=0.6.1`, so don't we want to make it `autoray==0.6.1`?

maliasadi: I think `autoray < 0.6.10` should be enough as a temporary solution which only affects developers.

vincentmr (approved):
LGTM, thanks @maliasadi .

trbromley:
Shall we revert this PR now that autoray `0.6.11` is out?

vincentmr:
I think so, plus @dwierichs has a patch PR in PennyLane which should fix the issue in all versions.

dwierichs:
Ah, wait! 0.6.10 will not be fixed, the PR just fixes a complementary problem in 0.6.10 that is not fixed by 0.6.11 itself!

PR #729, opened 2024-05-15
Fix ambiguous function overload with newer compiler when using streaming AVX ops

mlxd:
…ing ops

### Before submitting

**Context:** This PR fixes an issue with failed overload acceptance using newer GCC toolchains.

**Description of the Change:** Adds more overloaded load/store/stream operations for the kernel definitions.

**Benefits:** Supports newer GCC (13.2+)

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/avx_common/AVX512Concept.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/avx_common/AVX512Concept.hpp
@@ -47,6 +47,17 @@ template <typename T> struct AVX512Concept {
     PL_FORCE_INLINE
     static auto load(std::complex<PrecisionT> *p) -> IntrinsicType {
         if constexpr (std::is_same_v<PrecisionT, float>) {
+            return _mm512_load_ps(reinterpret_cast<PrecisionT *>(p));
+        } else if (std::is_same_v<PrecisionT, double>) {
+            return _mm512_load_pd(reinterpret_cast<PrecisionT *>(p));
+        } else {
+            static_assert(std::is_same_v<PrecisionT, float> ||
+                          std::is_same_v<PrecisionT, double>);
+        }
+    }
+    PL_FORCE_INLINE
+    static auto load(PrecisionT *p) -> IntrinsicType {
+        if constexpr (std::is_same_v<PrecisionT, float>) {
             return _mm512_load_ps(p);
         } else if (std::is_same_v<PrecisionT, double>) {
             return _mm512_load_pd(p);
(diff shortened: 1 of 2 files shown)


mlxd:
[sc-63398]

.github/workflows/tests_linux_cpp.yml
```diff
@@ -46,12 +46,12 @@ jobs:
       matrix:
         pl_backend: ["lightning_qubit"]
         enable_kernel_omp: ["OFF", "ON"]
-        enable_kernel_avx_stream: ["OFF", "ON"]
+        enable_kernel_avx_streaming: ["OFF", "ON"]
         exclude:
         - enable_kernel_omp: OFF
           enable_kernel_avx_stream: ON
```
vincentmr, line 52:
```suggestion
          enable_kernel_avx_streaming: ON
```

vincentmr (changes requested):
I think there is an `exclude` directive that is broken, but otherwise ready to approve.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/avx_common/AVX512Concept.hpp
```diff
@@ -118,7 +141,11 @@ template <typename T> struct AVX512Concept {
 
     PL_FORCE_INLINE
     static void store(std::complex<PrecisionT> *p, IntrinsicType value) {
-        store(reinterpret_cast<PrecisionT *>(p), value);
+#ifdef PL_LQ_KERNEL_AVX_STREAMING
+        store_(reinterpret_cast<PrecisionT *>(p), value);
```
multiphaseCFD, line 145: Just want to make sure. Are we going to use `_mm512_store` when turning on  `PL_LQ_KERNEL_AVX_STREAMING`?

multiphaseCFD (commented):
Thanks @mlxd . Just one question.

pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/avx_common/AVX512Concept.hpp
mlxd: We want to use stream instead of store when on. Stream and store do the same thing, except one bypasses cache invalidation and update. Here, when `PL_LQ_KERNEL_AVX_STREAMING` is not set, we want to use store, which updates the cache. When not set, we use stream.

mlxd: Oh, now I see it, you are right, this should be if not defined. My bad --- good catch

multiphaseCFD (approved):
Thanks @mlxd for that!

vincentmr (approved):
LGTM, thanks @mlxd .

PR #725, opened 2024-05-10
Bump Kokkos version to 4.3.01

mlxd:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [x] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:** Building Catalyst currently causes issues with Kokkos path resolution. This PR aims to update the Kokkos version to offer a potential route to rectify this issue.

**Description of the Change:** Updates kokkos to 4.3.01

**Benefits:** New support for Kokkos architectures and performance gains. Potential fix for Catalyst runtime Kokkos failure.

**Possible Drawbacks:**

**Related GitHub Issues:** https://github.com/PennyLaneAI/catalyst/pull/732

PR #734, opened 2024-05-17
Fix PennyLane LightningKokkos and LightningQubits tests for stable/stable

AmintorDusko:
**Context:** PennyLane LightningKokkos and LightningQubits tests for stable/stable configuration are currently failing.
This PR looks to solve these problems.

The LightningKokkos and LightningQubits stable configurations (v0.36.0) were wrongly being tested with the most recent suite of tests.

PR #725 changed the default Kokkos version used by the LKokkos device. Currently, we build all wheels from scratch in our Plugin Test Matrix workflows, the sum of these two factors led me to expand our Kokkos caching functionality to build and cache any number of Kokkos versions. This is now configurable by an input variable `kokkos_version`.

I brought the change proposed by @vincentmr to this PR so we can have everything in a single place, to see if it is possible to merge it sooner than later.

**Benefits:** Tests are now right.

[sc-63561]

### DIFF
--- a/.github/workflows/tests_lkcpu_python.yml
+++ b/.github/workflows/tests_lkcpu_python.yml
@@ -48,6 +48,7 @@ jobs:
     uses: ./.github/workflows/build_and_cache_Kokkos_linux.yml
     with:
       os: ubuntu-22.04
+      kokkos_version: ${{ inputs.lightning-version == 'latest' && '4.2.00' || '4.3.01' }}
 
   build_lightning_kokkos_wheels:
     needs: [determine_runner, build_and_cache_Kokkos]
@@ -168,7 +169,7 @@ jobs:
           fetch-tags: true
           path: main
 
-      - name: Switch to release build of Lightning
+      - name: Switch to release tag of Lightning
         if: inputs.lightning-version == 'release'
         run: |
           cd main
@@ -179,6 +180,16 @@ jobs:
         if: inputs.lightning-version == 'stable'
         run: |
           python -m pip install -U pennylane-lightning --no-deps
+
+      - name: Switch to stable tag of Lightning
+        if: inputs.lightning-version == 'stable'
+        run: |
+          cd main
+          git fetch --tags --force
+          git checkout $(git tag | sort -V | tail -1)
+          git log -1 --format='%H'
+          git status
+
 
       - uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/tests_lkcpu_python.yml
+++ b/.github/workflows/tests_lkcpu_python.yml
@@ -48,7 +48,7 @@ jobs:
     uses: ./.github/workflows/build_and_cache_Kokkos_linux.yml
     with:
       os: ubuntu-22.04
-      kokkos_version: ${{ inputs.lightning-version == 'latest' && '4.2.00' || '4.3.01' }}
+      kokkos_version: ${{ inputs.lightning-version == 'stable' && '4.2.00' || '4.3.01' }}
 
   build_lightning_kokkos_wheels:
     needs: [determine_runner, build_and_cache_Kokkos]
--- a/.github/workflows/tests_lkcpu_python.yml
+++ b/.github/workflows/tests_lkcpu_python.yml
@@ -81,7 +81,7 @@ jobs:
         run: |
           cd main
           git fetch --tags --force
-          git checkout $(git tag | sort -V | tail -1)
+          git checkout latest_release
           git log -1 --format='%H'
           git status
 
@@ -186,7 +186,7 @@ jobs:
         run: |
           cd main
           git fetch --tags --force
-          git checkout $(git tag | sort -V | tail -1)
+          git checkout latest_release
           git log -1 --format='%H'
           git status
 
--- a/.github/workflows/tests_lkcpu_python.yml
+++ b/.github/workflows/tests_lkcpu_python.yml
@@ -48,7 +48,7 @@ jobs:
     uses: ./.github/workflows/build_and_cache_Kokkos_linux.yml
     with:
       os: ubuntu-22.04
-      kokkos_version: ${{ inputs.lightning-version == 'stable' && '4.2.00' || '4.3.01' }}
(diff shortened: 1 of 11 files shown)


.github/workflows/tests_lkcpu_python.yml
```diff
@@ -48,6 +48,7 @@ jobs:
     uses: ./.github/workflows/build_and_cache_Kokkos_linux.yml
     with:
       os: ubuntu-22.04
+      kokkos_version: ${{ inputs.lightning-version == 'stable' && '4.2.00' || '4.3.01' }} #To be updated next Release.
```
mlxd, line 51: Note that with the recent changes, 4.2 will likely fail if hit, as new functionality was added in 4.3 `e.g. kokkos_swap`. Though, correct me if wrong.

.github/workflows/tests_lkcpu_python.yml
```diff
@@ -180,6 +181,16 @@ jobs:
         run: |
           python -m pip install -U pennylane-lightning --no-deps
 
+      - name: Switch to stable tag of Lightning
+        if: inputs.lightning-version == 'stable'
+        run: |
+          cd main
+          git fetch --tags --force
+          git checkout $(git tag | sort -V | tail -1)
```
mlxd, line 189: Isn't this what `latest_release` was set to? The last released tag?

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -158,7 +158,7 @@ jobs:
           mv ${{ github.workspace }}/wheel_${{ matrix.pl_backend }}-${{ matrix.blas }}.whl ${{ github.workspace }}/$WHEEL_NAME
           python -m pip install -r requirements-dev.txt
           python -m pip install openfermionpyscf
-          python -m pip install ${{ github.workspace }}/$WHEEL_NAME --no-deps
+          python -m pip install ${{ github.workspace }}/$WHEEL_NAME --no-deps --force-reinstall
```
mlxd, line 161: Was this being skipped?

mlxd (commented):
Thanks @AmintorDusko 
Quick round of questions

.github/workflows/tests_lqcpu_python.yml
AmintorDusko: Thi one came from @vincentmr.

.github/workflows/tests_lkcpu_python.yml
AmintorDusko: 4.2.00 will only be used if calling the stable version. Currently, the stable version is based on the 4.2.00 version.

.github/workflows/tests_lkcpu_python.yml
AmintorDusko: The stable version is the one we push to pypi.
This version is based on the release branch.
The release branch has the tests, which should be for the stable version.

.github/workflows/tests_lkcpu_python.yml
mlxd: Ah, got it. :+1: 

Should this become 
```suggestion
      kokkos_version: ${{ (inputs.lightning-version == 'stable' && '4.2.00') || '4.3.01' }} #To be updated next Release.
```
for clarity then? Or will that influence the parsing?

.github/workflows/tests_lkcpu_python.yml
mlxd: I'm confused --- if we create a new tag, for any prupose, won't this influence this? I still don't see why `latest_release` cannot be used here. They give the same result, and the `latest_release` won't be moved if new tags are added:
```
git checkout $(git tag | sort -V | tail -1)
git checkout latest_release
```
Both give the same hash for me.

.github/workflows/tests_lkcpu_python.yml
AmintorDusko: I don't remember if this will break something or not. GitHub is kind of nitty pick with how we code workflows.
I'm following here their [prescription](https://docs.github.com/en/actions/learn-github-actions/expressions#example) for a ternary operator.

mlxd: Ok, I'll abide by that.

.github/workflows/tests_lkcpu_python.yml
AmintorDusko: Only tags that look like a release tag will be returned by this, and always the most recent.
But I agree that latest_release is equivalent.

mlxd: Just created a new tag starting with w though z and it superseded the expected order here.

mlxd: Yea, let's use `latest_release` instead, since that's depended upon for this purpose.

AmintorDusko: Sure thing. I didn't give too much thought to that, because of time and the fact that this piece of code has been running around for some time. But you are right. I will also update it in all other workflows. Thanks!

.github/workflows/tests_lqcpu_python.yml
AmintorDusko: Please feel free to unresolve if sensible.

mlxd (approved):
Thanks @AmintorDusko

.github/workflows/tests_lkcpu_python.yml
```diff
@@ -286,4 +297,4 @@ jobs:
         with:
           fail_ci_if_error: true
           verbose: true
-          token: ${{ secrets.CODECOV_TOKEN }}
+          token: ${{ secrets.CODECOV_TOKEN }}
```
multiphaseCFD, line 300: Do we want to add a blank line after to resolve the complain?

multiphaseCFD (approved):
Thanks @AmintorDusko ! Looks good to me!

.github/workflows/tests_lkcpu_python.yml
AmintorDusko: I don't see a complaint anywhere. Let's merge it like that for now.

PR #735, opened 2024-05-17
ISWAP Method Added

sungwooahn:
**Context:**
I conducted this ISWAP addition as part of the Xanadu applicant coding test. The following is the ISWAP repository that I added and verified.

I implemented handling for 7 bullets in the issue, enabling the ISWAP gate, which was not previously available, for use in PennyLane Lightning.

**Benefits:**
Even now, due to Lightning's nature, unnecessary computations are minimized, resulting in significant speed improvements. Furthermore, I will describe a method for achieving even faster execution using parallel programming in Bullet 4. We are limited to processing only one SWAP operation at a time using a single thread. However, by employing multiple threads, we can pair them up without overlap, allowing for the simultaneous update of two states. 

**Possible Drawbacks:**
The most crucial factor is how far apart the memory is. If the memory gap is significant (which occurs when the indices of the qubits to swap are large), there will be a larger difference in memory addresses accessed by each thread. This means that the memory accessed by all multiple threads will be located in different spaces. This will be the most significant difference.

**Related GitHub Issues:**
Nothing at the moment.

### DIFF
--- a/pennylane_lightning/core/_version.py
+++ b/pennylane_lightning/core/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.37.0-dev16"
+__version__ = "0.37.0-dev17"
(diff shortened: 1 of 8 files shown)


sungwooahn:
It seems multiple errors are present. I'll make updates when the tests are complete.

PR #737, opened 2024-05-21
Removing the `quimb` backend from `lightning.tensor`

PietropaoloFrisoni:
**Context:** This PR removes the `quimb` backend and the corresponding interface from the `lightning.tensor` device. Instead, a new `default.tensor` device [has been introduced in PennyLane](https://github.com/PennyLaneAI/pennylane/pull/5699). Due to requirements change, it has been decided to create a `default.tensor` device and move it to the PennyLane repository. 

**Benefits:** Having two separate devices simplifies the distinction between the C++ and the Python components. The relationship between `lightning.tensor` and `default.tensor` will be similar to that between `lightning.qubit` and `default.qubit`.

**Possible Drawbacks:** As we are removing an interface from an existing device, all the codes using the `quimb` interface in `lightning.tensor` now should use `default.tensor`, which lives in a different repository.

**Related GitHub Issues:** None.

**Related Shortcut Stories:** 
[sc-63277]
[sc-62257]


pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -211,9 +177,10 @@ def preprocess(
         * Does not support vector-Jacobian products.
         """
 
-        config = self._setup_execution_config(execution_config)
-        program = self._interface.preprocess()
-        return program, config
+        # TODO: remove comments when cuTensorNet MPS backend is available as a prototype
+        # config = self._setup_execution_config(execution_config)
+        # program = self._interface.preprocess()
+        # return program, config
```
mlxd, line 183: Can we uncomment these now? Do they need to be commented out?

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -230,7 +197,8 @@ def execute(
             TensorLike, tuple[TensorLike], tuple[tuple[TensorLike]]: A numeric result of the computation.
         """
 
-        return self._interface.execute(circuits, execution_config)
+        # TODO: remove comment when cuTensorNet MPS backend is available as a prototype
+        # return self._interface.execute(circuits, execution_config)
```
mlxd, line 201: Same question as above?

mlxd (commented):
Thanks @PietropaoloFrisoni 
Just a quick q or 2:

pennylane_lightning/lightning_tensor/lightning_tensor.py
PietropaoloFrisoni: Right now, the `interface` variable cannot have a non-trivial value since we still don't have the `cutensornet` MPS backend available (or, at least, not the corresponding Python layer). If we uncomment this we have an error since the `interface` variable does not 'link' to a `preprocess()` function. We can also remove these lines as they seem odd right now, but hopefully, it will be updated soon. What do you think?

pennylane_lightning/lightning_tensor/lightning_tensor.py
PietropaoloFrisoni: Same as above, but with `execute` instead of `preprocess`

mlxd:
Once the PR is update and green, I'll approve.

PietropaoloFrisoni:
@mlxd I am updating it now, but I am afraid it will never be green since our friend @codecov will fail anyway (although all lines are covered by tests)

mlxd (approved):
Thanks @PietropaoloFrisoni

multiphaseCFD (approved):
Thanks @PietropaoloFrisoni ! Looks good to me

PR #744, opened 2024-05-28
Patch Measurements.probs(wires) to Measurements.probs() when called with all wires

maliasadi:
### Before submitting

**Context:**
All calls to `qml.probs` to calculate the probs of the whole system in the analytic mode, patches to the C++ `Measurements.probs(wires, ...)` method that's implemented for computing the probabilities of a subset of wires. There is a more dedicated method (`Measurements.probs()`) that skips part of the logic in `Measurements.probs(wires)` which is indeed required to deal with subsystems. We should call `Measurements.probs()` when computing the probs of the whole system instead. This PR ensures that we patch to `Measurements.probs()` resulting a ~12% performance boost for qubits > 10.

- Update the C++ `Measurements.probs(wires, ...)` method to call `Measurements.probs()` in LQ and LK when requesting the probs of all wires.
- Tidy up the src code and docs

**Benefits:**
This will skip the unnecessary steps required for calculating the probs of a subsystem by calling the proper C++ `probs` implementation in LQ and LK. 

**Possible Drawbacks:**
Alternatively, we could fix this by calling to the proper `probs` C++ implementation in Python Bindings instead. I need to do a bit of investigation to find the better solution here. Any thoughts would be appreciated :)   

**Related GitHub Issues:**
[sc-63167]

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -48,9 +48,9 @@
 * Changed the name of `lightning.tensor` to `default.tensor` with the `quimb` backend.
   [(#719)](https://github.com/PennyLaneAI/pennylane-lightning/pull/719)
 
-* Dispatch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when requesting the probabilities for all wires.
-  This will trigger a more optimized implementation for calculating the probabilities of the whole system.
-  [(#)]()
+* Patch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when called with all wires.
+  This will trigger a more optimized implementation for calculating the probabilities of the entire system.
+  [(#744)](https://github.com/PennyLaneAI/pennylane-lightning/pull/744)
 
 ### Documentation
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -45,6 +45,10 @@
 * Changed the name of `lightning.tensor` to `default.tensor` with the `quimb` backend.
   [(#719)](https://github.com/PennyLaneAI/pennylane-lightning/pull/719)
 
+* Dispatch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when requesting the probabilities for all wires.
+  This will trigger a more optimized implementation for calculating the probabilities of the whole system.
+  [(#)]()
+
 ### Documentation
 
 ### Bug fixes
--- a/pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
@@ -80,11 +80,12 @@ class Measurements final
      * observables.
      *
      * @tparam functor_t Expectation value functor class for Kokkos dispatcher.
-     * @tparam nqubits Number of wires.
+     * @tparam num_wires Number of wires.
      * @param wires Wires to apply the observable to.
      */
     template <template <class> class functor_t, int num_wires>
-    PrecisionT applyExpValNamedFunctor(const std::vector<std::size_t> &wires) {
+    auto applyExpValNamedFunctor(const std::vector<std::size_t> &wires)
+        -> PrecisionT {
         if constexpr (num_wires > 0)
             PL_ASSERT(wires.size() == num_wires);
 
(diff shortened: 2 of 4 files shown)


.github/CHANGELOG.md
```diff
@@ -48,6 +48,10 @@
 * Changed the name of `lightning.tensor` to `default.tensor` with the `quimb` backend.
   [(#719)](https://github.com/PennyLaneAI/pennylane-lightning/pull/719)
 
+* Dispatch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when requesting the probabilities for all wires.
+  This will trigger a more optimized implementation for calculating the probabilities of the whole system.
+  [(#)]()
```
maliasadi, line 53:
```suggestion
* Patch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when called with all wires.
  This will trigger a more optimized implementation for calculating the probabilities of the entire system.
  [(#744)](https://github.com/PennyLaneAI/pennylane-lightning/pull/744)
```

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -506,11 +507,18 @@ class Measurements final
...
         PL_ABORT_IF_NOT(
             std::is_sorted(wires.cbegin(), wires.cend()),
             "LightningKokkos does not currently support out-of-order wire "
             "indices with probability calculations");
+
+        // If all wires are requested, dispatch to `this->probs()`
+        if (wires.size() == this->_statevector.getNumQubits()) {
+            return this->probs();
+        }
+
```
AmintorDusko, line 521: We have proper [Python bindings](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/bindings/Bindings.hpp#L439-L447) for probs(). The problem is that we always provide wires in the devices' class even if this is an all-wires situation. The more straightforward solution would be to have a test somewhere (like you did here) and redirect when possible. The best place for that is in the C++ layer. Also, this is a cheap check.

AmintorDusko (approved):
As I said before, we have proper [Python bindings](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/bindings/Bindings.hpp#L439-L447) for probs(). The problem is that we always provide wires in the devices' class even if this is an all-wires situation. The more straightforward solution would be to have a test somewhere (like you did in this PR) and redirect when possible. The best place for that is in the C++ layer. This is a cheap check. I'm happy to approve.

.github/CHANGELOG.md
```diff
@@ -48,6 +48,10 @@
 * Changed the name of `lightning.tensor` to `default.tensor` with the `quimb` backend.
   [(#719)](https://github.com/PennyLaneAI/pennylane-lightning/pull/719)
 
+* Patch the C++ `Measurements.probs(wires)` method in Lightning-Qubit and Lighnting-Kokkos to `Measurements.probs()` when called with all wires.
```
vincentmr, line 51: What about Lightning-GPU?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -80,11 +80,12 @@ class Measurements final
...
      *
      * @tparam functor_t Expectation value functor class for Kokkos dispatcher.
-     * @tparam nqubits Number of wires.
+     * @tparam num_wires Number of wires.
      * @param wires Wires to apply the observable to.
      */
     template <template <class> class functor_t, int num_wires>
-    PrecisionT applyExpValNamedFunctor(const std::vector<std::size_t> &wires) {
+    auto applyExpValNamedFunctor(const std::vector<std::size_t> &wires)
+        -> PrecisionT {
```
vincentmr, line 88: Is this for style? If so, do we want to use it everywhere from now on?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -506,11 +507,18 @@ class Measurements final
...
-    std::vector<PrecisionT> probs(const std::vector<std::size_t> &wires) {
+    auto probs(const std::vector<std::size_t> &wires)
+        -> std::vector<PrecisionT> {
         PL_ABORT_IF_NOT(
             std::is_sorted(wires.cbegin(), wires.cend()),
             "LightningKokkos does not currently support out-of-order wire "
             "indices with probability calculations");
+
+        // If all wires are requested, dispatch to `this->probs()`
+        if (wires.size() == this->_statevector.getNumQubits()) {
```
vincentmr, line 518: Do we want to also defer in this case?
```suggestion
        if (wires.empty() || wires.size() == this->_statevector.getNumQubits()) {
```

vincentmr (commented):
Thanks @maliasadi , just a couple questions but I'm basically ready to approve. Nice formatting and docstring edits.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
maliasadi: This is mainly for style and we have followed this in many Lightning src files.

.github/CHANGELOG.md
maliasadi: The `probs()` method in LGPU calls `probs(wires)` following the cuTensorNet API. This patch is not useful there. 

https://github.com/PennyLaneAI/pennylane-lightning/blob/81870b7bc427ce2362e29a64d77f33aa80062104/pennylane_lightning/core/src/simulators/lightning_gpu/measurements/MeasurementsGPU.hpp#L156-L162

vincentmr (approved):
LGTM, cheers @maliasadi .

PR #745, opened 2024-05-29
[Prototype] Generate Shared Libraries for LKokkos

AmintorDusko:
**Context:** Building LKokkos in C++ may be computationally costly. 

**Description of the Change:** This PR prototype provides the LKokkos library as a C++ package.

**Benefits:** We expect the building time for other C++ packages (as Catalyst) to improve.

**Checking the prototype locally:** Let's start with simple, because simple is good.

To simulate a more similar environment a final user would face, in a fresh environment with the requirements installed, let's build wheels for LQubit (compilation not required) and LKokkos:
```
make clean
SKIP_COMPILATION=True PL_BACKEND=lightning_qubit python setup.py bdist_wheel
CMAKE_PREFIX_PATH=$path_to_kokkos PL_BACKEND=lightning_kokkos python setup.py bdist_wheel
```
One can install both packages with pip and define the path for Kokkos' library (`path_to_kokkos`) and the path to the pennylane-lightning library (`path_to_lightning`).

Consider a simple piece of code where we initialize the StateVector Kokkos and normalize it (`main.cpp`):
```c++
#include <complex>
#include <iostream>
#include <vector>

#include <Kokkos_Core.hpp>

#include "LightningKokkos.hpp"

namespace {
using namespace Pennylane::LightningKokkos;
} // namespace

int main()
{
    using PrecisionT = double;
    using ComplexT = std::complex<PrecisionT>;

    constexpr ComplexT one{PrecisionT{1.0}, PrecisionT{0.0}};
    constexpr ComplexT half{PrecisionT{0.5}, PrecisionT{0.0}};
    constexpr ComplexT zero{PrecisionT{0.0}, PrecisionT{0.0}};

    std::vector<ComplexT> init_state = {one, zero, one, one,
                                        zero, zero, one, zero};

    std::vector<ComplexT> expected_state = {half, zero, half, half,
                                            zero, zero, half, zero};

    StateVectorKokkos<PrecisionT> state_vector(
        reinterpret_cast<ComplexT *>(init_state.data()), init_state.size());
    state_vector.normalize();

    std::vector<Kokkos::complex<PrecisionT>> data = state_vector.getDataVector();
    for (const auto& term : data){
        std::cout << term << std::endl;
    }

    return 0;
}
```

Having these installations, definitions and `main.cpp` in place, we can make:
```
g++ main.cpp -std=c++20 -llightning_kokkos_shared -L$path_to_lightning -I$path_to_lightning/include -fopenmp -lkokkoscore -L$path_to_kokkos/lib -I$path_to_kokkos/include
```

We can now compile code directly against the `lightning_kokkos_shared` library.

**Possible Drawbacks:** The wheel size will increase because we are building static and shared libraries. The Python bindings library is linked against the static library, providing the best performance. The shared library is delivered for developers to access the LKokkos class directly, without re-compilation.

**Related GitHub Issues:**


pennylane_lightning/core/src/utils/config.h
```diff
@@ -17,7 +17,8 @@
  * Config file for the path to scipy.libs at compile time.
  */
 
+
 #ifndef CONFIG_H
 #define CONFIG_H
-#define SCIPY_LIBS_PATH ""
+#define SCIPY_LIBS_PATH "/home/amintor/Projects/pennylane-lightning/venv/lightning_310/lib/python3.10/site-packages/scipy.libs"
```
AmintorDusko, line 23:
```suggestion
#define SCIPY_LIBS_PATH ""
```

pennylane_lightning/core/src/utils/config.h
```diff
@@ -17,6 +17,7 @@
  * Config file for the path to scipy.libs at compile time.
  */
 
+
```
AmintorDusko, line 20:
```suggestion
```

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ target_link_libraries(${PL_TENSOR} PUBLIC  lightning_compile_options
                                             ${PL_BACKEND}_tensornetBase
                                             ${PL_BACKEND}_utils
                                             )
-    
+
```
AmintorDusko, line 45:
```suggestion
    
```

pennylane_lightning/core/src/simulators/lightning_gpu/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ target_link_libraries(${PL_BACKEND} PUBLIC  lightning_compile_options
                                             ${PL_BACKEND}_utils
                                             ${PL_BACKEND}_gates
                                             )
-    
+
```
AmintorDusko, line 60:
``` suggestion
    
```

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.cpp
```diff
@@ -19,5 +19,5 @@
 
 #include "StateVectorKokkos.hpp"
 
-template class Pennylane::LightningKokkos::StateVectorKokkos<float>;
-template class Pennylane::LightningKokkos::StateVectorKokkos<double>;
+extern template class Pennylane::LightningKokkos::StateVectorKokkos<float>;
+extern template class Pennylane::LightningKokkos::StateVectorKokkos<double>;
```
mlxd, line 23: This likely shouldn't be extern --- it is telling the library that the objects are built elsewhere, but this is the file that we want to use to compile them for the .so

PR #749, opened 2024-06-02
assignment: add ISWAP functionality and unit tests

luotianwu:
### Before submitting

**Context:**

Add ISWAP functionality in BasicGateFunctors.hpp
Add ISWAP unit tests in : 
   StateVectorKokkos::applyMatrix/Param-Operation
  StateVectorKokkos::applyISWAP
Passed updated/new unit tests.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -372,6 +372,25 @@ void applySWAP(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
         });
 }
 
+template <class ExecutionSpace, class PrecisionT>
+void applyISWAP(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
+               const std::size_t num_qubits,
+               const std::vector<std::size_t> &wires,
+               [[maybe_unused]] const bool inverse = false,
```
vincentmr, line 379: If `inverse` is unused, does it mean ISWAP is an involutory matrix?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
```diff
@@ -28,6 +28,7 @@
 #include "TestHelpers.hpp"
 #include "TestHelpersWires.hpp"
 #include "Util.hpp"
+#include "iostream"
```
vincentmr, line 31: Why is that required?

pennylane_lightning/core/src/gates/Gates.hpp
```diff
@@ -1238,6 +1257,8 @@ std::vector<ComplexT<T>> getMatrix(const GateOperation gate_op,
         return getCZ<ComplexT, T>();
     case GateOperation::SWAP:
         return getSWAP<ComplexT, T>();
+    case GateOperation::ISWAP:
+        return getISWAP<ComplexT, T>();
```
vincentmr, line 1261: Is this correct regardless of the value of `inverse`?

vincentmr (commented):
Thanks for initiating this @luotianwu. Could you compare the performance of the ISWAP gate on `lightning.kokkos` between the SERIAL and OpenMP backends and explain how the ISWAP C++ Kokkos kernel is executed concurrently. Sweeping over target qubits you should see a hardware dependent pattern, can you explain where it comes from?

pennylane_lightning/core/src/gates/Gates.hpp
luotianwu: No, take inverse parameter into consideration, the adjoint matrix is different. It may take a parameter (inverse?-1:1) and multiply it to the image elements in the matrix.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
luotianwu: It is something I forgot to clean up, I was trying to print out something.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
luotianwu: ISWAP is not a involutory matrix. Its adjoint is not the same to itself. If inverse is true, that indicates to apply its adjoint. It requires additional code to handle that situation, which is missing here.

luotianwu:
The ISWAP kernel is launched by Kokkos::parallel_for in applyNC2Functor. The kernel will be executed in parallel over range 0-exp(num_qubits-2)-1. The kernel execution for different indices in the range can be assigned to multiple processing units. The state vector array has size exp(num_qubits), by operating two of the qubits, the state vector indices are grouped by 4 and thus needs to loop over exp(num_qubits-2) groups.
The hardware dependent pattern is passed by template parameter ExecutionSpace. The value passed for it is Kokkos::DefaultExecutionSpace. I think it get value from CMAKE configuration -DKokkos_ENABLE_OPENMP=ON, -DKokkos_ENABLE_CUDA=ON, -DKokkos_ENABLE_SERIAL=ON etc.

I tried to build with -DKokkos_ENABLE_OPENMP=ON and -DKokkos_ENABLE_CUDA=OFF. To see the performance, I added some time collection code in the kernel and also printed out the backend by KokkosExecSpace::name(). But somehow it always show CUDA as the backend. Unfortunately I didn't have more time to figure it out.

PR #778, opened 2024-06-26
Updated CI workflows to avoid race conditions on PRs with concurrency label

rashidnhm:
**Context:**
Currently there exists a race condition within the CI due to misconfiguration of the concurrency label.

This PR updates the affected workflows to fix the concurrency labels.

`${{ github.event }}` is not a string, but rather an object when you have, that in the concurrency label you get:

```yaml
# Label
group: my-workflow-${{ github.event }}-${{ github.ref }}

# Renders as
group: my-workflow-[object Object]-refs/heads/my-branch
```

This was causing working that had `pull_request` and `workflow_call` triggers to overstep on each other's runtime as they both rendered with the same concurrency label.

This PR changes the attribute to use the correct string name `event_name`.

**Benefits:**
Race conditions averted. No random job cancellations especially when PR needs wheels built.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None. [sc-66955](https://app.shortcut.com/xanaduai/story/66955/fix-bug-in-lightning-ci-causing-race-conditions-of-parallel-job-runs)

### DIFF
--- a/.github/workflows/tests_lgpu_cpp.yml
+++ b/.github/workflows/tests_lgpu_cpp.yml
@@ -30,7 +30,7 @@ env:
   TORCH_VERSION: 1.11.0+cpu
 
 concurrency:
-  group: tests_lgpu_cpp-${{ github.ref }}-${{ github.event }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
+  group: tests_lgpu_cpp-${{ github.ref }}-${{ github.event_name }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/update_dev_version.yml
+++ b/.github/workflows/update_dev_version.yml
@@ -4,6 +4,11 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+
 jobs:
   update-dev-version:
     name: Update version for development branches
--- a/.github/workflows/tests_lgpu_python.yml
+++ b/.github/workflows/tests_lgpu_python.yml
@@ -34,7 +34,7 @@ env:
   TORCH_VERSION: 1.11.0+cpu
 
 concurrency:
-  group: tests_lgpu_python-${{ github.ref }}-${{ github.event }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
+  group: tests_lgpu_python-${{ github.ref }}-${{ github.event_name }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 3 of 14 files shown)


vincentmr (approved):
LGTM, but do we want to target the RC branch?

rashidnhm:
> LGTM, but do we want to target the RC branch?

Probably best to have it in the default branch first. I'll open a subsequent PR to the rc branch

maliasadi (approved):
LGTM! Don't forget to update the changelog.

.github/workflows/update_dev_version.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
maliasadi, line 9: Temporarily changing this to false for checking CI jobs:
```suggestion
  cancel-in-progress: false
```

rashidnhm: Any reason for this @maliasadi ?

.github/workflows/update_dev_version.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
```
maliasadi, line 9:
```suggestion
  cancel-in-progress: true
```

.github/workflows/tests_lgpu_cpp.yml
```diff
@@ -30,7 +30,7 @@ env:
   TORCH_VERSION: 1.11.0+cpu
 
 concurrency:
-  group: tests_lgpu_cpp-${{ github.ref }}-${{ github.event }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
+  group: tests_lgpu_cpp-${{ github.ref }}-${{ github.event_name }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
```
maliasadi, line 33:
```suggestion
  group: tests_lgpu_cpp-${{ github.ref }}-${{ github.event }}-${{ inputs.lightning-version }}-${{ inputs.pennylane-version }}
```

rashidnhm:
@maliasadi @vincentmr 

Re-requesting a review as I've had to address an additional issue within the PR.

Scipy was being installed to the python venv on the self-hosted runner, but the build was using the system Python (and thus could not find the installed scipy). This was causing sporradic failures and bugs.


The solution was to export `Python_ROOT_DIR` and `Python3_ROOT_DIR` to hint cmake's `find_package` to use the correct venv python installation for workflows running on aws instances.

vincentmr (approved):
LGTM, thanks again @rashidnhm .

maliasadi (approved):
Thanks @rashidnhm! Looks good to merge now :)

PR #788, opened 2024-07-08
Do not run GPU tests and Docker workflows on release.

vincentmr:
### Before submitting

**Context:**
Some test workflows are run on release (GPU ones) and not others.
The Docker build workflow is also run, but it it missing required inputs.

Do not run the mentioned workflows on release. The Docker workflow in particular cannot be run because it requires the new PennyLane release which is not available when Lightning is released. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-68101]

### DIFF
--- a/.github/workflows/compat-docker-latest.yml
+++ b/.github/workflows/compat-docker-latest.yml
@@ -1,4 +1,4 @@
-name: Compat Check Docker - Lightning@stable
+name: Compat Check Docker - Lightning@latest
 
 on:
   schedule:
@@ -6,15 +6,15 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: docker_linux_x86_64-stable
+  group: docker_linux_x86_64-latest
   cancel-in-progress: true
 
 jobs:
-  docker_linux_x86_64_stable:
-    name: Docker stable - Linux::x86_64
+  docker_linux_x86_64_develop:
+    name: Docker develop - Linux::x86_64
     uses: ./.github/workflows/docker_linux_x86_64.yml
     with:
-      lightning-version: v0.37.0
-      pennylane-version: v0.37.0
+      lightning-version: latest
+      pennylane-version: latest
       push-to-dockerhub: false
     secrets: inherit # pass all secrets
--- a/.github/workflows/compat-docker-latest.yml
+++ b/.github/workflows/compat-docker-latest.yml
@@ -10,8 +10,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  docker_linux_x86_64_develop:
-    name: Docker develop - Linux::x86_64
+  docker_linux_x86_64_latest:
+    name: Docker latest - Linux::x86_64
     uses: ./.github/workflows/docker_linux_x86_64.yml
     with:
       lightning-version: latest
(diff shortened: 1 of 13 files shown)


.github/workflows/compat-docker-latest.yml
```diff
@@ -1,20 +1,20 @@
-name: Compat Check Docker - Lightning@latest
+name: Compat Check Docker - Lightning@stable
```
AmintorDusko, line 1: Stable because, our latest already diverged from stable, right?

AmintorDusko: Should we rename the script to stable? Or do we use it for latest too?

vincentmr: I realized that we refer to the latest official release as "stable" and the development (also our master branch) as "latest". I'm thus adopting the convention in the Docker workflow.

AmintorDusko: In this case, should we rename the script to:
.github/workflows/compat-docker-stable.yml ?

vincentmr: Good point. Will do.

.github/workflows/compat-docker-latest.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     name: Docker latest - Linux::x86_64
     uses: ./.github/workflows/docker_linux_x86_64.yml
     with:
-      lightning-version: v0.36.0
-      pennylane-version: v0.36.0
+      lightning-version: master
+      pennylane-version: master
```
AmintorDusko, line 18: Isn't the latest tag applicable here?

.github/workflows/compat-docker-stable.yml
```diff
@@ -1,20 +1,20 @@
...
-  docker_linux_x86_64_develop:
-    name: Docker develop - Linux::x86_64
+  docker_linux_x86_64_stable:
+    name: Docker stable - Linux::x86_64
     uses: ./.github/workflows/docker_linux_x86_64.yml
     with:
-      lightning-version: master
-      pennylane-version: master
+      lightning-version: v0.37.0
+      pennylane-version: v0.37.0
```
AmintorDusko, line 18: Isn't the stable tag applicable here?

AmintorDusko (commented):
some more questions.

.github/workflows/compat-docker-stable.yml
vincentmr: No, we do not maintain a `stable` tag. We could use `latest_release` for `lightning-version` but PennyLane does not have an equivalent tag it appears.

AmintorDusko (approved):
I have no blockers, but we could propose creating the necessary tags.
Having these tags in place would be a more maintainable solution.

vincentmr:
> I have no blockers, but we could propose creating the necessary tags. Having these tags in place would be a more maintainable solution.

Agreed. Let's do in a follow-up PR? I believe we'll soon improve on everything while including Catalyst into the mix, which will further motivate creating a more maintainable solution. We can then see with core and compilation what tags they would all like to use.

AmintorDusko:
> > I have no blockers, but we could propose creating the necessary tags. Having these tags in place would be a more maintainable solution.
> 
> Agreed. Let's do in a follow-up PR? I believe we'll soon improve on everything while including Catalyst into the mix, which will further motivate creating a more maintainable solution. We can then see with core and compilation what tags they would all like to use.

Agreed!

PR #796, opened 2024-07-12
Update toml files with sum_observables flag

lillian542:
**Context:**
Catalyst is adding `sum_observables` as a flag in the TOML file, and conditionally applying the `split_to_single_terms` transform in the QJIT device preprocessing if the backend doesn't allow sum observables (but doesn't need to apply `split_non_commuting`).

Update the lightning TOML files to indicate that sum observables are supported.

**Benefits:**
Lightning stays consistent with the changes in [this PR](https://github.com/PennyLaneAI/catalyst/pull/927)

### DIFF
--- a/pennylane_lightning/core/_version.py
+++ b/pennylane_lightning/core/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.38.0-dev6"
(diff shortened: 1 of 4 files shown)


lillian542:
[sc-60188] [sc-60187]

lillian542:
Never mind, we're going to infer this from the observables list instead of adding a flag

PR #819, opened 2024-07-26
Update kokkos files from catalyst repo to support MCM seeding

paul0403:
**Context:**
The lightning kokkos files in the Catalyst repo has changed since #770. We make a small update to track these changes.

The Catalyst PR that made the changes added seeding to qjit: https://github.com/PennyLaneAI/catalyst/pull/936

**Benefits:** unblocks kokkos with catalyst

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -15,8 +15,11 @@
 
 ### Improvements
 
+* Update the Catalyst-specific wrapping class for Lightning Kokkos to track Catalyst's new support for MCM seeding.
+  [(#819)](https://github.com/PennyLaneAI/pennylane-lightning/pull/819)
+
 * Shot batching is made more efficient by executing all the shots in one go on LightningQubit.
-  [#814](https://github.com/PennyLaneAI/pennylane-lightning/pull/814)
+  [(#814)](https://github.com/PennyLaneAI/pennylane-lightning/pull/814)
 
 * LightningQubit calls `generate_samples(wires)` on a minimal subset of wires when executing in finite-shot mode.
   [(#813)](https://github.com/PennyLaneAI/pennylane-lightning/pull/813)
@@ -50,9 +53,6 @@
 
 * Add a Catalyst-specific wrapping class for Lightning Kokkos.
   [(#770)](https://github.com/PennyLaneAI/pennylane-lightning/pull/770)
-
-* Update the Catalyst-specific wrapping class for Lightning Kokkos to track Catalyst's new support for MCM seeding.
-  [(819)](https://github.com/PennyLaneAI/pennylane-lightning/pull/819)
 
 ### Documentation
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -51,6 +51,9 @@
 * Add a Catalyst-specific wrapping class for Lightning Kokkos.
   [(#770)](https://github.com/PennyLaneAI/pennylane-lightning/pull/770)
 
+* Update the Catalyst-specific wrapping class for Lightning Kokkos to track Catalyst's new support for MCM seeding.
+  [(819)](https://github.com/PennyLaneAI/pennylane-lightning/pull/819)
+
 ### Documentation
 
 ### Bug fixes
@@ -74,7 +77,7 @@
 
 This release contains contributions from (in alphabetical order):
 
-Ali Asadi, Amintor Dusko, Vincent Michaud-Rioux, Shuli Shu
+Ali Asadi, Amintor Dusko, Vincent Michaud-Rioux, Shuli Shu, Paul Haochen Wang
 
 ---
 
--- a/pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
+++ b/pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
@@ -106,6 +106,10 @@ auto LightningKokkosSimulator::GetDeviceShots() const -> std::size_t {
     return this->device_shots;
 }
 
+void LightningKokkosSimulator::SetDevicePRNG(std::mt19937 *gen) {
+    this->gen = gen;
+}
+
 /// LCOV_EXCL_START
 void LightningKokkosSimulator::PrintState() {
     using std::cout;
@@ -480,7 +484,7 @@ auto LightningKokkosSimulator::Measure(QubitIdType wire,
     SetDeviceShots(device_shots);
 
     // It represents the measured result, true for 1, false for 0
-    bool mres = Lightning::simulateDraw(probs, postselect);
+    bool mres = Lightning::simulateDraw(probs, postselect, this->gen);
     auto dev_wires = getDeviceWires(wires);
     this->device_sv->collapse(dev_wires[0], mres ? 1 : 0);
     return mres ? this->One() : this->Zero();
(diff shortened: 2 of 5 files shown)


pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
```diff
@@ -63,6 +63,8 @@ class LightningKokkosSimulator final : public Catalyst::Runtime::QuantumDevice {
 
     std::size_t device_shots;
 
+    std::mt19937 *gen = nullptr;
```
maliasadi, line 66: Let's use uniform initialization over C-style init with assignment: 
```suggestion
    std::mt19937 *gen{nullptr};
```

vincentmr (changes requested):
So this seed is only used for the `measure` RNG. Is that intended? Otherwise looks good to me, but please update the change log.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -480,7 +484,7 @@ auto LightningKokkosSimulator::Measure(QubitIdType wire,
     SetDeviceShots(device_shots);
 
     // It represents the measured result, true for 1, false for 0
-    bool mres = Lightning::simulateDraw(probs, postselect);
+    bool mres = Lightning::simulateDraw(probs, postselect, this->gen);
```
maliasadi, line 487: Did you only need this for MCM? What about the rest of MPs?

maliasadi (changes requested):
Hmm... `SetDevicePRNG` is a virtual functions with a default implementation. The first method of this kind in the `QuantumDevice` API.

```
/home/runner/work/pennylane-lightning/pennylane-lightning/Build/pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/include/QuantumDevice.hpp:125:46: error: unused parameter ‘gen’ [-Werror=unused-parameter]
  125 |     virtual void SetDevicePRNG(std::mt19937 *gen){};
```

You need to use `[[maybe_unused]]` for `gen` to get all actions PASS.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
vincentmr: I would propose to upgrade `generate_samples` to take a `wires` argument, as in LQubit, and use that via a measurement object. Everything would then use Kokkos' `Random_XorShift64_Pool` RNG. But that is beyond the scope of this PR. If there is a motivation for it, I could quickly put that together however.

paul0403:
> So this seed is only used for the `measure` RNG. Is that intended? Otherwise looks good to me, but please update the change log.

This is intended (sampling could not be seeded from the catalyst repo as it is done through `Pennylane::LightningKokkos/Qubit::Measures::Measurements`)

https://github.com/PennyLaneAI/catalyst/blob/a49e0bc9f828a28aa8cae7d07c64d09d84713ef4/runtime/lib/backend/lightning/lightning_dynamic/LightningSimulator.cpp#L304

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
dime10: Absolutely it would be great for the MPs to also use the RNG instance, however it wasn't required to fix our flaky test. Since this requires additional internal device changes, we left it for a follow up.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
paul0403: done

paul0403:
> Hmm... `SetDevicePRNG` is a virtual function with default impl. The first method of this kind in the `QuantumDevice` API.
> 

This is because not all devices can be seeded, so in the device base class we need to supply a default implementation of doing nothing, and let the seedable devices override them.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
vincentmr: > Absolutely it would be great for the MPs to also use the RNG instance

To leverage Kokkos we need a parallel RNG (and to be straightforwardly performance portable we need Kokkos' RNG). So we could pass the seed to Kokkos' RNG, but not use `this->gen`.

maliasadi: > I would propose to upgrade generate_samples to take a wires argument, as in LQubit, and use that via a measurement object. Everything would then use Kokkos' Random_XorShift64_Pool RNG. 

@vincentmr :100:  but, let's accept this PR as is to unblock compilation, and we can get back to this when working on sc-64459.

.github/CHANGELOG.md
```diff
@@ -51,6 +51,9 @@
 * Add a Catalyst-specific wrapping class for Lightning Kokkos.
   [(#770)](https://github.com/PennyLaneAI/pennylane-lightning/pull/770)
 
+* Update the Catalyst-specific wrapping class for Lightning Kokkos to track Catalyst's new support for MCM seeding.
+  [(819)](https://github.com/PennyLaneAI/pennylane-lightning/pull/819)
```
maliasadi, line 55:
```suggestion
  [(#819)](https://github.com/PennyLaneAI/pennylane-lightning/pull/819)
```

maliasadi, line 55: Please add the new entry to the top of the list.

paul0403: ah, this was awkward, fixed now

paul0403: done

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
```diff
@@ -63,6 +63,8 @@ class LightningKokkosSimulator final : public Catalyst::Runtime::QuantumDevice {
 
     std::size_t device_shots;
 
+    std::mt19937 *gen{nullptr};
```
multiphaseCFD, line 66: Do we want to add the `<random>` header?

multiphaseCFD (commented):
Thanks @paul0403 ! LGTM.

vincentmr (approved):
LGTM, thanks @paul0403 .

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
paul0403: It isn't absolutely needed, since this file includes `QuantumDevice.hpp` which includes the random header https://github.com/PennyLaneAI/catalyst/blob/9e2f805024b98077472dc052aac1c715d8ab99eb/runtime/include/QuantumDevice.hpp#L20

But I have no objections to having it here as well for safety. The standard headers should have some sort of #pragma once or inclusion guards.

maliasadi (approved):
:raised_hands:

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
paul0403: done

PR #835, opened 2024-08-02
Adding semantic-version to workflow reqs

PietropaoloFrisoni:
This PR should fix [this test](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/10208461253/job/28245011292) in the plugin-test-matrix. The latest PL version does not use `semantic-version`, but the stable version still does, so we add such a package to the workflow requirements.


AmintorDusko (approved):
Thank you for the update!

maliasadi (approved):

PR #846, opened 2024-08-12
Use `decomposition` instead of `Operator.expand`

astralcai:
`Operator.expand` is deprecated (https://github.com/PennyLaneAI/pennylane/pull/5994). `Operator.decomposition` should be used instead.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -70,7 +70,7 @@
 * Move `setBasisState`, `setStateVector` and `resetStateVector` from `StateVectorLQubitManaged` to `StateVectorLQubit`.
   [(#841)](https://github.com/PennyLaneAI/pennylane-lightning/pull/841)
 
-* Remove use of the deprecated `Operator.expand` in favour of `Operator.decomposition`
+* Remove use of the deprecated `Operator.expand` in favour of `Operator.decomposition`.
   [(#846)](https://github.com/PennyLaneAI/pennylane-lightning/pull/846)
 
 ### Documentation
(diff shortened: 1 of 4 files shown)


pennylane_lightning/core/_serialize.py
```diff
@@ -404,7 +404,7 @@ def get_wires(operation, single_op):
                 uses_stateprep = True
                 continue
             if isinstance(operation, Rot):
-                op_list = operation.expand().operations
+                op_list = operation.decomposition()
```
maliasadi, line 407: Is it still using the same logic as `expand()` under the hood? What are the performance implications with using `decomposition` instead of `expand`?

maliasadi (commented):
LGTM, thanks @astralcai! Feel free to ping me again when you add the PR to `./.github/CHANGELOG.md`.

pennylane_lightning/core/_serialize.py
astralcai: `Operator.expand` calls `Operator.decomposition` under the hood and package the decomposed operations into a tape. In this case, creating a tape with a list of operations just to retrieve these operations from the tape is quite redundant. Using `decomposition` does literally the same thing.

.github/CHANGELOG.md
```diff
@@ -70,6 +70,9 @@
 * Move `setBasisState`, `setStateVector` and `resetStateVector` from `StateVectorLQubitManaged` to `StateVectorLQubit`.
   [(#841)](https://github.com/PennyLaneAI/pennylane-lightning/pull/841)
 
+* Remove use of the deprecated `Operator.expand` in favour of `Operator.decomposition`
```
maliasadi, line 73:
```suggestion
* Remove use of the deprecated `Operator.expand` in favour of `Operator.decomposition`.
```

maliasadi (approved):
:raised_hands:

AmintorDusko (approved):
Thank you very much, @astralcai!

PR #857, opened 2024-08-16
Replace toml with tomllib & tomli

maliasadi:
### Before submitting

**Context:**
Prefer `tomlkit` over `toml` for building Lightning wheels, and choose `tomli` and `tomllib` over `toml` when installing the package.

**Benefits:**
- `tomli` and `tomllib` (for python > 3.10) provides a read-only, lightweight and fast toml parser compare to `toml`
- `tomlkit` gets regular updates and provides support for all the toml updates and new standards. `toml` [hasn't been updated since Oct 2020](https://pypi.org/project/toml/).
**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-71707]


scripts/configure_pyproject_toml.py
```diff
@@ -51,7 +61,16 @@ def parse_args():
...
+    if not has_toml:
+        raise ImportError(
+            "A TOML parser is required to configure 'pyproject.toml'. "
+            "We support any of the following TOML parsers: [tomli, tomlkit, tomllib] "
+            "You can install either tomli via `pip install tomli`, "
+            "tomlkit via `pip install tomlkit`, or use Python 3.11 "
+            "or above which natively offers the tomllib library."
+        )
+
+    pyproject = tomllib.load(pyproject_path)
```
AmintorDusko, line 73: These other Python packages need a binary file object. Please update everywhere accordingly:
```suggestion
    with open(pyproject_path, "rb") as f:
        pyproject = tomllib.load(f)
```

scripts/configure_pyproject_toml.py
```diff
@@ -18,14 +18,24 @@
...
+from importlib.util import find_spec
 from backend_support import backend, device_name
 
 path_to_version = Path("pennylane_lightning").absolute() / "core" / "_version.py"
 with open(path_to_version, encoding="utf-8") as f:
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
 
+has_toml = False
+toml_libs = ["tomllib", "tomli", "tomlkit"]
```
mlxd, line 31: Note that we should test (locally is fine) that the syntax works against all of these different libraries.

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           cp -rf ${{ github.workspace }}/Kokkos_install/${{ matrix.exec_model }}/* Kokkos/
 
       - name: Install dependencies
-        run: python -m pip install cibuildwheel~=2.16.0 toml
+        run: python -m pip install cibuildwheel~=2.16.0 tomli
```
mlxd, line 123: If this is in the pyproject.toml (and setup.py as a runtime dep) there will be no need to set it here and everywhere.

mlxd (commented):
Thanks Ali

.github/workflows/wheel_linux_aarch64.yml
maliasadi: This is not in the pyproject.toml in favour of using tomllib builtin package for py3.11 and py3.12. But it'd be much cleaner to add tomli to pyproject that also preserves the toml parser consistency for all python versions.

scripts/configure_pyproject_toml.py
maliasadi: Good suggestion! I re-use the logic in PL here but worth testing the use cases for all different toml libs.

maliasadi: We need both `tomli` and `tomli-w` as `tomli` is read-only and `tomli-w` is required in `configure_pyproject_toml.py` to be able to update the setup config for other Lightning devices. I drop the `tomli` support as to keep the logic simple and straightforward.

.github/workflows/wheel_linux_aarch64.yml
maliasadi: A toml parser is required for configuring `pyproject.toml`. This configuration occurs before calling `setup.py` via `PL_BACKEND="lightning_sim" python scripts/configure_pyproject_toml.py`.

I had to replace `tomli` (w/o `tomli-w`) with `tomlkit` as the `configure_pyproject_toml.py` reads and writes into `pyproject.toml`, and `tomli` is a read-only toml parser.

scripts/configure_pyproject_toml.py
```diff
@@ -51,7 +65,16 @@ def parse_args():
...
     pyproject_path = os.path.join(parsed_args.path, "pyproject.toml")
 
-    pyproject = toml.load(pyproject_path)
+    if not has_toml:
+        raise ImportError(
+            "A TOML parser is required to configure 'pyproject.toml'. "
+            "We support any of the following TOML parsers: [tomlkit, tomllib] "
+            "You can install tomlkit via `pip install tomlkit`, "
+            "or use Python 3.11 or above which natively offers the tomllib library."
+        )
```
AmintorDusko, line 74:
```suggestion
        raise ImportError(
            "A TOML parser is required to configure 'pyproject.toml'. "
            "We support any of the following TOML parsers: " + str(toml_libs) +
            "You can install tomlkit via `pip install tomlkit`, "
            "or use Python 3.11 or above which natively offers the tomllib library."
        )
```
It would be nice for maintenance to have this or any variant.

scripts/configure_pyproject_toml.py
```diff
@@ -60,7 +86,7 @@ def parse_args():
         "cmake~=3.24.0",
         "ninja; platform_system!='Windows'",
         "setuptools>=42",
-        "toml",
+        "tomlkit",
```
vincentmr, line 89: If `tomlkit` is not found, do we want to ship with `toml`?

vincentmr (commented):
Why do we want to use `tomlkit` over `toml` again? Seems like we'll favor `tomlkit` so could you update the PR's title?

scripts/configure_pyproject_toml.py
maliasadi: We don't want to ship our packages with toml anymore. we should be able to remove `toml` from the list of `toml_libs` when PL drops support for `toml` in the future too. For now, we only rely on `tomlkit.load` when installing Lightning PyPI packages and it's available in every platform.

maliasadi:
> Why do we want to use tomlkit over toml again? Seems like we'll favor tomlkit so could you update the PR's title?

We initially aimed to transition away from using `toml` to more efficient, read-only toml parsers like `tomli` and `tomllib` (available in Python 3.11+). Both of these parsers offer standardized and lightweight toml parsing, which is more efficient compared to `toml`. This made them ideal for situations where only reading toml files was necessary. However, moving Lightning to the new build system, we need for modifying toml files, which required a more capable library like `tomlkit`. Unlike `toml`, `tomlkit` offers better support for structured files and is regularly updated to keep up with the latest standards.. I keep the `toml` logic as it's still a package requirement for PL and a fallback for Windows users.

We now use `tomlkit` (and `toml` only on Windows) for building wheels to handles modifications to toml files during the build process. `cibuildwheel` on Windows has a limited support for `tomlkit` and as such we are using `toml` as a fallback. For installation purposes, where only toml parsing is needed, we prefer `tomli` or `tomllib` over `toml`, as they provide a more efficient and standardized solution.

AmintorDusko:
> We now use `tomlkit` (and `toml` only on Windows) for building wheels to handles modifications to toml files during the build process. `cibuildwheel` on Windows has a limited support for `tomlkit` and as such we are using `toml` as a fallback.

@maliasadi, why the exception for Windows?

maliasadi:
> why the exception for Windows?

@AmintorDusko cibuildwheel on Windows expects toml polices to parse the `pyproject.toml` file instead of tomli or tomlkit. It doesn't have any implications on our build or install systems of Lightning.

.github/CHANGELOG.md
```diff
@@ -63,9 +66,6 @@ Amintor Dusko, Luis Alfredo Nuñez Meneses, Shuli Shu
 * Do not run GPU tests and Docker workflows on release.
   [(#788)](https://github.com/PennyLaneAI/pennylane-lightning/pull/788)
```
maliasadi, line 68:
```suggestion

* Update python packaging to follow PEP 517/518/621/660 standards.
  [(#832)](https://github.com/PennyLaneAI/pennylane-lightning/pull/832)

```

multiphaseCFD:
Thanks @maliasadi for the nice work. I'm wondering if it works with arm64?

setup.py
```diff
@@ -16,13 +16,39 @@
...
 from setuptools import setup, Extension, find_namespace_packages
 from setuptools.command.build_ext import build_ext
 
-project_name = toml.load("pyproject.toml")['project']['name']
+
+has_toml = False
+toml_libs = ["tomli", "tomllib", "tomlkit", "toml"]
+for pkg in toml_libs:
+    spec = find_spec(pkg)
+    if spec:
```
vincentmr, line 32:
```suggestion
    if find_spec(pkg):
```

vincentmr (approved):
LGTM, thanks @maliasadi .

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -125,7 +125,7 @@ jobs:
           cp -rf ${{ github.workspace }}/Kokkos_install/${{ matrix.exec_model }}/* Kokkos/
 
       - name: Install dependencies
-        run: python -m pip install cibuildwheel~=2.16.0 toml
+        run: python -m pip install cibuildwheel~=2.16.0 tomlkit
```
maliasadi, line 128: > I'm wondering if it works with arm64?

@multiphaseCFD yes, it works on arm64; the behaviour should be the same as both tomlkit and tomli/tomllib have consistent supports on arm64 and x86 arch

.github/CHANGELOG.md
AmintorDusko, line 1: You forgot your name.

AmintorDusko (approved):
Looks good!

setup.py
maliasadi: Let me keep this `find_spec` call before if-cond as it'd help debugging the logic specially when `find_spec` fails and you'd like to lookup/traverse these lines in the call stack.

.github/CHANGELOG.md
maliasadi: Good point! I'll add it in the next PR :smile:

multiphaseCFD (approved):
Great job!

PR #873, opened 2024-08-27
Chunk Hamiltonian, PauliSentence, LinearCombination [sc-65680]

vincentmr:
### Before submitting

      change and including a link back to the PR.

**Context:**
Parallelizing over observables can accelerate adjoint Jacobian calculations' backward pass. This PR revisits our implementation for L-Qubit and L-GPU which are the two devices that support it. Certain observables like Hamiltonian, PauliSentence, and LinearCombination can be split into many observables, enabling the distribution of the cost of expectation value computation. This strategy is initiated by the serializer which partitions the observables if `split_obs` is not `False`. The serializer proceeds to a complete partitioning, meaning a 1000-PauliWord PauliSentence is partitioned into a 1000 PauliWords. We note in passing that L-Qubit does not split observables since it does not pass a `split_obs` value to `_process_jacobian_tape`. This is wasteful because we end up with either of two situations: 

- The Jacobian is computed N processes (threads, devices, etc.) at a time which results in a lot of duplicate computation (forward/backward passes are repeated and the results combined);
- The Jacobian is parallelized over all observables, each of which requires a state vector copy which increases the memory requirements by as much. 

We explore chunking instead of full partitioning for LinearCombination-like objects, meaning a 1000-PauliWord PauliSentence is partitioned into four 250-PauliWords PauliSentences if we parallelize over 4 processes.

Modify the serializer to chunk LinearCombination-like objects if `self.split_obs` is truthy.
Correctly route `_batch_obs` such that L-Qubit splits observables.
Enhance/adapt tests.

**Analysis:**
**Lightning-Qubit**

`applyObservable` is a bottleneck for somewhat large linear combinations (say 100s or 1000s of terms). Chunking isn't helpful for a circuit like
```   
    @qml.qnode(dev, diff_method="adjoint")
    def c(weights):
        qml.templates.AllSinglesDoubles(weights, wires, hf_state, singles, doubles)
        return qml.expval(ham)
```
because L-Qubit's `applyObservable` method is parallelized over terms for a single `Hamiltonian` observable. Chunking in this case is counter-productive because it requires extra state vectors, extra backward passes, etc.

For a circuit like however
```    
    @qml.qnode(dev, diff_method="adjoint")
    def c(weights):
        qml.templates.AllSinglesDoubles(weights, wires, hf_state, singles, doubles)
        return np.array([qml.expval(ham), qml.expval(qml.PauliZ(0))])
```
`applyObservable` is parallelized over observables, which only scales up to 2 threads, and with poor load-balance. In this case, it is better to split the observable, which is what the current changes do.

| mol | master-serial | master-batched | chunk-serial | chunk-batched |
| --- | ------------- | -------------- | ------------ | ------------- |
| CH4 | 1.793e+01     | 1.330e+01      | 1.819e+01    | 8.040e+00     |
| Li2 | 5.333e+01     | 3.354e+01      | 5.289e+01    | 1.839e+01     |
| CO  | 9.817e+01     | 5.945e+01      | 9.619e+01    | 2.559e+01     |
| H10 | 1.220e+02     | 7.317e+01      | 1.182e+02    | 3.305e+01     |

So for this circuit the current PR yields speeds-up ranging from 1.5x to >2x by using obs-batching + chunking (compared with the previous obs-batching).

**Lightning-GPU** 

Lightning-GPU splits the observables as soon as `batch_obs` is true. The current code splits a Hamiltonian into all its individual terms, which is quite inefficient and induces a lot of redundant backward passes. This is visible benchmarking the circuit
```   
    @qml.qnode(dev, diff_method="adjoint")
    def c(weights):
        qml.templates.AllSinglesDoubles(weights, wires, hf_state, singles, doubles)
        return qml.expval(ham)
```

| mol | master-serial | master-batched | chunk-serial | chunk-batched |
| --- | ------------- | -------------- | ------------ | ------------- |
| CH4 | 1.463e+01     | forever        | 5.583e+00    | 3.405e+00     |
| Li2 | 1.201e+01     | forever        | 5.284e+00    | 2.658e+00     |
| CO  | 2.357e+01     | forever        | 4.716e+00    | 4.577e+00     |
| H10 | 2.992e+01     | forever        | 5.476e+00    | 5.469e+00     |
| HCN | 8.622e+01     | forever        | 3.144e+01    | 2.452e+01     |

The batched L-GPU runs are using 2 x A100 GPUs on ISAIC. The speed-ups for batched versus serial are OK, but most important is the optimization of `Hamiltonian::applyInPlace` which brings about nice speed-ups between master and this PR.
 
**Related GitHub Issues:**

### DIFF
--- a/pennylane_lightning/core/_serialize.py
+++ b/pennylane_lightning/core/_serialize.py
@@ -48,12 +48,6 @@ PAULI_NAME_MAP = {
 }
 
 
-def _chunk_iterable(iteration, num_chunks):
-    """Lazy-evaluated chunking of given iterable from https://stackoverflow.com/a/22045226"""
-    iteration = iter(iteration)
-    return iter(lambda: tuple(islice(iteration, num_chunks)), ())
-
-
 class QuantumScriptSerializer:
     """Serializer class for `pennylane.tape.QuantumScript` data.
 
--- a/pennylane_lightning/core/_serialize.py
+++ b/pennylane_lightning/core/_serialize.py
@@ -14,7 +14,8 @@
 r"""
 Helper functions for serializing quantum tapes.
 """
-from typing import List, Sequence, Tuple
+from itertools import islice
+from typing import List, Sequence, Tuple, Union
 
 import numpy as np
 import pennylane as qml
@@ -47,6 +48,14 @@ PAULI_NAME_MAP = {
 }
 
 
+def _chunk_iterable(iteration, num_chunks):
+    """Lazy-evaluated chunking of given iterable from https://stackoverflow.com/a/22045226
+    Replicated from lightning_base to avoid circular import
+    """
+    iteration = iter(iteration)
+    return iter(lambda: tuple(islice(iteration, num_chunks)), ())
+
+
 class QuantumScriptSerializer:
     """Serializer class for `pennylane.tape.QuantumScript` data.
 
@@ -54,7 +63,7 @@ class QuantumScriptSerializer:
     device_name: device shortname.
     use_csingle (bool): whether to use np.complex64 instead of np.complex128
     use_mpi (bool, optional): If using MPI to accelerate calculation. Defaults to False.
-    split_obs (bool, optional): If splitting the observables in a list. Defaults to False.
+    split_obs (Union[bool, int], optional): If splitting the observables in a list. Defaults to False.
 
     """
 
@@ -214,9 +223,29 @@ class QuantumScriptSerializer:
         obs = observable.obs if isinstance(observable, Tensor) else observable.operands
         return self.tensor_obs([self._ob(o, wires_map) for o in obs])
 
+    def _chunk_ham_terms(self, coeffs, ops, wires_map: dict = None, split_num: int = 1) -> List:
+        "Create split_num sub-Hamiltonians from a single high term-count Hamiltonian"
+        num_terms = len(coeffs)
+        step_size = num_terms // split_num + bool(num_terms % split_num)
+        c_coeffs = list(_chunk_iterable(coeffs, step_size))
+        c_ops = list(_chunk_iterable(ops, step_size))
+        return c_coeffs, c_ops
+
     def _hamiltonian(self, observable, wires_map: dict = None):
         coeffs, ops = observable.terms()
         coeffs = np.array(unwrap(coeffs)).astype(self.rtype)
+        if self.split_obs:
+            ops_l = []
+            for t in ops:
+                term_cpp = self._ob(t, wires_map)
+                if isinstance(term_cpp, Sequence):
+                    ops_l.extend(term_cpp)
+                else:
+                    ops_l.append(term_cpp)
+            c, o = self._chunk_ham_terms(coeffs, ops_l, wires_map, self.split_obs)
+            hams = [self.hamiltonian_obs(c_coeffs, c_obs) for (c_coeffs, c_obs) in zip(c, o)]
+            return hams
+
         terms = [self._ob(t, wires_map) for t in ops]
         # TODO: This is in case `_hamiltonian` is called recursively which would cause a list
         # to be passed where `_ob` expects an observable.
@@ -282,6 +311,11 @@ class QuantumScriptSerializer:
         terms = [self._pauli_word(pw, wires_map) for pw in pwords]
         coeffs = np.array(coeffs).astype(self.rtype)
 
+        if self.split_obs:
+            c, o = self._chunk_ham_terms(coeffs, terms, wires_map, self.split_obs)
+            psentences = [self.hamiltonian_obs(c_coeffs, c_obs) for (c_coeffs, c_obs) in zip(c, o)]
+            return psentences
+
         if len(terms) == 1 and coeffs[0] == 1.0:
             return terms[0]
 
@@ -326,16 +360,16 @@ class QuantumScriptSerializer:
         """
 
         serialized_obs = []
-        offset_indices = [0]
+        offset_indices = []
 
-        for observable in tape.observables:
+        for i, observable in enumerate(tape.observables):
             ser_ob = self._ob(observable, wires_map)
             if isinstance(ser_ob, list):
                 serialized_obs.extend(ser_ob)
-                offset_indices.append(offset_indices[-1] + len(ser_ob))
+                offset_indices.extend([i] * len(ser_ob))
             else:
                 serialized_obs.append(ser_ob)
-                offset_indices.append(offset_indices[-1] + 1)
+                offset_indices.append(i)
         return serialized_obs, offset_indices
 
     def serialize_ops(self, tape: QuantumTape, wires_map: dict = None) -> Tuple[
--- a/pennylane_lightning/core/_serialize.py
+++ b/pennylane_lightning/core/_serialize.py
@@ -14,7 +14,6 @@
 r"""
 Helper functions for serializing quantum tapes.
 """
-from itertools import islice
 from typing import List, Sequence, Tuple
 
 import numpy as np
--- a/pennylane_lightning/core/_serialize.py
+++ b/pennylane_lightning/core/_serialize.py
@@ -249,9 +249,6 @@ class QuantumScriptSerializer:
         # to be passed where `_ob` expects an observable.
         terms = [t[0] if isinstance(t, Sequence) and len(t) == 1 else t for t in terms]
 
-        if self.split_obs:
-            return [self.hamiltonian_obs([c], [t]) for (c, t) in zip(coeffs, terms)]
-
         return self.hamiltonian_obs(coeffs, terms)
 
     def _sparse_hamiltonian(self, observable, wires_map: dict = None):
@@ -317,8 +314,6 @@ class QuantumScriptSerializer:
         if len(terms) == 1 and coeffs[0] == 1.0:
             return terms[0]
 
-        if self.split_obs:
-            return [self.hamiltonian_obs([c], [t]) for (c, t) in zip(coeffs, terms)]
         return self.hamiltonian_obs(coeffs, terms)
 
     # pylint: disable=protected-access, too-many-return-statements
--- a/pennylane_lightning/core/_serialize.py
+++ b/pennylane_lightning/core/_serialize.py
@@ -360,17 +360,17 @@ class QuantumScriptSerializer:
         """
 
         serialized_obs = []
-        offset_indices = []
+        obs_indices = []
 
         for i, observable in enumerate(tape.observables):
             ser_ob = self._ob(observable, wires_map)
             if isinstance(ser_ob, list):
                 serialized_obs.extend(ser_ob)
-                offset_indices.extend([i] * len(ser_ob))
+                obs_indices.extend([i] * len(ser_ob))
             else:
                 serialized_obs.append(ser_ob)
-                offset_indices.append(i)
-        return serialized_obs, offset_indices
+                obs_indices.append(i)
+        return serialized_obs, obs_indices
 
     def serialize_ops(self, tape: QuantumTape, wires_map: dict = None) -> Tuple[
         List[List[str]],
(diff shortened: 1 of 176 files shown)


pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -633,8 +634,9 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
         # Check adjoint diff support
         self._check_adjdiff_supported_operations(tape.operations)
 
+        split_obs = self._dp.getTotalDevices() if self._batch_obs else False
```
mlxd, line 637: Any concerns here if this is hit on a node with MPI enabled? Should we allow the user to define the number instead, as we do with batch_obs?

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -653,68 +655,30 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
-                    processed_data["ops_serialized"],
-                    trainable_params,
-                )
-
+            jac = adjoint_jacobian.batched(
+                self._gpu_state,
+                processed_data["obs_serialized"],
+                processed_data["ops_serialized"],
+                trainable_params,
+            )
```
mlxd, line 663: With this change won't we lose the ability to parallelise over multiple different observable types with adjoint? Such as the case when a transform can generate different outputs for a single circuit?

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -214,41 +219,32 @@ def calculate_jacobian(self, tape: QuantumTape):
...
-                )
-                jac.extend(jac_local)
-        else:
-            jac = self._jacobian_lightning(
-                processed_data["state_vector"],
-                processed_data["obs_serialized"],
-                processed_data["ops_serialized"],
-                trainable_params,
-            )
+        jac = self._jacobian_lightning(
```
mlxd, line 231: Same comment here as earlier.

pennylane_lightning/core/_serialize.py
```diff
@@ -47,14 +48,22 @@
 }
 
 
+def _chunk_iterable(iteration, num_chunks):
```
mlxd, line 51: Are we removing this from its other defined location? Could we import it instead?

mlxd (commented):
Thanks @vincentmr 
A few things I'd like to understand first.

Also, given the questions, I'd like to also better understand the impact of this on previously run workloads, especially when using multiple GPUs, potential validation of the MPI4PY workload in the paper, and non-H workloads with many terms before considered a merge.

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -30,6 +30,7 @@
 from pennylane.measurements import Expectation, State
 from pennylane.ops.op_math import Adjoint
 from pennylane.wires import Wires
+from scipy.sparse import coo_matrix
```
AmintorDusko, line 33:
```suggestion
from scipy.sparse import csr_matrix
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -653,68 +655,30 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
...
-                        :,
-                    ]
-
+        jac = np.array(jac)
+
+        num_obs = len(np.unique(processed_data["obs_indices"]))
+        rows = processed_data["obs_indices"]
+        cols = np.arange(len(rows), dtype=int)
+        data = np.ones(len(rows))
+        red_mat = coo_matrix((data, (rows, cols)), shape=(num_obs, len(rows)))
```
AmintorDusko, line 677:
```suggestion
        red_mat = csr_matrix((data, (rows, cols)), shape=(num_obs, len(rows)))
```

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -23,9 +23,9 @@
 from pennylane.measurements import Expectation, MeasurementProcess, State
 from pennylane.operation import Operation
 from pennylane.tape import QuantumTape
+from scipy.sparse import coo_matrix
```
AmintorDusko, line 26:
```suggestion
from scipy.sparse import csr_matrix
```

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -214,41 +219,32 @@ def calculate_jacobian(self, tape: QuantumTape):
...
+            processed_data["ops_serialized"],
+            trainable_params,
+        )
         jac = np.array(jac)
+
+        num_obs = len(np.unique(processed_data["obs_indices"]))
+        rows = processed_data["obs_indices"]
+        cols = np.arange(len(rows), dtype=int)
+        data = np.ones(len(rows))
+        red_mat = coo_matrix((data, (rows, cols)), shape=(num_obs, len(rows)))
```
AmintorDusko, line 243:
```suggestion
        red_mat = csr_matrix((data, (rows, cols)), shape=(num_obs, len(rows)))
```

AmintorDusko (commented):
Just flying by.
CSR typically outperforms the COO sparse representation.

vincentmr (commented):
@mlxd Let's focus on L-Qubit first for simplicity, then I'll port whatever solution we end up with to L-GPU(+MPI). 

> Also, given the questions, I'd like to also better understand the impact of this on previously run workloads

I think at some point we might have lacked the ability to pass an entire Hamiltonian (or LinearCombination) down to the C++ layer. A way around this is to compute the Jacobian of the individual terms (which are simpler objects like TensorProducts) and sum them up. This is embarrassingly parallel, but it requires a lot of memory and computation. Not sure since when, but now we can pass a LinearCombination directly to the adjoint pipeline. I think the only part that is effectively parallelizable upon splitting is `applyObservables`, but this may seldom be a bottleneck.

pennylane_lightning/core/_serialize.py
```diff
@@ -214,17 +221,35 @@ def _tensor_ob(self, observable, wires_map: dict = None):
         obs = observable.obs if isinstance(observable, Tensor) else observable.operands
         return self.tensor_obs([self._ob(o, wires_map) for o in obs])
 
+    def _chunk_ham_terms(self, coeffs, ops, split_num: int = 1) -> List:
+        "Create split_num sub-Hamiltonians from a single high term-count Hamiltonian"
+        iperm = np.argsort(np.array([len(op.get_wires()) for op in ops]))
```
vincentmr, line 220: Distribute terms cyclically over resources in increasing number of wires.

pennylane_lightning/core/src/simulators/lightning_gpu/observables/ObservablesGPU.hpp
```diff
@@ -194,10 +194,11 @@ class Hamiltonian final : public HamiltonianBase<StateVectorT> {
             std::make_unique<DataBuffer<CFP_t>>(sv.getDataBuffer().getLength(),
                                                 sv.getDataBuffer().getDevTag());
         buffer->zeroInit();
+        StateVectorT tmp(sv);
 
         for (std::size_t term_idx = 0; term_idx < this->coeffs_.size();
              term_idx++) {
-            StateVectorT tmp(sv);
```
vincentmr, line 200: Creating SV's in a loop can be a bottleneck.

AmintorDusko:
I will re-trigger your CIs as there was some problem with the test pypi.

AmintorDusko (commented):
Looks good. I will come back later to check the CIs, after you merge master.

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -214,41 +214,39 @@ def calculate_jacobian(self, tape: QuantumTape):
...
-            )
+        jac = self._jacobian_lightning(
+            processed_data["state_vector"],
+            processed_data["obs_serialized"],
+            processed_data["ops_serialized"],
+            trainable_params,
+        )
         jac = np.array(jac)
-        jac = jac.reshape(-1, len(trainable_params)) if len(jac) else jac
+        has_shape0 = bool(len(jac))
```
multiphaseCFD, line 239: What does the `0` mean here?

tests/test_adjoint_jacobian.py
```diff
@@ -1327,29 +1329,23 @@ def test_integration_custom_wires(returns):
 
     def circuit(params):
         circuit_ansatz(params, wires=custom_wires)
-        return qml.expval(returns), qml.expval(qml.PauliY(custom_wires[1]))
+        return np.array([qml.expval(returns), qml.expval(qml.PauliY(custom_wires[1]))])
```
multiphaseCFD, line 1332: Does `qml.expval(returns), qml.expval(qml.PauliY(custom_wires[1])) `work?Do users have to change their python code to get this feature?

multiphaseCFD (commented):
Thanks @vincentmr for the nice work! Just a few questions.

pennylane_lightning/core/src/simulators/lightning_kokkos/observables/ObservablesKokkos.hpp
```diff
@@ -189,9 +189,10 @@ class Hamiltonian final : public HamiltonianBase<StateVectorT> {
     void applyInPlace(StateVectorT &sv) const override {
         StateVectorT buffer{sv.getNumQubits()};
         buffer.initZeros();
+        StateVectorT tmp{sv};
         for (std::size_t term_idx = 0; term_idx < this->coeffs_.size();
              term_idx++) {
-            StateVectorT tmp{sv};
+            tmp.updateData(sv.getView());
```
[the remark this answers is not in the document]
vincentmr: Yes.

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
vincentmr: It stands for `shape[0]`.

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -14,7 +14,7 @@
 r"""
 Internal methods for adjoint Jacobian differentiation method.
 """
-from os import getenv
+import os
```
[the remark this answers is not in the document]
vincentmr: Hmm, let me revert it.

pennylane_lightning/core/_serialize.py
```diff
@@ -214,17 +214,40 @@ def _tensor_ob(self, observable, wires_map: dict = None):
         obs = observable.obs if isinstance(observable, Tensor) else observable.operands
         return self.tensor_obs([self._ob(o, wires_map) for o in obs])
 
+    def _chunk_ham_terms(self, coeffs, ops, split_num: int = 1) -> List:
+        "Create split_num sub-Hamiltonians from a single high term-count Hamiltonian"
+        num_terms = len(coeffs)
+        iperm = np.argsort(np.array([len(op.get_wires()) for op in ops]))
+        coeffs = [coeffs[i] for i in iperm]
+        ops = [ops[i] for i in iperm]
```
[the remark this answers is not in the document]
vincentmr: We could do something like `pairs = [(coeffs[i], ops[i]) for i in iperm]`, but I'm not sure how that would simplify the code. What do you have in mind?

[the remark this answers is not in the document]
multiphaseCFD: Current implementation is clear and concise. But what I would like to suggest is loop iperm for once instead of twice.

tests/test_adjoint_jacobian.py
multiphaseCFD: I see `casted_to_array_def` was removed

pennylane_lightning/core/_serialize.py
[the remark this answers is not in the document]
vincentmr: I see, introduce a good old `for` loop :) Let's keep it as is and revisit if this shows up in a perf profile.

multiphaseCFD (approved):
Great job! Thanks @vincentmr !

PR #887, opened 2024-09-04
Update/actions

AmintorDusko:
**Context:** GitHub Actions need to be updated.

**Description of the Change:** GH detected a [high-severity](https://github.com/advisories/GHSA-cxww-7g56-2vh6) vulnerability, with the download artifact action.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-72698]

### DIFF
--- a/.github/workflows/post_release_version_bump.yml
+++ b/.github/workflows/post_release_version_bump.yml
@@ -29,7 +29,7 @@ jobs:
           --changelog_path "./.github/CHANGELOG.md" --post_release
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v3
+        uses: peter-evans/create-pull-request@v7
         with:
           commit-message: post release version bump
           title: Version Bump
--- a/.github/workflows/tests_lgpu_python.yml
+++ b/.github/workflows/tests_lgpu_python.yml
@@ -237,7 +237,7 @@ jobs:
       - name: Upload code coverage results
         uses: actions/upload-artifact@v4
         with:
-          name: ubuntu-codecov-results-python
+          name: ubuntu-codecov-results-python-${{ matrix.pl_backend }}
           path: ./main/coverage-${{ github.job }}-${{ matrix.pl_backend }}.xml
           retention-days: 1
           if-no-files-found: error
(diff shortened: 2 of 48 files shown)


.github/workflows/post_release_version_bump.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           --changelog_path "./.github/CHANGELOG.md" --post_release
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v3
+        uses: peter-evans/create-pull-request@v7
```
vincentmr, line 32: Are we sure this won't break the workflow?

.github/workflows/tests_lgpu_python.yml
```diff
@@ -235,11 +235,13 @@ jobs:
           mv coverage.xml coverage-${{ github.job }}-${{ matrix.pl_backend }}.xml
 
       - name: Upload code coverage results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ubuntu-codecov-results-python
           path: ./main/coverage-${{ github.job }}-${{ matrix.pl_backend }}.xml
```
vincentmr, line 241: Is there no conflict between backends here?

vincentmr (commented):
Thanks @AmintorDusko , disk space is a real nuisance so I hope `retention-days: 1` will help.

maliasadi (approved):
Nice work! Thanks @AmintorDusko :partying_face: Feel free to merge it after updating the changelog.

.github/workflows/post_release_version_bump.yml
AmintorDusko: Sure, Sure we don't. I'm optimistic, though. I can undo the change if you think it is too risky.

.github/workflows/tests_lgpu_python.yml
AmintorDusko: We have a single Backend here. I will expand the name, though.

.github/workflows/post_release_version_bump.yml
vincentmr: I'm ok either way. Let's see how it does next release.

vincentmr (approved):
LGTM, thanks again @AmintorDusko .

PR #908, opened 2024-09-11
Add (check) full support for sampling in full parity with Lightning.

LuisAlfredoNu:
### Before submitting

**Context:**
Migrate LightningGPU to the new device API

Add all the functionality of the new device API to LGPU.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
## **Freezzed PR** :warning: :snowflake: 
To make a smooth integration of LightningGPU with the new device API, we set the branch `gpuNewAPI_backend` as the base branch target for future developments related to this big task. 

The branch `gpuNewAPI_backend` has the mock of all classes and methods necessary for the new API. Also, several tests were disabled with 
``` python
if device_name == "lightning.gpu":
    pytest.skip("LGPU new API in WIP.  Skipping.",allow_module_level=True)
```
However, these tests will unblocked as the implementation progresses.

After all the developments for integrating LightningGPU with the new API have been completed then the PR will be open to merge to `master`

[sc-70946]


tests/new_api/test_device.py
```diff
@@ -652,6 +679,11 @@ def test_derivatives_single_expval(
         self, theta, phi, dev, obs, execute_and_derivatives, batch_obs
     ):
         """Test that the jacobian is correct when a tape has a single expectation value"""
+        if isinstance(obs, qml.SparseHamiltonian) and dev.dtype == np.complex64:
+            pytest.skip(
+                reason="The conversion from qml.Hamiltonian to SparseHamiltonian is only possible with np.complex128"
+            )
```
LuisAlfredoNu, line 685: This limitation comes from a hard code in Pennylane [ref](https://github.com/PennyLaneAI/pennylane/blob/060bb9a7479dcf6735671d809da19d500c4b7a89/pennylane/ops/qubit/hamiltonian.py#L522)

AmintorDusko (approved):
Looks good!

tests/new_api/test_device.py
```diff
@@ -66,6 +63,25 @@
         validate_observables,
     )
 
+if device_name == "lightning.gpu":
```
maliasadi, line 66: should we use `elif` instead of `if` and raise an error or skip for `else`?

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -377,7 +489,13 @@ def supports_derivatives(
             Bool: Whether or not a derivative can be calculated provided the given information
 
         """
-        return 0
+        if execution_config is None and circuit is None:
+            return True
+        if execution_config.gradient_method not in {"adjoint", "best"}:
+            return False
+        if circuit is None:
+            return True
```
maliasadi, line 497:
```suggestion
        if circuit is None or (execution_config is None and circuit is None):
            return True
        if execution_config.gradient_method not in {"adjoint", "best"}:
            return False
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -358,7 +458,19 @@ def execute(
         Returns:
             TensorLike, tuple[TensorLike], tuple[tuple[TensorLike]]: A numeric result of the computation.
         """
-        return 0
+        results = []
+        for circuit in circuits:
+            if self._wire_map is not None:
+                [circuit], _ = qml.map_wires(circuit, self._wire_map)
```
maliasadi, line 464: Why do we need this list promotion here?
```suggestion
                circuit, _ = qml.map_wires(circuit, self._wire_map)
```

maliasadi (approved):
Nice work @LuisAlfredoNu!

tests/new_api/test_device.py
LuisAlfredoNu: @maliasadi 
instead of `skip`. Would it not be better to throw an error? :thinking:

pennylane_lightning/lightning_gpu/lightning_gpu.py
LuisAlfredoNu: Looks like this is necessary

pennylane_lightning/lightning_gpu/lightning_gpu.py
LuisAlfredoNu: I try this change but break something in the unit test. :sweat_smile:  
The explanation is the following:
> if `circuit` is None but `execution_config` != None then the logics need to reach the last `return`. With this suggestion, the return always will be `True` and never reach the last `return`

PR #859, opened 2024-08-19
Add MPO support  to `lightning.tensor`

multiphaseCFD:
### Before submitting

Please complete the following checklist when submitting a PR:

- [ ] All new features must include a unit test.
      If you've fixed a bug or added code that should be tested, add a test to the
      [`tests`](../tests) directory!

- [ ] All new functions and code must be clearly commented and documented.
      If you do make documentation changes, make sure that the docs build and
      render correctly by running `make docs`.

- [ ] Ensure that the test suite passes, by running `make test`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the
      change, and including a link back to the PR.

- [x] Ensure that code is properly formatted by running `make format`. 

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

[SC-70902]

MPO(Matrix Product Operator) is added to `lightning.tensor` for all gates support. Any 2 more wires gates and 1 more target wire controlled gates can be supported via `MPO`. Current implementation only accepts the decomposed gate data.  For the python users, `MPO` decomposition is conducted in the python layer with `numpy`. Further improvement can be done by conducting MPO decomposition with `cusolver` for large unitary matrices. 

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #909, opened 2024-09-16
Update ``lightning.tensor`` docs

multiphaseCFD:
### Before submitting

**Context:**

[SC-65786] and [SC-68333]

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/lightning_tensor/device.rst
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
 The ``lightning.tensor`` device dispatches all operations to be performed on a CUDA-capable GPU of generation SM 7.0 (Volta)
 and greater.
 
-.. note:: 
-    Given the inherent parallelism of GPUs, simulations with intensive parallel computation, such as those with larger maximum
-    bond dimensions, stand to gain the most from the computational power offered by GPU and those simulations can benifit from the 
-    ``lightning.tensor`` device.  It's worth noting that if the bond dimension used in the simulation is small, the ``lightning.tensor`` 
-    device with ``MPS`` running a GPU may perform slower compared to a ``default.tensor`` device with ``MPS`` running on a CPU. For more details
-    on how bond dimension affects the simulation performance, please refer to the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
+Some tips on the usage of the ``lightning.tensor`` device:
```
tomlqc, line 22: Please integrate these points here
> This device supports both analytical and finite shots measurements. Currently, the supported differentiation methods are ``parameter-shift`` and ``finite-diff``. Note that MPS backend of ``lightning.tensor`` supports multi-wires gate via Matrix Product Operators (MPO). Disable the new arithmetic operator dunders is recommended since only 1-wire Hermitian is supported by ``lightning.tensor``. It is recommended to use the shot measurements for the ``lightning.tensor`` device. The calculation of analytical probabilities and variances may result in excessive memory usage for large-scale quantum simulations or become impractical due to prohibitive computational costs. It's also recommended to disable ``new_opmath`` for the ``lightning.tensor`` device since ``lightning.tensor`` only supports 1-wire Hermitian observables.

doc/lightning_tensor/device.rst
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
...
+    - ``lightning.tensor`` performs better for the maximum bond dimension MPS calculation. Given the inherent parallelism of GPUs, simulations with intensive parallel computation, such as those with larger maximum bond dimensions, stand to gain the most from the computational power offered by GPU and those simulations can benifit from the ``lightning.tensor`` device.  It's worth noting that if the bond dimension used in the simulation is small, the ``lightning.tensor`` device with ``MPS`` running a GPU may perform slower compared to a ``default.tensor`` device with ``MPS`` running on a CPU. For more details on how bond dimension affects the simulation performance, please refer to the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
```
tomlqc, line 23:
```suggestion
    - ``lightning.tensor`` performs better for the maximum bond dimension MPS calculation. Given the inherent parallelism of GPUs, simulations with intensive parallel computation, such as those with larger maximum bond dimensions, stand to gain the most from the computational power offered by GPU and those simulations can benefit from the ``lightning.tensor`` device.  It's worth noting that if the bond dimension used in the simulation is small, the ``lightning.tensor`` device with ``MPS`` running a GPU may perform slower compared to a ``default.tensor`` device with ``MPS`` running on a CPU. For more details on how bond dimension affects the simulation performance, please refer to the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
```

README.rst
```diff
@@ -49,7 +49,7 @@ PennyLane-Lightning high performance simulators include the following backends:
...
...: is a tensor network simulator based on the `NVIDIA cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`_ (requires NVIDIA GPUs with SM 7.0 or greater). This device is designed to simulate large-scale quantum circuits using tensor networks. For small circuits, state-vector simulator plugins may be more suitable. The supported method is Matrix Product State (MPS). This device supports both analytical and finite shots measurements. Currently, the supported differentiation methods are ``parameter-shift`` and ``finite-diff``. Note that MPS backend of ``lightning.tensor`` supports multi-wires gate via Matrix Product Operators (MPO). Disable the new arithmetic operator dunders is recommended since only 1-wire Hermitian is supported by ``lightning.tensor``. It is recommended to use the shot measurements for the ``lightning.tensor`` device. The calculation of analytical probabilities and variances may result in excessive memory usage for large-scale quantum simulations or become impractical due to prohibitive computational costs. It's also recommended to disable ``new_opmath`` for the ``lightning.tensor`` device since ``lightning.tensor`` only supports 1-wire Hermitian observables.
```
tomlqc, line 52:
```suggestion
* ``lightning.tensor``: is a tensor network simulator based on the `NVIDIA cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`_ (requires NVIDIA GPUs with SM 7.0 or greater). The supported method is Matrix Product State (MPS).

File: doc/lightning_tensor/device.rst (hunk 2 above)
Comment (tomlqc) [reply doc/lightning_tensor/device.rst:23]: Just a typo

File: doc/lightning_tensor/device.rst [hunk 4]
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
...
+    - The ``lightning.tensor`` device is optimized for large-scale quantum simulations. For small-scale quantum simulations, the overhead of transferring data between the CPU and GPU may outweigh the benefits of GPU acceleration.
+    - For the ``lightning.tensor`` device, it is recommended to use shot-based ``probs()`` measurements. The analytical calculation of ``prob()`` can lead to excessive memory usage or become impractical due to high computational costs for large-scale quantum simulations.
+    - Similarly, shot-based ``var()`` measurements are recommended for the ``lightning.tensor`` device. The analytical calculation of ``var()`` may also result in excessive memory usage or be impractical due to prohibitive computational costs for large-scale quantum simulations.
+    - It is advisable to disable ``new_opmath`` for the ``lightning.tensor`` device, as it only supports 1-wire Hermitian observables.
```
tomlqc, line 27:
```suggestion
    - It is advisable to disable ``new_opmath`` for the ``lightning.tensor`` device, as the device only supports 1-wire Hermitian observables.
```

tomlqc: Just to make 100% clear that's the device and not the new_opmath option that supports only the 1-wire observables

tomlqc (changes requested):
Awesome @multiphaseCFD

.github/CHANGELOG.md
```diff
@@ -66,6 +66,9 @@
 
 ### Documentation
 
+*  Update ``lightning.tensor`` documentation on all the new features.
```
Qottmann, line 69: Perhaps specify which features those are, I could imagine looking at this entry in one year and not knowing what was meant

README.rst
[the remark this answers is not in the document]
Qottmann: +1

Qottmann, line 52: Concerning the contents of this section though:
> Disable the new arithmetic operator dunders is recommended since only 1-wire Hermitian is supported by ``lightning.tensor``.

Is this referring to disable new opmath? If yes, would be good to explicitly give the command.

> It is recommended to use the shot measurements for the ``lightning.tensor`` device.

Why!??? Sounds very counter intuitive to me. Running with finite shots should be much more expensive I'd guess

> Disable the new arithmetic operator dunders is recommended since only 1-wire Hermitian is supported by ``lightning.tensor``.

I dont think we should generally recommend to disable new opmath just because `qml.Hermitian` is not supported for multi-wires (yet). Rather, I'd phrase it as that it is currently not supported but if they really _need_ it they can disable new opmath

> The calculation of analytical probabilities and variances may result in excessive memory usage for large-scale quantum simulations or become impractical due to prohibitive computational costs.

What do you mean with "analytical probabilities"? Are you referring to returning the state as a dense vector / and or all probabilities (i.e. absolute value squared of state amplitudes)? If yes, I would rephrase it that returning the state vector `qml.state()` (or similarly, `qml.probs`, returns the full dense state vector at the moment, and is thus only suitable for small systems. Speaking of which: are we planning to add support for returning the unterlyding MPS state?

doc/lightning_tensor/device.rst
Qottmann: Can we shorten this?

```suggestion
    - ``lightning.tensor`` performs better for the larger bond dimensions in MPS calculations, see the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
```

doc/lightning_tensor/device.rst
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
 The ``lightning.tensor`` device dispatches all operations to be performed on a CUDA-capable GPU of generation SM 7.0 (Volta)
```
Qottmann, line 19: > ``cutoff`` (singular value truncation threshold) defaults to ``0`` .

We are currently not actually using this, right? Perhaps worth a small comment that this is actually not supported yet

Qottmann: Or was it just for MPOs where this isnt used?

Qottmann: > ``cutoff_mode`` (singular value truncation mode) defaults to ``abs`` , considering the absolute values of the singular values; Alternatively, users can opt to set ``cutoff_mode`` to ``rel`` to consider the relative values of the singular values. 

Singular values are always non-negative and real, why do we need this at all?

doc/lightning_tensor/device.rst
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
...
+    - The ``lightning.tensor`` device is optimized for large-scale quantum simulations. For small-scale quantum simulations, the overhead of transferring data between the CPU and GPU may outweigh the benefits of GPU acceleration.
+    - For the ``lightning.tensor`` device, it is recommended to use shot-based ``probs()`` measurements. The analytical calculation of ``prob()`` can lead to excessive memory usage or become impractical due to high computational costs for large-scale quantum simulations.
```
Qottmann, line 25: This really confuses me, can you elaborate what you are recommending here and what is the underlying issue?

doc/lightning_tensor/device.rst
```diff
@@ -19,12 +19,12 @@ The default setup for the MPS tensor network approximation is:
...
+    - Similarly, shot-based ``var()`` measurements are recommended for the ``lightning.tensor`` device. The analytical calculation of ``var()`` may also result in excessive memory usage or be impractical due to prohibitive computational costs for large-scale quantum simulations.
```
Qottmann, line 26: Same as above, why is shot based stuff better?

doc/lightning_tensor/device.rst
Qottmann: same recommendation as above

doc/lightning_tensor/device.rst
```diff
@@ -58,7 +58,7 @@ Check out the :doc:`/lightning_tensor/installation` guide for more information.
 Operations and observables support
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The "lightning.tensor" supports 1- and 2-wire gate operations and all other operations that can be decomposed to that by PennyLane.
+The "lightning.tensor" supports all gate operations supported by PennyLane.
```
Qottmann, line 61: 🙌

Qottmann (changes requested):
Great work @multiphaseCFD !

I am bit confused about some of the recommendations you are making, perhaps I am missing something so it would be good to clarify that to me (and to the user)

doc/lightning_tensor/device.rst
multiphaseCFD: > >cutoff (singular value truncation threshold) defaults to 0 .

> We are currently not actually using this, right? Perhaps worth a small comment that this is actually not supported yet

``cutoff`` is supported by the ``lightning.tensor`` device. A short description on how the ``cutensornet`` backend reduce the bond extent dimensions is added.

multiphaseCFD: > Or was it just for MPOs where this isnt used?
It is for both `MPS` and `MPO` in the `cutensornet` context as you can find out the details in the newly added links.

multiphaseCFD: > > ``cutoff_mode`` (singular value truncation mode) defaults to ``abs`` , considering the absolute values of the singular values; Alternatively, users can opt to set ``cutoff_mode`` to ``rel`` to consider the relative values of the singular values. 

> Singular values are always non-negative and real, why do we need this at all?

This is a feature offered by ``cutensornet``. You can get more info [here](https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/api/types.html#cutensornettensorsvdconfigattributes-t).

doc/lightning_tensor/device.rst
multiphaseCFD: > The ``lightning.tensor`` device is recommended for large-scale quantum simulations. For small-scale simulations, the cost of SVD decomposition and MPS-MPO multiplication might outweigh than the benifits of MPS.

Is the statement above more clear now?

doc/lightning_tensor/device.rst
multiphaseCFD: Good question! The updated context should be more clear.

README.rst
multiphaseCFD: Please check my replies in the ``device.rst``

.github/CHANGELOG.md
multiphaseCFD: Done.

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,18 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
+    - ``cutoff_mode`` (singular value truncation mode) defaults to ``abs`` , considering the absolute values of the singular values; Alternatively, users can opt to set ``cutoff_mode`` to ``rel`` to consider the relative values of the singular values.
+Note that the ``cutensornet`` will automatically determin the reduced extent of the bond dimension based on the lowest among the multiple truncation cutoffs (``max_bond_dim``, ``cutoff-abs`` and ``cutoff-rel``). For more details on how the ``cutoff`` works, please check it out the `cuQuantum documentation <https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/api/types.html#cutensornettensorsvdconfigattributes-t>`__. All swap and decomposition operations in MPS-MPO multiplication will follow the same constraints set by SVD configurations (``cutoff``) and target extents setups. For more details, please refer to the `cuQuantum documentation <https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/api/types.html#cutensornetstatempoapplication-t>`__.
```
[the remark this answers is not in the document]
Qottmann: > All swap and decomposition operations in MPS-MPO multiplication will follow the same constraints set by SVD configurations (``cutoff``) and target extents setups.

Is this really true? Perhaps I misunderstood your reply [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/859#discussion_r1761592785) but to me this sounded like the cutoff truncation is not performed on the MPO? (just the MPS after the truncation)

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,18 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
 
 The ``lightning.tensor`` device dispatches all operations to be performed on a CUDA-capable GPU of generation SM 7.0 (Volta)
-and greater.
+and greater. This device supports both analytical and finite shots measurements. Currently, the supported differentiation methods are parameter-shift and finite-diff. Note that MPS backend of lightning.tensor supports multi-wires gate via Matrix Product Operators (MPO). Disable the new arithmetic operator dunders is recommended since only 1-wire Hermitian is supported by lightning.tensor. It is recommended to use the shot measurements for the lightning.tensor device. The calculation of analytical probabilities and variances may result in excessive memory usage for large-scale quantum simulations or become impractical due to prohibitive computational costs. It's also recommended to disable new_opmath for the lightning.tensor device since lightning.tensor only supports 1-wire Hermitian observables.
```
Qottmann, line 21:
```suggestion
and greater. This device supports both exact and finite shots measurements. Currently, the supported differentiation methods are parameter-shift and finite-diff. Note that the MPS backend of lightning.tensor supports multi-wire gates via Matrix Product Operators (MPO).

Note that measurements of ``qml.probs()`` or ``qml.state()`` return dense vectors of dimension :math:`2^{n_\text{qubits}}`, so they should only be used for small systems. When using finite shots, ``qml.probs()`` can be used to get individual samples.

.. note:: ``qml.Hermitian`` is currently only supported for single wires. You can use ``qml.pauli_decompose`` on smaller matrices to obtain a compatible Pauli decomposition in the meantime.
```

To me "analytic" is orthogonal to numerical, do you mean "exact"? Alternatively, you can say "infinite shots" here, specifically.

I think it would be unwise to recommend people to disable new opmath - but perhaps I am misunderstanding the current limitations?

The comments about analytic probabilities is confusing, it sounds like finite shots is better / faster in general. I tried to rephrase it a bit.

Qottmann: > It is recommended to use the shot measurements for the lightning.tensor device.

Do you really mean generally? Is this what you are observing? This sounds very strange 🤔

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,18 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
+Some tips on the usage of the ``lightning.tensor`` device:
+    - ``lightning.tensor`` performs better for the larger bond dimensions in MPS calculations, see the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
+    - The ``lightning.tensor`` device is recommended for large-scale quantum simulations. For small-scale simulations, the cost of SVD decomposition and MPS-MPO multiplication might outweigh than the benifits of MPS.
+    - It is recommended to use shot-based ``probs()`` measurements for the ``lightning.tensor`` device when the number of target/projected wires is large. The analytical calculation of ``prob()`` target at ``30+`` wires can lead to excessive memory usage. The analytical calculation of ``prob()`` that targets at small number of subset wires but projects a larger number of wires could become impractical due to high computational costs for large-scale quantum simulations. If the number of projected wires is ``n``, ``2**n`` times of ``cutensornet`` API calls should be made to get the analytical ``probs()``, which could be prohibitive if ``n`` is large.
```
Qottmann, line 26: see recommendation above. if you want to repeat it, try to shorten it a bit

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,18 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
+    - It is recommended to use shot-based ``probs()`` measurements for the ``lightning.tensor`` device when the number of target/projected wires is large. The analytical calculation of ``prob()`` target at ``30+`` wires can lead to excessive memory usage. The analytical calculation of ``prob()`` that targets at small number of subset wires but projects a larger number of wires could become impractical due to high computational costs for large-scale quantum simulations. If the number of projected wires is ``n``, ``2**n`` times of ``cutensornet`` API calls should be made to get the analytical ``probs()``, which could be prohibitive if ``n`` is large. 
+    - The analytical calculation of ``var()`` measurements could be impractical for Hamiltonian observables with many terms. The computational cost of current ``var`` implementation scales with ``n**2``, where ``n`` is the number of terms in the Hamiltonian observable.
+    - It is advisable to disable ``new_opmath`` for the ``lightning.tensor`` device, as the device only supports 1-wire Hermitian observables.
```
Qottmann, line 28: are you referring to supporting only single wire `qml.Hermitian` or more generally only supporting Hermitian operators with one wire, i.e. `qml.X(0) @ qml.Y(1)` is not supported?

Qottmann (changes requested):
Thanks for the update @multiphaseCFD 👍 

Still some doubts and confusion around the following points:

- recommending to disable new opmath. This should be a very last resort as it will likely cause all kinds of unforseeable problems. If there are things not supported right now, we should just note that and aim at fixing that asap.
- recommending to use finite shots generally. For returning things like qml.probs and getting finite samples - sure, but for expectation values? Do we really get faster performance when using finite shots? Would be _very_ strange if that was the case 🤔

doc/lightning_tensor/device.rst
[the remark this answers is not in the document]
multiphaseCFD: Hey @Qottmann , I haven't had a chance to check how the `cutoff` works under the hood.  The statement above is based on the cuquantum docs below.

>When multiple truncation cutoffs (CUTENSORNET_TENSOR_SVD_CONFIG_ABS_CUTOFF, CUTENSORNET_TENSOR_SVD_CONFIG_REL_CUTOFF, CUTENSORNET_TENSOR_SVD_CONFIG_DISCARDED_WEIGHT_CUTOFF) or maximal extent in input [cutensornetTensorDescriptor_t](https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/api/types.html#types_8h_1a35bd3aace8877207eae09420ac9cdc9e) are specified. The runtime reduced extent will be determined as the lowest among all.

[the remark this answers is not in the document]
Qottmann: > I haven't had a chance to check how the cutoff works under the hood. 

That's fine, I'd just remove that sentence then:

> All swap and decomposition operations in MPS-MPO multiplication will follow the same constraints set by SVD configurations (``cutoff``) and target extents setups.

doc/lightning_tensor/device.rst
multiphaseCFD: Nice one!

doc/lightning_tensor/device.rst
multiphaseCFD: You suggestions above are accepted and don't have to repeat it here.

doc/lightning_tensor/device.rst
Qottmann: Oh, that is indeed quite severe 😬 
And with opmath disabled, this works? Where is this coming from?

AmintorDusko (approved):
Looks good!

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,15 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
-    device with ``MPS`` running a GPU may perform slower compared to a ``default.tensor`` device with ``MPS`` running on a CPU. For more details
-    on how bond dimension affects the simulation performance, please refer to the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
+The ``lightning.tensor`` device is designed for expectation value calculations. Measurements of ``qml.probs()`` or ``qml.state()`` return dense vectors of dimension :math:`2^{n_\text{qubits}}`, so they should only be used for small systems. When using finite shots, ``qml.probs()`` can be used to get individual samples.
+
+.. note:: ``qml.Hermitian`` is currently only supported for single wires. You can use ``qml.pauli_decompose`` on smaller matrices to obtain a compatible Pauli decomposition in the meantime. As a result, it is advisable to disable ``new_opmath`` for the ``lightning.tensor`` device to prevent the implicit conversion of Pauli words to multi-wire Hermitian observables. This limitation will be addressed once multi-wires Hermitian observables can be supported with ``cutensornet``.
```
Qottmann, line 25: I still dont understand this limitation. Are we restricted specifically for ``qml.Hermitian`` for more than one wire, or is it that generally we dont support observables with operators on more than one wire (i.e. also not supporting things like `qml.X(0) @ qml.X(1)`? From your previous comment here, it sounded like the latter is the case, but this here again reads like it is only a restriction for ``qml.Hermitian`` - which one is it?

And more generally what is the root of this restriction? How come this works with legacy opmath and not new opmath? Can we fix this before release? I would deem this high priority, as such a suggestion and limitation is quite severe.

Qottmann (changes requested):

doc/lightning_tensor/device.rst
```diff
@@ -14,17 +14,15 @@ By default, the device represents the quantum state approximated as a Matrix Pro
...
 
-.. note:: 
-    Given the inherent parallelism of GPUs, simulations with intensive parallel computation, such as those with larger maximum
-    bond dimensions, stand to gain the most from the computational power offered by GPU and those simulations can benifit from the 
-    ``lightning.tensor`` device.  It's worth noting that if the bond dimension used in the simulation is small, the ``lightning.tensor`` 
-    device with ``MPS`` running a GPU may perform slower compared to a ``default.tensor`` device with ``MPS`` running on a CPU. For more details
-    on how bond dimension affects the simulation performance, please refer to the ``Approximate Tensor Network Methods`` section in the `cuQuantum SDK <https://developer.nvidia.com/cuquantum-sdk>`__.
+The ``lightning.tensor`` device is designed for expectation value calculations. Measurements of ``qml.probs()`` or ``qml.state()`` return dense vectors of dimension :math:`2^{n_\text{qubits}}`, so they should only be used for small systems. When using finite shots, ``qml.probs()`` can be used to get individual samples.
```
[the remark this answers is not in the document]
Qottmann: > Shots-based calculation does not require an (2**n_wires*8 bytes) additional GPU memoery

Can you please help me understand this a better - why does shot-based calculation use less memory than the exact case?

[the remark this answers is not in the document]
multiphaseCFD: To make it more clear, less memory requirement for shot measurements here means that we don't have to allocate GPU memory (`2**num_wires`) in our C++ impelementations to store the full/subset state, which is required for the prob calculations. Shot-measurements return the generated samples to CPU memory directly and the required memory for the generate samples is (`num_shots*num_qubits*8`) bytes. Hope it makes it easier to understand. We can schedule a call if needed.

doc/lightning_tensor/device.rst
multiphaseCFD: This ticket is added to Q4!

Qottmann (approved):

PR #920, opened 2024-09-20
Patch tests after deprecation

AmintorDusko:
**Context:**
Patching stable lightning tests for the latest pennylane.

[stable/latest](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/10965680477)

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/tests_lqcpu_python.yml
+++ b/.github/workflows/tests_lqcpu_python.yml
@@ -69,6 +69,13 @@ jobs:
         with:
           fetch-tags: true
 
+      - name: Patching tests following deprecation (remove after v0.39)
+        if: inputs.lightning-version == 'stable' && inputs.lightning-version != 'latest'
+        run: |
+          cd tests
+          grep -rl "default.qubit.legacy" . | xargs sed -i "s|default.qubit.legacy|default.qubit|g";
+          cd ..
+
       - name: Switch to release build of Lightning
         if: inputs.lightning-version == 'release'
         run: |
@@ -82,13 +89,6 @@ jobs:
           git checkout latest_release
           git log -1 --format='%H'
           git status
-
-      - name: Patching tests following deprecation (remove after v0.39)
-        if: inputs.lightning-version == 'stable' && inputs.lightning-version != 'latest'
-        run: |
-          cd tests
-          grep -rl "default.qubit.legacy" . | xargs sed -i "s|default.qubit.legacy|default.qubit|g";
-          cd ..
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/tests_lqcpu_python.yml
+++ b/.github/workflows/tests_lqcpu_python.yml
@@ -198,7 +198,7 @@ jobs:
       - name: Patching tests following deprecation (FIXME: remove after v0.39)
         if: inputs.lightning-version == 'stable' && inputs.pennylane-version != 'stable'
         run: |
-          cd tests
+          pushd tests
           grep -rl "default.qubit.legacy" . | xargs sed -i "s|default.qubit.legacy|default.qubit|g";
           cd ..
 
--- a/.github/workflows/tests_lqcpu_python.yml
+++ b/.github/workflows/tests_lqcpu_python.yml
@@ -196,7 +196,7 @@ jobs:
           git status
 
       - name: Patching tests following deprecation FIXME(remove after v0.39)
-        if: inputs.lightning-version == 'stable' && inputs.pennylane-version == 'stable'
+        if: inputs.lightning-version == 'stable' && inputs.pennylane-version == 'latest'
         run: |
           pushd tests
           grep -rl "default.qubit.legacy" . | xargs sed -i "s|default.qubit.legacy|default.qubit|g";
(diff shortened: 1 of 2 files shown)


.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
           git log -1 --format='%H'
           git status
 
+      - name: Patching tests following deprecation (remove after v0.39)
```
maliasadi, line 198:
```suggestion
      - name: Patching tests following deprecation (FIXME: remove after v0.39)
```

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
...
+        if: inputs.lightning-version == 'stable' && inputs.pennylane-version != 'stable'
+        run: |
+          cd tests
```
maliasadi, line 201:
```suggestion
          pushd tests
```

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
...
+          grep -rl "default.qubit.legacy" . | xargs sed -i "s|default.qubit.legacy|default.qubit|g";
+          cd ..
```
maliasadi, line 203:
```suggestion
          popd
```

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
           git log -1 --format='%H'
           git status
 
+      - name: Patching tests following deprecation (remove after v0.39)
+        if: inputs.lightning-version == 'stable' && inputs.pennylane-version != 'stable'
```
maliasadi, line 199: Shouldn't pennylane version be 'latest' as we're trying to fix the issue with stable/latest?

maliasadi (commented):
Thanks @AmintorDusko! 🦅

multiphaseCFD (approved):
LGTM! Thanks @AmintorDusko .

.github/workflows/tests_lqcpu_python.yml
AmintorDusko: I didn't have much time to think about that.
This non-pervasive patch only does something if something needs to be done.
I can change it if this is important for you, though.
Just let me know. 🙂

maliasadi: I only wonder about the correctness of this condition. We need to ensure that it won't break other checks in the plugin-matrix. If you're confident about it and have already fully checked the plugin-matrix, I won't have any concern with this as we will remove it in the next release.

maliasadi: Do you know why we still get lgpu file-name errors in stable/stable? 
- https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/10998801618
- https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/10998790742

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
           git log -1 --format='%H'
           git status
 
+      - name: Patching tests following deprecation (FIXME: remove after v0.39)
+        if: inputs.lightning-version == 'stable' && inputs.pennylane-version != 'stable'
```
AmintorDusko, line 199:
```suggestion
        if: inputs.lightning-version == 'stable' && inputs.pennylane-version == 'stable'
```

.github/workflows/tests_lqcpu_python.yml
AmintorDusko: I was avoiding triggering the CIs without need. But I have no problem changing.

AmintorDusko: Regarding the lgpu file names I have no idea.

AmintorDusko: @maliasadi, I checked the lgpu name situation. I'm not touching these names here, but apparently, GH thinks these files still have the wrong name in this branch. I'm positive (but not 100% sure) that if we merge it to master no change will propagate. I never saw these before.

maliasadi: Thanks for checking @AmintorDusko! How did you test the support 3 days ago though? :thinking: 
https://github.com/PennyLaneAI/pennylane-lightning/actions/workflows/compat-check-stable-latest.yml
(I'd like us to double-check if there are more issues that should be fixed)

maliasadi: [08f9487](https://github.com/PennyLaneAI/pennylane-lightning/pull/920/commits/08f94879b39c4a5a5b4752f37f4e35375df520b9) :partying_face:

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -195,6 +195,13 @@ jobs:
           git log -1 --format='%H'
           git status
 
+      - name: Patching tests following deprecation FIXME(remove after v0.39)
+        if: inputs.lightning-version == 'stable' && inputs.pennylane-version == 'stable'
```
AmintorDusko, line 199:
```suggestion
        if: inputs.lightning-version == 'stable' && inputs.pennylane-version == 'latest'
```

maliasadi (approved):
Nice work! :raised_hands:

PR #927, opened 2024-09-30
Allow `lightning.qubit/kokkos::generate_samples` to take in seeds to make the generated samples deterministic

paul0403:
### Before submitting

**Context:**
[A while ago](https://github.com/PennyLaneAI/catalyst/pull/936) a new `seed` option to `qjit` was added. The seed was used to make measurement results deterministic, but samples were still probabilistic. This is because within a `qjit` context, [measurements were controlled from the catalyst repo](https://github.com/PennyLaneAI/catalyst/blob/a580bada575793b780d5366aa77dff6157cd4f93/runtime/lib/backend/common/Utils.hpp#L274) , but samples were not.

To resolve stochastically failing tests (i.e. flaky tests) in catalyst, we add seeding for samples in lightning.

When `qjit(seed=...)` receives a (unsigned 32 bit int) seed value from the user, the seed gets propagated through mlir and [generates a `std::mt19937` rng instance in the catalyst execution context](https://github.com/PennyLaneAI/catalyst/blob/934726fe750043886415953dbd89a4c4ddeb9a80/runtime/lib/capi/ExecutionContext.hpp#L268). This rng instance eventually becomes a field of the `Catalyst::Runtime::Simulator::LightningSimulator` (and kokkos) class [catalyst/runtime/lib/backend/lightning/lightning_dynamic/LightningSimulator.hpp](https://github.com/PennyLaneAI/catalyst/blob/a580bada575793b780d5366aa77dff6157cd4f93/runtime/lib/backend/lightning/lightning_dynamic/LightningSimulator.hpp#L54). 

To seed samples, catalyst uses this device rng instance on the state vector's `generate_samples` methods: https://github.com/PennyLaneAI/catalyst/pull/1164.

In lightning, the `generate_samples` method now takes in a seeding number. The catalyst devices pass in a seed into the lightning `generate_samples`; this seed is created deterministically from the aforementioned already seeded catalyst context rng instance. This makes the generated samples deterministc.

**Benefits:** 
Fewer (hopefully no) stochatically failing frontend tests in catalyst.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-72878]

### DIFF
--- a/pennylane_lightning/core/_version.py
+++ b/pennylane_lightning/core/_version.py
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.39.0-dev37"
+__version__ = "0.39.0-dev38"


pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -576,27 +576,28 @@ class Measurements final
...
      * @param wires Sample are generated for the specified wires.
      * @param num_samples The number of samples to generate.
+     * @param catalyst_rng The seeded random number generator from catalyst.
      * @return 1-D vector of samples in binary, each sample is
      * separated by a stride equal to the number of qubits.
      */
     std::vector<std::size_t>
     generate_samples(const std::vector<std::size_t> &wires,
-                     const std::size_t num_samples) {
+                     const std::size_t num_samples, std::mt19937 *catalyst_rng = nullptr) {
```
paul0403, line 597: don't use raw pointers

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -76,10 +76,15 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+     * If catalyst already provides a random generator, use the provided one
      *
-     * @param seed Seed
+     * @param catalyst_rng Seed
      */
-    void setRandomSeed() {
+    void setRandomSeed(std::mt19937 *catalyst_rng = nullptr) {
+        if (catalyst_rng){
+            this->rng = *catalyst_rng;
+            return;
```
multiphaseCFD, line 86: Do we need `return` here?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -76,10 +76,15 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
      * @brief Randomly set the seed of the internal random generator
+     * If catalyst already provides a random generator, use the provided one
      *
-     * @param seed Seed
+     * @param catalyst_rng Seed
      */
-    void setRandomSeed() {
+    void setRandomSeed(std::mt19937 *catalyst_rng = nullptr) {
+        if (catalyst_rng){
+            this->rng = *catalyst_rng;
```
multiphaseCFD, line 85: Is `this->` necessary?

multiphaseCFD (commented):
Thanks @paul0403 ! Just a few Q.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: You are right, I think we don't need `this->`

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: I think it is, we don't want to change the `mt19937` object once it's set (the `setSeed` below will change it)

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -573,30 +573,35 @@ class Measurements final
...
      *
      * @param num_samples The number of samples to generate.
+     * @param catalyst_rng The seeded random number generator from catalyst.
      * @return 1-D vector of samples in binary, each sample is
      * separated by a stride equal to the number of qubits.
      */
-    std::vector<std::size_t> generate_samples(const std::size_t num_samples) {
+    std::vector<std::size_t>
+    generate_samples(const std::size_t num_samples,
+                     const std::mt19937 *&catalyst_rng = nullptr) {
```
multiphaseCFD, line 582: Idem.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -76,10 +76,15 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 
     /**
      * @brief Randomly set the seed of the internal random generator
+     * If catalyst already provides a random generator, use the provided one
      *
-     * @param seed Seed
+     * @param catalyst_rng Seed
      */
-    void setRandomSeed() {
+    void setRandomSeed(const std::mt19937 *&catalyst_rng = nullptr) {
```
multiphaseCFD, line 83: Idem.

multiphaseCFD (commented):
Thanks @paul0403 .

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -573,30 +573,35 @@ class Measurements final
...
      * @param num_samples The number of samples to generate.
+     * @param catalyst_rng The seeded random number generator from catalyst.
      * @return 1-D vector of samples in binary, each sample is
      * separated by a stride equal to the number of qubits.
      */
     std::vector<std::size_t>
     generate_samples(const std::vector<std::size_t> &wires,
-                     const std::size_t num_samples) {
+                     const std::size_t num_samples,
+                     const std::mt19937 *&catalyst_rng = nullptr) {
```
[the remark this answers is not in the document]
paul0403: Sorry, this was WIP (I needed to push to lightning repo to call the make script from catalyst).

I'll convert to draft and reopen for review once I think everything's ready!

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: No longer relevant now that the two functionality is split into two functions

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -77,13 +77,19 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
      */
     void setRandomSeed() {
         std::random_device rd;
         setSeed(rd());
     }
 
+    /**
+     * @brief Set the internal random generator to an already existing instance
+     *
+     * @param catalyst_rng Seed
```
paul0403, line 89: update param docs

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: Actually I would like to keep `this->` just to be explicit, if that's fine : ) 

(Also the variable name `catalyst_rng` has changed to `rng`)

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -77,13 +77,19 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
         std::random_device rd;
         setSeed(rd());
     }
 
+    /**
+     * @brief Set the internal random generator to an already existing instance
+     *
+     * @param rng An already existing instance of a random number generator
+     */
+    void setRNG(std::mt19937 rng) { this->rng = rng; }
```
dime10, line 91: This makes a copy of the object

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -573,30 +574,38 @@ class Measurements final
...
-    generate_samples(const std::vector<std::size_t> &wires,
-                     const std::size_t num_samples) {
+    std::vector<std::size_t> generate_samples(
+        const std::vector<std::size_t> &wires, const std::size_t num_samples,
+        const std::optional<std::mt19937> &catalyst_rng = std::nullopt) {
         const std::size_t n_wires = wires.size();
         std::vector<std::size_t> samples(num_samples * n_wires);
-        this->setRandomSeed();
+        if (catalyst_rng.has_value()) {
+            this->setRNG(catalyst_rng.value());
```
dime10, line 605: Since we decided it was fine to seed the kokkos rng instance with a seed taken from the catalyst rng, would it not be simpler, and more independent (e.g. we definitely shouldn't call this variable "catalyst"_rng), to provide a seed to the sample function?
Something along the lines of:
```c++
std::vector<std::size_t> generate_samples(
        const std::vector<std::size_t> &wires, const std::size_t num_samples,
        const std::size_t seed = 0) {
    ...
    if (seed) {
        this->setSeed(seed);
    } else {
        this->setRandomSeed();
    }
```

This eliminates the need for the `setRNG` function entirely, and will be uniform for both qubit and kokkos.

dime10 (commented):
Nice seeing this will be pretty straightforward 👍

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: This method is no longer used now that we pass the seed number instead of the rng instance

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
paul0403: This function no longer exists

pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
```diff
@@ -20,6 +20,8 @@ using Pennylane::Util::isApproxEqual;
 } // namespace
 /// @endcond
 #include <algorithm>
+#include <optional>
+#include <random>
```
paul0403, line 24: no need for `<random>` header

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -674,10 +677,12 @@ class Measurements final
             });
 
         // Sampling using Random_XorShift64_Pool
-        Kokkos::Random_XorShift64_Pool<> rand_pool(
-            std::chrono::high_resolution_clock::now()
-                .time_since_epoch()
-                .count());
+        Kokkos::Random_XorShift64_Pool<> rand_pool =
+            seed.has_value() ? Kokkos::Random_XorShift64_Pool<>((seed.value()))
```
paul0403, line 681: one unnecessary layer of `()`

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -649,13 +649,16 @@ class Measurements final
...
+     * @param seed Seed to generate the samples from
      *
      * @return std::vector<std::size_t> to the samples.
      * Each sample has a length equal to the number of qubits. Each sample can
      * be accessed using the stride sample_id*num_qubits, where sample_id is a
      * number between 0 and num_samples-1.
      */
-    auto generate_samples(std::size_t num_samples) -> std::vector<std::size_t> {
+    auto generate_samples(std::size_t num_samples,
+                          const std::optional<std::size_t> &seed = std::nullopt)
```
paul0403, line 660: `<optional>` header

pennylane_lightning/core/_version.py
```diff
@@ -16,4 +16,4 @@
    Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.39.0-dev34"
+__version__ = "0.39.0-dev35"
```
paul0403, line 19: changelog

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -342,13 +342,24 @@ void LightningKokkosSimulator::PartialProbs(
...
-                                      std::size_t shots) {
+std::vector<size_t> LightningKokkosSimulator::GenerateSamples(size_t shots) {
+    // generate_samples is a member function of the Measures class.
     Pennylane::LightningKokkos::Measures::Measurements<StateVectorT> m{
         *(this->device_sv)};
+
     // PL-Lightning-Kokkos generates samples using the alias method.
     // Reference: https://en.wikipedia.org/wiki/Inverse_transform_sampling
-    auto li_samples = m.generate_samples(shots);
+    std::cout << "hi! " << ((this->gen) == nullptr) << std::endl;
```
paul0403, line 352: remove debug prints

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -342,13 +342,22 @@ void LightningKokkosSimulator::PartialProbs(
...
     // Reference: https://en.wikipedia.org/wiki/Inverse_transform_sampling
-    auto li_samples = m.generate_samples(shots);
+    if (this->gen) {
+        return m.generate_samples(shots, (*(this->gen))());
+    }
+    return m.generate_samples(shots);
+}
+
+void LightningKokkosSimulator::Sample(DataView<double, 2> &samples,
+                                      std::size_t shots) {
```
multiphaseCFD, line 359: Do we want to set `shots` as `const`?

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
```diff
@@ -174,6 +174,7 @@ class LightningKokkosSimulator final : public Catalyst::Runtime::QuantumDevice {
     auto CacheManagerInfo()
         -> std::tuple<std::size_t, std::size_t, std::size_t,
                       std::vector<std::string>, std::vector<ObsIdType>>;
+    auto GenerateSamples(size_t shots) -> std::vector<size_t>;
```
multiphaseCFD, line 177: Idem.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -342,13 +342,22 @@ void LightningKokkosSimulator::PartialProbs(
...
+std::vector<size_t> LightningKokkosSimulator::GenerateSamples(size_t shots) {
+    // generate_samples is a member function of the Measures class.
     Pennylane::LightningKokkos::Measures::Measurements<StateVectorT> m{
         *(this->device_sv)};
+
     // PL-Lightning-Kokkos generates samples using the alias method.
     // Reference: https://en.wikipedia.org/wiki/Inverse_transform_sampling
-    auto li_samples = m.generate_samples(shots);
+    if (this->gen) {
+        return m.generate_samples(shots, (*(this->gen))());
```
mlxd, line 353: Given the coverage warning, is there a way to ensure this line is hit by the Lightning test-suite, and not just by Catalyst? Or is this just a CodeCov reporting issue?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -649,13 +650,16 @@ class Measurements final
...
      * @param num_samples Number of Samples
+     * @param seed Seed to generate the samples from
      *
      * @return std::vector<std::size_t> to the samples.
      * Each sample has a length equal to the number of qubits. Each sample can
      * be accessed using the stride sample_id*num_qubits, where sample_id is a
      * number between 0 and num_samples-1.
      */
-    auto generate_samples(std::size_t num_samples) -> std::vector<std::size_t> {
+    auto generate_samples(std::size_t num_samples,
```
multiphaseCFD, line 660: also `const` here?

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -674,10 +678,12 @@ class Measurements final
             });
 
         // Sampling using Random_XorShift64_Pool
-        Kokkos::Random_XorShift64_Pool<> rand_pool(
-            std::chrono::high_resolution_clock::now()
-                .time_since_epoch()
-                .count());
+        Kokkos::Random_XorShift64_Pool<> rand_pool =
```
multiphaseCFD, line 681: can we use `auto` to replace `Kokkos::Random_XorShift64_Pool<>` here?

multiphaseCFD (commented):
Thanks @paul0403 . Nice job. No major blocker from my side.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
```diff
@@ -649,13 +650,16 @@ class Measurements final
...
+     * @param seed Seed to generate the samples from
      *
      * @return std::vector<std::size_t> to the samples.
      * Each sample has a length equal to the number of qubits. Each sample can
      * be accessed using the stride sample_id*num_qubits, where sample_id is a
      * number between 0 and num_samples-1.
      */
-    auto generate_samples(std::size_t num_samples) -> std::vector<std::size_t> {
+    auto generate_samples(std::size_t num_samples,
+                          const std::optional<std::size_t> &seed = std::nullopt)
```
mlxd, line 661: Since the contained type is a POD, do we need the `optional` to be a constant reference, or will a value be sufficient?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -573,30 +574,39 @@ class Measurements final
...
      *
      * @param num_samples The number of samples to generate.
+     * @param seed Seed to generate the samples from
      * @return 1-D vector of samples in binary, each sample is
      * separated by a stride equal to the number of qubits.
      */
-    std::vector<std::size_t> generate_samples(const std::size_t num_samples) {
+    std::vector<std::size_t>
+    generate_samples(const std::size_t num_samples,
+                     const std::optional<std::size_t> &seed = std::nullopt) {
```
mlxd, line 583: Same question here as earlier?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -573,30 +574,39 @@ class Measurements final
...
      * @param num_samples The number of samples to generate.
+     * @param seed Seed to generate the samples from
      * @return 1-D vector of samples in binary, each sample is
      * separated by a stride equal to the number of qubits.
      */
     std::vector<std::size_t>
     generate_samples(const std::vector<std::size_t> &wires,
-                     const std::size_t num_samples) {
+                     const std::size_t num_samples,
+                     const std::optional<std::size_t> &seed = std::nullopt) {
```
mlxd, line 602: And again

mlxd (commented):
Nice work @paul0403 
Just a few quick questions:

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
paul0403: The issue with a default argument value is, what would the value be? It should not be zero since zero is a valid seed value (seed is unsigned 32bit int), so the default argument value would be -1. This makes the argument to be a type "bigger" than `uint32_t`, and the argument type would mismatch its purpose a bit.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
paul0403: Oops, mistake from my end. Forgot to add catalyst tests. I will add it.

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
paul0403: Good suggestion, since now the type is specified in the initializing values!

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
paul0403: done

multiphaseCFD (approved):
LGTM ! Thanks @paul0403 !

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
```diff
@@ -174,6 +174,7 @@ class LightningKokkosSimulator final : public Catalyst::Runtime::QuantumDevice {
     auto CacheManagerInfo()
         -> std::tuple<std::size_t, std::size_t, std::size_t,
                       std::vector<std::string>, std::vector<ObsIdType>>;
+    auto GenerateSamples(const size_t shots) -> std::vector<size_t>;
```
dime10, line 177: This seems like it should be a private method if it's purpose is being a helper function

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -342,13 +342,22 @@ void LightningKokkosSimulator::PartialProbs(
...
+    }
+    return m.generate_samples(shots);
```
dime10, line 355: Question to the C++ experts, would an additional function like this (as opposed to calling `m.generate_samples` directly in the caller) ever lead to additional copies (from the local object in this scope to the object in the caller scope), or do optimizations pretty much guarantee this doesn't happen?

paul0403: (Note that [lightning.qubit in catalyst has this helper function](https://github.com/PennyLaneAI/catalyst/blob/934726fe750043886415953dbd89a4c4ddeb9a80/runtime/lib/backend/lightning/lightning_dynamic/LightningSimulator.cpp#L302); not sure if the performance folks inlined it back when porting kokkos)

dime10: Ah, that function even explicitly mentions return value optimization!

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
paul0403: done

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.hpp
paul0403: good point

paul0403: done

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
paul0403: > This makes the argument to be a type "bigger" than `uint32_t`

Let's call this "bigger type" `int`, for illustration's sake.

This leads to the following problem: the seed argument would need to have type `int` just for the default value of -1. Some other piece of code interfacing with this function (say, e.g., some other rng is generating the seed to pass in here) wouldn't know that the actual underlying expected type is `size_t`, and could potentially just pass in -1 (or -2, ...). This is bad because:
- a user can pass in -1, expecting a seeded run, and getting an unseeded run (since that was the default value we would check against)
- [std::mt19937 has undefined behavior if it's not initialized with unsigned](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine#:~:text=The%20effect%20is%20undefined%20if%20this%20is%20not%20one%20of%20unsigned%20short%2C%20unsigned%20int%2C%20unsigned%20long%2C%20or%20unsigned%20long%20long.)

These problems would not be present if we explicitly specify it as `std::optional<size_t>`, avoiding the default value / signed-ness problem altogether.

(Also, having this argument as `size_t` will make sure the underlying rng is initialized with an unsigned value.)

paul0403: @mlxd I suddenly realized you were talking about `const std::optional<size_t> &seed` vs `std::optional<size_t> seed, not vs `int seed = -1` :sweat_smile: 

I don't think there's that much of a difference whether it's a const reference or not. I would suggest keep it as is just so we don't have to wait for CI again : ))))

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
mlxd: Usually copy elision should come for free here, so should be all good (https://en.cppreference.com/w/cpp/language/copy_elision the section on prvalue semantics here should match the above case).

pennylane_lightning/core/src/simulators/lightning_kokkos/measurements/MeasurementsKokkos.hpp
mlxd: Sure, fine by me ;)

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+        MemRefT<double, 2> buffer{
+            samples.data(), samples.data(), 0, {shots, 1}, {1, 1}};
+        DataView<double, 2> view(buffer.data_aligned, buffer.offset,
+                                 buffer.sizes, buffer.strides);
+
+        std::vector<double> samples1(shots * 4);
+        MemRefT<double, 2> buffer1{
+            samples1.data(), samples1.data(), 0, {shots, 1}, {1, 1}};
+        DataView<double, 2> view1(buffer1.data_aligned, buffer1.offset,
+                                  buffer1.sizes, buffer1.strides);
```
mlxd, line 1798: If we allocate the samples vector(s) outside the loop, the memory addresses will remain the same, and so the memref and dataview objects should not be invalidated on each iteration. With the above moved out, we can move these out too and reuse them.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+
+        std::mt19937 gen(37);
```
mlxd, line 1800: Since we can generate 2 RNGs and reuse them here, we can also put these outside the loop.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+        sim->SetDevicePRNG(&gen);
+        std::vector<intptr_t> Qs;
+        Qs.reserve(1);
+        Qs.push_back(sim->AllocateQubit());
+        sim->NamedOperation("Hadamard", {}, {Qs[0]}, false);
+        sim->NamedOperation("RX", {0.5}, {Qs[0]}, false);
+        sim->Sample(view, shots);
```
mlxd, line 1807: As this alg defines how to execute the workload, we can put this into a lambda, and accept the appropriate arguments for both simulation cases.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+        std::vector<intptr_t> Qs1;
+        Qs1.reserve(1);
+        Qs1.push_back(sim1->AllocateQubit());
+        sim1->NamedOperation("Hadamard", {}, {Qs1[0]}, false);
+        sim1->NamedOperation("RX", {0.5}, {Qs1[0]}, false);
+        sim1->Sample(view1, shots);
+
+        for (std::size_t i = 0; i < shots * 4; i++) {
+            CHECK((samples[i] == samples1[i]));
+        }
```
mlxd, line 1820: Lastly, we can compare the recorded samples in the outer-stored vectors with their latest updated values.

mlxd (changes requested):
Thanks @paul0403 
Last review from me --- I'll be happy once the following changes go in.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
         CHECK(*m == *m1);
     }
 }
+
+TEST_CASE("Sample with a seeded device", "[Measures]") {
```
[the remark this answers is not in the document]
paul0403: Thanks! Definitely a lot of common "furniture" can be factored out of the loop. I will make these changes.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+    for (std::size_t _ = 0; _ < 5; _++) {
```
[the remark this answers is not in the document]
paul0403: done

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -404,20 +408,14 @@ void LightningKokkosSimulator::PartialSample(
 
 void LightningKokkosSimulator::Counts(DataView<double, 1> &eigvals,
                                       DataView<int64_t, 1> &counts,
-                                      std::size_t shots) {
+                                      const std::size_t shots) {
```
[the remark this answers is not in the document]
mlxd: Yea, none of these need to be const. If the outer API requires it, then fine, otherwise it doesn't seem necessary. Fine to leave it be, but in future I'd say no need to make pass-by-value entries const in cases like this.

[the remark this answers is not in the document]
joeycarter: The C++ Core Guidelines [F.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const) prefer passing cheaply copied types by value, without `const`. Although IMHO, the more we can mark as `const` the better.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1777,3 +1777,46 @@ TEST_CASE("Measurement with a seeded device", "[Measures]") {
...
+        std::unique_ptr<LKSimulator> sim = std::make_unique<LKSimulator>();
+        std::unique_ptr<LKSimulator> sim1 = std::make_unique<LKSimulator>();
```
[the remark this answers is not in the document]
paul0403: Why `std::array<std::unique_ptr<LKSimulator>, 2> sims;` for the overall container instead of `std::vector`?

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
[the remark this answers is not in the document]
paul0403: > The C++ Core Guidelines [F.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const)
> void f3(int x);            // OK: Unbeatable

I think as long as clang-tidy is happy everything's fine here, but I like how they straight up say "unbeatable" :laughing: 
I will go with the true and tried unbeatable recommended by the guideline then.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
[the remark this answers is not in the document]
paul0403: All done. I also did this for another seeding test above (for qjitted kokkos measurement) that can also be refactored.

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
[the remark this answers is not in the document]
dime10: Note that the guide compares const _ref_ to by value, which is an overhead for ints. `const` vs `const int` makes no difference in performance, although personally I don't particular like littering the source code with const annotations unless its with intention.

mlxd (approved):

Issue #935, opened 2024-10-07
Lightning utilities failing to hit correct inline function with +OMP -BLAS under clang


mlxd:
#### Before posting an issue

Search existing GitHub issues to make sure the issue does not already exist:
https://github.com/xanaduai/pennylane-lightning/issues

If posting a PennyLane-Lightning issue, delete everything above the dashed line, and fill
in the template. If the issue is a bug, start the title of the issue with [BUG].

If making a feature request, delete the following template and describe, in detail,
the feature and why it is needed.

For general technical details check out our documentation:
https://docs.pennylane.ai/projects/lightning

-------------------------------------------------------------------------------------------------------------

#### Issue description

Description of the issue - include code snippets and screenshots here
if relevant. You may use the following template below

* *Expected behavior:* Compilation enabling OpenMP, disabling BLAS, disabling Python and using clang as:  
`CXX=clang++-18 $(which cmake) -BBuild -DBUILD_TESTS=1 -DENABLE_PYTHON=0 . -G Ninja -DCMAKE_BUILD_TYPE="Release"`

* *Actual behavior:* (What actually happens)
Compilation failed as 
```bash
/tmp/pennylane-lightning/pennylane_lightning/core/src/utils/Util.hpp:46:12: error: no matching constructor for initialization of 'std::complex<float>'
   46 |     return {a * b.real(), a * b.imag()};
```

* *Reproduces how often:* (What percentage of the time does it reproduce?)

* *System information:* (post the output of `import pennylane as qml; qml.about()`)

#### Source code and tracebacks

Please include any additional code snippets and error tracebacks related
to the issue here.

#### Additional information

The above failure can be fixed by restricting the offending call to scalar arithmetic types.

PR #936, opened 2024-10-07
Ensure inline templated func hit with arith types only

mlxd:
### Before submitting

**Context:** Closes https://github.com/PennyLaneAI/pennylane-lightning/issues/935

**Description of the Change:** Adds concept restrictions to the `ConstMult` function definition for scalar-time-complex only instantiations. 

**Benefits:** Removes the given build failure.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -107,6 +107,9 @@
 
 ### Bug fixes
 
+* Add concept restriction to ensure `ConstMult` inline function only hit with arithmetic-values times complex values. Fixes build failures with +Test+OpenMP~BLAS~Python under clang.
+  [(#936)](https://github.com/PennyLaneAI/pennylane-lightning/pull/936)
+
 * Bug fix for `applyMatrix` in `lightning.tensor`. Matrix operator data is not stored in the `cuGateCache` object to support `TensorProd` obs with multiple `Hermitian` obs.
   [(#932)](https://github.com/PennyLaneAI/pennylane-lightning/pull/932)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -43,6 +43,9 @@
 
 ### Improvements
 
+* Add `setStateVector(state, wire)` support to the `lightning.gpu` C++ layer.
+  [(#930)](https://github.com/PennyLaneAI/pennylane-lightning/pull/930)
+
 * Add zero-state initialization to both `StateVectorCudaManaged` and `StateVectorCudaMPI` constructors to remove the `reset_state` in the python layer ctor and refactor `setBasisState(state, wires)` in the C++ layer.
   [(#933)](https://github.com/PennyLaneAI/pennylane-lightning/pull/933)
   
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -40,6 +40,9 @@
 
 ### Improvements
 
+* The `generate_samples` methods of lightning.{qubit/kokkos} can now take in a seed number to make the generated samples deterministic. This can be useful when, among other things, fixing flaky tests in CI.
+  [(#927)](https://github.com/PennyLaneAI/pennylane-lightning/pull/927)
+
 * Always decompose `qml.QFT` in Lightning.
   [(#924)](https://github.com/PennyLaneAI/pennylane-lightning/pull/924)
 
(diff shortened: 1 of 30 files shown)


pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -41,6 +41,7 @@ namespace Pennylane::Util {
  * @return constexpr std::complex<T>
  */
 template <class T, class U = T>
+    requires std::integral<U> || std::floating_point<U>
```
maliasadi, line 44: These restrictions could be applied to the rest of compile-time complex scalars. This will preserve the support for the future use cases. Could you add the required concept to the rest of scalars functions in `Utils`?

.github/CHANGELOG.md
```diff
@@ -107,6 +107,9 @@
 
 ### Bug fixes
 
+* Add concept restriction to ensure `ConstMult` inline function only hit with arithmetic-values times complex values. Fixes build failures with +Test+OpenMP~BLAS~Python under clang.
```
maliasadi, line 110: This doesn't render nicely. Will you update the format of `+Test+OpenMP~BLAS~Python`?

maliasadi (commented):
Thanks @mlxd! Happy to approve after resolving the comments

AmintorDusko (approved):
Concepts are beatiful! Thank you, @mlxd!

pennylane_lightning/core/src/utils/Util.hpp
mlxd: I don't think this line needs to be added anywhere else --- is there a specific set of arith * complex functions that would think is needed? This seems to be the only real * complex in this file, no?

.github/CHANGELOG.md
mlxd: Done.

pennylane_lightning/core/src/utils/Util.hpp
maliasadi, line 1: Could you also add `concepts` to the list of included headers?

pennylane_lightning/core/src/utils/Util.hpp
maliasadi: I meant whether we need to use concepts for the rest of scalar methods in this file (and maybe everywhere) to prevent such template issues in the future. For example, we can update `ConstMult` and similar ones as follows:

```
template <class T, class U = T>
    requires (std::integral<T> || std::floating_point<T>) && 
             (std::integral<U> || std::floating_point<U>)
inline static constexpr auto ConstMult(std::complex<U> a, std::complex<T> b)
    -> std::complex<T> {
// ...
}
```

mlxd: Given that the the problem here was U being mapped to `std:complex<A>` or `A`, causing the function call to fail, I think the above suggestion is a little different. The std::complex type already validates the incoming `U/T` types, and will prevent issues with the selection and compilation. I don't think it is necessary for the others in this PR.

pennylane_lightning/core/src/utils/Util.hpp
mlxd: I'll add `std::integral` and `std::floating_point` as comments to the `type_traits` import. The rest is C++20 language support, so no other headers required.

maliasadi: Both `std::integral` and `std::floating_point` belong to [the concepts header](https://en.cppreference.com/w/cpp/header/concepts).

`type_traits` is a C++11 header.. I don't think it has support for `std::integral` and `std::floating_point`. Let's be explicit about the header files in this file similar to the rest of the code base.

pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -24,7 +24,7 @@
 #include <numbers>
 #include <numeric> // transform_reduce
 #include <set>
-#include <type_traits> // is_same_v
+#include <type_traits> // integral, floating_point, is_same_v
```
maliasadi, line 27: I think `std::integral` and `std::floating_point` are in `concepts` 
```suggestion
#include <type_traits> // is_same_v
```
reference: https://en.cppreference.com/w/cpp/header/type_traits

pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -24,7 +24,7 @@
...
 #include <vector>
```
maliasadi, line 28:
```suggestion
#include <vector>
#include <concepts> // integral, floating_point
```
reference: https://en.cppreference.com/w/cpp/header/concepts

pennylane_lightning/core/src/utils/Util.hpp
mlxd: Ah, I was looking at `https://en.cppreference.com/w/cpp/types/is_integral` which is type_traits. Ok, I'll add the header.

pennylane_lightning/core/src/utils/Util.hpp
mlxd: The concepts std::integral is syntactic sugar around `std::is_integral` from type_traits. But yep, you're right, they are in concepts, so fine by me.

maliasadi (approved):
:partying_face:

PR #943, opened 2024-10-11
Optimize memory peak for `_preprocess_state_vector` in LightningTensor

LuisAlfredoNu:
### Before submitting

**Context:**
After profiling Lightning Tensor with [scalene profiler](https://github.com/plasma-umass/scalene). We found a bottleneck of memory in the function `_preprocess_state_vector` ([code](https://github.com/PennyLaneAI/pennylane-lightning/blob/4945ed08d5475d04add8d69a3cf5978ba31d1b39/pennylane_lightning/lightning_tensor/_tensornet.py#L208-L240)) which allocates 3 arrays with dimension 2 ** wires * wires.

Optimize the cartesian product to reduce the amount of memory necessary to set the `StatePrep` with LTensor

**Benefits:**
Reduce by half the peak of memory for large systems close to 30 qubit
image

Benchmark code
``` python
    import pennylane as qml
    import numpy as np

    wires = 27

    state = np.random.rand(2**(wires-1))
    state = state / np.linalg.norm(state)

    dev = qml.device('lightning.tensor', wires=wires)

    dev_wires = dev.wires.tolist()

    @qml.qnode(dev)
    def circuit(state=state,dev_wires=dev_wires):
        qml.StatePrep(state, wires=dev_wires[1:])
        return qml.expval(qml.Z(0)), qml.state()

    return  circuit(state, dev_wires)
```

**Possible Drawbacks:**
This change reduces readability but with a good improvement. 

**Related GitHub Issues:**
[sc-75692]

### DIFF
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -221,7 +221,8 @@ class LightningTensorNet:
         if len(device_wires) == self._num_wires and Wires(sorted(device_wires)) == device_wires:
             return np.reshape(state, output_shape).ravel(order="C")
 
-        local_dev_wires = np.array(device_wires.tolist())[::-1]
+        local_dev_wires = device_wires.tolist().copy()
+        local_dev_wires = local_dev_wires[::1]
 
         # generate basis states on subset of qubits via broadcasting as substitute of cartesian product.
 
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -240,8 +240,8 @@ class LightningTensorNet:
 
             two_n = 2 ** (i + 1)
             base = base.reshape(-1, two_n * 2)
-            swaper_A = two_n // 2
-            swaper_B = swaper_A + two_n
+            swapper_A = two_n // 2
+            swapper_B = swapper_A + two_n
 
             base[:, swaper_A:swaper_B] = base[:, swaper_A:swaper_B][:, ::-1]
             base = base.reshape(-1)
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -223,27 +223,37 @@ class LightningTensorNet:
 
         local_dev_wires = np.array(device_wires.tolist())[::-1]
 
-        # generate basis states on subset of qubits via broadcasting
+        # generate basis states on subset of qubits via broadcasting as substitute of cartesian product.
+
+        # Allocate a single row as a base to avoid a large array allocation with
+        # the cartesian product algorithm.
+        # Initialize the base with the pattern [0 1 0 1 ...].
         base = np.tile([0, 1], 2 ** (len(local_dev_wires) - 1)).astype(dtype=np.int64)
+        # Allocate the array where it will accumulate the value of the indexes depending on
+        # the value of the basis.
         indexes = np.zeros(2 ** (len(local_dev_wires)), dtype=np.int64)
 
         max_dev_wire = self._num_wires - 1
 
-        # get basis states to alter on full set of qubits
+        # Iterate over all device wires.
         for i, wire in enumerate(local_dev_wires):
 
-            # get indices for which the state is changed to input state vector elements
+            # Accumulate indexes from the basis.
             indexes += base * 2 ** (max_dev_wire - wire)
 
             if i == len(local_dev_wires) - 1:
                 continue
 
-            two_n = 2 ** (i + 1)
+            two_n = 2 ** (i + 1)  # Compute the value of the base.
+
+            # Update the value of the base without reallocating a new array.
+            # Reshape the basis to swap the internal columns.
             base = base.reshape(-1, two_n * 2)
             swapper_A = two_n // 2
             swapper_B = swapper_A + two_n
 
             base[:, swapper_A:swapper_B] = base[:, swapper_A:swapper_B][:, ::-1]
+            # Flatten the base array
             base = base.reshape(-1)
 
         # get full state vector to be factorized into MPS
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -222,7 +222,7 @@ class LightningTensorNet:
             return np.reshape(state, output_shape).ravel(order="C")
 
         local_dev_wires = device_wires.tolist().copy()
-        local_dev_wires = local_dev_wires[::1]
+        local_dev_wires = local_dev_wires[::-1]
 
         # generate basis states on subset of qubits via broadcasting as substitute of cartesian product.
 
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -243,7 +243,7 @@ class LightningTensorNet:
             swapper_A = two_n // 2
             swapper_B = swapper_A + two_n
 
-            base[:, swaper_A:swaper_B] = base[:, swaper_A:swaper_B][:, ::-1]
+            base[:, swapper_A:swapper_B] = base[:, swapper_A:swapper_B][:, ::-1]
             base = base.reshape(-1)
 
         # get full state vector to be factorized into MPS
--- a/pennylane_lightning/lightning_tensor/_tensornet.py
+++ b/pennylane_lightning/lightning_tensor/_tensornet.py
@@ -223,20 +223,35 @@ class LightningTensorNet:
         if len(device_wires) == self._num_wires and Wires(sorted(device_wires)) == device_wires:
             return np.reshape(state, output_shape).ravel(order="C")
 
-        # generate basis states on subset of qubits via the cartesian product
-        basis_states = np.array(list(product([0, 1], repeat=len(device_wires))))
+        local_dev_wires = np.array(device_wires.tolist())[::-1]
 
+        # generate basis states on subset of qubits via broadcasting 
+        base = np.tile([0,1], 2**(len(local_dev_wires)-1)).astype(dtype=np.int64)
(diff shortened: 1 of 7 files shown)


pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -223,20 +221,35 @@ def _preprocess_state_vector(self, state, device_wires):
         if len(device_wires) == self._num_wires and Wires(sorted(device_wires)) == device_wires:
             return np.reshape(state, output_shape).ravel(order="C")
 
-        # generate basis states on subset of qubits via the cartesian product
-        basis_states = np.array(list(product([0, 1], repeat=len(device_wires))))
+        local_dev_wires = np.array(device_wires.tolist())[::-1]
```
mlxd, line 224: Why is this cast to a numpy array?
Is it preferable to do so, rather than use native Python?

mlxd, line 224: Also, if the device wires are not passed in as integers, but strings/labels, do we have an issue with this?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -223,20 +221,35 @@ def _preprocess_state_vector(self, state, device_wires):
...
+            # get indices for which the state is changed to input state vector elements
+            indexes += base * 2 ** (max_dev_wire - wire)
+
+            if i == len(local_dev_wires) - 1:
+                continue
+
+            two_n = 2 ** (i + 1)
+            base = base.reshape(-1, two_n * 2)
+            swaper_A = two_n // 2
+            swaper_B = swaper_A + two_n
```
mlxd, line 244:
```suggestion
            swapper_A = two_n // 2
            swapper_B = swapper_A + two_n
```

Maybe?

mlxd, line 244: I guess it's not fully clear to me why we are doing what we are doing here? Could you walk me through it?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -285,7 +298,7 @@ def _apply_MPO(self, gate_matrix, wires):
             None
         """
         # TODO: Discuss if public interface for max_mpo_bond_dim argument
-        max_mpo_bond_dim = 2 ** len(wires)  # Exact SVD decomposition for MPO
+        max_mpo_bond_dim = self._max_bond_dim
```
mlxd, line 301: :+1:

Good call

mlxd (commented):
Feel free to ignore me if these comments are being a blocker. Otherwise, just looking for some clarifications :)

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: An explanation of the change:

The current algorithm is the following:

1. Generate basis states on a subset of qubits via the cartesian product
``` python
basis_states = np.array(list(product([0, 1], repeat=len(dev_wires))))
```
The line above produces the following pattern (Transposed)
```
[[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1]
 [0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1]
 [0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1]
 [0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1]
 [0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1]]
```
 For a small system, it is not a problem. However, here we are creating a list of `2 ** dev_wires * dev_wires` which will overload the memory if we try to work with 30 qubits. 

2. Create a new array with size `2 ** dev_wires * num_wires` to allocate each basis row to the `state prep` row, producing a second big allocation.  

``` python
unravelled_indices = np.zeros((2 ** len(device_wires), self._num_wires), dtype=int)
unravelled_indices[:, device_wires] = basis_states
unravelled_indices.T
```

3.  Multiply each row by `2**n` where `n` goes from `num_wires` to 1 and then sum each column to get indices for which the state is changed to input state vector elements. This can be done with the following line:

``` python
ravelled_indices = np.ravel_multi_index(unravelled_indices.T, [2] * self._num_wires)
```

4. With the indices from `ravelled_indices` the new state `full_state` is filled. 
``` python
 full_state = np.zeros(2**self._num_wires, dtype=self.dtype)
 for i, value in enumerate(state):
    full_state[ravelled_indices[i]] = value
```

The solution that we propose is the following: 
 
1. Create a single row of  `len(device_wires)` length with the pattern `[0 1 0 1 ....]`.  
2. Multiply by the basis `2**n`  and accumulate the value of the indices in the array `indexes`. 
3. Start a for loop for the next row of `base`. Instead of reallocating the `base` array, it is faster to swap columns because it only changes the viewer for the array.  
4. With the indices from `indexes` the new state `full_state` is filled. 
``` python 
        base = np.tile([0, 1], 2 ** (len(local_dev_wires) - 1)).astype(dtype=np.int64)
        indexes = np.zeros(2 ** (len(local_dev_wires)), dtype=np.int64)

        for i, wire in enumerate(local_dev_wires):
            two_n = 2 ** (i + 1)
            indexes += base * 2 ** (max_dev_wire - wire)
            
            if i == len(local_dev_wires) - 1:
                continue

            base = base.reshape(-1, two_n * 2)
            swaper_A = two_n // 2
            swaper_B = swaper_A + two_n

            base[:, swaper_A:swaper_B] = base[:, swaper_A:swaper_B][:, ::-1]
            base = base.reshape(-1)
```

The swap works as follows:
1. From a flattened array, it reshaped to a number of columns of `2**(n+1)` 
2. Swapped the internal range from `2**(n)` to `2**(n+2)` 
3. Flatten the matrix. 

This is an example of the evolution of the algorithm 
```
First allocation:
[0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1]

Reshape
[[0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]
 [0 1 0 1]]

Swap internal columns
[[0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]
 [0 0 1 1]]

Flatten
[0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0 1 1]

Start again. Reshape
[[0 0 1 1 0 0 1 1]
 [0 0 1 1 0 0 1 1]
 [0 0 1 1 0 0 1 1]
 [0 0 1 1 0 0 1 1]]

Swap internal columns
[[0 0 0 0 1 1 1 1]
 [0 0 0 0 1 1 1 1]
 [0 0 0 0 1 1 1 1]
 [0 0 0 0 1 1 1 1]]

Flatten
[0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1 0 0 0 0 1 1 1 1]
....
```

AmintorDusko (approved):
Thank you, @LuisAlfredoNu, for the very comprehensive explanation and in-office discussion. Happy to approve!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: No problem because the `dev_wires` is mapped before to reach this point

LuisAlfredoNu: I was looking to avoid modifying the list `device_wires` by copying the data. Do you suggest to use `list`?

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: I will add more documentation to the code to make more readable this change.

pennylane_lightning/lightning_tensor/_tensornet.py
mlxd: I think it is often better to stick to native types, unless there is a reason to use numpy here of course. If there's no appreciable runtime difference between the two, I'd use the non-numpy option. If numpy gives a performance benefit (assuming we are setting the dtype to int), then fine to use it with that reason in mind.

pennylane_lightning/lightning_tensor/_tensornet.py
mlxd: Thanks for the explanation, this is really clear.

As a follow on, do we even need the single allocation, single we are seemingly just generating parity/gray patterns iteratively? Couldn't we just build these on the fly for each iteration and map/select the values without building the full array?

LuisAlfredoNu: Yes could be a good approach but I think this will be slower because we will use one or two extra Python `for` loops to compute the value for the `indexes` but I am not sure. I suggest keeping this implementation and if in the future we need to optimize this section then apply this iteration on the fly. What do you think? :thinking:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: I changed to Python `List` :zap:

pennylane_lightning/lightning_tensor/_tensornet.py
mlxd: Sounds good. I think it'll be faster at a certain threshold, as you do less memory ops than the columnar swaps, but that could be anywhere in the 21-27 qubits range. Are you ok to track a backlog story to investigate this approach at a later date?

mlxd (approved):
:100:

PR #952, opened 2024-10-18
(LK-C-1) Add controlled 1-qubit named-gate (e.g. NCPauliX) support to Lightning Kokkos

josephleekl:
### Before submitting

**Context:**

This PR adds support for named controlled 1-qubit gates (e.g. PauliX / RX / PhaseShift / Hadamard / Rot / GlobalPhase). These are defined in `BasicGateFunctors.hpp`, and are applied through `applyNC1Functor` defined in the same file. 

**Benefits:**
Performance benchmarks for gates are shown here: https://www.notion.so/xanaduai/Lightning-Kokkos-Native-Controlled-Operation-Gate-Benchmarks-12ebc6bd17648017a2dcd237748b24fe

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-76773]


pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,52 @@ class StateVectorKokkos final
...
+        } else if (array_contains(controlled_gate_names,
+                                  std::string_view{opName})) {
+            const std::size_t num_qubits = this->getNumQubits();
+            const ControlledGateOperation gateop =
+                reverse_lookup(controlled_gate_names, std::string_view{opName});
+            applyNCNamedOperation<KokkosExecSpace>(
+                gateop, *data_, num_qubits, controlled_wires, controlled_values,
+                wires, inverse, params);
+        } else {
+            PL_ABORT_IF(gate_matrix.empty(),
```
multiphaseCFD, line 489: Do we need this line? Would it always fail at this point?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,52 @@ class StateVectorKokkos final
...
+                        std::string("Operation does not exist for ") + opName +
+                            std::string(" and no matrix provided."));
+            PL_ABORT("Controlled matrix operation not yet supported");
```
multiphaseCFD, line 492: Do we want to add a test for this line?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -491,23 +511,6 @@ class StateVectorKokkos final
         applyMultiQubitOp(matrix_, wires, inverse);
     }
 
-    /**
-     * @brief Apply a given matrix directly to the statevector.
-     *
-     * @param matrix Matrix data (in row-major format).
-     * @param wires Wires to apply gate to.
-     * @param inverse Indicate whether inverse should be taken.
```
multiphaseCFD, line 499: Any reason to remove this method?

pennylane_lightning/core/src/simulators/lightning_kokkos/bindings/LKokkosBindings.hpp
```diff
@@ -48,6 +48,35 @@ using StateVectorBackends =
     Pennylane::Util::TypeList<StateVectorKokkos<float>,
                               StateVectorKokkos<double>, void>;
 
+template <class StateVectorT, class PyClass>
+void registerControlledGate(PyClass &pyclass) {
+    using PrecisionT =
+        typename StateVectorT::PrecisionT; // Statevector's precision
+    using ParamT = PrecisionT;             // Parameter's data precision
```
multiphaseCFD, line 55: Does ```
using ParamT =
        typename StateVectorT::PrecisionT; work?
```

File: pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp [hunk 5]
```diff
@@ -83,15 +182,33 @@ void applyPauliY(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+    [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
+    auto core_function =
         KOKKOS_LAMBDA(Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
                       const std::size_t i0, const std::size_t i1) {
-            const auto v0 = arr(i0);
-            const auto v1 = arr(i1);
-            arr(i0) = Kokkos::complex<PrecisionT>{imag(v1), -real(v1)};
-            arr(i1) = Kokkos::complex<PrecisionT>{-imag(v0), real(v0)};
-        });
+        [[maybe_unused]] const auto i0_ = i0;
```
multiphaseCFD, line 201: why unused here? do we need it?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -83,15 +182,33 @@ void applyPauliY(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+        arr(i1) *= -1.0;
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
```
multiphaseCFD, line 205: Would like to understand if there is any performance regression with this branch?

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -100,13 +217,38 @@ void applyPauliZ(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+    const std::vector<std::size_t> &wires,
+    [[maybe_unused]] const bool inverse = false,
+    [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
+    auto core_function =
         KOKKOS_LAMBDA(Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
                       const std::size_t i0, const std::size_t i1) {
-            [[maybe_unused]] const auto i0_ = i0;
-            arr(i1) *= -1.0;
-        });
+        [[maybe_unused]] const auto i0_ = i0;
```
multiphaseCFD, line 236: Idem

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -100,13 +217,38 @@ void applyPauliZ(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+        const Kokkos::complex<PrecisionT> v0 = arr(i0);
+        const Kokkos::complex<PrecisionT> v1 = arr(i1);
+        arr(i0) = M_SQRT1_2 * v0 +
+                  M_SQRT1_2 * v1; // NOLINT(readability-magic-numbers)
+        arr(i1) = M_SQRT1_2 * v0 +
+                  -M_SQRT1_2 * v1; // NOLINT(readability-magic-numbers)
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
```
multiphaseCFD, line 245: Idem

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
                   : Kokkos::complex<PrecisionT>{0.0, 1.0};
-    applyNC1Functor(
-        ExecutionSpace{}, arr_, num_qubits, wires,
+    auto core_function =
         KOKKOS_LAMBDA(Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
                       const std::size_t i0, const std::size_t i1) {
-            [[maybe_unused]] const auto i0_ = i0;
-            arr(i1) *= shift;
-        });
+        [[maybe_unused]] const auto i0_ = i0;
```
multiphaseCFD, line 277: Idem.

josephleekl:
Missing code coverage due to new gates not tested, they are tested in https://github.com/PennyLaneAI/pennylane-lightning/pull/955 with the introduction of controlled qubitunitary.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
+        arr(i1) *= shift;
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
+            ExecutionSpace{}, arr_, num_qubits, wires, core_function);
+    } else {
+        applyNC1Functor<PrecisionT, decltype(core_function), true>(
```
multiphaseCFD, line 284: Codecov complains

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
                         0, static_cast<PrecisionT>(M_PI / 4)));
-    applyNC1Functor(
-        ExecutionSpace{}, arr_, num_qubits, wires,
+    auto core_function =
         KOKKOS_LAMBDA(Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
                       const std::size_t i0, const std::size_t i1) {
-            [[maybe_unused]] const auto i0_ = i0;
-            arr(i1) *= shift;
-        });
+        [[maybe_unused]] const auto i0_ = i0;
```
multiphaseCFD, line 314: Idem

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
+        arr(i1) *= shift;
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
+            ExecutionSpace{}, arr_, num_qubits, wires, core_function);
+    } else {
+        applyNC1Functor<PrecisionT, decltype(core_function), true>(
```
multiphaseCFD, line 321: Codecov complains

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
                   : exp(Kokkos::complex<PrecisionT>(0, angle));
-    applyNC1Functor(
-        ExecutionSpace{}, arr_, num_qubits, wires,
+    auto core_function =
         KOKKOS_LAMBDA(Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
                       const std::size_t i0, const std::size_t i1) {
-            [[maybe_unused]] const auto i0_ = i0;
-            arr(i1) *= shift;
-        });
+        [[maybe_unused]] const auto i0_ = i0;
```
multiphaseCFD, line 351: Idem

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,144 +257,249 @@ void applyHadamard(
...
+        arr(i1) *= shift;
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
+            ExecutionSpace{}, arr_, num_qubits, wires, core_function);
+    } else {
+        applyNC1Functor<PrecisionT, decltype(core_function), true>(
```
multiphaseCFD, line 358: Idem

multiphaseCFD (commented):
Thanks @josephleekl ! Nice work. A few Q for the 1st round.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/MatrixGateFunctors.hpp
```diff
@@ -399,17 +386,13 @@ template <class PrecisionT> struct apply5QubitOpFunctor {
 
     KokkosComplexVector arr;
     KokkosComplexVector matrix;
-    KokkosIntVector wires;
     KokkosIntVector parity;
     KokkosIntVector rev_wire_shifts;
-    const std::size_t n_wires = 5;
```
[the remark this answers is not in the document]
josephleekl: These variables are never used in the functor body, especially  `wires` is not used so no need to be copied to device

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
josephleekl: There is a similar method below, just with `const std::vector<ComplexT> &matrix` instead of `std::vector<ComplexT> &matrix`, so the one without const here shouldn't be necessary?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
josephleekl: test added

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
```diff
@@ -661,6 +665,45 @@ TEMPLATE_TEST_CASE("StateVectorKokkos::applyToffoli",
...
+
+        SECTION("Apply using dispatcher") {
+            SECTION("CSWAP [0,1,2]|+10> -> |010> + |101>") {
+                const std::vector<ComplexT> expected_results = {z, z, i, z,
+                                                                z, i, z, z};
+
+                StateVectorKokkos<TestType> svdat012{num_qubits};
+                Kokkos::deep_copy(svdat012.getView(), ini_sv);
+
+                svdat012.applyOperation("CSWAP", {0, 1, 2}, false);
```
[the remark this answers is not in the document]
josephleekl: CSWAP is a 3-qubit (non-controlled) gate

tests/test_gates.py
```diff
@@ -414,6 +414,76 @@ def circuit():
                 assert np.allclose(circ(), circ_def(), tol)
 
 
+@pytest.mark.skipif(
+    device_name != "lightning.kokkos",
+    reason="Controlled Single Qubit gate supported by Lightning Kokkos",
+)
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
josephleekl: reused, thanks!

pennylane_lightning/lightning_kokkos/_state_vector.py
```diff
@@ -281,10 +284,9 @@ def _apply_lightning(
             elif method is not None:  # apply specialized gate
                 param = operation.parameters
                 method(wires, invert_param, param)
-            elif isinstance(operation, qml.ops.Controlled) and isinstance(
-                operation.base, qml.GlobalPhase
+            elif (
```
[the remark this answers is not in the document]
josephleekl: I've now added support back

pennylane_lightning/lightning_kokkos/_state_vector.py
```diff
@@ -193,15 +192,19 @@ def _apply_lightning_controlled(self, operation):
...
-        inv = False
-        param = operation.parameters[0]
-        wires = self.wires.indices(operation.wires)
-        matrix = global_phase_diagonal(param, self.wires, control_wires, control_values)
-        state.apply(name, wires, inv, [[param]], matrix)
+        target_wires = list(operation.target_wires)
+        if method is not None:  # apply n-controlled specialized gate
+            inv = False
+            param = operation.parameters
+            method(control_wires, control_values, target_wires, inv, param)
```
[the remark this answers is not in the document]
josephleekl: There is some issue with codecov's reporting (It doesn't pick up Lightning Kokkos python coverage). I have run lcov locally and this is covered.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: removed

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: removed

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
```diff
@@ -410,16 +622,34 @@ TEMPLATE_TEST_CASE("StateVectorKokkosManaged::applyControlledGlobalPhase",
...
+            CHECK((imag(result_sv[j])) == Approx(imag(tmp)));
+        }
+    }
+
+    SECTION("Controlled GlobalPhase") {
+        const TestType pi2 = 1.5707963267948966;
+        auto sv_data = createRandomStateVectorData<TestType>(re, num_qubits);
+        StateVectorKokkos<TestType> kokkos_sv(
+            reinterpret_cast<ComplexT *>(sv_data.data()), sv_data.size());
+        kokkos_sv.applyOperation("GlobalPhase", {0, 1}, {0, 1}, {2}, inverse,
```
maliasadi, line 644: Do you have any plan to test these named gates with multiple controlled wires and values?

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -106,6 +106,17 @@
...
+        "C(PauliY)",
+        "C(PauliZ)",
+        "C(Hadamard)",
+        "C(S)",
+        "C(T)",
+        "C(PhaseShift)",
+        "C(RX)",
+        "C(RY)",
+        "C(RZ)",
+        "C(Rot)",
```
maliasadi, line 119: Where do we test these controlled wires with 1 and multiple controlled wires and values?

pennylane_lightning/lightning_kokkos/_state_vector.py
```diff
@@ -281,10 +284,26 @@ def _apply_lightning(
...
+                        qml.PauliY,
+                        qml.PauliZ,
+                        qml.Hadamard,
+                        qml.S,
+                        qml.T,
+                        qml.PhaseShift,
+                        qml.RX,
+                        qml.RY,
+                        qml.RZ,
+                        qml.Rot,
```
LuisAlfredoNu, line 302: Do you think will be better to define this list at the top of the file or in `lightning_kokkos.py`?

tests/test_gates.py
```diff
@@ -415,8 +415,8 @@ def circuit():
 
 
 @pytest.mark.skipif(
-    device_name not in ("lightning.qubit", "lightning.tensor"),
-    reason="N-controlled operations only implemented in lightning.qubit and lightning.tensor.",
+    device_name not in ("lightning.qubit", "lightning.tensor", "lightning.kokkos"),
+    reason="N-controlled operations only implemented in lightning.qubit, lightning.tensor and lightning.kokkos.",
```
LuisAlfredoNu, line 419: Could you invert this to something like
```python 
device_name in ("lightning.gpu"), 
reason="N-controlled operation is not supported by lightning.gpu"
```

LuisAlfredoNu (commented):
Thank you @josephleekl 
Good work adding N-Controlled gates  support in LKokkos
A few comments.
:rocket:

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: I have tested the performance for these non-controlled gates and did not observe any difference

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,54 +255,105 @@ void applyHadamard(
...
+              const std::size_t num_qubits,
+              const std::vector<std::size_t> &controlled_wires,
+              const std::vector<bool> &controlled_values,
+              const std::vector<std::size_t> &wires, const bool inverse = false,
+              [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
     const Kokkos::complex<PrecisionT> shift =
         (inverse) ? conj(exp(Kokkos::complex<PrecisionT>(
                         0, static_cast<PrecisionT>(M_PI / 4))))
                   : exp(Kokkos::complex<PrecisionT>(
                         0, static_cast<PrecisionT>(M_PI / 4)));
```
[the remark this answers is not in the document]
josephleekl: good catch!

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -190,15 +190,11 @@ void LightningKokkosSimulator::MatrixOperation(
     const std::vector<std::complex<double>> &matrix,
     const std::vector<QubitIdType> &wires, bool inverse,
     const std::vector<QubitIdType> &controlled_wires,
-    const std::vector<bool> &controlled_values) {
```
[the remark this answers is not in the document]
josephleekl: Controlled QubitUnitary is not supported

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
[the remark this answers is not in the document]
josephleekl: Good catch!

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -106,6 +106,17 @@
         "CRX",
         "CRY",
         "CRZ",
+        "C(PauliX)",
+        "C(PauliY)",
+        "C(PauliZ)",
+        "C(Hadamard)",
```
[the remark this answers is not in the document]
josephleekl: I don't think CH is a gate available here, should it be added? 
https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/src/gates/GateOperation.hpp

pennylane_lightning/lightning_kokkos/_state_vector.py
josephleekl: This list is temporary, and will be removed when a later PR when all gates are supported

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -27,13 +27,64 @@ namespace {
...
+using Pennylane::LightningKokkos::Util::reverseWires;
 using Pennylane::LightningKokkos::Util::vector2view;
 } // namespace
 /// @endcond
 
 namespace Pennylane::LightningKokkos::Functors {
 
-template <class PrecisionT, class FuncT> class applyNC1Functor {
+template <class PrecisionT, class FuncT, bool has_controls = true>
+class applyNC1Functor {};
```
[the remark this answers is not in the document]
josephleekl: These are needed so it can be specialized into ones with and without controls later on

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -64,17 +115,65 @@ template <class PrecisionT, class FuncT> class applyNC1Functor {
...
+        kokkos_swap(arr(i0), arr(i1));
+    };
+    if (controlled_wires.empty()) {
+        applyNC1Functor<PrecisionT, decltype(core_function), false>(
+            ExecutionSpace{}, arr_, num_qubits, wires, core_function);
+    } else {
+        applyNC1Functor<PrecisionT, decltype(core_function), true>(
+            ExecutionSpace{}, arr_, num_qubits, controlled_wires,
+            controlled_values, wires, core_function);
+    }
```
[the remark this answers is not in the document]
josephleekl: This is because `applyPauliX` calls `applyNCPauliX` but with empty `controlled_wires`

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -27,13 +27,64 @@ namespace {
...
+    Kokkos::View<Kokkos::complex<PrecisionT> *> arr;
+    const FuncT core_function;
+    KokkosIntVector indices;
+    KokkosIntVector parity;
+    KokkosIntVector rev_wires;
+    KokkosIntVector rev_wire_shifts;
+
+  public:
+    template <class ExecutionSpace>
+    applyNC1Functor([[maybe_unused]] ExecutionSpace exec,
```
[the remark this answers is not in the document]
josephleekl: So this `ExecutionSpace` is passed from https://github.com/PennyLaneAI/pennylane-lightning/blob/5f70b06e286447afb6703ce98df05d379108e83f/pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp#L77 , which will depend on what device execution space Kokkos is built for

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -100,13 +216,37 @@ void applyPauliZ(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+    [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
+    auto core_function = KOKKOS_LAMBDA(
+        Kokkos::View<Kokkos::complex<PrecisionT> *> arr,
+        [[maybe_unused]] const std::size_t i0, const std::size_t i1) {
+        const Kokkos::complex<PrecisionT> v0 = arr(i0);
+        const Kokkos::complex<PrecisionT> v1 = arr(i1);
+        arr(i0) = M_SQRT1_2 * v0 +
+                  M_SQRT1_2 * v1; // NOLINT(readability-magic-numbers)
+        arr(i1) = M_SQRT1_2 * v0 +
+                  -M_SQRT1_2 * v1; // NOLINT(readability-magic-numbers)
```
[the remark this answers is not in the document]
AmintorDusko: It is nice to keep the original formulation to guide other developers, but we also need to consider performance.
I suggest we implement the code following @LuisAlfredoNu suggestion and add the original formulation as a comment.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,49 @@ class StateVectorKokkos final
...
     }
 
+    /**
+     * @brief Apply a controlled-single gate to the state vector.
+     *
+     * @param opName Name of gate to apply.
+     * @param controlled_wires Control wires.
+     * @param controlled_values Control values (false or true).
+     * @param wires Wires to apply gate to.
+     * @param inverse Indicates whether to use adjoint of gate.
```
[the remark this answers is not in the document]
AmintorDusko: Hey @josephleekl, I see several comments here marked as resolved, but it is unclear to me what the resolution was. Could you please add a comment? It is not clear if you are applying locally or considering it a nonsensical suggestion.

AmintorDusko (commented):
Hey @josephleekl, could you please clarify in the resolved conversation which was the decided resolution? Or maybe commit the changes if still local.
Thanks.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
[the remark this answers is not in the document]
josephleekl: ah good point, most of these are fixed locally. I'll add a comment, thanks for the reminder!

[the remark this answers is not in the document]
josephleekl: added in docstring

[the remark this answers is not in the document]
AmintorDusko: Thanks!

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
josephleekl: Thanks, I've removed this now.

pennylane_lightning/core/src/simulators/lightning_kokkos/bindings/LKokkosBindings.hpp
josephleekl: Edited, thanks!

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: tested and found no performance regression

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: Actually after discussion with Ali, these are necessary for both GCC and Clang. These will be added back in later commit

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: will fix in later commit (issue with GCC/Clang)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: Related to Clang/GCC complain issue, will fix in later commit

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: Codecov tested locally which covers these region. (There is issue with LK codecov collection on CI)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: CI has codecov collection issue with LK, tested locally and is covered.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: will fix (GCC/Clang issue)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: Codecov issue with LK and CI.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,49 @@ class StateVectorKokkos final
...
+     * @param params Optional parameter list for parametric gates.
+     * @param gate_matrix Optional std gate matrix if opName doesn't exist.
```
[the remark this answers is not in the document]
josephleekl: applied.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,49 @@ class StateVectorKokkos final
...
+    void applyOperation(const std::string &opName,
+                        const std::vector<std::size_t> &controlled_wires,
+                        const std::vector<bool> &controlled_values,
+                        const std::vector<std::size_t> &wires,
+                        bool inverse = false,
+                        const std::vector<fp_t> &params = {},
+                        const std::vector<ComplexT> &gate_matrix = {}) {
+        PL_ABORT_IF_NOT(
+            areVecsDisjoint<std::size_t>(controlled_wires, wires),
+            "`controlled_wires` and `target wires` must be disjoint.");
```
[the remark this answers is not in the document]
josephleekl: applied.

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -710,4 +710,4 @@ class StateVectorKokkos final
     inline static bool is_exit_reg_ = false;
 };
 
-}; // namespace Pennylane::LightningKokkos
+}; // namespace Pennylane::LightningKokkos
```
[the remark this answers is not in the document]
josephleekl: applied. Will double check before next commit

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -27,13 +27,64 @@ namespace {
...
+        indices = vector2view(indices_);
+        Kokkos::parallel_for(
+            Kokkos::RangePolicy<ExecutionSpace>(
+                0, exp2(num_qubits - controlled_wires.size() - wires.size())),
+            *this);
+    }
+    KOKKOS_FUNCTION void operator()(const std::size_t k) const {
+        const std::size_t offset = parity_2_offset(parity, k);
+        std::size_t i0 = indices(0B00);
+        std::size_t i1 = indices(0B01);
```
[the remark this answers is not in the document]
josephleekl: applied

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -27,13 +27,64 @@ namespace {
...
+using Pennylane::LightningKokkos::Util::parity_2_offset;
+using Pennylane::LightningKokkos::Util::reverseWires;
 using Pennylane::LightningKokkos::Util::vector2view;
 } // namespace
 /// @endcond
 
 namespace Pennylane::LightningKokkos::Functors {
 
-template <class PrecisionT, class FuncT> class applyNC1Functor {
+template <class PrecisionT, class FuncT, bool has_controls = true>
```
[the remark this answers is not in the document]
josephleekl: default value removed

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -64,17 +115,65 @@ template <class PrecisionT, class FuncT> class applyNC1Functor {
...
+            ExecutionSpace{}, arr_, num_qubits, controlled_wires,
+            controlled_values, wires, core_function);
+    }
+}
+
 template <class ExecutionSpace, class PrecisionT>
 void applyPauliX(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
                  const std::size_t num_qubits,
                  const std::vector<std::size_t> &wires,
                  [[maybe_unused]] const bool inverse = false,
```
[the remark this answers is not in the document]
josephleekl: applied

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -83,15 +182,32 @@ void applyPauliY(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
```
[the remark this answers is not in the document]
josephleekl: applied

pennylane_lightning/core/src/simulators/lightning_kokkos/tests/Test_StateVectorLKokkos.cpp
```diff
@@ -417,4 +418,4 @@ TEMPLATE_TEST_CASE("StateVectorKokkos::normalize", "[StateVectorKokkos]", float,
                               expected_state.data(), expected_state.size(),
                               eps));
     }
-}
+}
```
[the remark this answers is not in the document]
josephleekl: applied

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
josephleekl: the argument `exec` is not actually used in the body, so it will complain if we remove `[[maybe_unused]]`. The argument is needed just to infer the class type for `ExecutionSpace`

pennylane_lightning/core/src/simulators/lightning_kokkos/bindings/LKokkosBindings.hpp
```diff
@@ -48,6 +48,33 @@ using StateVectorBackends =
...
+    using Pennylane::Util::for_each_enum;
+    namespace Constant = Pennylane::Gates::Constant;
+
+    for_each_enum<ControlledGateOperation>(
+        [&pyclass](ControlledGateOperation gate_op) {
+            using Pennylane::Util::lookup;
+            const auto gate_name =
+                std::string(lookup(Constant::controlled_gate_names, gate_op));
+            const std::string doc = "Apply the " + gate_name + " gate.";
+            auto func = [gate_name = gate_name](
```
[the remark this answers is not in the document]
josephleekl: updated

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
josephleekl: will do!

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -115,54 +255,105 @@ void applyHadamard(
...
+
+template <class ExecutionSpace, class PrecisionT>
+void applyNCT(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
+              const std::size_t num_qubits,
+              const std::vector<std::size_t> &controlled_wires,
+              const std::vector<bool> &controlled_values,
+              const std::vector<std::size_t> &wires, const bool inverse = false,
+              [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
     const Kokkos::complex<PrecisionT> shift =
         (inverse) ? conj(exp(Kokkos::complex<PrecisionT>(
```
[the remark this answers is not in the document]
josephleekl: updated.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -262,15 +504,77 @@ void applyRot(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+
+template <class ExecutionSpace, class PrecisionT>
+void applyNCGlobalPhase(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
+                        const std::size_t num_qubits,
+                        const std::vector<std::size_t> &controlled_wires,
+                        const std::vector<bool> &controlled_values,
+                        [[maybe_unused]] const std::vector<std::size_t> &wires,
+                        const bool inverse = false,
+                        const std::vector<PrecisionT> &params = {}) {
+    const Kokkos::complex<PrecisionT> phase = Kokkos::exp(
```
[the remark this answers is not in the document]
josephleekl: For now I will add `Kokkos::` namespace to calls

tests/test_gates.py
josephleekl: edited

pennylane_lightning/core/src/simulators/lightning_kokkos/utils/UtilKokkos.hpp
```diff
@@ -115,4 +116,123 @@ inline auto wires2Parity(const std::size_t num_qubits,
...
+inline auto generateBitPatterns(const std::vector<std::size_t> &wires,
+                                const std::size_t num_qubits)
+    -> std::vector<std::size_t> {
+    std::vector<std::size_t> indices;
+    indices.reserve(Pennylane::Util::exp2(wires.size()));
+    indices.emplace_back(0);
+
+    for (auto index_it = wires.rbegin(); index_it != wires.rend(); index_it++) {
+        const std::size_t value =
+            Pennylane::Util::maxDecimalForQubit(*index_it, num_qubits);
```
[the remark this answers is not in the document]
josephleekl: It does work and I have changed it, and included `<ranges>` , maybe others can comment if we like using this?

pennylane_lightning/core/src/simulators/lightning_kokkos/utils/UtilKokkos.hpp
```diff
@@ -115,4 +116,123 @@ inline auto wires2Parity(const std::size_t num_qubits,
...
+    std::vector<std::size_t> all_wires;
+    all_wires.reserve(nw_tot);
+    all_wires.insert(all_wires.begin(), wires.begin(), wires.end());
+    all_wires.insert(all_wires.begin() + wires.size(), controlled_wires.begin(),
+                     controlled_wires.end());
+
+    std::vector<std::size_t> rev_wires_(nw_tot);
+    for (std::size_t k = 0; k < nw_tot; k++) {
+        rev_wires_[k] = (num_qubits - 1) - all_wires[(nw_tot - 1) - k];
+    }
```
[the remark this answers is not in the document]
josephleekl: updated

pennylane_lightning/core/src/simulators/lightning_kokkos/utils/UtilKokkos.hpp
```diff
@@ -115,4 +116,123 @@ inline auto wires2Parity(const std::size_t num_qubits,
...
+    KokkosIntVector rev_wires;
+
+    const std::size_t n_contr = controlled_wires.size();
+    const std::size_t n_wires = wires.size();
+    const std::size_t nw_tot = n_contr + n_wires;
+    std::vector<std::size_t> all_wires;
+    all_wires.reserve(nw_tot);
+    all_wires.insert(all_wires.begin(), wires.begin(), wires.end());
+    all_wires.insert(all_wires.begin() + wires.size(), controlled_wires.begin(),
+                     controlled_wires.end());
```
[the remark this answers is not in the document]
josephleekl: updated to 
```
std::copy(wires.begin(), wires.end(), std::back_inserter(all_wires));
std::copy(controlled_wires.begin(), controlled_wires.end(), std::back_inserter(all_wires));
```

pennylane_lightning/core/src/simulators/lightning_kokkos/utils/UtilKokkos.hpp
```diff
@@ -115,4 +116,123 @@ inline auto wires2Parity(const std::size_t num_qubits,
...
+    std::vector<std::size_t> rev_wires_(nw_tot);
+    for (std::size_t k = 0; k < nw_tot; k++) {
+        rev_wires_[k] = (num_qubits - 1) - all_wires[(nw_tot - 1) - k];
+    }
+    const std::vector<std::size_t> parity_ = revWireParity(rev_wires_);
+
+    Kokkos::View<const std::size_t *, Kokkos::HostSpace,
+                 Kokkos::MemoryTraits<Kokkos::Unmanaged>>
+        rev_wires_host(rev_wires_.data(), rev_wires_.size());
+    Kokkos::resize(rev_wires, rev_wires_host.size());
```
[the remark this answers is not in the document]
josephleekl: This is necessary because `rev_wires` (Kokkos view vector) does not have a size yet, only the host vector (`rev_wires_`) and the host view of the same vector (`rev_wires_host`), and this copy fails when they are not the same size.

pennylane_lightning/core/src/simulators/lightning_kokkos/utils/UtilKokkos.hpp
```diff
@@ -115,4 +116,123 @@ inline auto wires2Parity(const std::size_t num_qubits,
...
+    Kokkos::deep_copy(rev_wires, rev_wires_host);
+
+    Kokkos::View<const std::size_t *, Kokkos::HostSpace,
+                 Kokkos::MemoryTraits<Kokkos::Unmanaged>>
+        parity_host(parity_.data(), parity_.size());
+    Kokkos::resize(parity, parity_host.size());
```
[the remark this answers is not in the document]
josephleekl: same as above

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
josephleekl: I will add some tests for multiple control wires/values. More tests will be added when controlled matrix is supported

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
josephleekl: these are tested [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/4dd344ed5d1f02fecbd56be0c0910a78c601f701/tests/test_gates.py#L452) in python, c++ tests will be added

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
josephleekl: tests added.

josephleekl:
Thanks for first round of comments @maliasadi @LuisAlfredoNu @multiphaseCFD @AmintorDusko , I've now implemented the comments, and some new changes include:
- updated toml file for catalyst
- updated [[maybe_unused]] in kokkos lambda body instead of argument
- added more tests

(There's still issue with codecov & LK-python, so I will check locally that all changes are covered)

pennylane_lightning/lightning_kokkos/lightning_kokkos.toml
AmintorDusko, line 1: This file is more about catalyst capabilities. Would be nice to have the @PennyLaneAI/compilation team checking it. Could you folks double-check it?

AmintorDusko (commented):
Some comments, and suggestions.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -621,7 +929,7 @@ void applyIsingZZ(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
     const Kokkos::complex<PrecisionT> shift_0 = Kokkos::complex<PrecisionT>{
         std::cos(angle / 2),
         (inverse) ? std::sin(angle / 2) : -std::sin(angle / 2)};
-    const Kokkos::complex<PrecisionT> shift_1 = conj(shift_0);
+    const Kokkos::complex<PrecisionT> shift_1 = Kokkos::conj(shift_0);
```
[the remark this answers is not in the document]
LuisAlfredoNu: @AmintorDusko When I was reviewing I noticed that some functions have this namespace then I checked the [documentation](https://kokkos.org/kokkos-core-wiki/API/core/numerics/mathematical-functions.html#notes) this suggestion to use explicit qualification `Kokkos::[math_func]` but now we have the question if this only valid for all the code or just for Kokkos parallel region. :thinking: 

I suggest using it in all math calls because Kokkos will be aware of which math library link, e.g. [AMD Math Lib](https://www.amd.com/en/developer/aocl/libm.html)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
```diff
@@ -758,43 +801,281 @@ TEMPLATE_TEST_CASE("StateVectorKokkos::applyMultiQubitOp",
...
+        sv_gate.applyOperation("PauliY", control_wires, control_values,
+                               target_wire, inverse);
+        auto sv_gate_host = Kokkos::create_mirror_view_and_copy(
+            Kokkos::HostSpace{}, sv_gate.getView());
+
+        std::vector<ComplexT> expected_result{
+            ComplexT{0.35355339, 0.0},  ComplexT{0.35355339, 0.0},
+            ComplexT{0.35355339, 0.0},  ComplexT{0.35355339, 0.0},
+            ComplexT{0.0, -0.35355339}, ComplexT{0.35355339, 0.0},
+            ComplexT{0.0, 0.35355339},  ComplexT{0.35355339, 0.0},
```
LuisAlfredoNu, line 980: Could you explain where come this number and if it is possible to generate it analytically, please?

The same for the rest of the test

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
```diff
@@ -116,6 +116,378 @@ TEMPLATE_TEST_CASE("StateVectorKokkos::applyMatrix/Operation",
...
+        sv_gate.applyOperation("RX", control_wires, control_values, target_wire,
+                               inverse, {param});
+        auto sv_gate_host = Kokkos::create_mirror_view_and_copy(
+            Kokkos::HostSpace{}, sv_gate.getView());
+
+        std::vector<ComplexT> expected_result{
+            ComplexT{0.35355339, 0.0},         ComplexT{0.35355339, 0.0},
+            ComplexT{0.35355339, 0.0},         ComplexT{0.35355339, 0.0},
+            ComplexT{0.35113625, -0.04127144}, ComplexT{0.35355339, 0.0},
+            ComplexT{0.35113625, -0.04127144}, ComplexT{0.35355339, 0.0},
```
LuisAlfredoNu, line 387: idem

LuisAlfredoNu (commented):
Thank @josephleekl for applying the suggestion. This PR is almost done on my end. My only concern is regarding using `Kokkos::[math_funct]`, once this has been resolved I will give you the approval. :smile:

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -27,13 +27,64 @@ namespace {
...
+    template <class ExecutionSpace>
+    applyNC1Functor([[maybe_unused]] ExecutionSpace exec,
+                    Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
+                    std::size_t num_qubits,
+                    const std::vector<std::size_t> &controlled_wires,
+                    const std::vector<bool> &controlled_values,
+                    const std::vector<std::size_t> &wires, FuncT core_function_)
+        : arr(arr_), core_function(core_function_) {
+
+        std::tie(parity, rev_wires) =
```
maliasadi, line 64: Couldn't we use structured bindings instead of `std::tie`? Structured binding is more modern and concise I reckon.
```suggestion
        auto &&[parity, rev_wires] =
```

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -262,15 +508,77 @@ void applyRot(Kokkos::View<Kokkos::complex<PrecisionT> *> arr_,
...
+    };
+    std::size_t target{0U};
+    if (!controlled_wires.empty()) {
+        for (std::size_t i = 0; i < num_qubits; i++) {
+            if (std::find(controlled_wires.begin(), controlled_wires.end(),
+                          i) == controlled_wires.end()) {
+                target = i;
+                break;
+            }
+        }
```
maliasadi, line 562: These for-loops are really good candidates for C++20 `ranges`! What do you think about pros & cons of using `ranges` in Lightning? @PennyLaneAI/performance

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/MatrixGateFunctors.hpp
```diff
@@ -539,4 +522,4 @@ template <class PrecisionT> struct apply5QubitOpFunctor {
     }
 };
 
-} // namespace Pennylane::LightningKokkos::Functors
+} // namespace Pennylane::LightningKokkos::Functors
```
maliasadi, line 525: Could you double check these EOF rendering issues? There are easy to fix but will require some last checks...
```suggestion
} // namespace Pennylane::LightningKokkos::Functors

```

pennylane_lightning/core/src/utils/Util.hpp
```diff
@@ -592,4 +592,4 @@ bool areVecsDisjoint(const std::vector<T> &v1, const std::vector<T> &v2) {
     }
     return true;
 }
-} // namespace Pennylane::Util
+} // namespace Pennylane::Util
```
maliasadi, line 595:
```suggestion
} // namespace Pennylane::Util

```

maliasadi (approved):
Thanks @josephleekl! No major comments... just a few last chores before merging to the base class :parachute:

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -64,105 +115,249 @@ template <class PrecisionT, class FuncT> class applyNC1Functor {
...
+              const std::vector<bool> &controlled_values,
+              const std::vector<std::size_t> &wires, const bool inverse = false,
+              [[maybe_unused]] const std::vector<PrecisionT> &params = {}) {
     const Kokkos::complex<PrecisionT> shift =
-        (inverse) ? Kokkos::complex<PrecisionT>{0.0, -1.0}
-                  : Kokkos::complex<PrecisionT>{0.0, 1.0};
-    applyNC1Functor(
-        ExecutionSpace{}, arr_, num_qubits, wires,
+        (inverse) ? Kokkos::conj(exp(Kokkos::complex<PrecisionT>(
+                        0, static_cast<PrecisionT>(M_PI_4))))
```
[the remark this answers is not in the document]
maliasadi: @LuisAlfredoNu `static_cast` is a compile-time type conversion operator

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
maliasadi: It's mostly useful for parallel region. But I wonder if it adds any abstraction overhead to the code base if we replace all these math ops with Kokkos::func :thinking:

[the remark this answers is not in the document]
AmintorDusko: Feel free to resolve this conversation when sensible.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
LuisAlfredoNu: @maliasadi Thank you, you are right.  I was playing a bit with [compiler explorer](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxCDSAA6oCoRODB7evnoJSY4CQSHhLFEx0naYDilCBEzEBGk%2Bfly2mPY5DJXVBHlhkdGxtlU1dRmNCgOdwd2FvZIAlLaoXsTI7BzmAMzByN5YANQma24AbmVExPvYJhoAguub25h7B8gsTAQI55c315%2Bjr3jIO0OqDw6B2aCwXAgo3QIBAxwcJH2bnQCwi9A%2BZgAbDs4tVWApSGCBKMdhFUJ4dsFjsQlDM9gB2KzXHYsokMEkorxoh7mTGGYD0R4AERxeJYChMAFYLBopUL9kyrqy2RzUYKAfsRWgFBB%2BYKAFQ7X6OZAAfVEoyRnO55wgGgAdJKZjMFZ9lQB6d0qgg7a2ChTCnYQKnRWmPABiRoIMJASQYuqMBqjfzNFoIVrVmFtDqdLuZrM9yqLxZLpZZICjMbjEAAtHqHobjf9zUxLQc/Vm1tg7Y7na7vvT5d9rk2AUCQUSsGYodHYfDThmueiu7zRcR8YS0OyfWSKSGaZg6SZGW7WVvRphVHFiL7MzsEGIqIHRy228jM9ne/2lWfiT6O3sWL1oGuLruKUoynK37Kue/53hqaxaokCYCg296Pi6ayKh6XqwbeS4PAGmpBvuYb7JG0KwtWwGGg%2BtBUEe%2BYsoWZasWx5aVlRwS1jR6H0ZhirHkOVwcHMtCcJKvB%2BNwvCoJwbjWNYRoLEsPJmGsPCkAQmiiXMADWICSho%2BicJIUk6aQckcLwCggMZ2kcFocxwLASBoCwcR0NE5CUO5nn0DEwBSGYfB0AQoaUBEFkRME1QAJ6cJpMXMMQcUAPIRNoJyJbw7lsIIaUMLQCWObwWARF4wBuGItC2TJpBYC8iYrFoDV4MQJx4McdWtZeZReOFOXkIIzQWbQeAROuqUeFgFkEMQeAsEN1JkkoQqYE1ArBKApVzFQBjAAoABqeCYAA7mlcSMEN/CCCIYjsFIMiCIoKjqKVpC6I0BhGCgimWPoE22ZAcyoHErR1bJ1ILVgwMQHMpTlM4ECuEMDSkIEEwFEUmSJMkAho7j2QpF02O9CMzSdQI7SDJ49R6IjrQ0%2BM%2BQ9DEIxjITHMdKTbMSAjKnLPzJkcJJpDSa1Vk7KoAAcmI1pikg7MAyAAlI9pmEGuCECQgEaTMvAOU5%2BmGcZ4kcGZ4sWVZNl2VpOnOiLZjmR9Nv27tczUkkziSEAA%3D%3D%3D) and produce the same result.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -64,105 +115,249 @@ template <class PrecisionT, class FuncT> class applyNC1Functor {
...
+                  : exp(Kokkos::complex<PrecisionT>(
+                        0, static_cast<PrecisionT>(M_PI_4)));
```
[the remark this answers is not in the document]
LuisAlfredoNu: My only concern is to use the correct math function because is not the same using `std::sin` as `sin` and Kokkos takes care of this. Also, `sin`, `cos`, `exp`... only works with `double` then if we use `float` instead of `double` will be a slowdown because the compiler needs to convert the `float` -> `double` -> math func -> `double` - `float`. However, `std` the math function is templated to `sinf` or `sin` or `sinl` ([ref](https://en.cppreference.com/w/cpp/numeric/math/sin))

pennylane_lightning/core/src/simulators/lightning_kokkos/bindings/LKokkosBindings.hpp
```diff
@@ -48,6 +48,33 @@ using StateVectorBackends =
...
+    using Pennylane::Util::for_each_enum;
+    namespace Constant = Pennylane::Gates::Constant;
+
+    for_each_enum<ControlledGateOperation>(
+        [&pyclass](ControlledGateOperation gate_op) {
+            using Pennylane::Util::lookup;
+            const auto gate_name =
+                std::string(lookup(Constant::controlled_gate_names, gate_op));
+            const std::string doc = "Apply the " + gate_name + " gate.";
+            auto func = [gate_name](
```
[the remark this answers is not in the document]
josephleekl: Thanks, updated (locally)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
```diff
@@ -661,6 +665,45 @@ TEMPLATE_TEST_CASE("StateVectorKokkos::applyToffoli",
...
+                Kokkos::deep_copy(svdat012.getView(), ini_sv);
+
+                svdat012.applyOperation("CSWAP", {0, 1, 2}, false);
+
+                auto sv012 = Kokkos::create_mirror_view_and_copy(
+                    Kokkos::HostSpace{}, svdat012.getView());
+
+                for (std::size_t j = 0; j < exp2(num_qubits); j++) {
+                    CHECK(imag(expected_results[j]) == Approx(imag(sv012[j])));
+                    CHECK(real(expected_results[j]) == Approx(real(sv012[j])));
```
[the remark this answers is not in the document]
josephleekl: Thanks for the suggestion! Because `sv012[j]` is a `kokkos::complex` this doesn't seem to work, and I have tried various ways to cast `sv012[j]` into a `std::complex` to make this work, but I can't find any good solutions. I will keep this for now

pennylane_lightning/core/src/utils/Util.hpp
josephleekl: thanks, updated.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/MatrixGateFunctors.hpp
josephleekl: I've updated my vscode setting now to add new final line on save! Thanks :)

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
josephleekl: So I think for parallel region I will use `Kokkos::` for all func, and for non-parallel region I will use `std::` for standard (float/double) operations, and `Kokkos::` for operations on `kokkos::complex` types

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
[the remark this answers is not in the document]
AmintorDusko: Makes sense. Thank you for trying.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
josephleekl: `parity` and `rev_wires` are data members of the class (`parity` needs to be available to the `operator()`), and structured binding requires defining new variables. So if I use structured binding I will have to define a new variable and copy to the `parity` member, which would require an extra step?

pennylane_lightning/core/src/simulators/lightning_kokkos/StateVectorKokkos.hpp
```diff
@@ -474,6 +448,49 @@ class StateVectorKokkos final
...
+            areVecsDisjoint<std::size_t>(controlled_wires, wires),
+            "`controlled_wires` and target wires must be disjoint.");
+        PL_ABORT_IF_NOT(controlled_wires.size() == controlled_values.size(),
+                        "`controlled_wires` must have the same size as "
+                        "`controlled_values`.");
+        if (controlled_wires.empty()) {
+            return applyOperation(opName, wires, inverse, params, gate_matrix);
+        }
+        if (opName == "Identity") {
+            return; // No op
```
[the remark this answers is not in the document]
josephleekl: Good points!

Actually Python already checks for Identity, so it shouldn't be Identity by this point in C++. I'll change it to the suggested, thanks!

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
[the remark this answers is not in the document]
josephleekl: I will use `Kokkos::` for `kokkos::complex` types and in parallel regions, and `std::` for otherwise

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/BasicGateFunctors.hpp
```diff
@@ -64,105 +115,249 @@ template <class PrecisionT, class FuncT> class applyNC1Functor {
...
+    const PrecisionT &angle = params[0];
     const Kokkos::complex<PrecisionT> shift =
-        (inverse) ? conj(exp(Kokkos::complex<PrecisionT>(
-                        0, static_cast<PrecisionT>(M_PI / 4))))
-                  : exp(Kokkos::complex<PrecisionT>(
-                        0, static_cast<PrecisionT>(M_PI / 4)));
-    applyNC1Functor(
-        ExecutionSpace{}, arr_, num_qubits, wires,
+        (inverse) ? exp(-Kokkos::complex<PrecisionT>(0, angle))
+                  : exp(Kokkos::complex<PrecisionT>(0, angle));
```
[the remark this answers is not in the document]
josephleekl: updated

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_NonParam.cpp
josephleekl: These are generated from Pennylane. For this case it is possible to generate anaylytically, but for some gates (those with rotations) these are just numerical. I will add a comment in the code.

pennylane_lightning/core/src/simulators/lightning_kokkos/gates/tests/Test_StateVectorKokkos_Param.cpp
josephleekl: Comment added

LuisAlfredoNu (approved):
Thank you @josephleekl 
 Good feature addition!!!
:rocket:

pennylane_lightning/core/src/simulators/lightning_kokkos/bindings/LKokkosBindings.hpp
```diff
@@ -48,6 +48,33 @@ using StateVectorBackends =
...
+
+    using Pennylane::Gates::ControlledGateOperation;
+    using Pennylane::Util::for_each_enum;
+    namespace Constant = Pennylane::Gates::Constant;
+
+    for_each_enum<ControlledGateOperation>(
+        [&pyclass](ControlledGateOperation gate_op) {
+            using Pennylane::Util::lookup;
+            const auto gate_name =
+                std::string(lookup(Constant::controlled_gate_names, gate_op));
```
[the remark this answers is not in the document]
josephleekl: For L71 we need a `std::string` for `gate_name` in `applyOperation`, I can't think of a neat way to avoid converting to a `std::string` for that?

PR #961, opened 2024-10-24
Fix PTM stable latest

LuisAlfredoNu:
### Before submitting

**Context:**
The PTM has an unexpected behaviour for Stable Lightnin - Latest PennyLane because on CIs looks like Latest PL reference to the last release instead of to `master`

**Benefits:**
Correct behavior for the PTM

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-76772]


multiphaseCFD:
Thanks @LuisAlfredoNu . Could you please give more context in the description?

.github/workflows/tests_gpu_python.yml
```diff
@@ -150,13 +150,21 @@ jobs:
...
         uses: actions/checkout@v4
         with:
           path: pennylane
           repository: PennyLaneAI/pennylane
 
+      - name: Switch to latest build of PennyLane
+        if: inputs.pennylane-version == 'latest'
+        run: |
+          cd pennylane
+          git fetch --all
```
multiphaseCFD, line 164: Does it work with `git fetch origin master`?

multiphaseCFD (commented):
Thanks @LuisAlfredoNu ! LGTM. Just one comment

.github/workflows/tests_gpu_python.yml
mudit2812: Wondering the same thing.

mudit2812 (approved):
Thanks @LuisAlfredoNu . Looks good to me. Just one comment about the `git fetch --all` command you used, but it's not a blocker.

multiphaseCFD (approved):
Nice work @LuisAlfredoNu ! Please get it merged once all CIs passed

.github/workflows/tests_gpu_python.yml
```diff
@@ -150,13 +150,21 @@ jobs:
...
         with:
           path: pennylane
           repository: PennyLaneAI/pennylane
 
+      - name: Switch to latest build of PennyLane
+        if: inputs.pennylane-version == 'latest'
+        run: |
+          cd pennylane
+          git fetch origin master
+          git checkout master
```
mudit2812, line 165: Actually now that I think about it this might be unnecessary. The `actions/checkout@v4` action checks out the master branch by default, so we don't need to `git fetch` and `git checkout` as we should already be on the master branch. We need to `fetch --all` for the release build because we need to first find the RC branch.

LuisAlfredoNu: I suggest to keep `git checkout master` to make clear what is going on. I will remove `git fetch origin master`

mudit2812: Good idea :)

PR #983, opened 2024-11-06
Optimize `lightning.tensor` by adding direct MPS sites data set

LuisAlfredoNu:
**Context:**
Optimize `lightning.tensor` by adding direct MPS sites data set

Adding the  `MPSPrep` gate to be able to pass an MPS directly to the Tensor Network.
The `MPSPrep` gate frontend was developed on this [PR](https://github.com/PennyLaneAI/pennylane/pull/6431)

**Benefits:**
Avoid the decomposition from state vector to MPS sites which are expensive and inefficient 

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-74709]


pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -118,6 +120,56 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
     return mpos, sorted_wires
 
 
+def set_bond_dims(num_qubits: int, max_bond_dim: int) -> List:
```
multiphaseCFD, line 123: Could we move those check to C++ layer?

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -252,6 +252,31 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
+
+    void MPSShapeCheck(
+        const std::vector<std::vector<std::size_t>> &MPS_shape_source) {
+        bool sameShape = sitesExtents_ == MPS_shape_source;
+        if (!sameShape) {
+            auto MPS_shape_source_str =
+                Pennylane::Util::vector2DToString<std::size_t>(
+                    MPS_shape_source);
+            auto MPS_shape_dest_str =
+                Pennylane::Util::vector2DToString<std::size_t>(sitesExtents_);
```
LuisAlfredoNu, line 270: Convert the MPS shape array to string to print together with `PL_ABORT_IF_NOT` and give a hit to the user in case of failure.

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -252,6 +252,31 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
     }
 
+    /**
+     * @brief Check if the provided MPS has the correct dimension for C++
+     * backend.
+     *
+     * @param MPS_shape_source Dimension list of incoming MPS.
+     */
+
+    void MPSShapeCheck(
```
multiphaseCFD, line 262: Let's move this method to `util`.

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -444,6 +444,10 @@ def apply_operations(self, operations):
             elif isinstance(operations[0], BasisState):
                 self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
                 operations = operations[1:]
+            elif isinstance(operations[0], MPSPrep):
+                mps = operations[0].mps
+                self._tensornet.updateMPSSitesData(mps)
```
multiphaseCFD, line 449: Please add unit tests for this new feature and error message should be thrown out if the method is `tn`

multiphaseCFD (commented):
Thanks @LuisAlfredoNu ! Nice work, just a couple of comments

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: I checked this suggestion but It will be difficult because `MPSShapeCheck` need access to the private variable `sitesExtents_`. If we want to move  `MPSShapeCheck` to `utils` then we need to create a getter function and duplicate the `sitesExtents_` vector.  What do you think?

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
josephleekl: I agree it would be good to move this to util, and having a getter function to return the `sitesExtens_` vector reference makes sense, right?

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -252,6 +252,31 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
+        const std::vector<std::vector<std::size_t>> &MPS_shape_source) {
+        bool sameShape = sitesExtents_ == MPS_shape_source;
+        if (!sameShape) {
```
josephleekl, line 265: can we just combine these to `if (sitesExtents != MPS_shape_source)`

LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -444,6 +444,14 @@ def apply_operations(self, operations):
             elif isinstance(operations[0], BasisState):
                 self._apply_basis_state(operations[0].parameters[0], operations[0].wires)
                 operations = operations[1:]
+            elif isinstance(operations[0], MPSPrep):
+                if self.method == "tn":
```
multiphaseCFD, line 448: Do we want to do two `if` checks here?

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -169,6 +170,9 @@ def stopping_condition(op: Operator) -> bool:
     if isinstance(op, qml.ControlledQubitUnitary):
         return True
 
+    if isinstance(op, qml.MPSPrep):
```
multiphaseCFD, line 173: Is `StatePrep` missed here?

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -71,6 +71,7 @@
     {
         "Identity",
         "BasisState",
+        "MPSPrep",
```
multiphaseCFD, line 74: Is `StatePrep` missed here?

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,29 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
 
+/**
+ * @brief Check if the provided MPS has the correct dimension for C++
+ * backend.
+ *
+ * @param MPS_shape_dest Dimension list of destination MPS.
+ * @param MPS_shape_source Dimension list of incoming MPS.
+ */
+inline void
+MPSShapeCheck(const std::vector<std::vector<std::size_t>> &MPS_shape_dest,
```
multiphaseCFD, line 205: Not clear to me. To me, this API should have a `maxBondDim` signature.

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -499,6 +499,16 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
 
+    /**
+     * @brief Get a const vector reference of sitesExtents_.
+     *
+     * @return const std::vector<std::vector<std::size_t>>
+     */
+    [[nodiscard]] auto getSitesExtents()
+        -> const std::vector<std::vector<std::size_t>> & {
+        return sitesExtents_;
+    }
```
maliasadi, line 510: ?
```suggestion
    [[nodiscard]] auto getSitesExtents() const
        -> const std::vector<std::vector<std::size_t>> & {
        return sitesExtents_;
    }
```

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -499,6 +499,16 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
+
   protected:
```
maliasadi, line 512: Could you merge protected and private sections to avoid the class scoping fragmentation?

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
```diff
@@ -33,13 +33,14 @@
...
 #include "cuda_helpers.hpp"
+#include "tncuda_helpers.hpp"
 
 /// @cond DEV
 namespace {
 using namespace Pennylane;
 using namespace Pennylane::Bindings;
 using namespace Pennylane::LightningGPU::Util;
-using Pennylane::LightningTensor::TNCuda::MPSTNCuda;
+using namespace Pennylane::LightningTensor::TNCuda::Util;
```
maliasadi, line 43: nice catch!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
```diff
@@ -137,6 +138,20 @@ void registerBackendClassSpecificBindingsMPS(PyClass &pyclass) {
...
             [](TensorNet &tensor_network, std::vector<np_arr_c> &tensors) {
+                // Extract the incoming MPS shape
+                std::vector<std::vector<std::size_t>> MPS_shape_source;
+                for (std::size_t idx = 0; idx < tensors.size(); idx++) {
+                    py::buffer_info numpyArrayInfo = tensors[idx].request();
+                    auto MPS_site_source_shape = numpyArrayInfo.shape;
+                    std::vector<std::size_t> MPS_site_source(
+                        MPS_site_source_shape.begin(),
+                        MPS_site_source_shape.end());
+                    MPS_shape_source.push_back(std::move(MPS_site_source));
```
maliasadi, line 149: emplace_back ?
```suggestion
                    MPS_shape_source.emplace_back(std::move(MPS_site_source));
```

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
```diff
@@ -137,6 +138,20 @@ void registerBackendClassSpecificBindingsMPS(PyClass &pyclass) {
...
+                }
+
+                const auto &MPS_shape_dest = tensor_network.getSitesExtents();
+                MPSShapeCheck(MPS_shape_dest, MPS_shape_source);
```
maliasadi, line 153: I'm not sure if this check can be inlined? I also argue this check should be checked before calling the binding or within the implementation of updating sites data.

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,29 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
+              const std::vector<std::vector<std::size_t>> &MPS_shape_source) {
+
+    if (!(MPS_shape_dest == MPS_shape_source)) {
```
maliasadi, line 208: what's the difference with `!=`?

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,29 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
+        auto MPS_shape_source_str =
+            Pennylane::Util::vector2DToString<std::size_t>(MPS_shape_source);
+        auto MPS_shape_dest_str =
+            Pennylane::Util::vector2DToString<std::size_t>(MPS_shape_dest);
```
maliasadi, line 212: Why do we really need these vec2str? It seems redundant if users can print these shapes from the user program in case of the "Incorrect incoming MPS sites data shape" error?

pennylane_lightning/core/src/utils/tests/Test_Util.cpp
```diff
@@ -233,3 +233,15 @@ TEST_CASE("Util::areVecsDisjoint", "[Util][LinearAlgebra]") {
...
+    SECTION("Test for convert 2Dvector to string") {
+        std::vector<std::vector<std::size_t>> vec{
+            {2, 2, 4}, {4, 2, 8}, {8, 2, 8}};
+        std::string ref_str{"[[2, 2, 4], [4, 2, 8], [8, 2, 8]]"};
+
+        std::string vec2str = vector2DToString<std::size_t>(vec);
+
+        REQUIRE(ref_str == vec2str);
+    }
+}
```
maliasadi, line 247:
```suggestion
}

```

pennylane_lightning/lightning_tensor/lightning_tensor.py
maliasadi: Same question.

maliasadi (commented):
Nice work! :christmas_tree:

pennylane_lightning/lightning_tensor/lightning_tensor.py
LuisAlfredoNu: I am not sure why `StatePrep` is not here. I checked the full history of the file and never was added to `_operation` list :thinking:

pennylane_lightning/lightning_tensor/lightning_tensor.py
LuisAlfredoNu: I don't think so because the test for `StatePrep` past :thinking:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Yes, because, just like `StatePrep,` the method 'tn' doesn't support `MPSPrep` due to the limitation of the max bond dimension of 2. Your concert might be related to performance, but usually, we call this gate only once, such as `StatePrep`.
Additionally, for the method `tn`, `StatePrep` and `MPSPrep` will be equivalent.

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: Nice catch!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
LuisAlfredoNu: nice catch!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
LuisAlfredoNu: The initial implementation that I developed was in the Python layer before calling the binding. However, @multiphaseCFD [suggest](https://github.com/PennyLaneAI/pennylane-lightning/pull/983/#discussion_r1878564969) 
 moving this check to the cpp layer. :thinking: 
I think keeping this check on the cpp layer prevents a double computation of `sitesExtents` 

It is possible to move this check into `updateSiteData` but an early testing will prevent to compute additional steps such as `ZeroInit` or `MoveHost2Device`  which happen inside `updateSiteData`

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: shame on me :face_with_thermometer: 
I will update

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: I thought being friendly with the user because this print would help the user find easily the problem due to the MPS format for cutensornet is too specific.  However, if you request to remove `vector2DToString` I will delete it.  :smile:

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: It is not necessary the  `maxBondDim` because the vector `MPS_shape_dest` already has it :smile:

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: Unfortunately, it is not possible to merge protected and private because `MPSTNCuda` and `ExactTNCuda` are inherited from TNCuda class  :confused:

pennylane_lightning/lightning_tensor/_tensornet.py
multiphaseCFD: It's more  likely the function will go to the second branch right?

LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,28 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
+/**
+ * @brief Check if the provided MPS has the correct dimension for C++
+ * backend.
+ *
+ * @param MPS_shape_dest Dimension list of destination MPS.
+ * @param MPS_shape_source Dimension list of incoming MPS.
+ */
+inline void
+MPSShapeCheck(const std::vector<std::vector<std::size_t>> &MPS_shape_dest,
+              const std::vector<std::vector<std::size_t>> &MPS_shape_source) {
```
josephleekl, line 206: Could this function not be simplified to something like
```
PL_ABORT_IF_NOT(MPS_shape_dest == MPS_shape_source, "The incoming MPS..." + Pennylane::Util::vector2DToString<std::size_t>(MPS_shape_source) + "... " )

So we don't need to check `MPS_shape_dest == MPS_shape_source` twice, and also don't need to introduce these temporary strings?

File: pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/tests/Tests_MPSTNCuda.cpp [hunk 19]
```diff
@@ -323,4 +323,64 @@ TEMPLATE_TEST_CASE("MPOTNCuda::getBondDims()", "[MPOTNCuda]", float, double) {
 
         CHECK(bondDims == expected_bondDims);
     }
-}
\ No newline at end of file
+}
+
+TEMPLATE_TEST_CASE("MPSTNCuda::getSitesExtents()", "[MPSTNCuda]", float,
+                   double) {
+    SECTION("Check if sitesExtents retrun is correctly with 3 qubits") {
```
josephleekl, line 330:
```suggestion
    SECTION("Check if sitesExtents retrun is correct with 3 qubits") {
```

tests/lightning_tensor/test_lightning_tensor.py
```diff
@@ -157,3 +158,88 @@ def test_execute_and_compute_vjp(self, method):
...
+
+    def circuit(MPS):
+        qml.MPSPrep(mps=MPS, wires=dev_wires)
+        return qml.state()
+
+    qnode_ltensor = qml.QNode(circuit, dev)
+
+    try:
+        _ = qnode_ltensor(MPS)
+    except Exception as excinfo:
```
josephleekl, line 190: why do we need this try block? wouldn't the test just fail if this fails?

josephleekl (commented):
Just have some small comments!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
```diff
@@ -137,6 +138,20 @@ void registerBackendClassSpecificBindingsMPS(PyClass &pyclass) {
...
+                    std::vector<std::size_t> MPS_site_source(
+                        MPS_site_source_shape.begin(),
+                        MPS_site_source_shape.end());
+                    MPS_shape_source.emplace_back(std::move(MPS_site_source));
+                }
+
+                const auto &MPS_shape_dest = tensor_network.getSitesExtents();
+                MPSShapeCheck(MPS_shape_dest, MPS_shape_source);
+
                 for (std::size_t idx = 0; idx < tensors.size(); idx++) {
```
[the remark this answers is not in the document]
LuisAlfredoNu: No, because it will be faster to evaluate the `MPSCheck` and find an issue with the MPS than waiting for each evaluation and move of each MPS site to the device memory.

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
maliasadi: No, I meant can we have one single `private` and one `protected` instead of multiple in this class?

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
maliasadi: Agree -- thanks!

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
maliasadi: Let's be consistent with other runtime error messages! I know this is more friendly but the sort of info can be easily pulled up from the user code.

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: Nice catch

tests/lightning_tensor/test_lightning_tensor.py
LuisAlfredoNu: Yes I thought the same but I felt the test was too empty :sweat_smile:  I will remove the try block :smile:

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -499,7 +473,27 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
+     */
+    [[nodiscard]] auto getSitesExtents() const
+        -> const std::vector<std::vector<std::size_t>> & {
+        return sitesExtents_;
+    }
+
   protected:
+    // Note both maxBondDim_ and bondDims_ are used for both MPS and Exact
+    // Tensor Network. Per Exact Tensor Network, maxBondDim_ is 1 and bondDims_
+    // is {1}. Per Exact Tensor Network, setting bondDims_ allows call to
```
josephleekl, line 489:
```suggestion
    // is {1}. For Exact Tensor Network, setting bondDims_ allows call to
```

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,21 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
+    if (MPS_shape_dest != MPS_shape_source) {
```
josephleekl, line 207: We don't need this `if`, right? the `PL_ABORT_IF_NOT` will already check for this if

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
```diff
@@ -194,4 +194,21 @@ inline auto create_swap_wire_pair_queue(const std::vector<std::size_t> &wires)
...
+        PL_ABORT_IF_NOT(MPS_shape_dest == MPS_shape_source,
```
josephleekl, line 208: and in fact, if this `MPSShapeCheck` is just a `PL_ABORT_IF_NOT` , maybe we can just replace all instance of `MPSShapeCheck` to the `PL_ABORT_IF_NOT`  in `bindings`? it's only used once and it's quite short

josephleekl:
Added some small comments, happy to approve straight away once they're dealt with! @LuisAlfredoNu

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
```diff
@@ -499,7 +473,27 @@ class TNCuda : public TNCudaBase<PrecisionT, Derived> {
...
+     * @return const std::vector<std::vector<std::size_t>>
+     */
+    [[nodiscard]] auto getSitesExtents() const
+        -> const std::vector<std::vector<std::size_t>> & {
+        return sitesExtents_;
+    }
+
   protected:
+    // Note both maxBondDim_ and bondDims_ are used for both MPS and Exact
+    // Tensor Network. Per Exact Tensor Network, maxBondDim_ is 1 and bondDims_
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/tncuda/TNCuda.hpp
LuisAlfredoNu: Done!

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: I prefer to keep this function here due to Shuli's [suggestion](https://github.com/PennyLaneAI/pennylane-lightning/pull/983#discussion_r1888589332).

tests/lightning_tensor/test_lightning_tensor.py
```diff
@@ -157,3 +158,85 @@ def test_execute_and_compute_vjp(self, method):
...
+    dev = LightningTensor(wires=wires, method="mps", max_bond_dim=max_bond)
+    dev_wires = dev.wires.tolist()
+
+    def circuit(MPS):
+        qml.MPSPrep(mps=MPS, wires=dev_wires)
+        return qml.state()
+
+    qnode_ltensor = qml.QNode(circuit, dev)
+
+    _ = qnode_ltensor(MPS)
```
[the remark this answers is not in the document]
LuisAlfredoNu: No without creating an extra binding. I think this test is enough.

pennylane_lightning/core/src/simulators/lightning_tensor/utils/tncuda_utils/tncuda_helpers.hpp
LuisAlfredoNu: Done!

josephleekl (approved):
LTGM, thanks @LuisAlfredoNu

pennylane_lightning/lightning_tensor/lightning_tensor.py
maliasadi: Let's get back to this in a followup PR -- I have this on my list for Q1

maliasadi (approved):
Happy to approve :partying_face:

PR #985, opened 2024-11-06
Delete references to the removed `QubitStateVector` class

andrijapau:
**Context:**

Completing the deprecation cycle of `QubitStateVector` (see https://github.com/PennyLaneAI/pennylane/pull/6525)

Removed all references to the deprecated source code.

[sc-77482]


lillian542 (commented):
Left a few suggestions regarding the tests, see what you think :)

AmintorDusko (commented):
Thank you for your nice work. Could you please check your PR formatting?
Also, please fill out the changelog. :slightly_smiling_face:

AmintorDusko (approved):
Great work! Thank you!

lillian542 (approved):
🎉

albi3ro:
Looking over the indirect changes on code cov, I think codecov is just misbehaving as it sometimes does.  We can try just rerunning the CI and seeing if that fixes it, and if not, we could even potentially override.

It makes it look like we've removed *All* lightning tensor device tests somehow.  Initializing `lightning.tensor` is somehow no longer covered.  Which seems wrong.  So I would say codecov is just being wonky in this case.

Issue #962, opened 2024-10-24
 Parallelize `collapse` method for `lightning.qubit` with `OpenMP`


tomlqc:
## Important Note
⚠️ This issue is part of an internal assignment and not meant for external contributors

## Context

Recently, Mid-Circuit Measurement (MCM) support was added to the `lightning.qubit` backend. The performance of the `collapse` method is very import for the  MCM support. Parallelize the `collapse` method with `OpenMP` would boost the performance of MCM simulations using `lightning.qubit`.

## Requirements

The `collapse` method with `OpenMP` support has to be implemented in the `lightning.qubit` C++ backend. For more information about the `collapse` method, please refer to [Pennylane-Lightning source code](https://github.com/PennyLaneAI/pennylane-lightning/blob/7c43ed690807b45834b72633bac1b0ef7158d088/pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp#L673).

- Parallelize the `collapse` method with `OpenMP` in [pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp](https://github.com/PennyLaneAI/pennylane-lightning/blob/7c43ed690807b45834b72633bac1b0ef7158d088/pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp#L673)
- The new implementation would allow users to choose whether the `collapse` method is built against `OpenMP` or not.
- Benchmark the new C++ implementation (`Num_Qubits` vs `OpenMP` threads number) and upload the results to the Pull-Request for further discussions.
- Create a pull-request in the [PennyLane Lightning](https://github.com/PennyLaneAI/pennylane-lightning/) repository and ensure to complete all the necessary steps outlined for creating the PR before marking it ready for review.

Feel free to ask any questions or raise any concerns regarding the issue. We'll be happy to discuss with you!

xiaohanzai:
Hi Thomas, for the benchmarking, should I just take the collapse function out and run it on my own laptop?  What's the possible range of values for Num_Qubits?

tomlqc:
Hi Xiaohan, yes, you may just benchmark `collapse()` separately, and please share with us how you did it. The upper limit for the number of qubits will be your laptop's memory, you can start with 10. And don't forget to compare different number of threads :slightly_smiling_face:

xiaohanzai:
[...]

PR #986, opened 2024-11-07
parallelized collapse function with openmp in StateVectorLQubit.hpp

xiaohanzai:
### Before submitting

**Context:**
Would like to improve performance of the `collapse` method since Mid-Circuit Measurement (MCM) support was added to the `lightning.qubit` backend.

Parallelized `collapse` function with `OpenMP`.

**Benefits:**
Faster execution with several threads.

Here's the strong scaling results on my mac and on a cluster at UofT.
To benchmark the implementation, I assumed `num_qubits = 10-30`.  The upper bound of 30 is chosen so that on my mac a complex float array of size `1<<30` can be held in my RAM.
The execution times are obtained by averaging the run time of the `collapse` function 10 or more times.  Attached please also see the code that I used for benchmarking.

Screenshot 2024-11-06 at 7 06 44 PM
Screenshot 2024-11-06 at 7 10 48 PM
Screenshot 2024-11-07 at 10 04 14 AM

**Possible Drawbacks:**

**Related GitHub Issues:**
[#962](https://github.com/PennyLaneAI/pennylane-lightning/issues/962)


AmintorDusko:
Hi @xiaohanzai, thank you for that! Could you please resolve the conflicts so we can run our CIs?

AmintorDusko:
Thank you, @xiaohanzai!

multiphaseCFD:
Thanks @xiaohanzai.  Could you please share your benchmark scripts (both cpp and shell script code) for discussion?

multiphaseCFD:
Thanks @xiaohanzai for sharing the scripts. A few questions come to my mind for discussion. Please feel free to leave your thoughts, thanks!

multiphaseCFD:
Since `normalize()` is a part of the `collapse()`, would you like to parallelize it as well?

multiphaseCFD:
Could you please explain why increasing the number of threads can improve the performance when `num_threads <=16`, while further increase of thread numbers would improve the performance  when `num_qubits > 20`?

multiphaseCFD:
Could you please explain why the `stride=vec_size/2` scales better than `stride=1` on the `mac`, while it's opposite on the `node of cluster`?

multiphaseCFD:
Is there any room to optimize this method? If yes, how would you like to further optimize this code? If not, why?

xiaohanzai:
Hi @multiphaseCFD 

I pushed another commit to parallelize `normalize`.  Here are my thoughts on your questions:

> Could you please explain why more threads lead performance regression when the number of qubit is small, that's say num_qubits = 10?

There's too few array elements to deal with, so the parallelization overhead is a lot larger than the actual workload.  It's not very beneficial to parallelize the code unless the array size is large enough.

> Could you please explain why increasing the number of threads can improve the performance when num_threads <=16, while further increase of thread numbers would lead the performance regression when num_qubits > 20?

My speculation is that 8 32KB L1 caches can hold about 2^15 complex float array elements, so when array size isn't too large they can be stored in the nearest caches.  But when array size gets too large e.g. 2^30 then they will have to be loaded from RAM.  Then memory latency gets in the way.

> Could you please explain why the stride=vec_size/2 scales better than stride=1 on the mac, while it's opposite on the node of cluster?

That I'm not sure...  I'd expect the performance of stride=1 to be worse in general because of cache misses, but I'm not sure why mac and cluster appear to be very different.

> Is there any room to optimize this method? If yes, how would you like to further optimize this code? If not, why?

I considered dynamic scheduling, static scheduling with a small chunk size, and loop tiling, but these all didn't seem to work better than the default scheduling.  I think the scheduling overhead is too large when array size is large and chunk size is small.  So I think on my side I don't have any further ways of optimizing it.

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -695,9 +695,12 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
         // **__**__ for stride 2
         // ****____ for stride 4
         const std::size_t k = branch ? 0 : 1;
+#if defined(_OPENMP)
+#pragma omp parallel for collapse(2) default(none) shared(arr, half_section_size, stride, k)
+#endif
         for (std::size_t idx = 0; idx < half_section_size; idx++) {
```
multiphaseCFD, line 701: Thanks @xiaohanzai . 
Could you try to fuse these two loops into one to see if the performance can be improved, especially for the `stride=1` case?

xiaohanzai:
@multiphaseCFD Thanks a lot for the suggestion!  I just updated the code in codeshare.  It seems like on my mac the results didn't get improved a lot, especially for stride = 1.  This plot is got by running `collapse` function without `normalize()`, so same as before:
Screenshot 2024-11-08 at 8 31 21 PM

This is after adding in a parallelized `normalize`:
Screenshot 2024-11-08 at 8 32 29 PM

And results on the cluster with `normalize`:
Screenshot 2024-11-08 at 8 38 17 PM

tomlqc:
Thanks @xiaohanzai

PR #994, opened 2024-11-12
Remove Tensor and Hamiltonian in python frontend for lightning

lillian542:
We are removing the Tensor and Hamiltonian classes, so we should remove their (minimal) remaining references in the Lightning python code. 

Blocks running CI for the [PL PR #6548](https://github.com/PennyLaneAI/pennylane/pull/6548) to remove legacy_opmath, since every test that uses `lightning.qubit` fails with Tensor removed.


maliasadi (approved):
Awesome! Thanks @lillian542 :partying_face:

multiphaseCFD (commented):
Thanks @lillian542 . Awesome. Just wondering if same action is required for `lightning_tensor`?

lillian542:
@multiphaseCFD thanks, good question! It doesn't import or any custom code for Hamiltonian or Tensor, so I had discounted it, but I went back and checked and "Hamiltonian" was still on the list of supported ops, which it preferably shouldn't be. I've updated that now; as far as I can see, nothing else in the code for that device is specific to the legacy opmath.

multiphaseCFD:
> @multiphaseCFD thanks, good question! It doesn't import or any custom code for Hamiltonian or Tensor, so I had discounted it, but I went back and checked and "Hamiltonian" was still on the list of supported ops, which it preferably shouldn't be. I've updated that now; as far as I can see, nothing else in the code for that device is specific to the legacy opmath.

Thanks @lillian542 . Could you please also add the `ci:use-gpu-runner` label to this PR and then trigger CIs? Then changed lines in `lightning.gpu` and `lightning.tensor` can be tested.

AmintorDusko (commented):
@maliasadi and @lillian542, we have some C++ tests that emulate the data structure of a Tensor and/or Hamiltonian. Are we renaming, updating or removing these tests? Or are we going to leave it for now?
Lilian, please refresh my mind, what will be substituting these in the new opmath?

lillian542:
@multiphaseCFD if we need to make changes in the C++ backend, would it be alight for that to happen in a different PR? Since PennyLane runs tests with `lightning.qubit`, the CI for the PennyLane PR that removes legacy operator arithmetic is blocked by this.

The `Tensor` class will now be fully replaced by `Prod`. The default has been `Prod` for a few months, but we have continued to support `Tensor` if users opt in.

The `Hamiltonian` class will now be replaced by `LinearCombination` and `Sum` (which one depends on how the user creates the operator). `LinearCombination` is a subclass of `Sum` with some minimal, generally user-facing differences.  Like with `Tensor`, this behaviour has been the default for a few months, and only users who 'opted out' of the new behaviour have still been using the legacy opmath support.

maliasadi:
> we have some C++ tests that emulate the data structure of a Tensor and/or Hamiltonian. Are we renaming, updating or removing these tests? Or are we going to leave it for now?

@AmintorDusko @multiphaseCFD @lillian542 We don't need any changes in C++ right now as Lightning and Catalyst continue serializing `qml.pauli.PauliWord` --> `TensorProduct` and  `qml.pauli.PauliSentence` --> `Hamiltonian`.
We will simplify [the observable serialization implementation](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/_serialize.py#L322) and extend the implementation of C++ observables to adhere the new PL obs abstractions in both Lightning and Catalyst.

multiphaseCFD (approved):
Thanks @lillian542 for the nice work! Happy to approve!

PR #999, opened 2024-11-19
Remove SVDynamic

rauletorresc:
**Context:** When the LQ plugin was migrated from Catalyst to Lightning, the `StateVectorLQDynamic` class was also migrated. Given the fact that we have `StateVectorLQManaged`, we can get rid of `StateVectorLQDynamic`.

**Description of the Change:** Replace the uses of  `StateVectorLQDynamic` with  `StateVectorLQManaged`. I basically reused the code from the LK simulator and removed the calls to the `StateVectorLQDynamic` related methods. Of course, I had to use `std::complex` instead of `kokkos::complex`.

**Benefits:** Using the same underlying class for LG, LK and LQ.

**NOTE:** I have deleted the only two tests that were failing:

1. **Qubit allocatation and deallocation** (https://github.com/PennyLaneAI/pennylane-lightning/pull/999/files#diff-4782c8d1aa88463716419ff1ef4f9f551d6241e43f5bc107d27ff0ed4299a3ebL78): the second call to `State(...)` showed a size of 2 for the given state and a size of 4 for the device state, which made the assertion fail.
2. **QuantumDevice object test** (https://github.com/PennyLaneAI/pennylane-lightning/pull/999/files#diff-4782c8d1aa88463716419ff1ef4f9f551d6241e43f5bc107d27ff0ed4299a3ebL193) The test was killed after 1 minute of running, probably because an infinite loop/recursion?

It seems to me these two tests do not comply with the way `StateVectorLQManaged` is structured but let the discussion resolve it.

[sc-77689]


.github/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 * Add a Catalyst-specific wrapping class for Lightning Qubit.
```
maliasadi, line 8:
```suggestion
* Add a Catalyst-specific wrapping class for Lightning-Qubit.
```

maliasadi (approved):
Nice work @rauletorresc! :partying_face:

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -22,8 +22,29 @@
 namespace Catalyst::Runtime::Simulator {
 
 auto LightningSimulator::AllocateQubit() -> QubitIdType {
```
dime10, line 24: @maliasadi Are we keeping this functionality in then, rather than only allowing a single allocation & deallocation?

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -45,14 +66,11 @@ auto LightningSimulator::AllocateQubits(size_t num_qubits)
...
 void LightningSimulator::ReleaseAllQubits() {
-    this->device_sv->clearData();
     this->qubit_manager.ReleaseAll();
+    this->device_sv = std::make_unique<StateVectorT>(0); // reset the device
 }
 
 void LightningSimulator::ReleaseQubit(QubitIdType q) {
-    if (this->qubit_manager.isValidQubitId(q)) {
-        this->device_sv->releaseWire(this->qubit_manager.getDeviceId(q));
-    }
```
dime10, line 55: If the answer to the above is yes, does the removal of this block break things, or will the statevector merely carry around a "logically dead" qubit that will never be addressed again but still needs to be updated by gate kernels?

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
maliasadi: There aren't any lowering to these function calls. These methods can be removed from the API if the dynamic qubit management will not be added in a foreseeable future.

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
maliasadi: @dime10 I don't understand your question.

PR #1001, opened 2024-11-20
Add support for Python 3.13 wheels

mlxd:
### Before submitting

**Context:** Add support for building Python 3.13 wheels.

**Benefits:** Support Python 3.13.

**Possible Drawbacks:** Additional CI overheads. 
        Since PennyLane requires Lightning to be installed, we cannot test PennyLane 3.13 until a Lightning wheel is available.
        As PennyLane also pins numpy<2.1, this pin needs to be relaxed as 2.1 is the first version to support Python 3.13. This will likely require a "fail-and-merge" process to get the wheels out for Lightning initially.
UPDATE: PennyLane now unpins the Numpy version, so this should be unblocked.
**Related GitHub Issues:**


mlxd:
[sc-78819]

mlxd:
To avoid any massive CI hit, all Changelog entries will be a final addition.

AmintorDusko (approved):
Looks good!

maliasadi (approved):
Thanks @mlxd! Could you also update `classifiers` in `pyproject.toml`?

.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           python-version: '3.10'
 
       - name: Install dependencies
-        run: python3.10 -m pip install cibuildwheel~=2.20.0 tomlkit
+        run: python3.10 -m pip install cibuildwheel~=2.21.0 tomlkit
```
maliasadi, line 140: Could we also remove this pinned version across our wheels recipes? Any concerns with using the [latest_release](https://github.com/pypa/cibuildwheel/releases) version of `cibuildwheel`?

maliasadi:
The issue with ARM::CUDA wheels will be fixed in https://github.com/PennyLaneAI/pennylane-lightning/pull/1105

.github/workflows/wheel_linux_aarch64.yml
mlxd: I think I'd rather we be explicit about x.y versions, but be unbounded in the bugfix x.y.z version.

AmintorDusko:
I merged the master here to manage the use of CI runners. Please double-check it and merge if it is sensible. Thanks!

mlxd:
Thanks @AmintorDusko 
I needed to remove jax tests from anything hitting 313 due to the unavailability of wheels. We can fix the install and setup of deps when jax is upgraded in Q2.

AmintorDusko:
> Thanks @AmintorDusko I needed to remove jax tests from anything hitting 313 due to the unavailability of wheels. We can fix the install and setup of deps when jax is upgraded in Q2.

That sounds good to me, as this is a subcase of Python 3.13 installation, and this version is quite new (in terms of Python releases).

PR #1003, opened 2024-11-22
Generalize seeding for measurements

rauletorresc:
**Context:** Currently, only `catalyst.measure()`, `qml.sample()` and `qml.counts()` can be seeded with qjit(seed=N). It would be nice to be able to seed all circuit readout methods, like `probs()`, `expval()`, `var()`, when executing a device with shots.

**Description of the Change:** Let the base measurement class store the device seed so it can be propagated when necessary to newly created derived measurements. Move the logic from the simulator to the base measurement class.

**Benefits:** All measurements using the `generate_samples` method will be seeded.

Measures to test locally:

- [x] `expval`
- [x] `var`
- [x] `probs`
- [x] `sample`
- [x] `counts`

Devices to modify

- [x] LQ
- [x] LK
- [x] LG

TODO

- [x] Simplify design
- [x] Document changes
- [x] Test probs using the LQ, LK and LG plugins
- [x] Test expval using the LQ, LK and LG plugins
- [x] Test var using the LQ, LK and LG plugins
- [x] Update changelog

[sc-77124]

### DIFF
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -57,7 +57,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
-    std::optional<std::size_t> DeviceSeed = std::nullopt;
+    const std::optional<std::size_t> &DeviceSeed = std::nullopt;
 
     std::mt19937 rng;
 
@@ -275,8 +275,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(
-                static_cast<Derived *>(this)->get_DeviceSeed());
+            measure.set_DeviceSeed(this->DeviceSeed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -285,8 +284,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(
-                static_cast<Derived *>(this)->get_DeviceSeed());
+            measure.set_DeviceSeed(this->DeviceSeed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -379,7 +377,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
      */
     auto sample(const std::size_t &num_shots) -> std::vector<std::size_t> {
         Derived measure(_statevector);
-        measure.set_DeviceSeed(static_cast<Derived *>(this)->get_DeviceSeed());
+        measure.set_DeviceSeed(this->DeviceSeed);
         return measure.generate_samples(num_shots);
     }
 
@@ -446,10 +444,6 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         this->DeviceSeed = deviceSeed;
     }
 
-    const std::optional<std::size_t> &get_DeviceSeed() {
-        return this->DeviceSeed;
-    }
-
     void set_DeviceSeedFromPRNG(std::mt19937 *gen) {
         if (gen != nullptr) {
             this->set_DeviceSeed((*gen)());
@@ -510,15 +504,13 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(
-                static_cast<Derived *>(this)->get_DeviceSeed());
+            measure.set_DeviceSeed(this->DeviceSeed);
             samples = measure.generate_samples(num_shots);
         } else {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(
-                static_cast<Derived *>(this)->get_DeviceSeed());
+            measure.set_DeviceSeed(this->DeviceSeed);
             samples = measure.generate_samples(num_shots);
         }
 
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -57,7 +57,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
-    const std::optional<std::size_t> &DeviceSeed = std::nullopt;
+    std::optional<std::size_t> DeviceSeed = std::nullopt;
 
     std::mt19937 rng;
 
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -275,7 +275,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(this->DeviceSeed);
+            measure.setDeviceSeed(this->DeviceSeed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -284,7 +284,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(this->DeviceSeed);
+            measure.setDeviceSeed(this->DeviceSeed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -377,7 +377,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
      */
     auto sample(const std::size_t &num_shots) -> std::vector<std::size_t> {
         Derived measure(_statevector);
-        measure.set_DeviceSeed(this->DeviceSeed);
+        measure.setDeviceSeed(this->DeviceSeed);
         return measure.generate_samples(num_shots);
     }
 
@@ -439,18 +439,18 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         return outcome_map;
     }
 
-    void set_DeviceSeed(
-        const std::optional<std::size_t> &deviceSeed = std::nullopt) {
+    void
+    setDeviceSeed(const std::optional<std::size_t> &deviceSeed = std::nullopt) {
         this->DeviceSeed = deviceSeed;
     }
 
-    void set_DeviceSeedFromPRNG(std::mt19937 *gen) {
+    void setDeviceSeedFromPRNG(std::mt19937 *gen) {
         if (gen != nullptr) {
-            this->set_DeviceSeed((*gen)());
+            this->setDeviceSeed((*gen)());
         }
     }
 
-    void activate_DeviceSeed() {
+    void activateDeviceSeed() {
         if (this->DeviceSeed.has_value()) {
             setSeed(this->DeviceSeed.value());
         } else {
@@ -504,13 +504,13 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(this->DeviceSeed);
+            measure.setDeviceSeed(this->DeviceSeed);
             samples = measure.generate_samples(num_shots);
         } else {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.set_DeviceSeed(this->DeviceSeed);
+            measure.setDeviceSeed(this->DeviceSeed);
             samples = measure.generate_samples(num_shots);
         }
 
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -17,6 +17,7 @@
  */
 #pragma once
 
+#include <optional>
 #include <random>
 #include <string>
 #include <vector>
@@ -56,6 +57,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
+    std::optional<std::size_t> DeviceSeed = std::nullopt;
+
     std::mt19937 rng;
 
   public:
@@ -66,22 +69,6 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
     explicit MeasurementsBase(const StateVectorT &statevector)
         : _statevector{statevector} {};
 #endif
-
-    /**
-     * @brief Set the seed of the internal random generator
-     *
-     * @param seed Seed
-     */
-    void setSeed(const std::size_t seed) { rng.seed(seed); }
-
-    /**
-     * @brief Randomly set the seed of the internal random generator
-     *
-     */
-    void setRandomSeed() {
-        std::random_device rd;
-        setSeed(rd());
-    }
 
     /**
      * @brief Calculate the expectation value for a general Observable.
@@ -454,7 +441,46 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         return outcome_map;
     }
 
+    void set_DeviceSeed(
+        const std::optional<std::size_t> &deviceSeed = std::nullopt) {
+        this->DeviceSeed = deviceSeed;
+    }
+
+    const std::optional<std::size_t> &get_DeviceSeed() {
+        return this->DeviceSeed;
+    }
+
+    void set_DeviceSeedFromPRNG(std::mt19937 *gen) {
+        if (gen != nullptr) {
+            this->set_DeviceSeed((*gen)());
+        }
+    }
+
+    void activate_DeviceSeed() {
+        if (this->DeviceSeed.has_value()) {
+            setSeed(this->DeviceSeed.value());
+        } else {
+            setRandomSeed();
+        }
+    }
+
   private:
+    /**
+     * @brief Set the seed of the internal random generator
+     *
+     * @param seed Seed
+     */
+    void setSeed(const std::size_t seed) { rng.seed(seed); }
+
+    /**
+     * @brief Randomly set the seed of the internal random generator
+     *
+     */
+    void setRandomSeed() {
+        std::random_device rd;
+        setSeed(rd());
+    }
+
     /**
      * @brief Return samples of a observable
      *
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -17,6 +17,7 @@
  */
 #pragma once
 
+#include <cstddef>
 #include <optional>
 #include <random>
 #include <string>
@@ -451,30 +452,21 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
     }
 
     void activateDeviceSeed() {
-        if (this->DeviceSeed.has_value()) {
-            setSeed(this->DeviceSeed.value());
-        } else {
-            setRandomSeed();
-        }
+        std::size_t seed = this->DeviceSeed.has_value()
+                               ? this->DeviceSeed.value()
+                               : this->generateRandomSeed();
+        rng.seed(seed);
     }
 
   private:
     /**
-     * @brief Set the seed of the internal random generator
-     *
-     * @param seed Seed
-     */
-    void setSeed(const std::size_t seed) { rng.seed(seed); }
-
-    /**
-     * @brief Randomly set the seed of the internal random generator
+     * @brief Generate a random seed
      *
      */
-    void setRandomSeed() {
+    std::size_t generateRandomSeed() {
         std::random_device rd;
-        setSeed(rd());
+        return rd();
     }
-
     /**
      * @brief Return samples of a observable
      *
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -17,7 +17,6 @@
  */
 #pragma once
 
-#include <cstddef>
 #include <optional>
 #include <random>
 #include <string>
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -288,6 +288,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
+            measure.set_DeviceSeed(
+                static_cast<Derived *>(this)->get_DeviceSeed());
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -296,6 +298,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
+            measure.set_DeviceSeed(
+                static_cast<Derived *>(this)->get_DeviceSeed());
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -388,8 +392,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
      */
     auto sample(const std::size_t &num_shots) -> std::vector<std::size_t> {
         Derived measure(_statevector);
-        return measure.generate_samples(num_shots,
-                                        static_cast<Derived *>(this));
+        measure.set_DeviceSeed(static_cast<Derived *>(this)->get_DeviceSeed());
+        return measure.generate_samples(num_shots);
     }
 
     /**
@@ -480,14 +484,16 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            samples = measure.generate_samples(num_shots,
-                                               static_cast<Derived *>(this));
+            measure.set_DeviceSeed(
+                static_cast<Derived *>(this)->get_DeviceSeed());
+            samples = measure.generate_samples(num_shots);
         } else {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            samples = measure.generate_samples(num_shots,
-                                               static_cast<Derived *>(this));
+            measure.set_DeviceSeed(
+                static_cast<Derived *>(this)->get_DeviceSeed());
+            samples = measure.generate_samples(num_shots);
         }
 
         if (!shot_range.empty()) {
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -94,14 +94,6 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
     }
 
     /**
-     * @brief Copy a seed for the internal random generator
-     *
-     */
-    void copySeed(const std::optional<std::size_t> &deviceSeed = std::nullopt) {
-        this->_deviceseed = deviceSeed;
-    }
-
-    /**
      * @brief Calculate the expectation value for a general Observable.
      *
      * @param obs Observable.
@@ -306,7 +298,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.copySeed(this->_deviceseed);
+            measure.setSeed(this->_deviceseed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -315,7 +307,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
-            measure.copySeed(this->_deviceseed);
+            measure.setSeed(this->_deviceseed);
             if (num_shots > std::size_t{0}) {
                 return measure.probs(obs_wires, num_shots);
             }
@@ -408,7 +400,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
      */
     auto sample(const std::size_t &num_shots) -> std::vector<std::size_t> {
         Derived measure(_statevector);
-        measure.copySeed(this->_deviceseed);
+        measure.setSeed(this->_deviceseed);
         return measure.generate_samples(num_shots);
     }
 
@@ -500,13 +492,13 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.copySeed(this->_deviceseed);
+            measure.setSeed(this->_deviceseed);
             samples = measure.generate_samples(num_shots);
         } else {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            measure.copySeed(this->_deviceseed);
+            measure.setSeed(this->_deviceseed);
             samples = measure.generate_samples(num_shots);
         }
 
--- a/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
+++ b/pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
@@ -57,9 +57,9 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
-    std::optional<std::size_t> DeviceSeed = std::nullopt;
+    std::optional<std::size_t> _deviceseed{std::nullopt};
 
-    std::mt19937 rng;
+    std::mt19937 _rng;
 
   public:
 #ifdef _ENABLE_PLGPU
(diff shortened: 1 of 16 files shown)


dime10 (commented):
This looks like a clean solution to me! 💯  Have you tested that the randomness is now completely fixed for all measurements (e.g. decimals do not vary in an expval etc)?

rauletorresc:
> This looks like a clean solution to me! 💯 Have you tested that the randomness is now completely fixed for all measurements (e.g. decimals do not vary in an expval etc)?

Sorry to ask, but how can I test this :)

paul0403:
> > This looks like a clean solution to me! 💯 Have you tested that the randomness is now completely fixed for all measurements (e.g. decimals do not vary in an expval etc)?
> 
> Sorry to ask, but how can I test this :)

https://github.com/PennyLaneAI/catalyst/blob/main/frontend/test/pytest/test_seeded_qjit.py

What I would do is on your local branch you can build catalyst first (make all ), this installs whatever lightning specified in wherever, then you can build your own lightning branch locally so it overwrites the previous installation

dime10:
> > > This looks like a clean solution to me! 💯 Have you tested that the randomness is now completely fixed for all measurements (e.g. decimals do not vary in an expval etc)?
> > 
> > 
> > Sorry to ask, but how can I test this :)
> 
> https://github.com/PennyLaneAI/catalyst/blob/main/frontend/test/pytest/test_seeded_qjit.py
> 
> What I would do is on your local branch you can build catalyst first (make all ), this installs whatever lightning specified in wherever, then you can build your own lightning branch locally so it overwrites the previous installation

In addition to running those tests for existing functionality, we should check that something like

```py
@qml.qnode(qml.device("lightning.qubit", wires=1, shots=100)
def circuit():
    qml.RY(0.3, wires=0)
```

produces the same decimals with repeated runs if seeded.

paul0403:
Acatully, I just remembered you can test this directly in lightning-catalyst, and no need to go through frontend https://github.com/PennyLaneAI/pennylane-lightning/pull/927/files#diff-934a38a9af759b117f260f29c5b2732bee701b7cd3b49a4bdf22da7f69973ad8

rauletorresc:
@dime10 @paul0403 after the last changes, `expval` does produce the same decimals with repeated runs if seeded. I'll perform similar checks with the rest of the measures. Keep you posted.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -479,12 +480,14 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
             Derived measure(sv);
-            samples = measure.generate_samples(num_shots);
+            samples = measure.generate_samples(num_shots,
+                                               static_cast<Derived *>(this));
         } else {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenValues, obs_wires);
             Derived measure(sv);
-            samples = measure.generate_samples(num_shots);
+            samples = measure.generate_samples(num_shots,
```
dime10, line 489: What about calling `set_PRNG` on `Derived` here rather than passing down `this`?

paul0403:
Can we have tests similar to https://github.com/PennyLaneAI/pennylane-lightning/pull/927/files#diff-934a38a9af759b117f260f29c5b2732bee701b7cd3b49a4bdf22da7f69973ad8? So tests with the lightning-catalyst interface, not just the lightning base classes.

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -580,11 +582,14 @@ class Measurements final
      */
     std::vector<std::size_t>
     generate_samples(const std::size_t num_samples,
-                     const std::optional<std::size_t> &seed = std::nullopt) {
+                     const std::optional<Measurements *> &m = std::nullopt) {
```
paul0403, line 585: Is it possible to not pass the entire measurement class into itself here?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
dime10: I'm not sure about your question, someone from perf would know how to answer it. 

> Calling set_PRNG is assuming all derived classes have implemented that method.

What if the base class implemented those methods?

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
dime10: Duplicate of my comment above

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: That would be the simplest solution if there is no problem with being explicit about the usage of PRNGs for generating samples at the base class. @maliasadi what do you think?

rauletorresc: Checking the measurement base class, I realized there exists a PRNG for the internal use of the measure. Adding a second PRNG would be confusing. So, I have simplified the design and added a device seed to the base measurement class. This member variable serves as the seed of the base class when provided. If not provided, a random seed is used. Therefore, there is no need anymore to have a copy of the PRNG in the derived measure class. With the new design, all methods calling the sample generation routine will take into account the base class seed. The simulator is the one setting the measure seed using its own PRNG right after creating it the measure. The sample generation routine in the derived measure class is the one activating the seed.

rauletorresc: And we are no longer passing the whole derived measurement class as an argument :)

pennylane_lightning/core/src/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
rauletorresc: See reply above

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -288,6 +275,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
+            measure.set_DeviceSeed(this->DeviceSeed);
```
paul0403, line 278: I think lightning C++ is all CamelCase, so words are chained via capitalized letters and overall first letter is capitalized for nouns and not capitalized for verbs, i.e.
`MyCustomObject`, `StateVector`; `doSomething`, `applyInPlaceShots`

paul0403:
```suggestion
            measure.setDeviceSeed(this->DeviceSeed);
```

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -296,6 +284,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             StateVectorT sv(_statevector);
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
+            measure.set_DeviceSeed(this->DeviceSeed);
```
paul0403, line 287:
```suggestion
            measure.setDeviceSeed(this->DeviceSeed);
```

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -388,6 +377,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
      */
     auto sample(const std::size_t &num_shots) -> std::vector<std::size_t> {
         Derived measure(_statevector);
+        measure.set_DeviceSeed(this->DeviceSeed);
```
paul0403, line 380:
```suggestion
        measure.setDeviceSeed(this->DeviceSeed);
```

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,42 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+        }
+    }
+
   private:
+    /**
+     * @brief Set the seed of the internal random generator
+     *
+     * @param seed Seed
+     */
+    void setSeed(const std::size_t seed) { rng.seed(seed); }
```
paul0403, line 467: Is there a reason why we don't use the existing seed setting methods? It is a bit confusing to have two suites of seed setters here. 

What about this: we update the existing `setSeed` to take in an optional seed integer, and depending on the option we either `rng.seed(seed)` or `std::random_device rd; rng.seed(rd())`?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,42 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+        this->DeviceSeed = deviceSeed;
+    }
+
+    void set_DeviceSeedFromPRNG(std::mt19937 *gen) {
+        if (gen != nullptr) {
+            this->set_DeviceSeed((*gen)());
+        }
+    }
+
+    void activate_DeviceSeed() {
```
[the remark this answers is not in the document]
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: You are right. I have replaced them like this:
```
void activateDeviceSeed() {
        std::size_t seed = this->DeviceSeed.has_value()
                               ? this->DeviceSeed.value()
                               : this->generateRandomSeed();
        rng.seed(seed);
    }

  private:
    /**
     * @brief Generate a random seed
     *
     */
    std::size_t generateRandomSeed() {
        std::random_device rd;
        return rd();
    }
```
What do you think?

rauletorresc: The reason I created a member variable with the seed info is because we need to be able to propagate the seed to newly created derived measures. E.g.:
```
// we are inside the base measure class, which already has a seed
Derived measure(_statevector);
measure.setDeviceSeed(this->DeviceSeed);
return measure.generate_samples(num_shots);
```
`setDeviceSeed` propagates the seed. `measure.generate_samples(...)` will actually apply it to the base rng.

rauletorresc: We set the value separately from applying it to the rng because not all the simulators will actually use the rng. LQ and LG do use it, but LK only uses the seed, it does not use the rng, it has its own way to generate samples.

rauletorresc: It did not change drastically. Instead of taking the seed from an argument, it takes it from the `DeviceSeed `member variable. This way, the LK simulator (in fact any simulator) does not need to check if the PRNG is set or not and choose which `generate_samples `signature to call. Instead, it calls the `generate_samples` method takes care of this logic.

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1831,48 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        sims[0] = std::make_unique<LQSimulator>();
+        sims[1] = std::make_unique<LQSimulator>();
+
+        for (std::size_t sim_idx = 0; sim_idx < sims.size(); sim_idx++) {
+            circuit(*(sims[sim_idx]), views[sim_idx], gens[sim_idx]);
+        }
+
+        for (std::size_t i = 0; i < probs[0].size(); i++) {
+            if (i == 4 || i == 6 || i == 12 || i == 14) {
+                CHECK(probs[0][i] == Approx(0.25).margin(1e-5));
```
paul0403, line 1871: I think here it suffices to check that the two runs of the circuit with the same seed gives the same result, and the actual numerical values do not need to be checked.

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1831,48 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+            } else {
+                CHECK(probs[0][i] == Approx(0.).margin(1e-5));
```
paul0403, line 1873: Also I feel a bit concerned here on whether the seeding even took effect, or is it just that seeding is bypassed altogether and we are returning the analytical probs value (aka zero shots). 

Can you add a shots to the test? I think with shots, `probs` result was always going to stay the same.

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1831,48 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+    std::vector<std::vector<double>> probs(2, std::vector<double>(16));
+
+    std::vector<DataView<double, 1>> views{DataView<double, 1>(probs[0]),
+                                           DataView<double, 1>(probs[1])};
+
+    std::vector<std::mt19937> gens{std::mt19937{37}, std::mt19937{37}};
+
+    auto circuit = [](LQSimulator &sim, DataView<double, 1> &view,
+                      std::mt19937 &gen) {
+        sim.SetDevicePRNG(&gen);
```
paul0403, line 1846:
```suggestion
        sim.SetDevicePRNG(&gen);
        sim.SetDeviceShots(1000);
```

paul0403: see my other comment

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Done

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Done

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,33 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         return outcome_map;
     }
 
+    void
+    setDeviceSeed(const std::optional<std::size_t> &deviceSeed = std::nullopt) {
+        this->DeviceSeed = deviceSeed;
+    }
+
+    void setDeviceSeedFromPRNG(std::mt19937 *gen) {
```
paul0403, line 447: I am still a bit uncomfortable with having two seed setter suites. As far as I can tell, the purpose of this second suite is:
- A measurement, say `LightningSimulator::Probs()` in catalyst, would call this `setDeviceSeedFromPRNG` on the measurement class. This adds a seed number to the measurement class.
- When the measurement class actually performs the measurements (e.g. in generate sameples), we `activateDeviceSeed` to produce the seeded innate prng object of the measurement class.

I still think this logic can be performed inside the two existing setters. The goal of the entire pipeline here is step 2, aka perform `rng.seed(seed)`. I think the catalyst measurements can just pass the seed number into the measurement class without passing the catalyst device prng object. This way we can just use the old `setSeed(size_t)` methods. In fact, there only needs to be one setseed method that handles everything:
```
void setSeed(optional int seed){
   if seed:
      rng.seed(seed);
      this.DeviceSeed = seed;
   else:
      rng.seed(rd());
}
``` 

And it can handle the various callers, whether giving a seed or not.

I'm fine with the measurement base saving the seed number for utility, e.g. using the seed number it saw for LK's special `generate_samples`.

What do you think? I think this would be a lot cleaner.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -56,6 +57,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
+    std::optional<std::size_t> DeviceSeed = std::nullopt;
```
maliasadi, line 60: For consistency with the rest of class variables:
```suggestion
    std::optional<std::size_t> _deviceseed{std::nullopt};
```

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,33 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+    void
+    setDeviceSeed(const std::optional<std::size_t> &deviceSeed = std::nullopt) {
+        this->DeviceSeed = deviceSeed;
+    }
+
+    void setDeviceSeedFromPRNG(std::mt19937 *gen) {
+        if (gen != nullptr) {
+            this->setDeviceSeed((*gen)());
+        }
+    }
```
maliasadi, line 451: Do we really need this method as part of the API? It seems redundant to `setDeviceSeed` as we also need to pass the `gen`. Removing this method in favour of using `setDeviceSeed((*gen)())` directly would avoid a function call and the additional if-condition check. What do you think?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,33 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+
+    void activateDeviceSeed() {
+        std::size_t seed = this->DeviceSeed.has_value()
```
maliasadi, line 454: it's better to specify it with `const`

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,33 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
         return outcome_map;
     }
 
+    void
+    setDeviceSeed(const std::optional<std::size_t> &deviceSeed = std::nullopt) {
```
maliasadi, line 443: could you add docstrings to the new methods?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -449,7 +439,33 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
...
+                               : this->generateRandomSeed();
+        rng.seed(seed);
+    }
+
   private:
+    /**
+     * @brief Generate a random seed
+     *
+     */
+    std::size_t generateRandomSeed() {
```
maliasadi, line 465: Lightning follows the tailing style for return data-types:
```suggestion
    auto generateRandomSeed() -> std::size_t {
```

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -56,6 +57,8 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
+    std::optional<std::size_t> DeviceSeed = std::nullopt;
+
     std::mt19937 rng;
```
maliasadi, line 62:
```suggestion
    std::mt19937 _rng;
```

pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
```diff
@@ -1257,6 +1257,7 @@ void testSamples(const std::optional<std::size_t> &seed = std::nullopt) {
         // This object attaches to the statevector allowing several
         // measurements.
         Measurements<StateVectorT> Measurer(statevector);
+        Measurer.setDeviceSeed(seed);
```
maliasadi, line 1260: Where is `seed` defined?

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
maliasadi, line 465: Could we also remove this method from the API? We can directly use leverage the random-device in `activateDeviceSeed`.

maliasadi (commented):
Nice work @rauletorresc! Happy to review the complete version of the support

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Removed

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Removed

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Done

pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp
rauletorresc: It comes as an argument of the `testSamples` function

rauletorresc: https://github.com/PennyLaneAI/pennylane-lightning/blob/f419d371810400b7cb8c6f3b8cd7cf959beb595b/pennylane_lightning/core/src/measurements/tests/Test_MeasurementsBase.cpp#L1307

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Removed

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
rauletorresc: Removed

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -56,7 +57,9 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
 #else
     const StateVectorT &_statevector;
 #endif
-    std::mt19937 rng;
+    std::optional<std::size_t> _deviceseed{std::nullopt};
```
paul0403, line 60:
```suggestion
    std::optional<std::size_t> _deviceSeed{std::nullopt};
```

paul0403: CamelCase, first letter upper case for nouns

paul0403: Oh, this came from Ali's suggestion. Ignore this then!

paul0403:
Looks good! Thanks @rauletorresc for working on this! 💯 

As a final sanity check, can you maybe build against the lightning in this branch, and do some catalyst frontend tests like https://github.com/PennyLaneAI/catalyst/blob/main/frontend/test/pytest/test_seeded_qjit.py?

For example, 
```python
    dev = qml.device("lightning.qubit", wires=2, shots=100)

    @qjit(seed=37)
    def workflow():
        @qml.qnode(dev)
        def circuit():
            qml.Hadamard(wires=[0])
            qml.RX(12.34, wires=[1])
            return qml.probs()

        return circuit(), circuit(), circuit(), circuit()

   trail1 = workflow()  # [probs10, probs11, probs12, probs13]
   trail2 = workflow() # [probs20, probs21, probs22, probs23]

   for (probs1i, probs2i) in zip(trail1, trail2):
      assert probs1i == probss2i // or whatever the equivalent pytest syntax is
```

Note that within the same `workflow()` call the four `circuit()` calls should give different results, i.e. the four `probs1*` and the four `probs2*` should be different among themselves.

(This should also be a PR in catalyst, after this lightning PR is merged)

rauletorresc:
> Looks good! Thanks @rauletorresc for working on this! 💯
> 
> As a final sanity check, can you maybe build against the lightning in this branch, and do some catalyst frontend tests like https://github.com/PennyLaneAI/catalyst/blob/main/frontend/test/pytest/test_seeded_qjit.py?
> 
> For example,
> 
> ```python
>     dev = qml.device("lightning.qubit", wires=2, shots=100)
> 
>     @qjit(seed=37)
>     def workflow():
>         @qml.qnode(dev)
>         def circuit():
>             qml.Hadamard(wires=[0])
>             qml.RX(12.34, wires=[1])
>             return qml.probs()
> 
>         return circuit(), circuit(), circuit(), circuit()
> 
>    trail1 = workflow()  # [probs10, probs11, probs12, probs13]
>    trail2 = workflow()  # [probs20, probs21, probs22, probs23]
> 
>    for (probs1i, probs2i) in zip(trail1, trail2):
>       assert probs1i == probs2i // or whatever the equivalent pytest syntax is
> ```
> 
> Note that within the same `workflow()` call the four `circuit()` calls should give different results, i.e. the four `probs1*` and the four `probs2*` should be different among themselves.
> 
> (This should also be a PR in catalyst, after this lightning PR is merged)

Something like that I am doing at the Catalyst side here: https://github.com/PennyLaneAI/catalyst/pull/1344
It works locally, I hope it will work in the CI once this PR gets merged.

pennylane_lightning/core/src/measurements/MeasurementsBase.hpp
```diff
@@ -288,6 +298,7 @@ template <class StateVectorT, class Derived> class MeasurementsBase {
             sv.updateData(data_storage.data(), data_storage.size());
             obs.applyInPlaceShots(sv, eigenvalues, obs_wires);
             Derived measure(sv);
+            measure.setSeed(this->_deviceseed);
```
paul0403, line 301: Isn't it circular here? i.e. the `setSeed` method will record the received seed into `this->_deviceseed`, but here the setter argument is also `this->_deviceseed`. I know in the catalyst path the measurement object's `_deviceseed` is set from the catalyst rng's produced number, but what sets the seed for the non-catalyst path? Because here it seems like the seed will always be the initialized value of `_deviceseed` when `measure` is created (which is nullopt).

rauletorresc: `this->_deviceseed` contains the seed given by the simulator. When we create the new Derived measure here, the seed is `nullopt `for that new measure. Hence, we need to set it with the current value of the seed. If we don't do this here, `expval`, `var` and `probs` will be unseeded.

paul0403: Ahh, the link I was missing was that the `Derived measure` is a separately created object whose seed need to be set again. Thanks!

paul0403 (approved):
:100: :partying_face: 

All good from my side! I'm unfamiliar with the lightning codecov so maybe someone from performance can suggest there.

pennylane_lightning/core/src/simulators/lightning_gpu/catalyst/tests/Test_LightningGPUMeasures.cpp
```diff
@@ -1823,3 +1823,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < probs[0].size(); i++) {
+            CHECK((probs[0][i] == probs[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Var with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1870:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_gpu/catalyst/tests/Test_LightningGPUMeasures.cpp
```diff
@@ -1823,3 +1823,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < vars[0].size(); i++) {
+            CHECK((vars[0][i] == vars[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Expval with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1916:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1822,3 +1822,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
         }
     }
 }
+
+TEST_CASE("Probs with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1827:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1822,3 +1822,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < probs[0].size(); i++) {
+            CHECK((probs[0][i] == probs[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Var with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1869:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
```diff
@@ -1822,3 +1822,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < vars[0].size(); i++) {
+            CHECK((vars[0][i] == vars[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Expval with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1915:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1832,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
         }
     }
 }
+
+TEST_CASE("Probs with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1837:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1832,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < probs[0].size(); i++) {
+            CHECK((probs[0][i] == probs[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Var with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1879:
```suggestion
    constexpr std::size_t shots = 1000;
```

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
```diff
@@ -1831,3 +1832,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
...
+        }
+
+        for (std::size_t i = 0; i < vars[0].size(); i++) {
+            CHECK((vars[0][i] == vars[1][i]));
+        }
+    }
+}
+
+TEST_CASE("Expval with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
maliasadi, line 1925:
```suggestion
    constexpr std::size_t shots = 1000;
```

maliasadi (approved):
Nice work @rauletorresc! Happy to approve :partying_face:

pennylane_lightning/core/src/simulators/lightning_gpu/catalyst/tests/Test_LightningGPUMeasures.cpp
```diff
@@ -1823,3 +1823,137 @@ TEST_CASE("Sample with a seeded device", "[Measures]") {
         }
     }
 }
+
+TEST_CASE("Probs with a seeded device", "[Measures]") {
+    std::size_t shots = 1000;
```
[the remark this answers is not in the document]
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_gpu/catalyst/tests/Test_LightningGPUMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_gpu/catalyst/tests/Test_LightningGPUMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_kokkos/catalyst/tests/Test_LightningKokkosMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Fixed

pennylane_lightning/core/src/simulators/lightning_qubit/catalyst/tests/Test_LightningMeasures.cpp
rauletorresc: Fixed

PR #1004, opened 2024-11-22
Fix Lightning GPU Rot adjoint bug and enable Rot in toml

josephleekl:
### Before submitting

**Context:**
Previously the implementation of Rot and CRot gates (e.g. `Rot(a, b, c)`) with adjoint in LGPU ended up multiplying `Rot(-a, -b, -c)` instead of `Rot(-c, -b, -a)`. The error was not found in tests since the C++ tests used the wrong adjoint rotation matrix, and since `Adjoint(Rot)` and `Adjoint(CRot)` were not in the frozenset, they were not executed in the Python tests. The error is found when adding the invertible capability for Rot and CRot and [the following test](https://github.com/PennyLaneAI/pennylane-lightning/blob/45a67d495552ca9aca3c05467b16b1b6d2e1d4ab/tests/test_apply.py#L1362) fails. 

This PR fixes the implementation and C++ tests. The toml file is also updated to enable `Adjoint(Rot)` and `Adjoint(CRot)` from Python.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-78931]

### DIFF
--- a/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
@@ -403,7 +403,9 @@ class StateVectorCudaMPI final
                 adjoint
                     ? cuGates::getRot<CFP_t>(-params[2], -params[1], -params[0])
                     : cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
-            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, false);
+            // Use default adjoint=false for applyDeviceMatrixGate() as reversed
+            // rot_matrix explicitly used
+            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts);
         } else if (opName == "Matrix") {
             applyDeviceMatrixGate(gate_matrix.data(), ctrls, tgts, adjoint);
         } else if (par_gates_.find(opName) != par_gates_.end()) {
--- a/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
@@ -401,9 +401,9 @@ class StateVectorCudaMPI final
         } else if (opName == "Rot" || opName == "CRot") {
             auto rot_matrix =
                 adjoint
-                    ? cuGates::getRot<CFP_t>(params[2], params[1], params[0])
+                    ? cuGates::getRot<CFP_t>(-params[2], -params[1], -params[0])
                     : cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
-            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, adjoint);
+            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, false);
         } else if (opName == "Matrix") {
             applyDeviceMatrixGate(gate_matrix.data(), ctrls, tgts, adjoint);
         } else if (par_gates_.find(opName) != par_gates_.end()) {
--- a/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
+++ b/pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
@@ -400,12 +400,8 @@ class StateVectorCudaMPI final
                                      adjoint);
         } else if (opName == "Rot" || opName == "CRot") {
             auto rot_matrix =
-                adjoint
-                    ? cuGates::getRot<CFP_t>(-params[2], -params[1], -params[0])
-                    : cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
-            // Use default adjoint=false for applyDeviceMatrixGate() as reversed
-            // rot_matrix explicitly used
-            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts);
+                cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
+            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, adjoint);
         } else if (opName == "Matrix") {
             applyDeviceMatrixGate(gate_matrix.data(), ctrls, tgts, adjoint);
         } else if (par_gates_.find(opName) != par_gates_.end()) {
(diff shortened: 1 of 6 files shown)


LuisAlfredoNu:
Do you think that we need to change something in the Python test to prevent this error again? :thinking:

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
```diff
@@ -401,9 +401,9 @@ class StateVectorCudaMPI final
         } else if (opName == "Rot" || opName == "CRot") {
             auto rot_matrix =
                 adjoint
-                    ? cuGates::getRot<CFP_t>(params[2], params[1], params[0])
+                    ? cuGates::getRot<CFP_t>(-params[2], -params[1], -params[0])
                     : cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
-            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, adjoint);
+            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, false);
```
maliasadi, line 406: it seems a bit odd to always pass through `adjoint=false` in  `applyDeviceMatrixGate`! But I guess there's no better way if we want to pass a reversed rot matrix to this method. Could we fix this by adding the matrix evaluation (line 402) within `applyDeviceMatrixGate` instead?

pennylane_lightning/lightning_gpu/lightning_gpu.toml
```diff
@@ -32,7 +32,7 @@ PhaseShift             = { properties = [ "invertible", "controllable", "differe
 RX                     = { properties = [ "invertible", "controllable", "differentiable" ] }
 RY                     = { properties = [ "invertible", "controllable", "differentiable" ] }
 RZ                     = { properties = [ "invertible", "controllable", "differentiable" ] }
-Rot                    = { properties = [               "controllable", "differentiable" ] }
+Rot                    = { properties = [ "invertible", "controllable", "differentiable" ] }
```
maliasadi, line 35: I repeat @LuisAlfredoNu's concern here! Could you provide a link to where this is tested in Python for the future reference?

maliasadi (approved):
Thanks @josephleekl for the prompt fix! 🙌

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
josephleekl: We could keep `rot_matrix` to be the one with adjoint=false, and let `applyDeviceMatrixGate` handle the adjoint (is this what you mean?), but I believe that would be less efficient than passing a reversed rot matrix directly. Thoughts @multiphaseCFD?

multiphaseCFD: I'm fine with both approaches. I'll leave you guys to decide. Please add a note if this hard-coded `false` is adopted. Is it possible to change to signature of `getRot` from `getRot(param0, param1, param2)` to (`getRot(params, adjoint)`)?

multiphaseCFD (approved):
LGTM! Thanks @josephleekl !

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
josephleekl: This implementation is similar to e.g. what's been used in[ LQ ](https://github.com/PennyLaneAI/pennylane-lightning/blob/45a67d495552ca9aca3c05467b16b1b6d2e1d4ab/pennylane_lightning/core/src/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp#L1141), so to keep it consistent I will use this. Actually `applyDeviceMatrixGate` also has `use_adjoint=false`, so I will remove the argument, and add a note.

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
```diff
@@ -401,9 +401,11 @@ class StateVectorCudaMPI final
         } else if (opName == "Rot" || opName == "CRot") {
             auto rot_matrix =
                 adjoint
-                    ? cuGates::getRot<CFP_t>(params[2], params[1], params[0])
+                    ? cuGates::getRot<CFP_t>(-params[2], -params[1], -params[0])
                     : cuGates::getRot<CFP_t>(params[0], params[1], params[2]);
-            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts, adjoint);
+            // Use default adjoint=false for applyDeviceMatrixGate() as reversed
+            // rot_matrix explicitly used
+            applyDeviceMatrixGate(rot_matrix.data(), ctrls, tgts);
```
maliasadi, line 408: I'm afraid this is not yet clear why we had to prepare the adjoint matrix for `Rot` manually! This issue can be re-produced for any multi-param gates. `custatevecApplyMatrixGetWorkspaceSize` and `custatevecApplyMatrix` API calls were supposed to take care of these sign updates, right? :thinking: cc: @multiphaseCFD

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
maliasadi: I'm okay with keeping this as is for the sake of consistency in this PR! However, I’m still not persuaded by [the reasoning behind this](https://github.com/PennyLaneAI/pennylane-lightning/pull/1004#discussion_r1856959574).

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
josephleekl: Indeed `custatevecApplyMatrixGetWorkspaceSize` and `custatevecApplyMatrix` can take care of the adjoint signs. I assumed supplying the adjoint matrix manually would give better performance. I have just tested both methods (supplying the adjoint matrix vs using the adjoint from the custatevec API), and there's no performance difference. I can change it to let custatevec deal with the adjoint now.

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
josephleekl: Just did some test and indeed the performance is the same, happy to let custatevec api deal with the adjoint, which does give a cleaner approach.

pennylane_lightning/core/src/simulators/lightning_gpu/StateVectorCudaMPI.hpp
maliasadi: There you go, you got it right! thanks 🙌

PR #1016, opened 2024-12-03
Update logic for enabling `grad_on_execution` in `lightning_*.py` devices

andrijapau:
**Context:**

In PennyLane, `grad_on_execution` is expected to be enabled when the gradient method is set to `adjoint`. Since this is automatically true for Lightning devices, the `use_device_gradient` flag can serve as an additional condition to decide whether to enable `grad_on_execution`.  

This discrepancy caused certain Lightning device tests to fail during a PennyLane workflow clean-up PR. The issue arose because PennyLane expected `grad_on_execution` to be `None` but received `True` instead. Further details can be found in the [PR workflow](https://github.com/PennyLaneAI/pennylane/actions/runs/12129490447/job/33818035152?pr=6657).  

Update logic for when `grad_on_execution` is turned on in `lightning` device configurations.

**Benefits:** Unblocks https://github.com/PennyLaneAI/pennylane/pull/6657

**Possible Drawbacks:** None.

[sc-72150]


pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -299,7 +299,7 @@ def _setup_execution_config(self, config):
             updated_values["gradient_method"] = "adjoint"
         if config.use_device_gradient is None:
             updated_values["use_device_gradient"] = config.gradient_method in ("best", "adjoint")
-        if config.grad_on_execution is None:
+        if config.use_device_gradient and config.grad_on_execution is None:
```
maliasadi, line 302: Thanks @andrijapau for fixing this issue! Does this change mean that we cannot set `config.use_device_gradient=True` for PL built-in gradient methods (i.e., `use_device_gradient==False`)? In other words, why do we need this additional condition to be able to updaet the `grad_on_execution` value?

.github/CHANGELOG.md
```diff
@@ -23,6 +23,9 @@
 
 ### Improvements
 
+* Updated the logic for enabling `grad_on_execution` during device execution.
```
maliasadi, line 26:
```suggestion
* Update the logic for enabling `grad_on_execution` during device execution.
```

pyproject.toml
maliasadi, line 1: minor: could you revert these changes?

andrijapau: Done! Resolving 😄.

pennylane_lightning/lightning_gpu/lightning_gpu.py
albi3ro: We are starting to use the execution config to control non-device behavior as well.  So now it can store gradient transforms, for example.  So if we are using a normal core gradient transform, we want to to have `grad_on_execution=False`.  

Actually, we should check both the config and the `updated_values`.

```suggestion
        if config.use_device_gradient or updated_values.get("use_device_gradient", False) and config.grad_on_execution is None:
```

maliasadi: I see! That adds up!

maliasadi (approved):
Thanks @andrijapau, LGTM!

albi3ro (approved):

albi3ro (approved):

PR #1025, opened 2024-12-13
Add Python CUDA dependencies as Lightning-GPU and Lightning-Tensor wheel dependencies

josephleekl:
### Before submitting

**Context:**

Added all [Python CUDA dependencies](https://github.com/PennyLaneAI/pennylane-lightning/blob/954a6e7d12befc043597769c6357a1003e8d588b/bin/auditwheel) to `pyproject.toml` for LGPU and LT. `INSTALL_RPATH` updated in CMake files.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-79647]

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,8 +16,8 @@ isort==5.13.2
 click==8.0.4
 cmake
 jax[cpu]==0.4.28
-custatevec-cu12
-cutensornet-cu12
+custatevec-cu12; sys_platform == "linux"
+cutensornet-cu12; sys_platform == "linux"
 pylint==2.7.4
 scipy-openblas32>=0.3.26
 git+https://github.com/PennyLaneAI/pennylane.git@master
--- a/scripts/configure_pyproject_toml.py
+++ b/scripts/configure_pyproject_toml.py
@@ -117,6 +117,9 @@ if __name__ == "__main__":
         "pennylane>=0.37",
         "scipy-openblas32>=0.3.26",
     ]
+    
+    if backend != "lightning_qubit":
+        dependencies += ["pennylane_lightning==" + version]
 
     if backend == "lightning_gpu":
         dependencies += ["custatevec-cu12"]
@@ -133,8 +136,6 @@ if __name__ == "__main__":
             "nvidia-cuda-runtime-cu12",
         ]
 
-    if backend != "lightning_qubit":
-        dependencies += ["pennylane_lightning==" + version]
 
     # Package requirements.
     pyproject["project"]["dependencies"] = dependencies
--- a/scripts/configure_pyproject_toml.py
+++ b/scripts/configure_pyproject_toml.py
@@ -126,7 +126,7 @@ if __name__ == "__main__":
         dependencies += ["custatevec-cu12"]
 
     if backend == "lightning_tensor":
-        dependencies += ["cutensornet-cu12", "cutensor-cu12>=2.0.2", "nvidia-cusolver-cu12"]
+        dependencies += ["cutensornet-cu12", "nvidia-cusolver-cu12"]
 
     if backend in ("lightning_gpu", "lightning_tensor"):
         dependencies += [
--- a/scripts/configure_pyproject_toml.py
+++ b/scripts/configure_pyproject_toml.py
@@ -126,13 +126,12 @@ if __name__ == "__main__":
         dependencies += ["custatevec-cu12"]
 
     if backend == "lightning_tensor":
-        dependencies += ["cutensornet-cu12", "cutensor-cu12>=2.0.2"]
+        dependencies += ["cutensornet-cu12", "cutensor-cu12>=2.0.2", "nvidia-cusolver-cu12"]
 
     if backend in ("lightning_gpu", "lightning_tensor"):
         dependencies += [
             "nvidia-nvjitlink-cu12",
             "nvidia-cusparse-cu12",
-            "nvidia-cusolver-cu12",
             "nvidia-cublas-cu12",
             "nvidia-cuda-runtime-cu12",
         ]
--- a/scripts/configure_pyproject_toml.py
+++ b/scripts/configure_pyproject_toml.py
@@ -129,6 +129,7 @@ if __name__ == "__main__":
         "pennylane>=0.37",
         "scipy-openblas32>=0.3.26",
     ]
+
     if backend != "lightning_qubit":
         dependencies += ["pennylane_lightning==" + version]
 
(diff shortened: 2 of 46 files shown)


scripts/configure_pyproject_toml.py
```diff
@@ -122,6 +122,18 @@ def parse_args():
         "scipy-openblas32>=0.3.26",
     ]
 
+    if backend == "lightning_tensor":
+        dependencies += ["cutensor-cu12>=2.0.2"]
+
+    if backend in ("lightning_gpu", "lightning_tensor"):
+        dependencies += [
```
maliasadi, line 129: We may need to include `nvidia-pyindex` to add the NVIDIA private python package index to the user's environment -- only consider this change if you run into any issues with missing NVIDIA  libraries.

maliasadi, line 129: Shall we add `custatevec-cu12` to the list of dependencies?

scripts/configure_pyproject_toml.py
```diff
@@ -122,6 +122,18 @@ def parse_args():
         "scipy-openblas32>=0.3.26",
     ]
 
+    if backend == "lightning_tensor":
+        dependencies += ["cutensor-cu12>=2.0.2"]
```
maliasadi, line 126: Shall we include this `cutensornet-cu12`?

josephleekl: `cutensornet-cu12` and `custatevec-cu12` are currently included as requirements for the Configure Build step (L98/100), should we include them in dependencies too?

josephleekl: I have added it for testing currently

scripts/configure_pyproject_toml.py
josephleekl: It seems ok currently, pip doesn't have any issue pulling these packages.

josephleekl: added

josephleekl:
> Thanks @josephleekl ! Would this new change let our users to run LT on google colab?

Possibly; here we explicitly set cusparse and nvjitlink as dependencies. The exact version is not specified and I'm not sure which pairing works on google colab!

scripts/configure_pyproject_toml.py
```diff
@@ -122,6 +122,20 @@ def parse_args():
         "scipy-openblas32>=0.3.26",
     ]
 
+    if backend == "lightning_gpu":
+        dependencies += ["custatevec-cu12"]
+
+    if backend == "lightning_tensor":
+        dependencies += ["cutensornet-cu12", "cutensor-cu12>=2.0.2", "nvidia-cusolver-cu12"]
```
josephleekl, line 129: cutensor 2.0.2 is a [requirement](https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/index.html) for cutensornet

multiphaseCFD: How could we ensure the version match across those libs here?

multiphaseCFD, line 129: Do we have to update `requirements.txt`?

multiphaseCFD (commented):
Thanks @josephleekl for the nice work!

scripts/configure_pyproject_toml.py
josephleekl: Perhaps we should add them to `requirements-dev.txt`? 

I am also not sure whether e.g. the cuda runtime libraries should be added; can we assume that if people are compiling LGPU from source, they will have the appropriate cuda-runtime libraries, e.g. from nvidia-toolkit? Or is it better to have them all here so we can make sure their versions match? I am also wondering if it's a good idea to have a `requirements-gpu.txt` separately from `requirements-dev.txt`, so that people who don't want to compile specifically for LGPU (e.g. if they just want LQ/LK) don't need to pull any of the nvidia libraries. What do we think? cc@maliasadi

josephleekl: I think we shouldn't pin the versions as long as they're `cu12`, and unless we know there are definitely broken old versions that do not work

josephleekl: So from discussions, `requirements-dev.txt` is mainly for people installing from source, and for people building for LGPU, the lower level libraries (e.g. cublas/nvjitlink) should already be available from the cuda-toolkit, which will be a better source to install them, rather than from our requirements.txt.

josephleekl: So I won't include them here in the `requirements.txt` or `requirements-dev.txt`. 

However, something else I will do is to make `custatevec-cu12` and `cutensornet-cu12` in our `requirements-dev.txt` dependent on Linux platforms only (e.g. not build on mac). This is because they're only built for linux, and cannot be pip installed on macs.

multiphaseCFD: > However, something else I will do is to make custatevec-cu12 and cutensornet-cu12 in our requirements-dev.txt dependent on Linux platforms only (e.g. not build on mac). This is because they're only built for linux, and cannot be pip installed on macs.

Good call! Thanks @josephleekl .

multiphaseCFD: I asked this question because I'm not sure how nvidia libs sync the release with each other. That's say cutensor released a new version, while the latest cutensornet still depends on the previous one. Do we have to pin the versions to fix it or do you have any other options in mind to make it work?

josephleekl: Actually `cutensor-cu12` is already a dependency of `cutensornet-cu12`:
```
$ pip install cutensornet-cu12
Collecting cutensornet-cu12
  Using cached cutensornet_cu12-2.6.0-py3-none-manylinux2014_aarch64.whl (2.2 MB)
Collecting cutensor-cu12<3,>=2.0.2
  Using cached cutensor_cu12-2.0.2-py3-none-manylinux2014_aarch64.whl (149.9 MB)
Installing collected packages: cutensor-cu12, cutensornet-cu12
Successfully installed cutensor-cu12-2.0.2 cutensornet-cu12-2.6.0
```

And this will take care of the `cutensor` version; perhaps we should remove `cutensor` all together?

requirements-dev.txt
```diff
@@ -16,8 +16,8 @@ isort==5.13.2
 click==8.0.4
 cmake
 jax[cpu]==0.4.28
-custatevec-cu12
-cutensornet-cu12
+custatevec-cu12; sys_platform == "linux"
```
multiphaseCFD, line 19: Nice one!

LuisAlfredoNu (approved):
LGTM
Thanks @josephleekl and @multiphaseCFD

multiphaseCFD (approved):
Nice one! Thank you @josephleekl .

PR #1043, opened 2025-01-21
[dyn-alloc-base] Lightning dynamically allocate qubit memory

josephleekl:
### Before submitting

**Context:**
This PR enables lightning devices to dynamically allocate memory for statevectors just before execution. The usual `qml.device("lightning.x", wires=y)` is still supported, but user can also not provide `wires`, i.e. `qml.device("lightning.x")`, and the qubits required for the statevector is provided by the circuit

For LQ/LK/LG - the state-vector is only created at the execute step, and not during device creation. For all devices, if wires is provided during device creation, this will be the number of wires allocated in the device state-vector/tensornet; otherwise, it will be determined from the circuit to be executed.

**Benefits:**

```
dev = qml.device("lightning.qubit")
@qml.qnode(dev)
def circuit():
    qml.Identity(wires=3)
    qml.Identity(wires=1)
    qml.Hadamard(wires=0)
    return qml.state()


@qml.qnode(dev)
def circuit1():
    qml.Identity(wires=1)
    qml.Hadamard(wires=0)
    return qml.state()

circuit()
> array([0.70710678+0.j, 0.70710678+0.j, 0.        +0.j, 0.        +0.j,
       0.        +0.j, 0.        +0.j, 0.        +0.j, 0.        +0.j])
 
circuit1()
> array([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j])
```

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-82702]


pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
```diff
@@ -64,7 +64,6 @@ class StateVectorLQubit : public StateVectorBase<PrecisionT, Derived> {
     const Threading threading_;
     const CPUMemoryModel memory_model_;
 
-  private:
```
AmintorDusko, line 67: Could you please summarize the reason for making it protected?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp
```diff
@@ -182,6 +182,18 @@ class StateVectorLQubitManaged final
...
+     * @brief Update the number of qubits in the statevector.
+     *
+     * @param num_qubits The number of qubits.
+     */
+    void updateNumQubits(std::size_t num_qubits) {
+        BaseType::num_qubits_ = num_qubits;
+        BaseType::setKernels(num_qubits, BaseType::threading_,
+                             BaseType::memory_model_);
+        data_.resize(exp2(num_qubits));
+    }
```
AmintorDusko, line 195: Is changing the number of qubits in place in the current scope?
If yes we may need to implement how we treat the data already there when we change the statevector's number of qubits. @maliasadi has experience with it, he might have a good intuition about this decision.

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -77,6 +77,16 @@ def state(self):
...
 
+    def update_num_qubits(self, num_wires: int):
+        """Update number of qubits in the state vector.
+
+        Args:
+            num_wires (int): _description_
+        """
+        self._num_wires = num_wires
+        self._qubit_state.updateNumQubits(num_wires)
+        self.reset_state()
```
AmintorDusko, line 88: Do we need to reset the state? Do we want to keep and treat the data already there?
If we want to proceed with resetting the state it might be sensible to reset it directly in the C++ layer when calling the function to update.

tests/lightning_qubit/test_state_vector_class.py
```diff
@@ -483,3 +483,23 @@ def test_get_final_state(tol, operation, input, expected_output, par):
...
     assert final_state.state.dtype == final_state.dtype
     assert final_state == state_vector
+
+
+@pytest.mark.parametrize("num_wires", range(2, 5))
+@pytest.mark.parametrize("dtype", [np.complex64, np.complex128])
+def test_update_num_qubit(num_wires, dtype):
+    """Tests that the state vector is correctly updated when the number of qubits is changed."""
+    if device_name != "lightning.qubit":
+        pytest.skip("Only Lightning Qubit allows dynamic qubit allocation")
```
AmintorDusko, line 493:
```suggestion
    if device_name != "lightning.qubit":
        pytest.skip("This device doesn't allow dynamic qubit allocation.")
```

tests/test_apply.py
```diff
@@ -730,14 +730,42 @@ def circuit():
         qnode = qml.QNode(circuit, dev, diff_method="best")
         assert isinstance(qnode.device, ld)
 
-    @pytest.mark.xfail(
-        device_name == "lightning.tensor", reason="lightning.tensor raises different errors"
+    @pytest.mark.skipif(
+        device_name != "lightning.qubit",
+        reason="Only Lightning Qubit support dynamic qubit allocation",
```
AmintorDusko, line 735:
```suggestion
    @pytest.mark.skipif(
        device_name != "lightning.qubit",
        reason="This device doesn't allow dynamic qubit allocation.",
```

tests/test_apply.py
```diff
@@ -730,14 +730,42 @@ def circuit():
...
+            qml.Identity(wires=3)
+            qml.Hadamard(wires=0)
+            return qml.state()
 
-        with pytest.raises(TypeError, match="missing 1 required positional argument: 'wires'"):
-            qml.device(device_name)
+        for circuit in [circuit0, circuit1, circuit2, circuit3]:
+            results = qml.qnode(dev)(circuit)()
+            expected = qml.qnode(dev_default)(circuit)()
+            assert np.allclose(results, expected, atol=tol, rtol=0)
```
AmintorDusko, line 768: What do you think about making the circuit cases a pytest parameter and providing the expected result explicitly?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
josephleekl: The reason is that when we update the number of qubits (updateNumQubits in the derived classes e.g. StateVectorLQubitManaged) we also need to update the Kernels with setKernels, hence moving from private to protected

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp
josephleekl: The reason I believe we need to update num qubits in place is that we want the same device `qml.device('lightning.qubit')` to be able to execute circuit with different sizes one after another, so resizing is needed.

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -313,10 +318,24 @@ def compute_derivatives(
...
+        for circuit in circuits:
+            if self.wires is None:  # Dynamic wires allocation
+                if self._statevector is None:
+                    self._statevector = self.LightningStateVector(
+                        num_wires=circuit.num_wires, dtype=self._c_dtype
+                    )
+                else:
+                    if self._statevector.num_wires != circuit.num_wires:
+                        self._statevector.update_num_qubits(circuit.num_wires)
+                circuit = circuit.map_to_standard_wires()
```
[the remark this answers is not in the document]
josephleekl: This is needed only when we determine the wires from the circuit dynamically to be consistent with what e.g. default.qubit does. There is some particular behaviour, e.g.:
- when the circuit wires are not consecutive e.g. [3, 1, 0] it gets mapped to [0, 1, 2]
- when the circuit wires are consecutive e.g. [2, 1, 0] it doesn't get remapped, and stays to [2, 1, 0]

When the number of wires are explicitly given, we don't need this kind remapping.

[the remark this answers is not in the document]
josephleekl: I absolutely should refactor this into a method. I'll do that!

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -333,12 +352,23 @@ def execute_and_compute_derivatives(
...
+
+            results.append(
+                self.simulate_and_jacobian(
+                    circuit, self._statevector, batch_obs=batch_obs, wire_map=self._wire_map
+                )
             )
-            for c in circuits
-        )
         return tuple(zip(*results))
```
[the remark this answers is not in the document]
josephleekl: Indeed it's a bit confusing to have the combinations like [execute_and_, nothing] x [compute_derivative, vjp, ...]. I'll look at if this could be simplified

pennylane_lightning/lightning_qubit/_state_vector.py
josephleekl: That's a good point. Actually in `simulate()` `vjp` and `jacobian` when it's executed, it resets the state first, so this is redundant. I'll remove it.

tests/test_apply.py
josephleekl: I am happy to use the explicit results later, just before we merge the PR; for now it's convenient to make sure the results are consistent with default.qubit (the results can be counter-intuitive depending on the wires!)

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitRaw.hpp
```diff
@@ -137,5 +137,17 @@ class StateVectorLQubitRaw final
...
+
+    /**
+     * @brief Update the number of qubits in the statevector.
+     *
+     * @param num_qubits The number of qubits.
+     */
+    void updateNumQubits(std::size_t num_qubits) {
+        BaseType::num_qubits_ = num_qubits;
+        length_ = exp2(num_qubits);
+        BaseType::setKernels(num_qubits, BaseType::threading_,
```
josephleekl, line 149: Need to double check whether I need to manually dealloc and alloc, or will this setKernels call handle the memory correctly

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp
AmintorDusko: We need to determine what is the expected behaviour for the data already at the statevector when resizing.

tests/test_apply.py
AmintorDusko: OK. So you are going to make the change when the PR is already fully reviewed, right?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitRaw.hpp
AmintorDusko: `setKernel` calls the method of a new instance created by `getInstance, and all the data stored in an `OperationKernelMap` instance is stored in stl classes with their memory management. Right?
Please double-check, though.

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -176,6 +181,24 @@ def supports_derivatives(
 
         """
 
+    def update_dynamic_wires(self, circuit):
```
josephleekl, line 184: actually I'll make this an abstractmethod, because how the statevectors are initialized are slightly different for each of the devices

josephleekl:
> Please feel free to retag me when the PR is ready for review again.

Will do, just waiting for another PR to make the tests pass again :)

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -313,10 +313,18 @@ def compute_derivatives(
...
         batch_obs = execution_config.device_options.get("batch_obs", self._batch_obs)
 
-        return tuple(
-            self.jacobian(circuit, self._statevector, batch_obs=batch_obs, wire_map=self._wire_map)
-            for circuit in circuits
-        )
+        results = []
+        for circuit in circuits:
+            if self.wires is None:  # Dynamic wires allocation
+                circuit = self.update_dynamic_wires(circuit)
```
maliasadi, line 319: Could it be a static class method in lightning_base?

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -313,10 +313,18 @@ def compute_derivatives(
         """
         batch_obs = execution_config.device_options.get("batch_obs", self._batch_obs)
 
-        return tuple(
-            self.jacobian(circuit, self._statevector, batch_obs=batch_obs, wire_map=self._wire_map)
-            for circuit in circuits
-        )
+        results = []
+        for circuit in circuits:
```
maliasadi, line 317: What about replacing line 316-327 as follows?

``` python
return tuple(
    self.jacobian(
        self.update_dynamic_wires(circuit) if self.wires is None else circuit,
        self._statevector,
        batch_obs=batch_obs,
        wire_map=self._wire_map,
    )
    for circuit in circuits
)
```

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -389,10 +401,17 @@ def compute_vjp(
           then the shape must be ``(batch_size,)``.
         """
         batch_obs = execution_config.device_options.get("batch_obs", self._batch_obs)
-        return tuple(
-            self.vjp(circuit, cots, self._statevector, batch_obs=batch_obs, wire_map=self._wire_map)
-            for circuit, cots in zip(circuits, cotangents)
-        )
+        results = []
```
maliasadi, line 404: We can avoid this list update following the previous implementation.

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -411,12 +430,16 @@ def execute_and_compute_vjp(
             Tuple, Tuple: the result of executing the scripts and the numeric result of computing the vector jacobian product
         """
         batch_obs = execution_config.device_options.get("batch_obs", self._batch_obs)
-        results = tuple(
-            self.simulate_and_vjp(
+
+        results = []
```
maliasadi, line 434: The same here.

pennylane_lightning/core/lightning_newAPI_base.py
maliasadi, line 319: Do we also need to use this method in `jacobian` and `simulate_and_jacobian`?

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
maliasadi, line 67: Do we need to also expose all these aliases when `setKernels` is only needed ? :thinking:

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp
maliasadi: Why do we need to use the same device? What are the implications with creating a new one instead? re-sizing the number of qubits must not be part of the Lightning state-vector Python/C++ APIs.

pennylane_lightning/core/src/simulators/lightning_qubit/bindings/LQubitBindings.hpp
```diff
@@ -183,6 +183,10 @@ void registerBackendClassSpecificBindings(PyClass &pyclass) {
         .def(py::init([](std::size_t num_qubits) {
             return new StateVectorT(num_qubits);
         }))
+        .def("updateNumQubits",
```
maliasadi, line 186: We don't need this method --  should prefer the creation of new instances if there isn't a performance degradation involved with this.

maliasadi (changes requested):
Great work @josephleekl! :rocket: Happy to take a second look after you've addressed the design suggestions

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitRaw.hpp
josephleekl: no longer used with new approach of just creating a new statevector

pennylane_lightning/core/lightning_newAPI_base.py
josephleekl: I haven't put this as a base class method because the arguments to initialize the statevector for the different devices (LQ/LK/LG) are slightly different. I can create an `@abstractmethod` in the base class, but I can only do this when I write the method for all the devices - I will do this in the final (LG) PR. Does this make sense? or is there a better way to do it?

pennylane_lightning/core/lightning_newAPI_base.py
josephleekl: updated

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubit.hpp
josephleekl: no longer required!

pennylane_lightning/core/src/simulators/lightning_qubit/StateVectorLQubitManaged.hpp
josephleekl: It's much simpler if we don't need to change the number of qubits and to just create new statevectors. We can test the performance of this approach.

pennylane_lightning/core/lightning_newAPI_base.py
josephleekl: `jacobian` and `simulate_and_jacobian` has the state as an argument, which requires the state to already be there. The new method `dynamic_wire_alloc` is called  in `compute_derivatives`  (which calls `jacobian` and supplies the state) and `execute_and_compute_derivatives` (which calls `simulate_and_jacobian` and supplies the state)

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -323,6 +327,24 @@ def _setup_execution_config(self, config):
 
         return replace(config, **updated_values, device_options=new_device_options)
 
+    def dynamic_wire_alloc(self, circuit):
```
josephleekl, line 330: an abstractmethod will be created for this in the base class when implementation for all devices are ready

pennylane_lightning/core/lightning_newAPI_base.py
maliasadi: For the sake of reviews, you could still create this as an `abstractmethod` in the base class and override it with dummy implementations in LK/LG/LT. This would give us a better understanding of required changes in other devices and the design in one single PR. It's not a blocker though and happy to accept this as is.

josephleekl: Sounds good, I can add it!

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -200,6 +200,8 @@ jobs:
       - name: Run PennyLane-Lightning integration tests
         run: |
           DEVICENAME=`echo ${{ matrix.pl_backend }} | sed "s/_/./g"`
+          # Remove source files here to ensure pytest collect source coverage from wheel package
```
josephleekl, line 203: IGNORE CHANGES TO THIS FILE

This is to get code coverage for lqcpu, which is from this PR https://github.com/PennyLaneAI/pennylane-lightning/pull/1053/files . This update is temporary

LuisAlfredoNu (commented):
LGTM, Thanks @josephleekl 
Just a few comments :eyeglasses:

.github/workflows/tests_lqcpu_python.yml
josephleekl: I've removed these changes now. I'll just wait for the PR to be merged

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -323,6 +327,24 @@ def _setup_execution_config(self, config):
...
+        Returns:
+            QuantumTape: The updated circuit with the wires mapped to the standard wire order.
+        """
+        if (self._statevector is None) or (self._statevector.num_wires != circuit.num_wires):
+            self._statevector = self.LightningStateVector(
+                num_wires=circuit.num_wires, dtype=self._c_dtype
+            )
+        circuit = (
+            circuit.map_to_standard_wires()
+        )  # Map to follow default.qubit wire order for dynamic wires
```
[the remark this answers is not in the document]
josephleekl: fixed

pennylane_lightning/core/lightning_newAPI_base.py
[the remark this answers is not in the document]
josephleekl: I think I won't tackle this now, maybe in a later PR

tests/test_vjp.py
```diff
@@ -50,9 +51,15 @@ def get_jacobian(device, tape):
...
+    if device_name == "lightning.qubit":
+        fixture_params = itertools.product([np.complex64, np.complex128], [None, 2])
+    else:
+        fixture_params = itertools.product([np.complex64, np.complex128], [2])
+
+    @pytest.fixture(params=fixture_params)
     def dev(self, request):
-        return qml.device(device_name, wires=2, c_dtype=request.param)
+        params = request.param
+        return qml.device(device_name, wires=params[1], c_dtype=params[0])
```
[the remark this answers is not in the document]
josephleekl: updated

tests/test_apply.py
josephleekl: updated

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -43,7 +43,7 @@ class LightningBase(Device):
     A class that serves as a base class for Lightning state-vector simulators.
 
     Args:
-        wires (int or list): number or list of wires to initialize the device with
+        wires (int or list): number or list of wires to initialize the device with. Defaults to ``None`` if not specified, and the device will allocate the number of wires depending on the circuit to execute.
```
maliasadi, line 46:
```suggestion
        wires (Optional[int, list]): number or list of wires to initialize the device with. Defaults to ``None`` if not specified, and the device will allocate the number of wires depending on the circuit to execute.
```

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -71,7 +71,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._c_dtype = c_dtype
         self._batch_obs = batch_obs
 
-        if isinstance(wires, int):
+        if isinstance(wires, int) or wires is None:
```
maliasadi, line 74: Do you also need to update line 63?

pennylane_lightning/core/lightning_newAPI_base.py
```diff
@@ -98,6 +98,17 @@ def _setup_execution_config(self, config):
         Update the execution config with choices for how the device should be used and the device options.
         """
 
+    @abstractmethod
+    def dynamic_wires_from_circuit(self, circuit):
+        """From a given circuit, determine the number of wires and allocate a statevector if applicable. Circuit wires will be mapped to Pennylane default.qubit standard wire order.
```
maliasadi, line 103: As we use this method for LT,
```suggestion
        """From a given circuit, determine the number of wires and allocate the underlying quantum state if applicable. Circuit wires will be mapped to Pennylane ``default.qubit`` standard wire order.
```

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -323,6 +327,23 @@ def _setup_execution_config(self, config):
 
         return replace(config, **updated_values, device_options=new_device_options)
 
+    def dynamic_wires_from_circuit(self, circuit):
+        """From a given circuit, determine the number of wires and allocate a statevector if applicable. Circuit wires will be mapped to Pennylane default.qubit standard wire order.
```
maliasadi, line 331:
```suggestion
        """From a given circuit, determine the number of wires and allocate a state-vector if applicable. Circuit wires will be mapped to Pennylane ``default.qubit`` standard wire order.
```

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -288,6 +288,18 @@ def _setup_execution_config(self, config):
 
         return replace(config, **updated_values, device_options=new_device_options)
 
+    def dynamic_wires_from_circuit(self, circuit):
+        """(DUMMY IMPLEMENTATION) From a given circuit, determine the number of wires and allocate a statevector if applicable. Circuit wires will be mapped to Pennylane default.qubit standard wire order.
```
maliasadi, line 292:
```suggestion
        """(DUMMY IMPLEMENTATION) From a given circuit, determine the number of wires and allocate a state-vector if applicable. Circuit wires will be mapped to Pennylane ``default.qubit`` standard wire order.
```

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -286,8 +286,12 @@ def __init__(  # pylint: disable=too-many-arguments
             "kernel_name": self._kernel_name,
         }
 
-        # Creating the state vector
-        self._statevector = self.LightningStateVector(num_wires=len(self.wires), dtype=c_dtype)
+        # Create the state vector if wires are specified
+        self._statevector = (
```
maliasadi, line 290: Could we always set `_statevector = None` for both cases? we can allocate the state-vector at execution using the same logic that you've implemented for wire=None.

pennylane_lightning/lightning_qubit/lightning_qubit.toml
```diff
@@ -118,7 +118,7 @@ runtime_code_generation = false
 # does not support mid-circuit measurements.
 supported_mcm_methods = [ "one-shot" ]
 # Whether the device supports dynamic qubit allocation/deallocation.
-dynamic_qubit_management = false
+dynamic_qubit_management = true
```
maliasadi, line 121: Hmm, this means a different thing! dynamic allocation here refers to partial allocation and release of qubits at runtime.
 ```suggestion
dynamic_qubit_management = false
```

**Review by maliasadi — COMMENTED** [review]
:raised_hands:

File: pennylane_lightning/lightning_qubit/lightning_qubit.toml (hunk 26 above)
Comment (josephleekl) [reply pennylane_lightning/lightning_qubit/lightning_qubit.toml:121]: I see...

File: pennylane_lightning/lightning_qubit/lightning_qubit.py (hunk 23 above)
Comment (josephleekl) [reply pennylane_lightning/lightning_qubit/lightning_qubit.py:331]: updated!

File: pennylane_lightning/lightning_qubit/lightning_qubit.py (hunk 25 above)
Comment (josephleekl) [reply pennylane_lightning/lightning_qubit/lightning_qubit.py:290]: Yes it should be possible. It makes sense to only create the state-vector at the last moment

File: pennylane_lightning/lightning_kokkos/lightning_kokkos.py (hunk 24 above)
Comment (josephleekl) [reply pennylane_lightning/lightning_kokkos/lightning_kokkos.py:292]: Updated.

File: pennylane_lightning/core/lightning_newAPI_base.py (hunk 21 above)
Comment (josephleekl) [reply pennylane_lightning/core/lightning_newAPI_base.py:74]: updated

File: pennylane_lightning/core/lightning_newAPI_base.py (hunk 22 above)
Comment (josephleekl) [reply pennylane_lightning/core/lightning_newAPI_base.py:103]: Good idea!

File: mpitests/test_apply.py [hunk 27]
```diff
@@ -371,8 +371,6 @@ def test_dev_reset(self, tol, dev_mpi):
         comm.Scatter(state_vector, local_state_vector, root=0)
         dev_cpu = qml.device("lightning.qubit", wires=num_wires, c_dtype=c_dtype)
 
-        dev_cpu._statevector.reset_state()
```
josephleekl, line 374: This is removed  because lightning.qubit device is created, but statevector is not created (only created during execute), so reset_state is not possible and unnecessary

mpitests/test_apply.py
AmintorDusko, line 1: `dev_mpi` is part of this class self.
It can be removed from the arguments and substituted by `self.dev_mpi` inside each test.

mpitests/test_device.py
```diff
@@ -39,6 +39,14 @@ def test_create_device():
         dev = qml.device(device_name, mpi=True, wires=4)
 
 
+def test_unsupported_dynamic_wires():
+    with pytest.raises(
+        DeviceError,
+        match="Lightning-GPU-MPI does not support dynamic wires allocation.",
```
AmintorDusko, line 45:
```suggestion
        match="does not support dynamic wires allocation.",
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -268,15 +268,24 @@ def __init__(  # pylint: disable=too-many-arguments
...
-        # Creating the state vector
-        self._mpi_handler = MPIHandler(mpi, mpi_buf_size, len(self.wires), c_dtype)
-
-        self._statevector = self.LightningStateVector(
-            num_wires=len(self.wires),
-            dtype=c_dtype,
-            mpi_handler=self._mpi_handler,
-            use_async=self._use_async,
-        )
+        self._mpi = mpi
```
AmintorDusko, line 271: I wonder if we could merge all MPI-specific variables in a single structure. I believe this would fix the codefactor warning. Nothing needs to be done now, though.

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -382,21 +379,19 @@ def c_dtype(self):
...
                 self._c_dtype,
                 device_name=self.name,
                 max_bond_dim=self._max_bond_dim,
                 cutoff=self._cutoff,
                 cutoff_mode=self._cutoff_mode,
             )
-        return LightningTensorNet(
-            self._num_wires, self._method, self._c_dtype, device_name=self.name
-        )
+        return LightningTensorNet(num_wires, self._method, self._c_dtype, device_name=self.name)
```
AmintorDusko, line 394: Do you know why that warning?

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -417,6 +412,18 @@ def _setup_execution_config(
...
+        """Map circuit wires to Pennylane ``default.qubit`` standard wire order.
+
+        Args:
+            circuit (QuantumTape): The circuit to execute.
+
+        Returns:
+            QuantumTape: The updated circuit with the wires mapped to the standard wire order.
+        """
+
+        return circuit.map_to_standard_wires() if self.num_wires is None else circuit
```
AmintorDusko, line 425: Is this a false warning?

pennylane_lightning/lightning_tensor/lightning_tensor.py
```diff
@@ -475,7 +482,14 @@ def execute(
         for circuit in circuits:
             if self._wire_map is not None:
                 [circuit], _ = qml.map_wires(circuit, self._wire_map)
-            results.append(simulate(circuit, self._tensornet()))
+            results.append(
```
AmintorDusko, line 485: Another coverage warning.

pennylane_lightning/lightning_tensor/lightning_tensor.py
josephleekl: Lots of codecov false warnings here! It is actually tested:
https://app.codecov.io/gh/PennyLaneAI/pennylane-lightning/pull/1043/blob/pennylane_lightning/lightning_tensor/lightning_tensor.py

L394 here

pennylane_lightning/lightning_tensor/lightning_tensor.py
josephleekl: Also false warning... 
https://app.codecov.io/gh/PennyLaneAI/pennylane-lightning/pull/1043/blob/pennylane_lightning/lightning_tensor/lightning_tensor.py

pennylane_lightning/lightning_tensor/lightning_tensor.py
josephleekl: False warning again

mpitests/test_apply.py
AmintorDusko: THis would solve many of the related codefactor warnings.

josephleekl: I'll try this to see if it gets rid of the codefactor warnings :)

josephleekl: Actually it looks like I can't use `self.dev_mpi` directly because it's a fixture with parameters - I've revert it back and i'll ignore codefactor for now

AmintorDusko (approved):
Thank you @josephleekl! Nice job!

tests/test_vjp.py
```diff
@@ -32,9 +33,11 @@
 class TestVectorJacobianProduct:
     """Tests for the `vjp` function"""
 
-    @pytest.fixture(params=[np.complex64, np.complex128])
+    fixture_params = itertools.product([np.complex64, np.complex128], [None, 2])
+
+    @pytest.fixture(params=fixture_params)
     def dev(self, request):
-        return qml.device(device_name, wires=2, c_dtype=request.param)
+        return qml.device(device_name, wires=request.param[1], c_dtype=request.param[0])
```
maliasadi, line 40: What's the issue with using `pytest.mark.parametrize` instead of this fixture?

pennylane_lightning/lightning_gpu/lightning_gpu.py
maliasadi, line 1: After removing `_mpi` and `_mpi_buf_size`, please take care of remaining codecov warnings. I can tell that there are not that many unit tests to check these cases.

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -316,6 +323,35 @@ def _setup_execution_config(self, config):
 
         return replace(config, **updated_values, device_options=new_device_options)
 
+    def dynamic_wires_from_circuit(self, circuit):
```
maliasadi, line 326: Could you add unit tests for `dynamic_wires_from_circuit` method in the test suite?

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -286,8 +286,7 @@ def __init__(  # pylint: disable=too-many-arguments
             "kernel_name": self._kernel_name,
         }
 
-        # Creating the state vector
-        self._statevector = self.LightningStateVector(num_wires=len(self.wires), dtype=c_dtype)
+        self._statevector = None
```
maliasadi, line 289: Could you add a unit test for this change?

tests/lightning_base/test_device.py
maliasadi, line 1: Could you add tests with None wires to `test_eval_jaxpr.py`?

maliasadi (commented):
Nice work @josephleekl! Just a few minor comments before merging the PR :tada:

tests/test_vjp.py
josephleekl: If we don't have this fixture, we need to have `qml.device` for each test. For `test_vjp` and `test_adjoint_jacobian` because almost all the tests use the same device, it's more convenient to use fixture to set the parametrized device that's run before every test?

pennylane_lightning/lightning_qubit/lightning_qubit.py
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/blob/ea427d8db23d9acc814d15eac4c66423d94dab9d/tests/lightning_base/test_device.py#L215

tests/lightning_base/test_device.py
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/blob/ea427d8db23d9acc814d15eac4c66423d94dab9d/tests/test_eval_jaxpr.py#L54

pennylane_lightning/lightning_gpu/lightning_gpu.py
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/blob/ea427d8db23d9acc814d15eac4c66423d94dab9d/tests/lightning_base/test_device.py#L267

josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/blob/ea427d8db23d9acc814d15eac4c66423d94dab9d/tests/lightning_base/test_device.py#L312

AmintorDusko:
Hey @josephleekl, super useful posting the links to the changes for each comment/suggestion. Much easier to follow, thank you! :pray:

pennylane_lightning/lightning_gpu/lightning_gpu.py
josephleekl: Codecov is now happy https://app.codecov.io/gh/PennyLaneAI/pennylane-lightning/pull/1043 , the warnings from files changed are false warnings

maliasadi (approved):
Awesome! Happy to approve :partying_face:

PR #1044, opened 2025-01-22
Replace the type checking using property `return_type` with direct is…

JerryChen97:
**Context:**
In the deprecations of PL0.41, the property `return_type` of class `MeasurementProcess` will be deprecated very soon https://github.com/PennyLaneAI/pennylane/pull/6841. Therefore, we would like to replace the use of `return_type` with the equivalent direct `isinstance` check. For individual string info use of these return types, we replace them with in-place strings.

3 files are changed with equivalents in-place:

 - pennylane_lightning/core/_adjoint_jacobian_base.py
 - - Private staticmethod `_get_return_type` of class `LightningBaseAdjointJacobian`  called `return_type` once, so replaced with equivalent `isinstance`
 - - The private staticmethod `_check_adjdiff_supported_measurements` was only called by another private helper `_process_jacobian_tape`, therefore the two Enum type returns were replaced by strings, along with their conditional values in `_process_jacobian_tape`.
 - - `_process_jacobian_tape` also called `return_type` once, therefore replace with equivalent `isinstance`.
 
 - pennylane_lightning/core/lightning_base.py
 - - The private staticmethod `_check_adjdiff_supported_measurements` was identified as not used by any code inside lightning. Therefore, we deleted it directly.
 
 - tests/test_measurements.py
 - - Two `return_type` usages were replaced with `isinstance`

**Benefits:**
Consistent with both v0.40 and v0.41 of PL

**Possible Drawbacks:**
Not aware of yet.

**Related GitHub Issues:**
[[sc-71892](https://app.shortcut.com/xanaduai/story/71892)]

### DIFF
--- a/pennylane_lightning/core/lightning_base.py
+++ b/pennylane_lightning/core/lightning_base.py
@@ -23,7 +23,6 @@ import numpy as np
 import pennylane as qml
 from pennylane import BasisState, StatePrep
 from pennylane.devices import QubitDevice
-from pennylane.measurements import ExpectationMP, MeasurementProcess, StateMP
 from pennylane.operation import Operation
 from pennylane.ops import Prod, Projector, SProd, Sum
 from pennylane.wires import Wires
--- a/pennylane_lightning/core/lightning_base.py
+++ b/pennylane_lightning/core/lightning_base.py
@@ -23,7 +23,7 @@ import numpy as np
 import pennylane as qml
 from pennylane import BasisState, StatePrep
 from pennylane.devices import QubitDevice
-from pennylane.measurements import Expectation, MeasurementProcess, State
+from pennylane.measurements import ExpectationMP, MeasurementProcess, StateMP
 from pennylane.operation import Operation
 from pennylane.ops import Prod, Projector, SProd, Sum
 from pennylane.wires import Wires
@@ -348,14 +348,14 @@ class LightningBase(QubitDevice):
         if not measurements:
             return None
 
-        if len(measurements) == 1 and measurements[0].return_type is State:
+        if len(measurements) == 1 and isinstance(measurements[0], StateMP):
             # return State
             raise qml.QuantumFunctionError(
                 "Adjoint differentiation does not support State measurements."
             )
 
         # The return_type of measurement processes must be expectation
-        if any(m.return_type is not Expectation for m in measurements):
+        if any(not isinstance(m, ExpectationMP) for m in measurements):
             raise qml.QuantumFunctionError(
                 "Adjoint differentiation method does not support expectation return type "
                 "mixed with other return types"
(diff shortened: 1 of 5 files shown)


pennylane_lightning/core/lightning_base.py
```diff
@@ -348,22 +348,22 @@ def _check_adjdiff_supported_measurements(measurements: List[MeasurementProcess]
...
             return None
 
-        if len(measurements) == 1 and measurements[0].return_type is State:
+        if len(measurements) == 1 and isinstance(measurements[0], StateMP):
             # return State
             raise qml.QuantumFunctionError(
                 "Adjoint differentiation does not support State measurements."
             )
 
         # The return_type of measurement processes must be expectation
```
josephleekl, line 357: You can remove this I think

pennylane_lightning/core/lightning_base.py
```diff
@@ -348,22 +348,22 @@ def _check_adjdiff_supported_measurements(measurements: List[MeasurementProcess]
...
-        if any(m.return_type is not Expectation for m in measurements):
+        if any(not isinstance(m, ExpectationMP) for m in measurements):
             raise qml.QuantumFunctionError(
                 "Adjoint differentiation method does not support expectation return type "
```
josephleekl, line 360: something like this?
```suggestion
                "Adjoint differentiation method does not support Expectation Measurement type "
```

pennylane_lightning/core/lightning_base.py
```diff
@@ -348,22 +348,22 @@ def _check_adjdiff_supported_measurements(measurements: List[MeasurementProcess]
         if not measurements:
             return None
 
-        if len(measurements) == 1 and measurements[0].return_type is State:
+        if len(measurements) == 1 and isinstance(measurements[0], StateMP):
```
josephleekl, line 351: can you help check whether this method `_check_adjdiff_supported_measurements` is used anywhere at all? if not, maybe we should remove it entirely

josephleekl (commented):
Thanks @JerryChen97 ! Looks good, just have some quick questions before approving

pennylane_lightning/core/lightning_base.py
JerryChen97: By simply searching inside lighting for this private static method I found no other occurrence. @maliasadi @AmintorDusko Shall we just del this?

JerryChen97:
@maliasadi As described in description of change, the lines affected here are all in-line equivalents, therefore they should be no-effect towards any other files. However, codecov patch has been all the time 0 inside the changes of `pennylane_lightning/core/_adjoint_jacobian_base.py`, which I believe is caused by the fact that we actually do not have a direct unit test file for it; we should add another separate unit test file against this submodule in the future.

pennylane_lightning/core/lightning_base.py
AmintorDusko: I think it is safe to remove it.

JerryChen97:
@AmintorDusko I removed ci build_wheels label and retriggered tests.

AmintorDusko (approved):
This is great! Thanks!

PR #1064, opened 2025-02-12
`MPSPrep` adding extra empty qubit

LuisAlfredoNu:
**Context:**
Lightning Tensor can use an MPS as input but should be passed through the circuit operation `qml.MPSPrep` but must have the same shape as the allocated MPS.

This PR adds the following features to Lightning Tensor:

* Allow `qml.MPSPrep` to accept an MPS with `len(MPS) = n_wires-1`  
* Expand the input MPS to fit with the allocated MPS. 

The following algorithm was implemented
- Padding with zeros the tensor $B_i$ to fit the tensor shape $A_{i+1}$ up to $i = N/2$ where $N$ is the total number of tensors in $B$
- Add the identity matrix with shape `(1,2,2)` at the beginning of $B$
- Restore the $B$ MPS into a right-canonical form to spread the new site information across the entire MPS.

**Benefits:**
More flexibility to work with LT and provide custom MPS. 

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-84509]


.github/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Improvements
 
+* In Lightning Tensor. Allow `qml.MPSPrep` to accept an MPS with len(MPS) = n_wires-1.
```
AmintorDusko, line 9:
```suggestion
* In Lightning Tensor, allow `qml.MPSPrep` to accept an MPS with len(MPS) = n_wires-1.
```

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
     return mpos, sorted_wires
 
 
+def check_canonical_form(mps, direction="left"):
+    """Check if the MPS is in the canonical form. Computation of expectation and matrix elements is simpler if the MPS is built from tensors
+    relating orthonormal spaces. The canonical form of the MPS is defined as the form where the tensors are orthonormal in the left or right direction.
+    """
```
AmintorDusko, line 163:
```suggestion
    """Check if the MPS is in the canonical form. The computation of expectation and matrix elements is simpler if the MPS is built from tensors relating orthonormal spaces. The canonical form of the MPS is defined as the form where the tensors are orthonormal in the left or right direction.
    """
```

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    return all(canon_values)
+
+
+def expand_mps_top(state_MPS, max_bond_dim=128):
+    """Expand the MPS to match the size of the target wires. Note: The expansion works only for adding a single wire at the beginning of the MPS."""
+
+    expanded_MPS = state_MPS.copy()
+
+    # Number of sites that should be changed from the first site
+    n_sites_change = (len(expanded_MPS) + 1) // 2
```
josephleekl, line 190: is there any reason we only allow adding a single wire at the beginning, and not multiple wires?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    # Check if the number of sites is odd
+    odd_n_sites = len(expanded_MPS) % 2 == 1
```
maliasadi, line 193:
```suggestion
    n_sites = len(state_MPS)
    n_sites_change = (n_sites + 1) // 2
    odd_n_sites = n_sites % 2 == 1
```

maliasadi (commented):
Thanks @LuisAlfredoNu! Nice work :zap:

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
 from pennylane.wires import Wires
 
 
-def svd_split(Mat, site_shape, max_bond_dim):
+def svd_split(Mat, site_shape, max_bond_dim, direction="right"):
```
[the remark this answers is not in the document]
LuisAlfredoNu: I will suggest keeping those names because these are the standard names ([ref](https://en.wikipedia.org/wiki/Matrix_product_state#Left-Canonical_Decomposition))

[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
...
     """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer."""
     # TODO: Check if cutensornet allows us to remove all zero (or < tol) singular values and the respective rows and columns of U and Vd
+
     U, S, Vd = np.linalg.svd(Mat, full_matrices=False)
-    U = U * S  # Append singular values to U
+
     bonds = len(S)
+    chi = min([bonds, max_bond_dim])
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
...
+
+    if direction == "right":
+        U = U * S  # Append singular values to U
+
+    if direction == "left":
+        Vd = S * Vd.T  # Append singular values to Vd
+        Vd = Vd.T
```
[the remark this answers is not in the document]
LuisAlfredoNu: Nice catch!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
...
 
     # keep only chi bonds
-    chi = min([bonds, max_bond_dim])
     U, Vd = U[..., :chi], Vd[:chi]
+
     return U, Vd
 
 
-def decompose_dense(psi, n_wires, site_shape, max_bond_dim):
+def decompose_dense(psi, n_wires, site_shape, max_bond_dim, direction="right"):
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
     return mpos, sorted_wires
 
 
+def check_canonical_form(mps, direction="left"):
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
...
-        )  # psi [site_len*bondL, -1] -> U [site_len, mu] Vd [mu, (2x2x2x..)]
-        Ms[i] = U
+        Ms[0] = Vd
+        bondL = U.shape[-1]
+        psi = U
 
-        psi = Vd
+        for i in range(1, n_wires - 1):
+            psi = np.reshape(psi, (-1, site_len * bondL))
+            U, Vd = svd_split(psi, site_shape, max_bond_dim, direction=direction)
```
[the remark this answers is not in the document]
LuisAlfredoNu: It is not necessary due to we using the arg `full_matrices=False`. Therefore, the `S` matrix is a 1D array ([ref](https://numpy.org/doc/2.2/reference/generated/numpy.linalg.svd.html))

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: For the moment I was only focusing on adding a single wire. I suspect that is possible to add more but I couldn't test N plus wires.  We can visit this option when sensitive. :rocket:

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    for site in mps:
+        site_p = np.tensordot(Vd, site, axes=[[-1], [0]])
+        site_p = site_p.reshape(-1, site.shape[-1])
+
+        U, S, Vd = np.linalg.svd(site_p, full_matrices=False)
+
+        bonds = len(S)
+
+        Vd = S * Vd
```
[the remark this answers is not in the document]
LuisAlfredoNu: It is not necessary due to we using the arg full_matrices=False. Therefore, the S matrix is a 1D array ([ref](https://numpy.org/doc/2.2/reference/generated/numpy.linalg.svd.html))

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
LuisAlfredoNu: I tried this option but I got an error. The reason is the following:
`S` is a 1D array with `n` elements, `Vd` is a matrix with `(n,m)` elements. Therefore, the operation `np.diag(S) @ A` is equivalent to `(S * A.T).T` because `S * A.T` is broadcasting of `S` over all columns of `A`

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+        U = U.reshape([-1] + site_shape + [bonds])
```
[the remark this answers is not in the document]
LuisAlfredoNu: I will use this suggestion in `svd_split` because here is not necessary to truncate the `U` matrix due to `site_p` has a square matrix. :rocket:

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,47 +34,81 @@
 from pennylane.wires import Wires
 
 
-def svd_split(Mat, site_shape, max_bond_dim):
+def svd_split(Mat, site_shape, max_bond_dim, direction="right"):
     """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer."""
     # TODO: Check if cutensornet allows us to remove all zero (or < tol) singular values and the respective rows and columns of U and Vd
+
     U, S, Vd = np.linalg.svd(Mat, full_matrices=False)
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +475,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
             M = decompose_dense(state, self._num_wires, mps_site_shape, self._max_bond_dim)
             self._tensornet.updateMPSSitesData(M)
 
+    def _apply_mps_state(self, mps: tuple, target_wires: Wires) -> None:
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+        canon_values.append(close)
+
+    # Return True if all the values of canon_values are True
+    return all(canon_values)
+
+
+def expand_mps_top(state_MPS, max_bond_dim=128):
+    """Expand the MPS to match the size of the target wires. Note: The expansion works only for adding a single wire at the beginning of the MPS."""
+
+    expanded_MPS = state_MPS.copy()
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +157,143 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+        # Vertical padding with zeros
+        target_l, _, target_r = state_MPS[i + 1].shape
+
+        if odd_n_sites:  # odd sites need to double the bond dimension
+            target_r = target_l * 2 if target_l * 2 < max_bond_dim else max_bond_dim
+
+        site_r = new_site.shape[-1]
+
+        new_site = new_site.reshape(target_l, 2, site_r)
+        new_site = np.pad(new_site, ((0, 0), (0, 0), (0, target_r - site_r)), mode="constant")
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+    """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer.
+
+    Args:
+        Mat (array): input matrix
+        site_shape (list): shape of the site tensor
+        max_bond_dim (int): maximum bond dimension
+        direction (str): direction of the SVD decomposition. Options: ["right", "left"]
+
+    Returns:
+        [array, array]: U and Vd matrices
```
josephleekl, line 49: `tuple[np.ndarray, np.ndarray]`

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+def svd_split(
+    Mat: np.ndarray, site_shape: list, max_bond_dim: int, direction: str
+) -> tuple[np.ndarray, np.ndarray]:
+    """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer.
+
+    Args:
+        Mat (array): input matrix
+        site_shape (list): shape of the site tensor
+        max_bond_dim (int): maximum bond dimension
+        direction (str): direction of the SVD decomposition. Options: ["right", "left"]
```
josephleekl, line 46: should we throw an error if direction is neither "right" nor "left"

josephleekl: can you also clarify what left/right here means? is this what you mean/is the convention: `left` means we keep the left matrix orthnormal, that's why we multiply S to the rhs matrix (Vd)?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+    direction: str = "right",
+) -> list[np.ndarray]:
+    """Decompose a dense state vector/gate matrix into MPS/MPO sites.
+
+    Args:
+        psi (array): input state vector or gate matrix
+        n_wires (int): number of wires
+        site_shape (list): shape of the site tensor
+        max_bond_dim (int): maximum bond dimension
+        direction (str): direction of the SVD decomposition. Options: ["right", "left"]
```
josephleekl, line 88: Default to right. (why?)

josephleekl: should this be direction of canonical form?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+
+    Returns:
+        list[array]: MPS/MPO sites
+    """
+
     Ms = [[] for _ in range(n_wires)]
```
josephleekl, line 94: why not just have an empty list and append to it every time? why do we need this

pennylane_lightning/lightning_tensor/_tensornet.py
josephleekl: should we also throw error for invalid direction?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+def gate_matrix_decompose(
+    gate_ops_matrix: np.ndarray,
+    wires: list[int],
+    max_mpo_bond_dim: int,
+    c_dtype: np.complex64 | np.complex128,
+) -> tuple[list[np.ndarray], list[int]]:
+    """Permute and decompose a gate matrix into MPO sites. This method return the MPO sites in the Fortran order of the ``cutensornet`` backend. Note that MSB in the Pennylane convention is the LSB in the ``cutensornet`` convention.
+
+    Args:
+        gate_ops_matrix (array): input gate matrix
```
josephleekl, line 155: same as above

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
     return mpos, sorted_wires
 
 
+def check_canonical_form(mps: list[np.ndarray], direction: str = "left") -> bool:
+    """Check if the MPS is in the canonical form. The computation of expectation and matrix elements is simpler if the MPS is built from tensors relating orthonormal spaces. The canonical form of the MPS is defined as the form where the tensors are orthonormal in the left or right direction.
```
josephleekl, line 206:
```suggestion
    """Check if the MPS is in the canonical form. The computation of expectation values and matrix elements is simpler if the MPS is built from orthonormal tensors, i.e. in canonical form (either in the left or right direction).
```

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+
+    Args:
+        mps (list): MPS sites
```
josephleekl, line 210: list[np.ndarray]?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    # Return True if all the values of canon_values are True
+    return True
+
+
+def expand_mps_top(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
+    """Expand the MPS to match the size of the target wires. Note: The expansion works only for adding a single wire at the beginning of the MPS.
+
+    Args:
+        state_MPS (list): MPS sites
```
josephleekl, line 239: list[np.ndarray]?

josephleekl: can we rename this MPS_state? and should the description be something like `original MPS state to be expanded`

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+        close = np.allclose(C, eye, atol=1e-10)
+
+        if not close:
+            return False
+
+    # Return True if all the values of canon_values are True
+    return True
+
+
+def expand_mps_top(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
```
josephleekl, line 235: why is it `top`? do you mean first site?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    expanded_MPS = state_MPS
+
+    # Number of sites that should be changed from the first site
+    n_sites = len(state_MPS)
+    n_sites_change = (n_sites + 1) // 2
+    odd_n_sites = n_sites % 2 == 1
+
+    for i in range(n_sites_change - 1):
+        # Create the new site for expanded_MPS
+        new_site = expanded_MPS[i]
```
josephleekl, line 255: why do we need this intermediate new_site?

josephleekl: perhaps here's an alternative (just an idea): if we want this intermediate new_site, we don't need the initial `expanded_MPS = state_MPS`. can we not just do `new_site = state_MP[i]` then edit the `new_site`, then append to expanded_MPS, then at the end just copy the rest of state_MPS to expanded_MPS. Feel free to ignore if this makes it less clean

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    # Assign the last new site
+    expanded_MPS[n_sites_change - 1] = new_site
+
+    # Add the initial site
+    expanded_MPS = [np.eye(2, dtype=state_MPS[0].dtype).reshape(1, 2, 2)] + expanded_MPS
+
+    return expanded_MPS
+
+
+def restore_left_canonical_form(mps: list[np.ndarray], site_shape: list[int]) -> list[np.ndarray]:
```
josephleekl, line 309: can't this be combined with restore_right_canonical_form?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    """Restore the left canonical form of the MPS. The left canonical form is defined as the form where the tensors are orthonormal in the left direction.
+
+    Args:
+        mps (list): MPS sites
```
josephleekl, line 313: MPS state

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +558,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
             M = decompose_dense(state, self._num_wires, mps_site_shape, self._max_bond_dim)
             self._tensornet.updateMPSSitesData(M)
 
+    def _apply_mps_state(self, mps: tuple, target_wires: Wires) -> None:
+
+        if len(target_wires) == self._num_wires and Wires(sorted(target_wires)) == target_wires:
+            self._tensornet.updateMPSSitesData(mps)
+            return
+
+        trgt_wires = target_wires.tolist().copy()
```
josephleekl, line 567: do we need copy  here?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +558,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
...
+
+        # Sort wires in ascending order
+        trgt_wires.sort()
+
+        # check if 0 is prensent in trgt_wires
```
josephleekl, line 572: present

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +558,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
...
+        zero_trgt = 0 in trgt_wires
+        single_wire = self._num_wires - len(trgt_wires) > 1
+
+        if single_wire and not zero_trgt:
```
josephleekl, line 576: can we do something like if not (0 in trgt_wires) and/or self._num_wires > len(trgt_wires) + 1:

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +558,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
...
+        new_mps = list(mps)
+
+        if len(mps[0].shape) != 3:
+            new_mps[0] = new_mps[0].reshape(1, 2, 2)
+
+        if len(mps[-1].shape) != 3:
+            new_mps[-1] = new_mps[-1].reshape(2, 2, 1)
+
+        # Check the canonical form of the MPS
+        canon_left = check_canonical_form(new_mps, direction="left")
```
josephleekl, line 590: we don't have to check both, right?
can we do:
```
if check_canonical_form(new_mps, direction="left"):
    new_mps = expand_mps_top(...)
    new_mps = restore_left_canonical_form(...)
elif check_canonical_form(new_mps,direction_"right"):
    new_mps = ...
else:
    raise ...
```

josephleekl (commented):
Thanks @LuisAlfredoNu for the nice work! I've left some comments, happy to review again when addressed :)

josephleekl:
also i think codecov has something to say

pennylane_lightning/lightning_tensor/_tensornet.py
AmintorDusko: Alternatively, if it is sensible with the rest of the code style, you can use a boolean `right` or `left` instead of a str direction (whichever is more sensible as a default). This way we only have the 2 expected cases.

pennylane_lightning/lightning_tensor/_tensornet.py
AmintorDusko: Same thing here. It might be better to operate with a boolean.

AmintorDusko (commented):
Nice work! My main suggestion, so far, is to change direction from a str to a boolean control variable. It is more sensible as we only have two cases anyway and we use it to evaluate conditions. What are your thoughts?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
 
-def svd_split(Mat, site_shape, max_bond_dim):
-    """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer."""
+def svd_split(
+    Mat: np.ndarray, site_shape: list, max_bond_dim: int, direction: str
+) -> tuple[np.ndarray, np.ndarray]:
+    """SVD decomposition of a matrix via numpy linalg. Note that this function is to be moved to the C++ layer.
+
+    Args:
+        Mat (array): input matrix
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,132 @@
...
+    psi: np.ndarray,
+    n_wires: int,
+    site_shape: list[int],
+    max_bond_dim: int,
+    direction: str = "right",
+) -> list[np.ndarray]:
+    """Decompose a dense state vector/gate matrix into MPS/MPO sites.
+
+    Args:
+        psi (array): input state vector or gate matrix
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +202,181 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    Args:
+        mps (list): MPS sites
+        site_shape (list): shape of the site tensor
+
+    Returns:
+        list: MPS sites in the right canonical form
+    """
+
+    mps.reverse()
```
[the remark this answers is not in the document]
LuisAlfredoNu: Nice catch! :rocket:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
AmintorDusko: I mean something like `is_right`, and if it is not, it'll be left.

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Nice catch!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: > should we throw an error if direction is neither "right" nor "left"

No, it should be one of them. 

> can you also clarify what left/right here means? 

Ok I will add more information 

> I mean something like `is_right`, and if it is not, it'll be left.

Nice idea, I will included

LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: > Default to right. (why?)

The previous version always uses the right-canonical but with this expansion now it is possible to choose. This does not have a huge impact on the internal code but after accepting external MPS becomes relevant. 

> should this be direction of canonical form?

Yes, I will replace the name to avoid confusion. 

> should we also throw error for invalid direction?

I don't think so. always should have a direction. I will use the @AmintorDusko recommendation of `use_right`

LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Changed

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: The intermediate new_site is only a variable to point to the ndarray object because it doesn't create or copy any memory. The @maliasadi's approach simplifies this issue of copying the full MPS_sites :rocket:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Yes but it would produce a large if statement that probably will generate more confusion. Creating a new function might improve readability
``` python
def restore_canonical_form(..., direction):
    if direction == 'left':
        restore_left_canonical_form(...)
    else: # right canonical
        restore_right_canonical_form(...)
```
What do you think?

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -304,6 +558,53 @@ def _apply_state_vector(self, state, device_wires: Wires):
             M = decompose_dense(state, self._num_wires, mps_site_shape, self._max_bond_dim)
             self._tensornet.updateMPSSitesData(M)
 
+    def _apply_mps_state(self, mps: tuple, target_wires: Wires) -> None:
```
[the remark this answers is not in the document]
LuisAlfredoNu: Fixed!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Nice catch!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: It is not necessary the `else` `raise` because could be that the input MPS does not have any particular canonical shape.

LuisAlfredoNu: Done!

multiphaseCFD (commented):
Thank you @LuisAlfredoNu for your excellent work. This enhancement will improve the performance of the `LT` process with `block2` data and provides a way to verify the canonical form of MPS site data.

My only question for discussion is whether `LT` is the most appropriate location for this feature. I'm concerned about the potential risk due to the current lack of safety checks.  If we keep this functionality within `LT`, there's a possibility of issues arising. For example, a user might intend to analyze an `n_wires` system and correctly provide `n_wires` MPS site data to `MPSPrep`. However, if they accidentally provide `n_wires-1` MPS site data, the LT process might not detect the discrepancy and throw an error beforehand.  I'd like to open this up for discussion with the team.

CC: @maliasadi @josephleekl @AmintorDusko

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +212,182 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    The computation of expectation values and matrix elements is simpler if the MPS is built from orthonormal tensors, i.e. in canonical form (either in the left or right direction).
+
+    Args:
+        mps (list[np.ndarray]): MPS state
+        direction (str): direction to check the canonical form. Options: ["left", "right"]
+
+    Returns:
+        bool: True if the MPS is in the canonical form specified by the direction
+    """
```
AmintorDusko, line 226: We need to refine the treatment we give to direction here.
You can transform it into a boolean, like in the other cases, or you need to check the word and raise an error if the word is different from these two. We don't want someone typoing 'lefty' and defaulting to right.

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +212,182 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+        # Compare C with the identity matrix
+        if not np.allclose(C, np.eye(C.shape[0], dtype=C.dtype), atol=1e-10):
+            return False
+
+    # Return True if all the values of canon_values are True
+    return True
+
+
+def expand_mps_fist_site(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
```
multiphaseCFD, line 243: Maybe I miss some context here. Could you please explain more about the algorithm used in this method? If I remember correctly, the MPS state is $\[0,1] \otimes |\psi\rangle$. If it's true, I'm wondering if merely set the added site as $\[0,1]\otimes[1,0]$ and the site next to it to be $\[1,0] \otimes site_1$ would work.

LuisAlfredoNu:
> My only question for discussion is whether `LT` is the most appropriate location for this feature. I'm concerned about the potential risk due to the current lack of safety checks. If we keep this functionality within `LT`, there's a possibility of issues arising. For example, a user might intend to analyze an `n_wires` system and correctly provide `n_wires` MPS site data to `MPSPrep`. However, if they accidentally provide `n_wires-1` MPS site data, the LT process might not detect the discrepancy and throw an error beforehand. I'd like to open this up for discussion with the team.

Thank you @multiphaseCFD for your review. The situation that you mention throws a proper error to the user, if they try to do that this LT will show the error: 
```
Error in PennyLane Lightning: The incoming MPS does not have the correct layout for lightning.tensor.
```
because the input MPS and the target MPS don't match in dimension.  I will add an extra test to clarify this situation.

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,142 @@
...
+    Returns:
+        tuple[np.ndarray, np.ndarray]: U and Vd matrices.
+    """
     # TODO: Check if cutensornet allows us to remove all zero (or < tol) singular values and the respective rows and columns of U and Vd
+
     U, S, Vd = np.linalg.svd(Mat, full_matrices=False)
-    U = U * S  # Append singular values to U
+
+    # Removing noise from singular values
+    S[S < 1e-10] = 0.0
```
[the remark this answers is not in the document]
LuisAlfredoNu: This was a suggestion from @maliasadi It is not too clear to me why we need to remove the singular values noises :thinking:

.github/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Improvements
 
+* In Lightning-Tensor, allow `qml.MPSPrep` to accept an MPS with `len(MPS) = n_wires-1`.
```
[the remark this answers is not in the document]
LuisAlfredoNu: Nop, the `0th` qubit is entangling with `|0>` state to rest of the MPS state

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,142 @@
...
+    This method return the MPO sites in the Fortran order of the ``cutensornet`` backend. Note that MSB in the Pennylane convention is the LSB in the ``cutensornet`` convention.
+
+    Args:
+        gate_ops_matrix (np.ndarray): input gate matrix
+        wires (list): list of wires
+        max_mpo_bond_dim (int): maximum bond dimension
+        c_dtype (np.complex64 | np.complex128): complex dtype
+
+    Returns:
+        [list[np.ndarray], list[int]]: MPO sites and sorted wires
```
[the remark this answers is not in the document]
LuisAlfredoNu: Yes because the return's function is  `return mpos, sorted_wires` therefore Python convert it into a tuple

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +212,182 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+    for sites in mps:
+
+        if direction == "left":
+            C = np.tensordot(sites.conj().T, sites, axes=[[-1, -2], [0, 1]])
+        else:  # direction == "right"
+            C = np.tensordot(sites, sites.conj().T, axes=[[-1, -2], [0, 1]])
+
+        # Compare C with the identity matrix
+        if not np.allclose(C, np.eye(C.shape[0], dtype=C.dtype), atol=1e-10):
```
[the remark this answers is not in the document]
LuisAlfredoNu: Yes, this was a magic number :sweat_smile:
Do you have an idea of how to remove this magic number? I think could be set based on the `dtype` of the MPS but, again, which number should selected? :thinking:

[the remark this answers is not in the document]
multiphaseCFD: Just wondering if there is any reference can support this choice.

pennylane_lightning/lightning_tensor/_tensornet.py
multiphaseCFD: Thanks @LuisAlfredoNu . Is there any reference for this padding approach?

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
maliasadi: It has to do with the floating errors!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,142 @@
...
     # TODO: Check if cutensornet allows us to remove all zero (or < tol) singular values and the respective rows and columns of U and Vd
+
     U, S, Vd = np.linalg.svd(Mat, full_matrices=False)
-    U = U * S  # Append singular values to U
+
+    # Removing noise from singular values
+    S[S < 1e-10] = 0.0
+
     bonds = len(S)
+    chi = min(bonds, max_bond_dim)
```
maliasadi, line 63: @LuisAlfredoNu @multiphaseCFD Too many codecov warnings..

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,142 @@
...
+    canonical_right: bool = True,
+) -> list[np.ndarray]:
+    """Decompose a dense state vector/gate matrix into MPS/MPO sites.
+
+    Args:
+        psi (np.ndarray): input state vector or gate matrix
+        n_wires (int): number of wires
+        site_shape (list[int]): shape of the site tensor
+        max_bond_dim (int): maximum bond dimension
+        canonical_right (bool): right-canonical form if True; left-canonical form if False
```
maliasadi, line 96:
```suggestion
        canonical_right (bool): right-canonical form if True; left-canonical form if False. Default is True. 
```

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Nop, that is way took me a long time to develop this solution :face_with_spiral_eyes:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Yes, but if you visit the codecov [page](https://app.codecov.io/gh/PennyLaneAI/pennylane-lightning/pull/1064) everything is :green_circle: :smile:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: yes, you are right I will update it

LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
LuisAlfredoNu: what do you think about this approach 
``` python 
    # Removing noise from singular values
    epsilons = np.finfo(Mat.dtype).eps * S[0] if S[0] > 1.0 else np.finfo(Mat.dtype).eps    
    S[S < epsilons] = 0.0
```

[reference](https://scicomp.stackexchange.com/a/355)

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,142 @@
...
+    Vd = Vd[:chi]
+
+    if is_right:  # Vd as orthonormal singular vectors
+        U = U * S  # Append singular values to U
+    else:  # U as orthonormal singular vectors
+        Vd = (S * Vd.T).T  # Append singular values to Vd, equivalent operation to np.diag(S) @ Vd
+
+    # keep only chi bonds and reshape to fit the bond dimension and site shape
+    Vd = Vd.reshape([chi] + site_shape + [-1])
+    U = U.reshape([-1] + site_shape + [chi])
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
multiphaseCFD: As this is sth we developed, it might be worth creating a notion page to illustrate the mathematical magics behind it. CC: @maliasadi @tomlqc

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
maliasadi: Sounds good! Please also include the SE link for the future reference.

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,184 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+        # Compare C with the identity matrix
+        if not np.allclose(C, np.eye(C.shape[0], dtype=C.dtype), atol=1e-10):
+            return False
+
+    # Return True if all the values of canon_values are True
+    return True
+
+
+def expand_mps_fist_site(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
```
maliasadi, line 224: fist?

pennylane_lightning/lightning_tensor/_tensornet.py
maliasadi: @multiphaseCFD :100: but we should include this in the docstrings

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,184 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    """Expand the MPS to match the size of the target wires.
+
+    This function modifies the original MPS state by adding a single wire at the beginning of the MPS state.
```
maliasadi, line 227: @LuisAlfredoNu Could you elaborate on the technique? Do you need to include any references?

maliasadi (commented):
Nice work @LuisAlfredoNu! :guitar: Happy to approve after updating docstrings :writing_hand:

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -34,53 +34,123 @@
...
+    """
+
+    Ms = []
     site_len = np.prod(site_shape)
-    psi = np.reshape(psi, (site_len, -1))  # split psi [2, 2, 2, 2...] to psi [site_len, -1]
 
-    U, Vd = svd_split(
-        psi, site_shape, max_bond_dim
-    )  # psi [site_len, -1] -> U [site_len, mu] Vd [mu, (2x2x2x..)]
+    # psi = np.reshape(psi, (-1, site_len)) if canonical_right else np.reshape(psi, (site_len, -1))
```
multiphaseCFD, line 109: Want to remove this line?

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: In the SC story, I post a notion page explaining this algorithm with a small example. :smile:

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,184 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    Returns:
+        bool: True if the MPS is in the canonical form specified by the direction
+    """
+
+    for sites in mps:
+
+        if is_right:  # check if the MPS state has a right canonical form
+            C = np.tensordot(sites, sites.conj().T, axes=[[-1, -2], [0, 1]])
+        else:  # check if the MPS state has a left canonical form
+            C = np.tensordot(sites.conj().T, sites, axes=[[-1, -2], [0, 1]])
```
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
multiphaseCFD: I think we can combine this two methods into one by a bit refactoring.

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: Done!

pennylane_lightning/lightning_tensor/_tensornet.py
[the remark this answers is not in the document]
LuisAlfredoNu: I used the following line to set the comparison precision:
```
atol=np.finfo(C.dtype).eps * 1e4
```
it is based on the precision machine and a gap to allow the numerical error from the SVD decomposition.

AmintorDusko (approved):
Nice job! Thank you!

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,193 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    return True
+
+
+def expand_mps_first_site(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
+    """Expand the MPS to match the size of the target wires.
+
+    This function modifies the original MPS state by adding a single wire at the beginning of the MPS state. The algorithm to expand the input MPS state to fit into the device MPS state is based on the following steps:
+
+    - Set the device MPS state as $B$ and the input MPS state as $A$.
+    - Padding with zeros the tensor $B_i$ to fit the tensor shape $A_{i+1}$ up to $i = N/2$ where $N$ is the total number of tensors in $B$.
```
multiphaseCFD, line 232: Thanks for adding more docstring here. It would be even better to add some mathematical explanation here, as it's still not that clear to me mathematically.  For example, it would be better to explain the math behind the statement here. It sounds like just matching up the shape but why and how this would change the underlying MPS data is not clearly clarified? `Padding with zeros the tensor $B_i$ to fit the tensor shape $A_{i+1}$ up to $i = N/2$ where $N$ is the total number of tensors in $B$.`

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,193 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+    Args:
+        mps (list[np.ndarray]): MPS state
+        site_shape (list[int]): shape of the site tensor
+
+    Returns:
+        list[np.ndarray]: MPS state in the right canonical form
+    """
+
+    new_mps = []
+    U = np.eye(1, dtype=mps[0].dtype)
```
multiphaseCFD, line 359: There are many `codecov` complaints around. Are those all false negative?

pennylane_lightning/lightning_tensor/_tensornet.py
```diff
@@ -123,6 +193,193 @@ def gate_matrix_decompose(gate_ops_matrix, wires, max_mpo_bond_dim, c_dtype):
...
+
+
+def expand_mps_first_site(state_MPS: list[np.ndarray], max_bond_dim: int = 128) -> list[np.ndarray]:
+    """Expand the MPS to match the size of the target wires.
+
+    This function modifies the original MPS state by adding a single wire at the beginning of the MPS state. The algorithm to expand the input MPS state to fit into the device MPS state is based on the following steps:
+
+    - Set the device MPS state as $B$ and the input MPS state as $A$.
+    - Padding with zeros the tensor $B_i$ to fit the tensor shape $A_{i+1}$ up to $i = N/2$ where $N$ is the total number of tensors in $B$.
+    - Add the identity matrix with shape `(1,2,2)` at the beginning of $B$.
```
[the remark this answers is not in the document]
LuisAlfredoNu: The Identity is to preserve the canonical form in case the input MPS has a left-canonical form. If the input MPS has a right-canonical form then the first element is transformed. Therefore, the first or last MPS in the expansion should be an identity matrix.

pennylane_lightning/lightning_tensor/_tensornet.py
LuisAlfredoNu: I do not have a formal mathematical proof.  However, the explanation is the following:
The tensor product $|0\rangle \otimes |\psi\rangle$ can expanded as follows:

$$
\begin{align*}
|0\rangle \otimes |\psi\rangle &= \left( 1|0\rangle + 0 |1\rangle\right) \otimes \sum_i \phi_i | \sigma_i\rangle \\
&= \sum_i \psi_i |0,\sigma_i\rangle + \sum_i 0 |1,\sigma_i\rangle \\
&=  \sum_i \psi_i |0, \sigma_i \rangle\\
& \text{if}\,  i\leq 2^{n-1} = \psi_i \\
&\text{else}\, i >  2^{n-1} = 0 \\
\end{align*}
$$

Therefore, we have a $\psi_i$ filled with half of zeros. Then, using the SVD decomposition will produce a $U$ or $Vd$ matrices, depending on the canonical form, with half of zeros too.  Following the algorithm to convert the SV to a MPS, show that the only values that contribute to each decomposition are the non-zeros resulting in MPS states that have also zeros. 

Consequently, the SVD decomposition of $|\psi\rangle$ and $|0\rangle \otimes |\psi\rangle$ will produce the same values for each site. The key to this algorithm was to find where and how to add these zeros on each site, which are in the first half of the MPS sites due to the first SVD decompositions and only need to fit the shape to the next site.

multiphaseCFD: Thanks @LuisAlfredoNu for more info. Let me know if we could put it this in the following way. The state vector passed to the `cutensornet` backend is:
$|psi \rangle = \|1 \rangle \otimes | \psi \rangle$ 
The `0th` site would be an identity matrix:
$|1 \rangle \langle 0| $
The rest sites would be:
$\langle 0| \otimes \psi_{0} \otimes \psi_{1} \otimes \... \otimes \psi_{n-1} \rangle $
So I'm wondering if we can just set the site next to 0th site as:
$\langle 0| \otimes \psi_{0} \rangle$
With that can we just pad the this site by another zeros and see if it works?

LuisAlfredoNu: This will not work because the sites after the operation $|0\rangle \otimes \Psi_0\rangle$ will have different bond dims that will not fit with the next $\Psi_1$

multiphaseCFD: > This will not work because the sites after the operation | 0 ⟩ ⊗ Ψ 0 ⟩ will have different bond dims that will not fit with the next Ψ 1

Could you please explain why the bonddim does not fit? The shape of the initial $|\psi_{0}\rangle$ is [2, bonddim]. Once $|\psi_{0}\rangle$ outer product with $|0\rangle$, the shape turns to `[2,2,bonddim]`. As in, it doesn't change the bonddim connecting $|\psi_{0}\rangle$ and $|\psi_{1}\rangle$. Feel free to correct me. Thanks

LuisAlfredoNu: If I understand your idea correctly, formally the outer product of  $|0\rangle \otimes \Psi_0\rangle$  will produce a tensor with shape [2,2,1,2,2] then we need to reshape it to 3 indexes which will produce [4, 2, 2] or [2,2,4]. However, the correct transformation should be [2,2,4] because we are trying to entangle $|0\rangle$ to $\Psi$. 

Additionally, the point of the MPS is to express the entanglement through different tensors and the bond dim, more bond dim more information of entanglement. If you crop the sites we will lose the entanglement information to the additional qubit $|0\rangle$ across the rest of the cubits and only the qubit $|\Psi_0\rangle$ will notice the addition of the new qubit.

multiphaseCFD: Sounds good to me. Thanks @LuisAlfredoNu . It would be nice to add a follow-up story to track the mathematical representation.

LuisAlfredoNu: The story has been added to the SC. @multiphaseCFD 
:sunglasses:

maliasadi: I have a draft proof for this to share in the story after our offline discussions yesterday. We'll get back to this :raised_hands:

maliasadi (approved):
Thanks @LuisAlfredoNu! Happy to approve the impl :raised_hands:

PR #1073, opened 2025-02-27
Adding new workflow to allow triggering benchmark from pr

rashidnhm:
**Context:**
Allow us to trigger benchmark from a pull request 

Adds a net new workflow which would enable us to benchmark a pull request.

Expected workflow:
- User with at least *write* permissions would need to comment `/benchmark` on a pull request
- Workflow triggers, checks the comment and users permission using GitHub API
- If checks pass, a webhook is sent to pennylane-benchmark repo
- The workflow will react to the original comment, indicating that webhook was sent and the workflow (scoped to lightning) has completed successfully.

On the Benchmarking side, currently this will trigger a place holder workflow, which sleeps for a few seconds and replies back with a comment of "results".

**Benefits:**
Ability to trigger benchmark from a PR.

**Possible Drawbacks:**
This is more scaffolding. The PL benchmarking team would need to update the workflow that receives this webhook and ensure benchmarks actually run.

**Related GitHub Issues:**
None. [sc-85092](https://app.shortcut.com/xanaduai/story/85092/setup-pl-benchmarks-trigger-for-prs-on-pennylane-lightning)

### DIFF
--- a/.github/workflows/trigger_benchmarks.yml
+++ b/.github/workflows/trigger_benchmarks.yml
@@ -1,0 +1,100 @@
+name: Trigger Benchmarks for Pull Request
+on:
+  issue_comment:
+    types:
+      - created
+
+env:
+  BENCHMARK_TRIGGER_COMMAND: /benchmark
+
+jobs:
+  trigger_benchmark_webhook:
+    runs-on: ubuntu-latest
+
+    # Only run for pull request comments (newly created)
+    if: github.event.issue.pull_request
+
+    steps:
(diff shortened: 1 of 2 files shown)


AmintorDusko (approved):
Great! Thanks! I assume we can only test it after merging, right?

rashidnhm:
> Great! Thanks! I assume we can only test it after merging, right?

Yup, but this workflow has been tested on lightning so hoping no issues arise

PR #1089, opened 2025-03-12
Implement PSWAP in Lightning-Kokkos

jzaia18:
**Context:**
PSWAP does not have a custom implementation within Lightning-Kokkos. This results in it using the implementation from PennyLane-Core whenever the gate is used. By implementing the gate directly in Lightning-Kokkos, performance of the PSWAP gate can be improved.

 Implements the PSWAP gate natively in Lightning-Kokkos.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-85797]


AmintorDusko (approved):
Nice work! Thank you for that! Let me know if you need any support here.

PR #1100, opened 2025-03-18
Add semgrep/bandit security scanning for pull requests

rashidnhm:
**Context:**
This pull request adds a new workflow to enable Code Security scanning to pennylane using [bandit](https://github.com/PyCQA/bandit) and [semgrep](https://semgrep.dev/).

Both tools use a pre-existing set of rules to scan the code heuristically and flag anything that could potentially be a security issue.

The workflow introduced uses a re-usable action introduced in XanaduAI/cloud-actions#67.

Bandit scans source code and creates an AST which is then matched against it's rules.
Semgrep uses heuristic matching, but is able to support multiple languages.

The configuration is setup so that CI will fail if there is an issue flagged by either tool with a severity of "HIGH".

**Benefits:**
Basic security code scanning introduced. 

**Possible Drawbacks:**
PRs can be blocked if a false positive HIGH severity issue is flagged. However, both tools provide ways to suppress false positives at the line or file level.

**Related GitHub Issues:**
None. [sc-86314]

### DIFF
--- a/.github/workflows/code_security_scan.yml
+++ b/.github/workflows/code_security_scan.yml
@@ -15,7 +15,7 @@ jobs:
     with:
       github-repository: ${{ github.repository }}
       repository-ref: ${{ github.ref }}
-      runner_name: pl-4-core-large-runner
+      runner_name: ubuntu-latest
       bandit-config-file: pyproject.toml
       scan_directory: pennylane_lightning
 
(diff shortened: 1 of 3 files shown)


mlxd (approved):
:+1:

LuisAlfredoNu (approved):
Thank you @rashidnhm 
A quick question. Why de we exclude tests directory?

rashidnhm:
> Thank you @rashidnhm A quick question. Why do we exclude tests directory?

Since the scope of the security issue is code that will execute when someone installs lightning from pypi. Unless you are a contributor, tests don't run.

But if the team *really wants* we can add tests to scope, but I don't feel the need to.

PR #1146, opened 2025-04-30
fix use of num_wires

albi3ro:
**Context:**

In [(#7312)](https://github.com/PennyLaneAI/pennylane/pull/7312) we switched the default of `Operator.num_wires` to `None` instead of a special enum value. The special enums were "int-like" and could be compared against ints.

Update handling of `Operator.num_wires` in tests.

**Benefits:**

CI passes again.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -32,8 +32,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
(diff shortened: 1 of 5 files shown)


maliasadi (approved):
Thanks @albi3ro! Please update the changelog before merge.

AmintorDusko (approved):

PR #1152, opened 2025-05-08
Updates for supporting JAX 0.5.3+

mudit2812:
* Replace `jax.core` with `jax.extend.core`.
* Change handling of dtype casting for `Zero` tangents.
* Update usage of `pure_callback` to remove the use of the deprecated `vectorized` argument.

[sc-90945]

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -32,6 +32,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Update type hinting to be compatible with JAX version 0.5.3+.
+  [(#1152)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1152)
+
 - The `LightningBaseStateVector`, `LightningBaseAdjointJacobian`, `LightningBaseMeasurements`, `LightningInterpreter` and `QuantumScriptSerializer` base classes now can be found at `pennylane_lightning.lightning_base`.
 
   The new `lightning_base` module further enables the relocation of core files from `pennylane_lightning/core/src/*` to `pennylane_lightning/core/*`.
@@ -71,6 +74,7 @@ Christina Lee,
 Joseph Lee,
 Anton Naim Ibrahim,
 Luis Alfredo Nuñez Meneses,
+Mudit Pandey,
 Andrija Paurevic,
 
 ---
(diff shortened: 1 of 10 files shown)


pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -590,11 +590,11 @@ def shape(var):
             return var.aval
 
         shapes = [shape(var) for var in jaxpr.outvars]
-        return jax.pure_callback(evaluator, shapes, consts, *args, vectorized=False)
+        return jax.pure_callback(evaluator, shapes, consts, *args, vmap_method="sequential")
```
josephleekl, line 593: FYI I ran the tests locallyl - these codecov warnings are false, the lines are in fact covered

.github/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 <h3>Improvements 🛠</h3>
 
+- PennyLane-Lightning is compatible with JAX version 0.5.3+.
```
josephleekl, line 7: Does this support all versions >0.5.3? Will we need another upgrade for e.g. 0.6.0 or above? 
```suggestion
- PennyLane-Lightning is compatible with JAX version 0.5.3.
```

PietropaoloFrisoni: > Does this support all versions >0.5.3? Will we need another upgrade for e.g. 0.6.0 or above?

This is actually a WIP in PennyLane : )

PietropaoloFrisoni (approved):
I agree with @josephleekl's suggestion since here we are simply testing JAX 0.5.3, but happy to approve otherwise : )

Edit: actually, I am 99% sure this will work with JAX 0.6.0 as well

PietropaoloFrisoni:
This PR is needed in Lightning for the CI on [this PR](https://github.com/PennyLaneAI/pennylane/pull/7299) to be green

pennylane_lightning/lightning_base/lightning_base.py
mudit2812: I find that to usually be the case with codecov failures whenever I push to lightning 😆

.github/CHANGELOG.md
mudit2812: My instinct is to say that JAX 0.6.0 should just work with these changes, but I also know about Murphy's Law 😆.

I suggest we keep this entry as is, and if more changes are needed, I can open another PR, after which this changelog entry should be valid anyway.

PR #1156, opened 2025-05-12
Fix Docker container for ROCM

LuisAlfredoNu:
**Context:**
The Docker image for Lightning-Kokkos is [failing](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/14987894472) due to linux-headers-6.11.0-1012-azure incompatibility on this OS. We used to build the image against linux-headers-6.8.0-1021-azure

The solution is updating the Ubuntu version from 22.04 to 24.04
Add the CMAKE argument 
```
DCMAKE_CXX_FLAGS='--gcc-install-dir=/usr/lib/gcc/x86_64-linux-gnu/11/'
```
when using `hipcc` compiler.

Correct build for Docker [images](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/14984154575)

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-91269]

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,9 +15,9 @@
 # Define global build defaults
 ARG PENNYLANE_VERSION=master
 
-# Create basic runtime environment base on Ubuntu 22.04 (jammy)
+# Create basic runtime environment base on Ubuntu 24.04
 # Create and activate runtime virtual environment
-FROM ubuntu:jammy AS base-runtime
+FROM ubuntu:24.04 AS base-runtime
 ARG AMD_ARCH=AMD_GFX90A
 ARG CUDA_ARCH=AMPERE80
 ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.2.2/local_installers/cuda_12.2.2_535.104.05_linux.run
@@ -25,7 +25,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG GCC_VERSION=11
 ARG LIGHTNING_VERSION=master
 ARG PENNYLANE_VERSION
-ARG ROCM_INSTALLER=https://repo.radeon.com/amdgpu-install/5.7/ubuntu/jammy/amdgpu-install_5.7.50700-1_all.deb
+ARG ROCM_INSTALLER=https://repo.radeon.com/amdgpu-install/6.4/ubuntu/noble/amdgpu-install_6.4.60400-1_all.deb
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
     apt-utils \
@@ -123,7 +123,7 @@ RUN CMAKE_ARGS="-DKokkos_ENABLE_SERIAL:BOOL=ON -DKokkos_ENABLE_OPENMP:BOOL=ON -D
 
 # Install python3 and setup runtime virtual env in CUDA-12-runtime image (includes CUDA runtime and math libraries)
 # Install lightning-kokkos CUDA backend
-FROM nvidia/cuda:12.2.0-base-ubuntu22.04 AS wheel-lightning-kokkos-cuda
+FROM nvidia/cuda:12.2.0-base-ubuntu24.04 AS wheel-lightning-kokkos-cuda
 ARG PENNYLANE_VERSION
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -178,10 +178,12 @@ RUN pip install --no-cache-dir git+https://github.com/PennyLaneAI/pennylane.git@
 # Install ROCm in build venv image
 FROM base-build-python AS base-build-rocm
 RUN wget --progress=dot:giga ${ROCM_INSTALLER}
+# FIXME: the linux-headers-6.11.0-1021-azure package is not available in the repo then is hardcoded
+# to linux-headers-6.8.0-1021-azure, this should be fixed in the future
```
maliasadi, line 182: Is there any plans for this library to be included in the repo? If not, I wonder if this FIXME makes much sense and can be removed here.

docker/Dockerfile
```diff
@@ -178,10 +178,12 @@ RUN pip install --no-cache-dir git+https://github.com/PennyLaneAI/pennylane.git@
...
 RUN apt-get update \
-    && apt-get install --no-install-recommends -y \
-    "linux-headers-$(uname -r)" "linux-modules-extra-$(uname -r)" \
-    ./$(echo ${ROCM_INSTALLER} | xargs -I {} basename {}) \
+    && apt-get install --no-install-recommends -y "linux-headers-$(uname -r)" \
+    && apt-get install --no-install-recommends -y "linux-modules-extra-$(uname -r)" \
+    && apt-get install --no-install-recommends -y ./$(echo ${ROCM_INSTALLER} | xargs -I {} basename {}) \
```
maliasadi, line 186: would you like to revert this change in favour of installing all these packages at once in one apt call?

docker/Dockerfile
```diff
@@ -194,10 +196,10 @@ ENV CXX=hipcc
 ENV PL_BACKEND=lightning_kokkos
 RUN pip uninstall -y pennylane-lightning
 RUN python scripts/configure_pyproject_toml.py || true
-RUN CMAKE_ARGS="-DKokkos_ENABLE_SERIAL:BOOL=ON -DKokkos_ENABLE_OPENMP:BOOL=ON -DKokkos_ENABLE_HIP:BOOL=ON -DKokkos_ARCH_${AMD_ARCH}=ON" python -m build --wheel
+RUN CMAKE_ARGS="-DKokkos_ENABLE_SERIAL:BOOL=ON -DKokkos_ENABLE_OPENMP:BOOL=ON -DKokkos_ENABLE_HIP:BOOL=ON -DKokkos_ARCH_${AMD_ARCH}=ON -DCMAKE_CXX_FLAGS='--gcc-install-dir=/usr/lib/gcc/x86_64-linux-gnu/11/'" python -m build --wheel
```
maliasadi, line 199: Why is this required? What sort of error messaged would be resulted here otherwise?

maliasadi (commented):
Thanks @LuisAlfredoNu! Just a few minor comments before approving the PR

docker/Dockerfile
josephleekl: It's the same error message as I saw on Frontier: https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/14983467031/job/42092704492#step:5:9900

and this is the same fix I used there.

josephleekl: I think it's because we're upgrading ROCm from 5.7 to 6.3.4, and the compiler it ships with is also updated

josephleekl: @LuisAlfredoNu why did you choose 6.3.4?

docker/Dockerfile
```diff
@@ -15,17 +15,17 @@
...
 
-# Create basic runtime environment base on Ubuntu 22.04 (jammy)
+# Create basic runtime environment base on Ubuntu 24.04
 # Create and activate runtime virtual environment
-FROM ubuntu:jammy AS base-runtime
+FROM ubuntu:24.04 AS base-runtime
 ARG AMD_ARCH=AMD_GFX90A
 ARG CUDA_ARCH=AMPERE80
-ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.2.2/local_installers/cuda_12.2.2_535.104.05_linux.run
+ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.9.0/local_installers/cuda_12.9.0_575.51.03_linux.run
```
josephleekl, line 23: why are we updating to 12.9? what's the minimum required?

docker/Dockerfile
LuisAlfredoNu: I will removed

docker/Dockerfile
LuisAlfredoNu: Sure! 
However, I think is better to keep 
```
apt-get install --no-install-recommends -y ./$(echo ${ROCM_INSTALLER} | xargs -I {} basename {})
``` 
as an independent line for readability

docker/Dockerfile
LuisAlfredoNu: Because the ROCM docker images don't have support for 5.7 with Ubuntu 24.04[list](https://hub.docker.com/r/rocm/dev-ubuntu-24.04/tags) and, silly answer, that was the version with higher minor changes, which, I would like to believe, have more corrections thus more stable

docker/Dockerfile
LuisAlfredoNu: Because the Nvidia Docker hub don't have support for cuda 12.2 with Ubutnu 24.4, [list](https://hub.docker.com/r/nvidia/cuda/tags?name=0-base-ubuntu24), then the newest is 12.9

Issue #987, opened 2024-11-07
`seed` in `qml.device("lightning.qubit", ...)` seemingly does nothing


isaacdevlugt:
#### Issue description

The `seed` kwarg in `device` doesn't do anything for `lightning.qubit`. The results below are not fixed by `seed`.

* *Expected behavior:* Since all sampling behaviour in lightning qubit happens in the C++ layer and PennyLane doesn't use the seed argument to configure that at all, lightning qubit shouldn't accept a `seed` keyword argument (it should be removed).

* *Actual behavior:* It accepts `seed` and does nothing with it.

* *Reproduces how often:* 1000%

* *System information:* 

```
Name: PennyLane
Version: 0.39.0
Summary: PennyLane is a cross-platform Python library for quantum computing, quantum machine learning, and quantum chemistry. Train a quantum computer the same way as a neural network.
Home-page: https://github.com/PennyLaneAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: [/Users/isaac/.virtualenvs/pennylane-catalyst/lib/python3.11/site-packages](https://file+.vscode-resource.vscode-cdn.net/Users/isaac/.virtualenvs/pennylane-catalyst/lib/python3.11/site-packages)
Requires: appdirs, autograd, autoray, cachetools, networkx, numpy, packaging, pennylane-lightning, requests, rustworkx, scipy, toml, typing-extensions
Required-by: PennyLane-Catalyst, PennyLane_Lightning

Platform info:           macOS-15.1-arm64-arm-64bit
Python version:          3.11.9
Numpy version:           1.26.4
Scipy version:           1.12.0
Installed devices:
- default.clifford (PennyLane-0.39.0)
- default.gaussian (PennyLane-0.39.0)
- default.mixed (PennyLane-0.39.0)
- default.qubit (PennyLane-0.39.0)
- default.qutrit (PennyLane-0.39.0)
- default.qutrit.mixed (PennyLane-0.39.0)
- default.tensor (PennyLane-0.39.0)
- null.qubit (PennyLane-0.39.0)
- reference.qubit (PennyLane-0.39.0)
- nvidia.custatevec (PennyLane-Catalyst-0.9.0)
- nvidia.cutensornet (PennyLane-Catalyst-0.9.0)
- oqc.cloud (PennyLane-Catalyst-0.9.0)
- softwareq.qpp (PennyLane-Catalyst-0.9.0)
[...]

PR #1171, opened 2025-05-29
Set Random seed for Lightning devices

josephleekl:
### Before submitting

**Context:**
Currently the shots measurement seed can be set from C++, which is already used by catalyst. However, there is no Python binding to set the seed from Python.

This PR adds support to set `seed` from setting `qml.device(seed=...)`, similar to other Pennylane devices. This will enable deterministic results from shots measurements.

**Benefits:**
Deterministic shots measurements will allow us to remove flaky tests.

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane-lightning/issues/987

[sc-64459] [sc-77742]

### DIFF
--- a/pennylane_lightning/lightning_base/lightning_base.py
+++ b/pennylane_lightning/lightning_base/lightning_base.py
@@ -88,8 +88,9 @@ class LightningBase(Device):
 
         self._c_dtype = c_dtype
         self._batch_obs = batch_obs
-        self._seed = np.random.randint(2**31 - 1) if seed == "global" else seed
-        self._rng = np.random.default_rng(self._seed)
+        self._rng = np.random.default_rng(
+            np.random.randint(2**31 - 1) if seed == "global" else seed
+        )
 
         # State-vector is dynamically allocated just before execution
         self._statevector = None


.github/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 <h3>Improvements 🛠</h3>
 
+- Lightning devices accepts a `seed` argument to enable deterministic shots measurements.
```
maliasadi, line 7:
```suggestion
- Lightning devices accept a `seed` argument to enable deterministic shots measurements.
```

pennylane_lightning/core/bindings/Bindings.hpp
```diff
@@ -474,23 +474,27 @@ void registerBackendAgnosticMeasurements(PyClass &pyclass) {
...
+                     result.data(), /* data as contiguous array  */
+                     sz,            /* size of one scalar        */
+                     py::format_descriptor<std::size_t>::format(), /* data type
+                                                                    */
+                     ndim,   /* number of dimensions      */
+                     shape,  /* shape of the matrix       */
+                     strides /* strides for each axis     */
+                     ));
+             })
+        .def("set_seed", [](Measurements<StateVectorT> &M, std::size_t seed) {
```
maliasadi, line 496: hmm.. `set_seed` is somewhat vague! In ML libraries, `set_seed` is typically used from `random` or `PRNG` modules, which makes the purpose of the method more explicit, I reckon. Catalyst also uses `seed` for simplicity. For the simulator bindings, we could be more explicit and use a name like `set_random_seed` or `set_rand_seed`. :thinking:

tests/test_native_mcm.py
```diff
@@ -355,3 +362,42 @@ def func(x):
...
+
+
+@pytest.mark.parametrize("shots", [20, [20, 20]])
+@pytest.mark.parametrize("postselect", [None, 0, 1])
+@pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+@pytest.mark.parametrize(
+    "meas_obj",
+    [qml.PauliZ(0), qml.PauliY(1), [0], [0, 1], [1, 0], "mcm", "composite_mcm", "mcm_list"],
+)
+def test_seeded_mcm(shots, measure_f, postselect, meas_obj):
```
maliasadi, line 374: Do you have a plan to fix the MCM stochastic tests in this PR, as the ultimate use case of this new feature?

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -71,12 +78,15 @@ def __init__(  # pylint: disable=too-many-arguments
...
         shots: Union[int, List],
+        seed: Union[str, None, int, ArrayLike, SeedSequence, BitGenerator, Generator],
         batch_obs: bool,
     ):
         super().__init__(wires=wires, shots=shots)
 
         self._c_dtype = c_dtype
         self._batch_obs = batch_obs
+        self._seed = np.random.randint(2**32 - 1) if seed == "global" else seed
+        self._rng = np.random.default_rng(self._seed)
```
maliasadi, line 89: Is this consistent with PL?

pennylane_lightning/lightning_qubit/_measurements.py
```diff
@@ -66,8 +66,8 @@ def __init__(
             self._kernel_name = "Local"
         if self._mcmc and not self._num_burnin:
             self._num_burnin = 100
-
         self._measurement_lightning = self._measurement_dtype()(qubit_state.state_vector)
+        self._measurement_lightning.set_seed(qubit_state._rng.integers(0, 2**32 - 1))
```
maliasadi, line 70: shouldn't this be rather optional? :thinking:

maliasadi (commented):
Thanks @josephleekl! All good! Just a few comments...

.github/CHANGELOG.md
josephleekl: updated

pennylane_lightning/core/bindings/Bindings.hpp
josephleekl: that's a good point. Let's go with `set_random_seed`

tests/test_native_mcm.py
josephleekl: I'm planning on having that as a separate PR, and have this branch as the base.

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: now matched with PL https://github.com/PennyLaneAI/pennylane/blob/c76f9b3aa2abc1c5c2116d66b7ed5722e71d4051/pennylane/devices/default_qubit.py#L485C16-L485C51

pennylane_lightning/lightning_qubit/_measurements.py
josephleekl: We can make this optional. The thing is that to be consistent with default.qubit, we have `seed="global"` as the default during device initialization, which says let numpy draw a random seed. If we have are happy to have `seed=None` as default, we can make this optional. Do we want to:
- keep `seed="global"` as default, and make this seed setting mandatory, or
- set `seed=None` as default, so we only explicitly set the seed when users ask for it?

I'm happy with either option, probably more towards option 2.

pennylane_lightning/core/bindings/Bindings.hpp
```diff
@@ -474,24 +474,27 @@ void registerBackendAgnosticMeasurements(PyClass &pyclass) {
...
+                 auto &&result = M.generate_samples(num_shots);
+                 const std::size_t ndim = 2;
+                 const std::vector<std::size_t> shape{num_shots, num_wires};
+                 constexpr auto sz = sizeof(std::size_t);
+                 const std::vector<std::size_t> strides{sz * num_wires, sz};
+                 // return 2-D NumPy array
+                 return py::array(py::buffer_info(
+                     result.data(), /* data as contiguous array  */
+                     sz,            /* size of one scalar        */
+                     py::format_descriptor<std::size_t>::format(), /* data type
```
jzaia18, line 489: `make format` being weird here?

tests/lightning_base/test_measurements_class.py
```diff
@@ -847,6 +848,83 @@ def circuit():
         results = qml.QNode(circuit, dev)()
         assert np.allclose(expected, results, tol)
 
+    @pytest.mark.skipif(
+        device_name in ("lightning.tensor"),
+        reason=f"{device_name} does not support seeding device.",
+    )
```
jzaia18, line 857: Any reason not to support seeding for `lightning.tensor` as well?

jzaia18 (commented):
Looking good! Just curious if there's any reason for not supporting the same for `lightning.tensor`

tests/lightning_base/test_measurements_class.py
josephleekl: the measurement class for tensor is different from the rest of the lightning devices, and does not have a seed member. The measurements there afaik do not currently support seeding, but maybe in the future we can add it.

pennylane_lightning/core/bindings/Bindings.hpp
josephleekl: yeah - make format pushes it to a new line

jzaia18 (approved):
Looking good, nice work!

pennylane_lightning/core/bindings/Bindings.hpp
```diff
@@ -493,7 +493,7 @@ void registerBackendAgnosticMeasurements(PyClass &pyclass) {
                      strides /* strides for each axis     */
                      ));
              })
-        .def("set_seed", [](Measurements<StateVectorT> &M, std::size_t seed) {
+        .def("set_random_seed", [](Measurements<StateVectorT> &M, std::size_t seed) {
```
mlxd: Do we know that this function name change doesn't affect any internal/external people?

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -85,7 +85,7 @@ def __init__(  # pylint: disable=too-many-arguments
 
         self._c_dtype = c_dtype
         self._batch_obs = batch_obs
-        self._seed = np.random.randint(2**32 - 1) if seed == "global" else seed
+        self._seed = np.random.randint(0, high=10000000) if seed == "global" else seed
```
mlxd: So we pull a random value, and then seed the generator with that?
Why limit this to 1e7?

pennylane_lightning/lightning_kokkos/_state_vector.py
```diff
@@ -63,10 +64,11 @@ def __init__(
         self,
         num_wires: int,
         dtype: Union[np.complex128, np.complex64] = np.complex128,
+        rng: Generator = None,
```
mlxd, line 67: Should the Kokkos RNG not defer to an (exposed) internal Kokkos implementation? https://kokkos.org/kokkos-core-wiki/API/algorithms/Random-Number.html

That way we get the GPU doing RNG when requested, rather than the CPU

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -213,6 +220,7 @@ def __init__(  # pylint: disable=too-many-arguments
         c_dtype: Union[np.complex128, np.complex64] = np.complex128,
         shots: Union[int, List] = None,
         batch_obs: bool = False,
+        seed: Union[str, None, int, ArrayLike, SeedSequence, BitGenerator, Generator] = "global",
```
mlxd, line 230: The positioning of this argument isn't going to cause any issues?
Should it come last in the sequence?

tests/conftest.py
```diff
@@ -217,7 +217,7 @@ def _statevector(num_wires):
     return _statevector
 
 
-def validate_counts(shots, results1, results2):
+def validate_counts(shots, results1, results2, rtol=20, atol=0.2):
```
mlxd: Is 20 not a bit too high? Since this is being updated now, can we reduce this tolerance to something reasonable?

tests/conftest.py
```diff
@@ -235,10 +235,10 @@ def validate_counts(shots, results1, results2):
     for key1, val1 in results1.items():
         val2 = results2[key1]
         if abs(val1 + val2) > 100:
-            assert np.allclose(val1, val2, rtol=20, atol=0.2)
+            assert np.allclose(val1, val2, rtol=rtol, atol=atol)
 
 
-def validate_samples(shots, results1, results2):
+def validate_samples(shots, results1, results2, rtol=20, atol=0.2):
```
mlxd: Same here as above?

tests/lightning_base/test_measurements_class.py
```diff
@@ -847,6 +848,83 @@ def circuit():
...
+            result_1 = []
+            result_2 = []
+            result_3 = []
+            for i in range(20):
+                m_1 = LightningMeasurements(statevector1)
+                m_2 = LightningMeasurements(statevector2)
+                m_3 = LightningMeasurements(statevector3)
+                result_1.append(measure_final_state(m_1, tape))
+                result_2.append(measure_final_state(m_2, tape))
+                result_3.append(measure_final_state(m_3, tape))
```
mlxd: Is there a way we can test this seeding works without explicit executions and measurements?

pennylane_lightning/core/bindings/Bindings.hpp
josephleekl: this python binding is new, so it shouldn't affect anyone

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: Indeed that's the idea. The value itself is for consistency with Pennylane's default devices e.g. [here](https://github.com/PennyLaneAI/pennylane/blob/7ee40a550fdb2ad156a1ab17435ba2ed575f2ef2/pennylane/devices/default_qubit.py#L485)

pennylane_lightning/lightning_kokkos/_state_vector.py
josephleekl: It could be useful to expose the Kokkos RNG, but I see some drawbacks:
- The only use case for this RNG right now is to seed the Kokkos RNG, so we're effectively just drawing one number. I wonder if the overhead of drawing on GPU then copying to host, then back to device would give any performance benefit.
- The current numpy RNG implementation is more unified across the devices.

I agree if we have cases where we need lots of random numbers on device, it makes sense to have a proper RNG class (from Kokkos for LK and cuRand for LG?) that we expose to Python. Does that make sense?

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
josephleekl: I don't think it'll cause any issue, it's a keyword argument after the *

tests/conftest.py
josephleekl: rtol=20 for counts is definitely too high - I have updated it to rtol=0.15 and atol=20, which seems reasonable. Pushing lower I'll start to get errors with the number of shots currently set.

tests/conftest.py
josephleekl: updated to rtol=0.15 , atol=20

tests/lightning_base/test_measurements_class.py
josephleekl: I have added a [separate test](https://github.com/PennyLaneAI/pennylane-lightning/blob/aa3c9d3e1da8f97a0fa40dd264fbae0fad44940e/tests/lightning_base/test_measurements_class.py#L856) to directly compare the RNG state . 

I think this test is still useful to guarantee the outputs are identical, and takes about 3s. Happy to remove/shorten this if it's helpful

pennylane_lightning/lightning_base/lightning_base.py
mlxd: I'm not sure that setting is a good precedent --- I'd rather we use the full limit we are allowed here.

pennylane_lightning/lightning_kokkos/_state_vector.py
mlxd: Sounds reasonable :+1:

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: Sure - I'll move to `2**31-1` to the maximum for 32bit dtype=long

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -74,12 +81,15 @@ def __init__(  # pylint: disable=too-many-arguments
         *,
         c_dtype: Union[np.complex64, np.complex128],
         shots: Union[int, List],
+        seed: Union[str, None, int, ArrayLike, SeedSequence, BitGenerator, Generator],
```
LuisAlfredoNu, line 84: Could you add test for each of these option in `lightning_base/test_device.py`, pls? Also I think we need a exception raise here for `seed:str != "global"`

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -593,7 +603,7 @@ def loop(i, y):
             raise NotImplementedError("Wires must be specified for integration with plxpr capture.")
 
         self._statevector = self.LightningStateVector(
-            num_wires=len(self.wires), dtype=self._c_dtype
+            num_wires=len(self.wires), dtype=self._c_dtype, rng=self._rng
```
LuisAlfredoNu, line 607: Do you think it's necessary to add the rng in the L685?

tests/lightning_base/test_measurements_class.py
```diff
@@ -847,6 +848,107 @@ def circuit():
...
+        assert statevector1._rng.bit_generator.state == statevector2._rng.bit_generator.state
+        assert statevector1._rng.bit_generator.state != statevector3._rng.bit_generator.state
+
+    @pytest.mark.skipif(
+        device_name in ("lightning.tensor"),
+        reason=f"{device_name} does not support seeding device.",
+    )
+    @pytest.mark.parametrize("dtype", [np.complex64, np.complex128])
+    @pytest.mark.parametrize("shots", [10, [10, 10]])
+    @pytest.mark.parametrize("measurement", [qml.expval, qml.probs, qml.var])
```
LuisAlfredoNu: Could you add `qml.counts` and `qml.samples` for a specific wire and for all the wires, pls?

tests/test_native_mcm.py
```diff
@@ -587,3 +594,41 @@ def func(x):
...
             assert r1 == r2
+
+    @pytest.mark.parametrize("shots", [20, [20, 20]])
+    @pytest.mark.parametrize("postselect", [None, 0, 1])
+    @pytest.mark.parametrize("measure_f", [qml.counts, qml.expval, qml.probs, qml.sample, qml.var])
+    @pytest.mark.parametrize(
+        "meas_obj",
+        [qml.PauliZ(0), qml.PauliY(1), [0], [0, 1], [1, 0], "mcm", "composite_mcm", "mcm_list"],
+    )
+    def test_seeded_mcm(self, shots, measure_f, postselect, meas_obj):
```
LuisAlfredoNu: Could you add for the other two mcm methods pls? That will not take long time. 
`tree-traversal` and `deferred`

tests/test_native_mcm.py
```diff
@@ -587,3 +594,41 @@ def func(x):
...
+
+        results1 = qml.QNode(func, dev_1, mcm_method="one-shot")(*params)
+        results2 = qml.QNode(func, dev_2, mcm_method="one-shot")(*params)
+
+        if measure_f is qml.counts:
+            validate_counts(shots, results1, results2, rtol=1e-05, atol=1e-08)
+        elif measure_f is qml.sample:
+            validate_samples(shots, results1, results2, rtol=1e-05, atol=1e-08)
+        else:
+            validate_others(shots, results1, results2, rtol=1e-05, atol=1e-08)
```
LuisAlfredoNu: Could you use `validate_measurements` instead of all this logics. For `counts` and `sample` has more sense use `atol=0` because the measures are integers

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: added

tests/test_native_mcm.py
josephleekl: I can change the atol/rtol = 0, but I can't change to validate_measurements because I cannot set the rtol/atol for all the functions in one go with validate_measurements.

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: I have now added tests to test_device.py . I don't need to to raise an exception, numpy already does that when supplying a seed that is of the wrong type (e.g. a string).

tests/test_native_mcm.py
josephleekl: added

tests/lightning_base/test_measurements_class.py
josephleekl: added counts and samples, also with wires. However I have removed the tests for different rng seeds - it's very difficult to test that the results diverge in all cases reliably.

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -74,12 +81,15 @@ def __init__(  # pylint: disable=too-many-arguments
...
         c_dtype: Union[np.complex64, np.complex128],
         shots: Union[int, List],
+        seed: Union[str, None, int, ArrayLike, SeedSequence, BitGenerator, Generator],
         batch_obs: bool,
     ):
         super().__init__(wires=wires, shots=shots)
 
         self._c_dtype = c_dtype
         self._batch_obs = batch_obs
+        self._seed = np.random.randint(2**31 - 1) if seed == "global" else seed
```
LuisAlfredoNu:
```suggestion
        _seed = np.random.randint(2**31 - 1) if seed == "global" else seed
```
If we not longer use `self._seed` it is better to no have it as member variable :smile:

pennylane_lightning/lightning_gpu/_measurements.py
```diff
@@ -67,6 +67,8 @@ def __init__(
             self._num_local_wires = qubit_state._mpi_handler.num_local_wires
 
         self._measurement_lightning = self._measurement_dtype()(qubit_state.state_vector)
+        if qubit_state._rng:
+            self._measurement_lightning.set_random_seed(qubit_state._rng.integers(0, 2**31 - 1))
```
LuisAlfredoNu, line 71: Won't affect this definition to the MPI version? :thinking:

LuisAlfredoNu (approved):
Thank you @josephleekl 
Nice work. I left you a suggestion and a question. 
:seedling:

pennylane_lightning/lightning_base/lightning_base.py
josephleekl: updated

pennylane_lightning/lightning_gpu/_measurements.py
josephleekl: It's ok for MPI - it simply has no effect since the sample generation for GPU MPI doesn't use the seed

PR #1172, opened 2025-06-03
Remove flaky tests and add local test seeding mechanism

josephleekl:
### Before submitting

**Context:**
Indeterministic tests (probabilistic tests from shots), even with using flaky test infrastructure, can some time trigger failures in CI, and requires re-running, which increases testing time.

Similar to work in [Pennylane](https://github.com/PennyLaneAI/pennylane/pull/6435) , we add the ability to control the seed for tests. The seed can be updated regularly to make sure we're not using lucky seeds. For relevant tests, we pass the seed to both the default and/or lightning devices to ensure we get deterministic results.

We also remove flaky tests. 

**Benefits:**
Reduced testing time and reduce CI failures.

**Possible Drawbacks:**
Need to regularly update rng_salt.

**Related GitHub Issues:**

[sc-85986]


.github/CHANGELOG.md
```diff
@@ -37,6 +37,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Remove flaky tests and add random seed to measurement tests.
+  [(#11XX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/11XX)
```
```suggestion
  [(#1172)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1172)
```

tests/conftest.py
```diff
@@ -207,15 +207,46 @@ def _device(wires, shots=None):
...
-        return LightningStateVector(num_wires=num_wires, dtype=request.param)
+        if seed:
+            rng = np.random.default_rng(seed)
+            return LightningStateVector(num_wires=num_wires, dtype=request.param, rng=rng)
+        _statevector = LightningStateVector(num_wires=num_wires, dtype=request.param)
 
     return _statevector
 
+@pytest.fixture(autouse=True)
+def restore_global_seed():
```
josephleekl, line 224: these fixtures are same as the ones introduced in [PL](https://github.com/PennyLaneAI/pennylane/pull/6435/files)

tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    local_salt(salt): adds a salt to the seed provided by the pytest-rng fixture
+rng_salt = v0.41.0
```
josephleekl:
```suggestion
rng_salt = v0.41.1
```

josephleekl:
I won't trigger the GPU tests for now, since testing still take very long.

tests/test_native_mcm.py
```diff
@@ -240,19 +239,19 @@ def func(x, y, z):
 
     results1 = qml.QNode(func, dev, mcm_method="one-shot")(*params)
     results2 = qml.QNode(func, dq, mcm_method="deferred")(*params)
-
-    validate_measurements(measure_f, shots, results1, results2)
+    validate_measurements(measure_f, shots, results1, results2, atol=0.02)
```
josephleekl: I needed to increases the `atol` a bit here (from 0.01) to compensate for the error (which previously we used flaky to hide)

tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    local_salt(salt): adds a salt to the seed provided by the pytest-rng fixture
+rng_salt = v0.41.1
```
jzaia18, line 4: Should this be changed with each new PL release? If so, is this documented anywhere?

tests/test_native_mcm.py
```diff
@@ -200,15 +199,15 @@ def obs_tape(x, y, z, reset=False, postselect=None):
     return m0, m1
 
 
-@flaky(max_runs=5)
+@pytest.mark.local_salt(42)
```
jzaia18: What's the purpose of using local salt for these tests?

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
```diff
@@ -1504,7 +1504,7 @@ template <typename TypeList> void testHamiltonianObsExpvalShot() {
 
         DYNAMIC_SECTION("With shots_range "
                         << StateVectorToName<StateVectorT>::name) {
-            std::size_t num_shots = 20000;
+            std::size_t num_shots = 100000;
```
josephleekl, line 1507: Previously, this failed 1 in 1000 times. After updating this failed 0 in 1000 times.

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
```diff
@@ -1556,7 +1556,7 @@ template <typename TypeList> void testHamiltonianObsExpvalShot() {
 
             auto ob = Hamiltonian<StateVectorT>::create({0.5, 0.5}, {Y0, Her1});
 
-            std::size_t num_shots = 100000;
+            std::size_t num_shots = 500000;
```
josephleekl, line 1559: Previously, this failed 21 in 1000 times. This has been seen in CI multiple times e.g. [1](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/14914177741/job/41895610271?pr=1153) [2](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/15327456521/job/43125510116#step:5:481) . Now it fails 0 in 1000 times.

tests/pytest.ini
josephleekl: Yeah I believe this should be changed with every release - I agree it should be added to some documentation. Should this go in the release guildeline @maliasadi ?

tests/test_native_mcm.py
josephleekl: so the seed is a combination of the rng_salt + local_salt . So we can adjust globally the rng_salt for all tests that uses seed, and if required can adjust for the local test too using the local_salt.

jzaia18: Gotcha, so has this just been added to tests which give a "bad" result with the default salt?

josephleekl: these are just there so we have more fine grained control over the tests when it fails in the future

tests/conftest.py
```diff
@@ -238,14 +271,14 @@ def validate_counts(shots, results1, results2, rtol=0.15, atol=20):
             assert np.allclose(val1, val2, rtol=rtol, atol=atol)
 
 
-def validate_samples(shots, results1, results2, rtol=0.15, atol=20):
+def validate_samples(shots, results1, results2, rtol=0.15, atol=30):
```
josephleekl, line 279: Because we have removed flaky, the tolerance has to be slightly increased for native_mcm tests

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
maliasadi: What impact does this change to `num_shots` have on execution time?

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
maliasadi: Such a dramatic update :smile: 
What impact does this change to `num_shots` have on execution time? I had hoped that modifying the circuit would resolve the problem

tests/pytest.ini
maliasadi: Yes, this should be added to the release guideline. I added it as a TODO to [this PR](https://github.com/PennyLaneAI/guidance-docs/pull/65).

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -483,7 +483,7 @@ def supports_derivatives(
             return True
         return _supports_adjoint(circuit=circuit)
 
-    def simulate(
+    def simulate(  # pylint: disable=arguments-differ
```
maliasadi, line 486: ha! Is this relevant for other simulators?

maliasadi (commented):
Thanks @josephleekl

pennylane_lightning/lightning_gpu/lightning_gpu.py
josephleekl: it only affects LGPU because it has an overriden `simulate` method, the others just use the base class method.

It's odd that codefactor only complained in this PR, but not earliler PRs - this discrepancy was from a while ago!

josephleekl:
Thanks @maliasadi !

> What impact does this change to num_shots have on execution time?

With both of the changes to num_shots in the cpp tests, the test time increased by 0.9s in total locally which is not very significant

tests/conftest.py
```diff
@@ -271,14 +271,14 @@ def validate_counts(shots, results1, results2, rtol=20, atol=0.2):
...
     """Compares two samples.
 
     If the results are ``Sequence``s, loop over entries.
 
-    Fails if the results do not have the same shape, within 15 percent plus ``20`` entries.
+    Fails if the results do not have the same shape, within 15 percent plus ``30`` entries.
     This is to handle cases when post-selection yields variable shapes.
-    Otherwise, fails if the sums of samples differ by more than 15 percent plus ``20``.
+    Otherwise, fails if the sums of samples differ by more than 15 percent plus ``30``.
```
jzaia18, line 286: Is there a principled reason for the 15% + 30 number, or does this just seem good in testing?

jzaia18 (approved):
Looks good, nice work!

tests/conftest.py
josephleekl: It's from empirical testing, i need the tests to pass for a reasonable range of seeds, having it at just 20 will make it very sensitive to seeds

tests/pytest.ini
LuisAlfredoNu: I suggest to create a GitHub Action that updates this value each certain time. e.g. each 15 days. Update this value each release looks a bit dangerous :confused:

tests/lightning_base/test_measurements_class.py
```diff
@@ -712,11 +711,11 @@ def test_single_return_value(self, shots, measurement, observable, lightning_sv,
             )
         else:
             # TODO Set better atol and rtol
-            dtol = max(tol, 1.0e-2)
+            dtol = max(tol, 2.0e-2)
             # allclose -> absolute(a - b) <= (atol + rtol * absolute(b))
             assert np.allclose(result, expected, rtol=dtol, atol=dtol)
 
-    @flaky(max_runs=5)
+    @pytest.mark.local_salt(42)
```
LuisAlfredoNu, line 718: Why do we need a new salt here? :thinking:

requirements.txt
```diff
@@ -7,6 +7,7 @@ scipy-openblas32>=0.3.26
 pytest>=7.1.2
 pytest-cov>=3.0.0
 pytest-mock>=3.7.0
+pytest-rng
```
LuisAlfredoNu, line 10: Could you remove `flaky>=3.7.0` here and the rest of requirements files and the option `--no-flaky-report` from our GitHub Actions files, please?

LuisAlfredoNu (commented):
Thank you @josephleekl 
Nice job :seedling: 

I will be happy to approve after removing the flaky packages from requirements and the options from our CIs.

tests/lightning_base/test_measurements_class.py
josephleekl: This local_salt is so that we can have local control over the seed, so the seed is basically global rng_salt + local_salt . We will update the global rng_salt, but if it happens to be a bad seed for a specific test, we can update the local_salt too.

requirements.txt
josephleekl: I can't remove flaky from the requirements, since `pl-device-test` requires flaky. 

However I think I can remove `--no-flaky-report` since pl-device-test by default have the same setting.

.github/workflows/tests_gpu_cpp.yml
```diff
@@ -22,7 +22,7 @@ on:
 
 env:
   CI_CUDA_ARCH: 86
-  COVERAGE_FLAGS: "--cov=pennylane_lightning --cov-report=term-missing --cov-report=xml:./coverage.xml --no-flaky-report -p no:warnings --tb=native"
+  COVERAGE_FLAGS: "--cov=pennylane_lightning --cov-report=term-missing --cov-report=xml:./coverage.xml -p no:warnings --tb=native"
```
josephleekl, line 25: by default `pl-device-test` already use `--no-flaky-report`, so this could be removed.

LuisAlfredoNu (approved):
LGTM 
Thank you @josephleekl 
:no_entry_sign: :game_die:

tests/pytest.ini
AmintorDusko: We don't need a GitHub Action for that. We can add a small script to [Ringo](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/.github/workflows/update_dev_version.yml) and make it do the work. All the permissions are already figured out and the version accessible. We just open this file as a dictionary in python and update the field.

tests/test_native_mcm.py
AmintorDusko: I think this is acceptable if we can have a good impact in reducing the flakyness.

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
AmintorDusko: Can we maybe reduce the number of shots in this case? What was the difference in time observed?

tests/lightning_base/test_measurements_class.py
```diff
@@ -622,7 +621,7 @@ def calculate_reference(tape, lightning_sv):
         m = LightningMeasurements(statevector)
         return measure_final_state(m, tape)
 
-    @flaky(max_runs=5)
+    @pytest.mark.local_salt(42)
```
AmintorDusko, line 624: This will make this test always follow the same salt, and not the one we'll keep changing in pytest.ini, right?
Is this a desirable behaviour?

AmintorDusko (commented):
In general very good work. There is something that confuses me a little bit, though.
Do we want to keep changing the rng salt, or fix it? This are two different approaches with pros and cons, but we currently have a mix of both. Is it possible to have a single approach everywhere? Or would this throw a wrench in the test suite?

tests/lightning_base/test_measurements_class.py
josephleekl: So the seed will combine the rng_salt in pytest.ini and the local salt here. Change either will change the seed in the test. With this mechanism we can update all the seeds in one go in pytest.ini, and still have fine grain control locally here.

pennylane_lightning/core/measurements/tests/Test_MeasurementsBase.cpp
josephleekl: The time difference for running the test is less than a second. I can try to use a slightly lower number (e.g. half of this) and see how it impacts the error rate?

tests/lightning_base/test_measurements_class.py
AmintorDusko: Sorry, I didn't see that this was a parameter :sweat: .

AmintorDusko:
> In general very good work. There is something that confuses me a little bit, though. Do we want to keep changing the rng salt, or fix it? This are two different approaches with pros and cons, but we currently have a mix of both. Is it possible to have a single approach everywhere? Or would this throw a wrench in the test suite?

Already answered above.

pennylane_lightning/lightning_gpu/lightning_gpu.py
maliasadi: aha, thanks! make sense

tests/conftest.py
```diff
@@ -207,24 +207,57 @@ def _device(wires, shots=None):
...
+    """An integer random number generator seed
+    This fixture overrides the ``seed`` fixture provided by pytest-rng, adding the flexibility
+    of locally getting a new seed for a test case by applying the ``local_salt`` marker. This is
+    useful when the seed from pytest-rng happens to be a bad seed that causes your test to fail.
+    .. code_block:: python
+        @pytest.mark.local_salt(42)
+        def test_something(seed):
+            ...
+    The value passed to ``local_salt`` needs to be an integer.
+    """
```
maliasadi:
```suggestion
    """An integer random number generator seed

    This fixture overrides the ``seed`` fixture provided by pytest-rng, adding the flexibility
    of locally getting a new seed for a test case by applying the ``local_salt`` marker. This is
    useful when the seed from pytest-rng happens to be a bad seed that causes your test to fail.

    .. code_block:: python

        @pytest.mark.local_salt(42)
        def test_something(seed):
            ...

    The value passed to ``local_salt`` needs to be an integer.

    """
```

PR #1173, opened 2025-06-04
Test `set_shots` compatibility

JerryChen97:
**Context:**
Along with the `set_shots` transform, `PennyLane` is going to decouple the `shots` information from `device` to the application of transform very soon. Before deprecating the `device.shots` we need to test if any other packages in our ecosys suffer from hidden issues/bugs as well.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


requirements.txt
```diff
@@ -10,7 +10,7 @@ pytest-mock>=3.7.0
 pytest-rng
 pytest-xdist>=2.5.0
 flaky>=3.7.0 # For pl-device-test
-pennylane>=0.41
+pennylane @ git+https://github.com/PennyLaneAI/pennylane.git
```
josephleekl: is this temporary for the CI? will you change it back?

JerryChen97: Yes we should change it back. CI will automatically install latest PL right?

josephleekl: you should change it back, then trigger latest/latest + stable/stable to check that it won't break anything

JerryChen97: Triggered, but they look queued forever @josephleekl

josephleekl: they're running, just waiting for them to complete:
[stable/stable](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/16032223646)
[latest/latest](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/16032226880)

JerryChen97: @josephleekl All passed!

josephleekl (approved):
LGTM, thanks!

PR #1176, opened 2025-06-05
[nanobind] Nanobind Feature Branch PR [Base]

AmintorDusko:
# Nanobind Integration and Backend Binding Refactor

This pull request introduces **Nanobind** as a new framework for Python bindings, supporting multiple quantum simulation backends (e.g., **LGPU**, **LKokkos**, **LTensor**, **LQubit**) and enhancing modularity and performance. It also includes significant improvements to the **build system** and a first round of **codebase simplification**.

---

## Nanobind Integration & Backend Support

- Integrated **Nanobind** for Python bindings, initially supporting agnostic (backend-independent) modules.
- Laid the foundation for backend-specific extensions (e.g., LGPU, LQubit), to be completed in follow-up PRs, aiming for this feature branch.
- Implemented initial state vector operations, observables, and measurement functions using Nanobind.
- Replace existing PyBind11-based bindings.

---

## Build System Improvements

- Updated `CMakeLists.txt` to support Nanobind configuration.
- Introduced **temporary Makefile targets**:
  - `build-nanobind`
  - `build-test-bindings`
  - `test-bindings`
- Updated setup system to support the dual binding strategy.

---

## Codebase Simplification

- Promoted shared functionality to backend-agnostic modules to reduce duplication.
- Updated backend binding interfaces for greater modularity and maintainability.

---

## Testing Infrastructure

- Added Nanobind-specific test fixtures in `conftest.py`.
- Created new test cases to validate agnostic and backend-specific bindings.
- Introduced precision-aware class resolution utilities for test flexibility.

---

## Performance & Binary Optimization

According to Nanobind documentation and internal profiling:

- Reduced binary size by approximately **30%**.
- Improved runtime performance of state vector operations.
- Decreased memory usage during binding layer execution.

---

## 📚 Documentation

- Added initial documentation for Nanobind-based APIs and binding structure.

---

## ✅ Benefits

- Improved modularity and backend extensibility.
- Faster and smaller bindings through Nanobind.
- Cleaner and more maintainable binding code.
- Maintains backward compatibility with PyBind11 during migration.

---

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -123,6 +123,7 @@ Joseph Lee,
 Luis Alfredo Nuñez Meneses,
 Andrija Paurevic,
 Jake Zaia.
+
 ---
 
 # Release 0.42.0


CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ if(ENABLE_PYTHON)
                         GIT_TAG        v2.13.5
     )
     FetchContent_MakeAvailable(pybind11)
+
+    FetchContent_Declare(nanobind
+                    GIT_REPOSITORY https://github.com/wjakob/nanobind.git
+                    GIT_TAG        v1.5.2
```
AmintorDusko:
```suggestion
                    GIT_TAG        v2.7.0       
```

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -137,21 +138,25 @@ NB_MODULE(
         .def("append", [](std::vector<int> &v, int x) { v.push_back(x); })
         .def("__len__", [](const std::vector<int> &v) { return v.size(); })
         .def("__getitem__", [](const std::vector<int> &v, size_t i) {
-            if (i >= v.size()) throw nb::index_error();
+            if (i >= v.size())
+                throw nb::index_error();
```
mlxd:
```suggestion
            if (i >= v.size()){
                throw nb::index_error();
            }
```

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -137,21 +138,25 @@ NB_MODULE(
...
-                                  const nb::ndarray<std::complex<double>, nb::numpy> &matrix,
-                                  const std::vector<std::size_t> &wires, 
-                                  bool inverse) {
-        const std::complex<double> *data_ptr = reinterpret_cast<const std::complex<double>*>(matrix.data());
-        sv.applyMatrix(data_ptr, wires, inverse);
-    }, "Apply a given matrix to wires.");
+    sv_class.def(
+        "applyMatrix",
+        [](SimpleStateVector &sv,
+           const nb::ndarray<std::complex<double>, nb::numpy> &matrix,
```
mlxd: Don't we need to use both FP64 and FP32 datatypes here?

pennylane_lightning/core/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
```diff
@@ -23,7 +23,6 @@
 
 #include "cuda.h"
 
-#include "BindingsBase.hpp"
```
jzaia18, line 26: Any reason for moving the bindings for lightning-tensor even though we're leaving them as pybind11 currently? Is this just in prep for a change to where they will live later?

tests/bindings/test_bindings.py
```diff
@@ -0,0 +1,92 @@
...
+)
+def test_module_importable(module_name):
+    """Test if module can be imported."""
+    result = get_module_attributes(module_name)
+
+    # Skip test if module is not available instead of failing
+    if not result["importable"]:
+        pytest.skip(f"Module {module_name} not available: {result.get('error')}")
+
+    assert result["importable"]
```
jzaia18: Do we need this assert statement given this condition? Is the plan to remove the if statement eventually and leave the assert?

tests/bindings/test_nanobindings.py
```diff
@@ -0,0 +1,195 @@
...
+)
+def test_nanobind_module_importable(module_name):
+    """Test if Nanobind module can be imported."""
+    result = get_module_attributes(module_name)
+
+    # Skip test if module is not available instead of failing
+    if not result["importable"]:
+        pytest.skip(f"Module {module_name} not available: {result.get('error')}")
+
+    assert result["importable"]
```
jzaia18: Same question as above

jzaia18 (commented):
Looks great so far (and builds well), thanks for building out the infrastructure!

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    if (a.shape(0) != b.shape(0)) {
+        throw std::runtime_error("Array sizes must match");
+    }
+
+    size_t n = a.shape(0);
+
+    // Allocate memory
+    PrecisionT *result_data = new PrecisionT[n];
```
mlxd: Can we use `unique_ptr`s instead of raw pointers here?
Nanobind should allow transfer of ownership semantics with the array creation. If not, declaraing this as `shared_ptr` is a reasonable fallback.

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
...
+    using ComplexT = std::complex<PrecisionT>;
+
+    SimpleStateVector(size_t num_qubits) : num_qubits_(num_qubits) {}
+
+    void applyMatrix(const ComplexT *matrix,
+                     const std::vector<std::size_t> &wires, bool inverse) {
+        // Placeholder implementation that uses parameters to avoid warnings
+        (void)matrix;  // Silence unused parameter warning
+        (void)wires;   // Silence unused parameter warning
+        (void)inverse; // Silence unused parameter warning
```
mlxd: Can we make these `static_cast` to avoid C-semantics?

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    void applyOperation(const std::string &gate_name,
+                        const std::vector<std::size_t> &wires, bool inverse,
+                        const std::vector<PrecisionT> &params) {
+        // Placeholder implementation that uses parameters to avoid warnings
+        (void)gate_name; // Silence unused parameter warning
+        (void)wires;     // Silence unused parameter warning
+        (void)inverse;   // Silence unused parameter warning
+        (void)params;    // Silence unused parameter warning
```
mlxd: Same here as above?

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
...
+    sv_class.def(nb::init<size_t>());
+
+    // Manually register a few methods
+    sv_class.def(
+        "applyMatrix",
+        [](SimpleStateVector &sv,
+           const nb::ndarray<std::complex<double>, nb::numpy> &matrix,
+           const std::vector<std::size_t> &wires, bool inverse) {
+            const std::complex<double> *data_ptr =
+                reinterpret_cast<const std::complex<double> *>(matrix.data());
```
mlxd: Maybe I'm wrong, but I thought we needed binder-specific casting rules to go from numpy to C++ to bindings? If no, `const std::complex<double> *data_ptr` could be `auto`-fied

mlxd (commented):
Thanks @AmintorDusko 

Just a few quick queries, but will leave to the others for a more detailed review :)

pennylane_lightning/core/simulators/lightning_tensor/tncuda/bindings/LTensorTNCudaBindings.hpp
AmintorDusko: yes

tests/bindings/test_bindings.py
AmintorDusko: Most of these tests were transients. I'm updating it in the next PR.

tests/bindings/test_nanobindings.py
AmintorDusko: Same answer.

CMakeLists.txt
```diff
@@ -147,38 +167,48 @@ target_include_directories(pennylane_lightning  INTERFACE "$<INSTALL_INTERFACE:$
 #####################################################
 if(ENABLE_PYTHON)
     message(STATUS "ENABLE_PYTHON is ON.")
+    # Existing PyBind11 module
     pybind11_add_module("${PL_BACKEND}_ops" "pennylane_lightning/core/bindings/Bindings.cpp")
-
+    
+    # Add Nanobind module with _nb suffix
+    nanobind_add_module("${PL_BACKEND}_nb" "pennylane_lightning/core/bindings/NanobindBindings.cpp")
```
maliasadi: I know this is an intermediate step and we are going to remove pb11 at the end. Any concerns, like deps conflicts or symbol resolution issues, with linking Lightning against both nb and pb11?

Makefile
```diff
@@ -233,3 +233,19 @@ docker-push-all:
 docker-all:
 	$(MAKE) docker-build-all
 	$(MAKE) docker-push-all
+
+.PHONY: build-nanobind
+build-nanobind:
+	$(MAKE) clean
+	ENABLE_NANOBIND=1 PL_BACKEND=$(PL_BACKEND) python scripts/configure_pyproject_toml.py
```
maliasadi: `ENABLE_NANOBIND=ON` instead of `1`?

pennylane_lightning/core/bindings/BindingsBase.hpp
```diff
@@ -77,60 +77,4 @@ void registerGatesForStateVector(PyClass &pyclass) {
...
-    std::vector<ComplexT> conv_matrix;
-    if (m_buffer.size) {
-        const auto m_ptr = static_cast<const ComplexT *>(m_buffer.ptr);
-        conv_matrix = std::vector<ComplexT>{m_ptr, m_ptr + m_buffer.size};
-    }
-    tensor_network.applyOperation("applyMatrix", wires, inverse, {},
-                                  conv_matrix);
-}
-
-// TODO: Unify registerGatesForTensorNet and registerMatrix
```
maliasadi, line 101: I believe the idea behind implementing these two methods in BindingsBase was to unify them with the existing ones. We don't need to update these implementation in this project in favour of simplifying the nanobind work. Could you add these removed TODOs to `LTensorTNCudaBindings.hpp` for the future reference though?

pennylane_lightning/core/bindings/NanobindBindings.cpp
maliasadi: We can also leverage their memory borrowing/stealing API. These are `shared_ptr` under the hood.  https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4I0EN8nanobind6borrowE1T6handle

scripts/configure_pyproject_toml.py
```diff
@@ -94,6 +94,11 @@ def parse_args():
         "setuptools>=75.8.1",
         "tomli",
     ]
+
+    # Add nanobind if enabled
+    if os.environ.get("ENABLE_NANOBIND", "0") == "1":
+        requires.append("nanobind>=2.7.0")
```
maliasadi, line 100: Is this temporary?

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["cmake", "ninja; platform_system!='Windows'", "setuptools>=75.8.1", "tomli"]
+requires = ["cmake", "ninja; platform_system!='Windows'", "setuptools>=75.8.1", "tomli", "nanobind>=2.7.0"]
```
maliasadi: We don't need nanobind here; similar to pybind11, we always fetch the nanobind src when building the C++ core modules [here](https://github.com/PennyLaneAI/pennylane-lightning/pull/1176/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR99).

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
+// Copyright 2018-2024 Xanadu Quantum Technologies Inc.
```
maliasadi: 2025 :smile:

pennylane_lightning/core/bindings/NanobindBindings.cpp
```diff
@@ -0,0 +1,169 @@
...
+        delete[] static_cast<PrecisionT *>(p);
+    });
+
+    return nb::ndarray<nb::numpy, PrecisionT>(result_data, 1, shape, capsule);
+}
+
+// Example StateVector class for testing
+class SimpleStateVector {
+  public:
+    using PrecisionT = double;
```
maliasadi: This will be fixed and used a template, right?

pennylane_lightning/core/bindings/NanobindBindings.cpp
maliasadi: I think this is just for prototyping the first end-to-end nanobind example and will be fixed

tests/bindings/test_bindings.py
maliasadi: The plan should be to keep this base PR as simple as possible so we can finalize it asap and prevent any direct commits to it. Adding `NanobindBindings` is a feature that could be included in a separate feature PR (based on this current branch)

tests/bindings/test_nanobindings.py
```diff
@@ -0,0 +1,195 @@
...
+    if not result["importable"]:
+        pytest.skip(f"Module {module_name} not available: {result.get('error')}")
+
+    module = result["module"]
+
+    # Check if SimpleStateVector class exists
+    if not hasattr(module, "SimpleStateVector"):
+        pytest.skip(f"SimpleStateVector class not found in {module_name}")
+
+    sv = module.SimpleStateVector(2)
```
maliasadi: any performance concerns or advancements using nb instead of py11?

AmintorDusko:
Hey folks! Thank you for your feedback here. Most of what we see on this PR is temporary, tests, or prototypes. I would like to ask you to focus your review on [PR 1180](https://github.com/PennyLaneAI/pennylane-lightning/pull/1180), where most concerns have already been addressed or are no longer relevant. I will tag you when the PR is ready for review. Thank you! For visibility: @mlxd, @maliasadi, @jzaia18

CMakeLists.txt
AmintorDusko: Not at all.

Makefile
AmintorDusko: Removed

pennylane_lightning/core/bindings/BindingsBase.hpp
AmintorDusko: done

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: removed

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: This function was a prototype. Not relevant anymore.

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: This is a proof of concept. removed

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: The only reason for that was not to trigger clang. removed

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: same as above. removed

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: outdated

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: yes, we need. This was only a prototype.

pennylane_lightning/core/bindings/NanobindBindings.cpp
AmintorDusko: this was really just to show it worked.

pyproject.toml
AmintorDusko: I understand your perspective, but I believe it's beneficial to keep this type of dependency included here, just as it appears in the requirements files. I can remove it if you prefer, but I would rather not. In fact, we should consider adding more dependencies to pyproject.toml. This is becoming the new standard.

scripts/configure_pyproject_toml.py
AmintorDusko: yes

tests/bindings/test_nanobindings.py
AmintorDusko: This will be addressed in the future.

AmintorDusko (commented):
Hey everyone, to prevent future issues with reviews of outdated code, I will merge the base bindings PR with this one. This will significantly reduce the amount of temporary code.

CMakeLists.txt
```diff
@@ -85,13 +85,28 @@ set(CMAKE_POLICY_DEFAULT_CMP0127 NEW) # To suppress pybind11 CMP0127 warning
...
     FetchContent_MakeAvailable(pybind11)
+
+    message(STATUS "Enabling Nanobind Python bindings")
+    
+    include(FetchContent)
+    FetchContent_Declare(nanobind
+                    GIT_REPOSITORY https://github.com/wjakob/nanobind.git
+                    GIT_TAG        v2.7.0
+    )
+    FetchContent_MakeAvailable(nanobind)
```
jzaia18: Is there a reason we do this rather than having nanobind as a dependency and relying on the installed version?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+#include "Util.hpp" // for_each_enum
+
+// Include backend-specific headers and define macros based on compile flags
+#ifdef _ENABLE_PLQUBIT
+#include "AdjointJacobianLQubit.hpp"
+#include "LQubitBindings_nb.hpp"
+#include "MeasurementsLQubit.hpp"
+#include "ObservablesLQubit.hpp"
+
+#define BACKEND_NAME "lightning.qubit"
```
jzaia18: Is there a reason to prefer #define to other approaches (such as a global constexpr) here?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+
+    // Create capsule with custom deleter
+    auto capsule =
+        nb::capsule(ptr, [](void *p) noexcept { Util::alignedFree(p); });
+
+    // Create shape array
+    size_t shape[1] = {size};
+
+    // Return ndarray with custom allocated memory
+    return nb::ndarray<T, nb::numpy>(ptr, 1, shape, capsule);
```
jzaia18:
```suggestion
    size_t ndim = 1;
    size_t shape[ndim] = {size};

    // Return ndarray with custom allocated memory
    return nb::ndarray<T, nb::numpy>(ptr, ndim, shape, capsule);
```

Thoughts on this for readability? I realize its a little bit extra, but it feels like we should avoid having a magic number even just for the number of dimensions

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+    info["compiler.name"] = compiler_name_str;
+    info["compiler.version"] = compiler_version_str;
+    info["AVX2"] = use_avx2;
+    info["AVX512F"] = use_avx512f;
+
+    return info;
+}
+
+template <class StateVectorT>
+auto createStateVectorFromNanobindData(
```
jzaia18: This function doesn't appear to be used anywhere atm. Is this meant to be bound by each device separately since each will have its own template argument for StateVectorT?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+namespace nb = nanobind;
+namespace Pennylane::NanoBindings {
+
+/**
+ * @brief Register applyMatrix.
+ */
+template <class StateVectorT>
+void registerMatrix(
+    StateVectorT &st,
+    const nb::ndarray<typename StateVectorT::ComplexT, nb::numpy> &matrix,
```
jzaia18: Is there a particular reason for the `nb::numpy` constraint here? If not, it might be better to leave it off which would allow this function to support all array implementations which support dlpack. For function inputs this gives us more flexibility (although it does also prevent us from using e.g. Python lists or other things that don't support dlpack as inputs).

This same comment goes for many places in the code, but I didn't want to flood the PR with identical comments.

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+ * @param data Vector containing the data to transfer
+ * @return nb::ndarray<T, nb::numpy> NumPy array with copied data
+ */
+template <typename T>
+nb::ndarray<T, nb::numpy>
+createNumpyArrayFromVector(const std::vector<T> &data) {
+    const std::size_t size = data.size();
+
+    // Create a new array with the right size
+    std::size_t shape[1] = {size};
```
jzaia18: Same suggestion as above about magic numbers

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+ * @return nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy> NumPy array
+ * with probabilities
+ */
+template <class StateVectorT>
+nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy>
+probsForWires(Measurements<StateVectorT> &M,
+              const std::vector<std::size_t> &wires) {
+    using PrecisionT = typename StateVectorT::PrecisionT;
+    auto probs_vec = M.probs(wires);
+    return createNumpyArrayFromVector<PrecisionT>(probs_vec);
```
jzaia18: Is it necessary to make a copy of the array for returning here? This seems like the kind of place we want to start using a zero-copy exchange

jzaia18: Spoke to Ali offline, it sounds like we certainly want to start using a zero-copy exchange for this. We'll need to modify `M.probs()` to return a pointer to heap alloc'd memory for this, and we'd then return an `nb::ndarray` that wraps this data directly.

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+ * @param M Measurements object
+ * @return nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy> NumPy array
+ * with probabilities
+ */
+template <class StateVectorT>
+nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy>
+probsForAllWires(Measurements<StateVectorT> &M) {
+    using PrecisionT = typename StateVectorT::PrecisionT;
+    auto probs_vec = M.probs();
+    return createNumpyArrayFromVector<PrecisionT>(probs_vec);
```
jzaia18: Same comment for zero-copy exchange

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+ * @param num_wires Number of wires
+ * @param num_shots Number of shots
+ * @return nb::ndarray<std::size_t, nb::numpy> 2D NumPy array with samples
+ */
+template <class StateVectorT>
+nb::ndarray<std::size_t, nb::numpy>
+generateSamples(Measurements<StateVectorT> &M, std::size_t num_wires,
+                std::size_t num_shots) {
+    auto result = M.generate_samples(num_shots);
+    return create2DNumpyArrayFromVector<std::size_t>(result, num_shots,
```
jzaia18: Same comment for zero-copy exchange

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,841 @@
...
+        },
+        "Create a list of operations from data.");
+}
+
+/**
+ * @brief Register backend agnostic state vector methods.
+ *
+ * @tparam StateVectorT
+ * @tparam PyClass
+ * @param pyclass Pybind11's state vector class to bind methods.
```
jzaia18:
```suggestion
 * @param pyclass Nanobind's state vector class to bind methods.
```

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         self._device_name = "lightning.qubit"
 
         # Initialize the state vector
-        self._qubit_state = self._state_dtype()(self._num_wires)
+        self._qubit_state = self._state_dtype()(num_wires)
```
jzaia18, line 69: This seems unrelated to the nanobind changes, was this (and the other similar changes) just a bug that you uncovered while testing?

tests/bindings/test_bindings_nb.py
```diff
@@ -0,0 +1,213 @@
...
+        module_attr = self.module_attributes[module_name]
+        if not module_attr["importable"]:
+            pytest.skip(f"Module {module_name} not available: {module_attr.get('error')}")
+        return module_attr
+
+    @pytest.mark.parametrize("module_name", nb_modules)
+    def test_module_importable(self, module_name: str):
+        """Test if module can be imported."""
+        module_attr = self._skip_if_module_not_importable(module_name)
+        assert module_attr["importable"]
```
jzaia18: Is this assertion needed? Seems like the prior line guarantees it will pass

tests/bindings/test_bindings_nb.py
```diff
@@ -0,0 +1,213 @@
...
+        """Test if LightningException class exists in the module."""
+        module_attr = self._skip_if_module_not_importable(module_name)
+
+        assert (
+            "LightningException" in module_attr["classes"]
+        ), f"LightningException not found in {module_name}"
+
+    @pytest.mark.xfail(reason="Expected to fail while we don't have backend-specific bindings.")
+    @pytest.mark.parametrize("module_name, pybind_module_name", zip(nb_modules, pb_modules))
+    def test_api_parity_with_pybind(self, module_name, pybind_module_name):
```
jzaia18: Is it possible to automatically check the function parameter types as well to ensure we don't accidentally have breaking API changes?

tests/bindings/test_measurements_nb.py
```diff
@@ -0,0 +1,403 @@
...
+
+        return _get_classes
+
+    @pytest.mark.parametrize("backend_name", backends)
+    def test_state_initialization(self, backend_name, precision, get_classes, nanobind_module):
+        """Test state vector initialization."""
+        module = nanobind_module(backend_name)
+
+        StateVectorClass, MeasurementsClass, _, _ = get_classes(module, precision)
+        # Create a numpy array representing |0> state with appropriate size
```
jzaia18: We should probably test with a different state to ensure that `sv.updateData` successfully modifies the state amplitudes (assuming the constructor initializes the statevector to |00...0> anyway)

tests/bindings/test_measurements_nb.py
```diff
@@ -0,0 +1,403 @@
...
+        assert samples.shape == (num_samples, 2)
+
+        # In a Bell state, the qubits should be perfectly correlated
+        # Count how many times we get |00⟩ and |11⟩
+        count_00 = np.sum(np.all(samples == [0, 0], axis=1))
+        count_11 = np.sum(np.all(samples == [1, 1], axis=1))
+
+        # We should have approximately 50% |00⟩ and 50% |11⟩
+        # and very few or no |01⟩ and |10⟩
+        assert count_00 + count_11 >= 0.95 * num_samples  # Allow for small statistical fluctuations
```
jzaia18, line 320:
```suggestion
        assert count_00 + count_11 == num_samples
```

Isn't lightning a noiseless simulator? It shouldn't be possible to get |01> or |10> at all

tests/bindings/test_measurements_nb.py
```diff
@@ -0,0 +1,403 @@
...
+        all_samples = meas.generate_samples(num_qubits, num_samples)
+        assert all_samples.shape == (num_samples, num_qubits)
+
+        # The results should be the same as when specifying all wires
+        count_00_all = np.sum(np.all(all_samples == [0, 0], axis=1))
+        count_11_all = np.sum(np.all(all_samples == [1, 1], axis=1))
+
+        # The counts should be similar to the previous test
+        assert abs(count_00 - count_00_all) < 0.1 * num_samples
+        assert abs(count_11 - count_11_all) < 0.1 * num_samples
```
jzaia18:
```suggestion
        assert 0.4 * num_samples <= count_00_all <= 0.6 * num_samples
        assert 0.4 * num_samples <= count_11_all <= 0.6 * num_samples
```

These counts shouldn't depend directly on the outcome of the prior test. (e.g. what if the prior test had 40% |00> and this test has 60% |00>?)

tests/bindings/test_statevector_nb.py
```diff
@@ -0,0 +1,364 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for StateVector classes in nanobind-based modules."""
+
+import numpy as np
+import pytest
+from conftest import backend
+
+if backend != "qubit":
+    pytest.skip("Skipping tests for binaries other than lightning_qubit .", allow_module_level=True)
```
jzaia18: Is this just meant to be temporary as we implement the bindings in various devices?

tests/bindings/test_statevector_nb.py
```diff
@@ -0,0 +1,364 @@
...
+        # Initialize with number of qubits first
+        sv = StateVectorClass(num_qubits)
+        # Then update data
+        sv.updateData(state_data)
+
+        # Check that the size is correct
+        assert sv.size() == 2**num_qubits
+
+    @pytest.mark.parametrize("backend_name", backends)
+    def test_statevector_methods(
```
jzaia18: Is this meant to do something extra compared to `test_statevector_initialization`?

tests/bindings/test_statevector_nb.py
```diff
@@ -0,0 +1,364 @@
...
+        assert True
+
+    @pytest.mark.parametrize("backend_name", backends)
+    def test_statevector_matrix_application(
+        self, backend_name, precision, nanobind_module, get_statevector_class
+    ):
+        """Test matrix application on StateVectorC64/128 classes."""
+        # Skip non-qubit backends for this test
+        if backend_name != "qubit":
+            pytest.skip(f"Matrix application test only supported for qubit backend")
```
jzaia18: Doesn't this check already happen at the module level?

tests/bindings/test_statevector_nb.py
```diff
@@ -0,0 +1,364 @@
...
+
+        # Initialize with number of qubits first
+        sv = StateVectorClass(num_qubits)
+        # Then update data
+        sv.updateData(state_data)
+
+        # Get the result
+        result = np.zeros(2**num_qubits, dtype=dtype)
+        sv.getState(result)
+
```
jzaia18:
```suggestion
        # Get the result
        result = np.zeros(2**num_qubits, dtype=dtype)
        sv.getState(result)
        assert np.allclose(result, state_data)
    
```

jzaia18 (commented):
Awesome work! I have a bit of feedback that I left, the two most important bits being about using `nb::numpy` on function parameters, and zero-copy exchange support for measurement outcomes.

CMakeLists.txt
AmintorDusko: That's a good point. Let me start a discussion about this in the channel.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: You are very much right. I think we can indeed improve on that in terms of constexpr. I will remove this though, as we don't have it in the Pybind11 bindings and we would have some problems with this approach when installing more than one backend.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: I agree.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: No, it's old code that I forgot to remove. Thanks.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: @jzaia18, that's a great idea! Would you be able to implement it? If so, please open a pull request (PR) targeting this branch. This change is significant and extends beyond just converting the Python binding, so a well-structured PR would be beneficial.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: Same as before. This can go in the same PR.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: Same as before. This can go in the same PR.

pennylane_lightning/lightning_qubit/_state_vector.py
AmintorDusko: Less of a bug, more like standardization.

tests/bindings/test_bindings_nb.py
AmintorDusko: Let me know if it looks good now.

tests/bindings/test_statevector_nb.py
AmintorDusko: yes

tests/bindings/test_statevector_nb.py
AmintorDusko: Was, but with time, they evolved to be the same. I just removed.

tests/bindings/test_statevector_nb.py
AmintorDusko: removed

tests/bindings/test_statevector_nb.py
AmintorDusko: I prefer comparing with the expected data, as below.

pennylane_lightning/core/simulators/lightning_gpu/bindings/LGPUBindings_nb.hpp
```diff
@@ -0,0 +1,100 @@
...
+namespace nb = nanobind;
+
+namespace Pennylane::LightningGPU::NanoBindings {
+
+/**
+ * @brief Define StateVector backends for lightning.gpu
+ */
+using StateVectorBackends = Pennylane::Util::TypeList<
+    StateVectorCudaManaged<float>,
+    Pennylane::Util::TypeList<StateVectorCudaManaged<double>, void>>;
```
jzaia18: Was there a reason for changing this line to have an extra `TypeList` during the switch to nanobind? This new version doesn't compile for me, but switching back to the old way that was used resolves the issue.

jzaia18: (Same question for LKokkos, but it looks like LQubit was left the old way)

AmintorDusko: This was a mistake. I will fix it.

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,840 @@
...
+        throw std::invalid_argument("Array must be 1-dimensional");
+    }
+
+    // Get data pointer and size
+    const ComplexT *data_ptr = static_cast<const ComplexT *>(data.data());
+    std::size_t size = data.shape(0);
+
+    // Update the state vector data
+    sv.updateData(data_ptr, size);
+}
```
jzaia18:
```suggestion
template <class StateVectorT>
void updateStateVectorData(
    StateVectorT &sv,
    const nb::ndarray<const typename StateVectorT::ComplexT, nb::numpy> &data) {
    using ComplexT = typename StateVectorT::ComplexT;

    // Check dimensions
    if (data.ndim() != 1) {
        throw std::invalid_argument("Array must be 1-dimensional");
    }

    // Get data pointer and size
    const ComplexT *data_ptr = data.data();
    std::size_t size = data.shape(0);

    // Update the state vector data
    sv.updateData(data_ptr, size);
}
```

pennylane_lightning/core/simulators/lightning_gpu/bindings/LGPUBindings_nb.hpp
```diff
@@ -0,0 +1,100 @@
...
+ */
+template <class StateVectorT, class PyClass>
+void registerBackendClassSpecificBindings(PyClass &pyclass) {}
+
+/**
+ * @brief Register backend specific measurements class functionalities.
+ *
+ * @tparam StateVectorT
+ * @tparam PyClass
+ * @param pyclass Pybind11's measurements class to bind methods.
```
jzaia18:
```suggestion
 * @param pyclass Nanobind's statevector class to bind methods.
```

I think? Or is this actually a measurements class getting passed here?

pennylane_lightning/core/bindings/Bindings_nb.hpp
jzaia18: Yes, will do alongside the LGPU bindings

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,836 @@
...
+/**
+ * @brief Register StateVector class.
+ *
+ * @tparam StateVectorT Statevector type to register
+ * @tparam PyClass Nanobind's class object type
+ *
+ * @param pyclass Nanobind's class object to bind statevector
+ */
+template <class StateVectorT, class PyClass>
+void registerGatesForStateVector(PyClass &pyclass) {
```
jzaia18: It makes sense that this function is here. But for some reason the controlled version is unique to each impl, is there some reason that registering controlled gates couldnt be unified as well?

pennylane_lightning/core/simulators/lightning_gpu/bindings/LGPUBindings_nb.hpp
AmintorDusko: outdated?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -189,12 +188,10 @@ auto alignedArray(Util::CPUMemoryModel memory_model, std::size_t size,
...
         nb::capsule(ptr, [](void *p) noexcept { Util::alignedFree(p); });
 
-    // Create shape array
-    size_t ndim = 1;
-    size_t shape[ndim] = {size};
+    std::vector<size_t> shape{size};
 
     // Return ndarray with custom allocated memory
-    return nb::ndarray<T>(ptr, ndim, shape, capsule);
+    return nb::ndarray<T, nb::c_contig>(ptr, 1, shape.data(), capsule);
```
jzaia18:
```suggestion
    return nb::ndarray<T, nb::c_contig>(ptr, shape.size(), shape.data(), capsule);
```

To help avoid magic numbers (this is minor, feel free to overrule me on this)

jzaia18: (Same comment applies to a few places)

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -96,7 +139,8 @@ void registerBackendSpecificStateVectorMethods(PyClass &pyclass) {
 
     pyclass.def(
         "getState",
-        [](const StateVectorT &sv, nb::ndarray<ComplexT, nb::numpy> &state) {
+        [](const StateVectorT &sv,
+           nb::ndarray<ComplexT, nb::numpy, nb::c_contig> &state) {
```
jzaia18:
```suggestion
           nb::ndarray<ComplexT, nb::c_contig> &state) {
```

Any reason this shouldn't support all array types?

jzaia18 (commented):
Added some comments for the new commits, will look over the code from the top again as well

CMakeLists.txt
jzaia18: https://xanaduhq.slack.com/archives/C01KA3FB86L/p1750073856344909

pennylane_lightning/core/bindings/Bindings_nb.hpp
jzaia18: This now has its own story [sc-93651]

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,794 @@
...
+template <class StateVectorT>
+void registerMatrix(
+    StateVectorT &st,
+    const nb::ndarray<std::complex<typename StateVectorT::PrecisionT>, nb::c_contig> &matrix,
+    const std::vector<std::size_t> &wires, bool inverse = false) {
+    using ComplexT = typename StateVectorT::ComplexT;
+
+    // Cast to raw pointer
+    const ComplexT *data_ptr = matrix.data();
+    st.applyMatrix(data_ptr, wires, inverse);
```
jzaia18:
```suggestion
    st.applyMatrix(matrix.data(), wires, inverse);
```

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,794 @@
...
+        alignedAlloc(getAlignment<T>(memory_model), sizeof(T) * size, zeroInit);
+
+    // Create capsule with custom deleter
+    auto capsule =
+        nb::capsule(ptr, [](void *p) noexcept { Util::alignedFree(p); });
+
+    std::vector<size_t> shape{size};
+
+    // Return ndarray with custom allocated memory
+    return nb::ndarray<T, nb::c_contig>(ptr, 1, shape.data(), capsule);
```
jzaia18:
```suggestion
auto alignedArray(Util::CPUMemoryModel memory_model, std::size_t size,
                  bool zeroInit) -> nb::ndarray<T, nb::numpy, nb::c_contig> {
    using Pennylane::Util::alignedAlloc;
    using Pennylane::Util::getAlignment;

    // Allocate aligned memory
    void *ptr =
        alignedAlloc(getAlignment<T>(memory_model), sizeof(T) * size, zeroInit);

    // Create capsule with custom deleter
    auto capsule =
        nb::capsule(ptr, [](void *p) noexcept { Util::alignedFree(p); });

    std::vector<size_t> shape{size};

    // Return ndarray with custom allocated memory
    return nb::ndarray<T, nb::numpy, nb::c_contig>(ptr, 1, shape.data(), capsule);
```

For *returned* arrays I think we do actually want to specify a concrete type, otherwise it'll return a generic dlpack object

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,794 @@
...
+        "var",
+        [](Measurements<StateVectorT> &M, const Observable<StateVectorT> &obs) {
+            return M.var(obs);
+        },
+        "Calculate variance for an observable.");
+
+    // Add generate_samples method
+    pyclass.def("generate_samples", &generateSamples<StateVectorT>,
+                "Generate samples for all wires.");
+}
```
jzaia18: Need to add binding for newly introduced `set_random_seed`: e.g. https://github.com/PennyLaneAI/pennylane-lightning/commit/eaf73caa8737460865c282312110bf4e0dcb28e9#diff-d3a839f5744662a4c49119085cd37a09ef0bd7c0ed9bc2a21b2789994b4d2338R496

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,794 @@
...
+void registerArrayAlignmentBindings(nb::module_ &m) {
+    using Pennylane::Util::bestCPUMemoryModel;
+    using Pennylane::Util::CPUMemoryModel;
+
+    // Register CPUMemoryModel enum
+    nb::enum_<CPUMemoryModel>(m, "CPUMemoryModel")
+        .value("Unaligned", CPUMemoryModel::Unaligned)
+        .value("Aligned256", CPUMemoryModel::Aligned256)
+        .value("Aligned512", CPUMemoryModel::Aligned512);
+
```
jzaia18: What's the reasoning behind removing the binding for `get_alignment`?

tests/bindings/test_measurements_nb.py
AmintorDusko: Thanks

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: A similar cast will be necessary due to Kokkos. This will become clearer when I finalize my changes.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: yes, indeed we do.

jzaia18 (approved):
Awesome work, looks great. Just make sure to resolve the last few comments and update the changelog.

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,793 @@
...
+ */
+template <class StateVectorT>
+void registerMatrix(
+    StateVectorT &st,
+    const nb::ndarray<const std::complex<typename StateVectorT::PrecisionT>,
+                      nb::c_contig> &matrix,
+    const std::vector<std::size_t> &wires, bool inverse = false) {
+    using ComplexT = typename StateVectorT::ComplexT;
+    using Pennylane::Util::PL_reinterpret_cast;
+
```
jzaia18: We should probably do a size check here, e.g. `assert(matrix.size() == 2^(wires.size()*2) )`

AmintorDusko:
@LuisAlfredoNu, you can start reviewing when you have the opportunity.

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: done!

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -715,6 +673,45 @@ void registerBackendAgnosticStateVectorMethods(PyClass &pyclass) {
...
 
+/**
+ * @brief Register controlled gate operations for a statevector.
+ *
+ * @tparam StateVectorT State vector type
+ * @tparam PyClass Nanobind class type
+ * @param pyclass Nanobind class to bind methods to
+ */
+template <class StateVectorT, class PyClass>
+void registerControlledGates(PyClass &pyclass) {
```
jzaia18: Small thing: for organization, would it be better to move this to be right underneath `registerGatesForStatevector()`

.github/CHANGELOG.md
```diff
@@ -53,6 +53,11 @@
 
 <h3>Internal changes ⚙️</h3>
   
+- Introduce Nanobind as a new framework for Python bindings, and enhance modularity and performance. 
+  It also include significant improvements to the build system and a first round of codebase simplification. 
+  Legacy PyBind11 code remains accessible for compatibility and will be incrementally phased out.
```
jzaia18:
```suggestion
- Introduce Nanobind as a new framework for Python bindings, and enhance modularity and performance of binding code. 
  This includes significant improvements to the build system and a first round of codebase simplification. 
  Legacy PyBind11 code remains accessible for compatibility and will be incrementally phased out.
```

jzaia18 (approved):
Looks good! Just 2 minor comments

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: I agree.

.github/workflows/tests_lqcpu_python.yml
```diff
@@ -219,6 +219,9 @@ jobs:
           pl-device-test --device ${DEVICENAME} --skip-ops --shots=10000 $COVERAGE_FLAGS --cov-append
           pl-device-test --device ${DEVICENAME} --shots=None --skip-ops $COVERAGE_FLAGS --cov-append
 
+      - name: Python packages
+        run: pip freeze
```
AmintorDusko, line 213: I'm letting this one here as this is useful for any future debug.

pennylane_lightning/core/algorithms/AdjointJacobianBase.hpp
```diff
@@ -207,6 +227,6 @@ template <class StateVectorT, class Derived> class AdjointJacobianBase {
     }
 
   public:
-    ~AdjointJacobianBase() = default;
+    virtual ~AdjointJacobianBase() = default;
```
LuisAlfredoNu: Could you apply the [google style](https://google.github.io/styleguide/cppguide.html#Declaration_Order) here, please?

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -0,0 +1,489 @@
...
+    StateVectorT &sv,
+    const nb::ndarray<const typename StateVectorT::ComplexT, nb::c_contig>
+        &data) {
+    using ComplexT = typename StateVectorT::ComplexT;
+
+    // Check dimensions
+    if (data.ndim() != 1) {
+        throw std::invalid_argument("Array must be 1-dimensional");
+    }
+
```
LuisAlfredoNu: Will not be better using PL_ABORT?

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -0,0 +1,489 @@
...
+/**
+ * @brief Update state vector data from an array
+ *
+ * This function accepts any array-like object that follows the buffer protocol,
+ * including NumPy arrays and JAX arrays (for example).
+ */
+template <class StateVectorT>
+void updateStateVectorData(
+    StateVectorT &sv,
+    const nb::ndarray<const typename StateVectorT::ComplexT, nb::c_contig>
```
LuisAlfredoNu: What if someone use a Fortran style array? :thinking:

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -0,0 +1,489 @@
...
+        "applyPauliRot",
+        [](StateVectorT &sv, const std::vector<std::size_t> &wires,
+           const bool inverse, const std::vector<PrecisionT> &params,
+           const std::string &word) {
+            sv.applyPauliRot(wires, inverse, params, word);
+        },
+        "Apply a Pauli rotation.");
+
+    // Fix constructor binding for nanobind.
+    pyclass.def(nb::init<std::size_t>(), "Initialize with number of qubits");
```
LuisAlfredoNu: Could you move this to the top of the definitions please?

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -0,0 +1,489 @@
...
+ *
+ * @tparam StateVectorT
+ * @param m Nanobind module
+ */
+template <class StateVectorT>
+void registerBackendSpecificObservables(nb::module_ &m) {
+    using PrecisionT = typename StateVectorT::PrecisionT;
+    using ComplexT = typename StateVectorT::ComplexT;
+
+    const std::string bitsize =
```
LuisAlfredoNu: Could be `constexpr`?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -0,0 +1,884 @@
...
+ * @return nb::ndarray<VectorT, nb::numpy, nb::c_contig>
+ */
+template <typename VectorT>
+auto alignedArray(CPUMemoryModel memory_model, std::size_t size, bool zeroInit)
+    -> nb::ndarray<VectorT, nb::numpy, nb::c_contig> {
+    using Pennylane::Util::alignedAlloc;
+    using Pennylane::Util::getAlignment;
+
+    // Allocate aligned memory
+    void *ptr = alignedAlloc(getAlignment<VectorT>(memory_model),
```
LuisAlfredoNu: Why we don't do the conditional 
```
getAlignment<T>(memory_model) > alignof(std::max_align_t)
```
as we do with pybind11? :thinking:

pennylane_lightning/core/algorithms/AdjointJacobianBase.hpp
AmintorDusko: Sure. I think it is weird to start with a protected section start a public section and come back to a protected section, though.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: Thank you for your suggestion.
We are migrating the bindings from pybind11 to nanobind. 
Pybind11 bindings don't use PL_ABORT, and I prefer not to deal with this at this moment.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: We should not worry about this at this moment.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: I can try.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: https://nanobind.readthedocs.io/en/latest/ndarray.html#array-constraints

.github/CHANGELOG.md
```diff
@@ -170,6 +187,9 @@ Jake Zaia.
 
 <h3>Internal changes ⚙️</h3>
 
+- Added Nanobind-based Python bindings for Lightning Qubit as an alternative to PyBind11.
+  [(#1198)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1198)
+
```
AmintorDusko:
```suggestion
```

.github/CHANGELOG.md
```diff
@@ -100,12 +117,12 @@ This release contains contributions from (in alphabetical order):
 
 Ali Asadi,
 Yushao Chen,
+Amintor Dusko,
 Christina Lee,
 Joseph Lee,
 Luis Alfredo Nuñez Meneses,
 Andrija Paurevic,
 Jake Zaia.
```
maliasadi: To fix the rendering issue:
```suggestion
Jake Zaia.

```

maliasadi (approved):
:partying_face:

PR #1187, opened 2025-06-18
[nanobind] Improve `probs` and `sample` to make use of zero-copy exchanges

jzaia18:
**Context:**
Probs and sample currently work by copying arrays from C++ to Python.

Updates probs and sample to instead directly share heap memory with Python through nanobind's `ndarray`, allowing for a zero-copy exchange.

**Benefits:**
Halves the memory used by probs and sample for storing measurement results.

**Possible Drawbacks:**
Python is now responsible for deallocating memory allocated in C++ (should be fully handled by nanobind).

**Related GitHub Issues:**
[sc-93651]


pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -459,23 +459,19 @@ void registerBackendAgnosticObservables(nb::module_ &m) {
  */
 template <typename VectorT>
 nb::ndarray<VectorT, nb::numpy, nb::c_contig>
-createArrayFromVector(const std::vector<VectorT> &data) {
-    const std::size_t size = data.size();
+createArrayFromVector(std::vector<VectorT> &&data) {
```
josephleekl: so you're using rvalue here, which will have problems with lvalue - you'll need perfect forwarding (std::forward) . I see that in the use-cases you only use it on rvalues, but i believe it's good to be flexible here. 
I think you have the options:
1) use std::forward
2) overload with a lvalue ref
3) use pass-by-value.
With 3, if you pass in an rvalue (as you do), it will use move instead of copy anyways. I think people generally use/recommend 3. Have a think about which is best :)

jzaia18: I chose this approach to avoid the possibility of accidentally creating a copy since this is very expensive and not normally something we'd want to do. If I'm understanding correctly, each of these 3 approaches opens up a way in which you could call this function and copy the input vector rather than stealing its resources (whereas the current version would give a compile-time error for this). CC: @AmintorDusko if you have any thoughts on if it would be better to additionally support copying as well

josephleekl: if it's intentional then fair - but are you sure that every Array we create this way are rvalues? If so that's ok to me

AmintorDusko: Can we have this in terms of smart pointers? Something like:
```python
template <typename VectorT>
nb::ndarray<VectorT, nb::numpy, nb::c_contig>
createArrayFromVector(std::vector<VectorT>&& data) {
    auto data_ptr = std::make_unique<std::vector<VectorT>>(std::move(data));
    auto* raw_ptr = data_ptr.get();
    auto capsule = nb::capsule(
        raw_ptr, [](void* p) noexcept {
            delete static_cast<std::vector<VectorT>*>(p);
        }
    );
    std::vector<size_t> shape{raw_ptr->size()};
   
    data_ptr.release();
    return nb::ndarray<VectorT, nb::numpy, nb::c_contig>(
        raw_ptr->data(), shape.size(), shape.data(), capsule
    );
```
Please double-check my code. I didn't test.
Also, you should update the base branch. Some of the changes I made will affect this.

jzaia18: Is there any benefit to using a smart pointer here? It seems like we're creating a unique pointer only to immediately give up ownership and go back to using a raw pointer

AmintorDusko: You are right. I believe we can go with something like:
```c++
auto* raw_ptr = new std::vector<VectorT>(std::move(data));
auto capsule = nb::capsule(
    raw_ptr, [](void* p) noexcept {
        delete static_cast<std::vector<VectorT>*>(p);
    }
);
std::vector<size_t> shape{raw_ptr->size()};
return nb::ndarray<VectorT, nb::numpy, nb::c_contig>(
    raw_ptr->data(), shape.size(), shape.data(), capsule
);
```

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
```diff
@@ -25,7 +25,7 @@ namespace Pennylane::NanoBindings::Utils {
  */
 template <typename VectorT>
 nb::ndarray<VectorT, nb::numpy, nb::c_contig>
-createNumpyArrayFromVector(const std::vector<VectorT> &data,
+createNumpyArrayFromVector(const std::vector<VectorT> &&data,
```
josephleekl: I think this shouldn't be `const` for `move` to not copy

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
```diff
@@ -78,9 +77,9 @@ createNumpyArrayFromVector(const std::vector<VectorT> &data) {
  */
 template <typename VectorT>
 nb::ndarray<VectorT, nb::numpy, nb::c_contig>
-createNumpyArrayFromVector(const std::vector<VectorT> &data, std::size_t rows,
+createNumpyArrayFromVector(const std::vector<VectorT> &&data, std::size_t rows,
```
josephleekl: can we eliminate this one, and for generate samples just call `createNumpyArrayFromVector(..., num_shots * num_wires)`?

josephleekl (commented):
Looking great, thanks @jzaia18 ! Just some small comments - happy to approve once updated!

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
jzaia18: We definitely could, I'm not sure which way feels cleaner. I'll defer to @AmintorDusko on this one

AmintorDusko: If we don't need it, feel free to eliminate.

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
jzaia18: Done in feff4de

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
AmintorDusko: Thank you, @mlxd! It is much better now.

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
```diff
@@ -25,7 +25,7 @@ namespace Pennylane::NanoBindings::Utils {
  */
 template <typename VectorT>
 nb::ndarray<VectorT, nb::numpy, nb::c_contig>
-createNumpyArrayFromVector(const std::vector<VectorT> &data,
+createNumpyArrayFromVector(std::vector<VectorT> &&data,
```
AmintorDusko, line 28: I was wondering if we could use a std::span here instead of std::vector. Does it make sense?

AmintorDusko (approved):
Super cool! Nice job updating and expanding usage for the createNumpyArrayFromVector.

AmintorDusko:
@josephleekl, can we have a final review here?

AmintorDusko:
Thank you, @josephleekl!

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
jzaia18: I don't think span would work here since we need to move-construct the vector to pilfer its memory and avoid a copy. I think if we use span it would have to copy-construct, wouldn't it?

PR #1193, opened 2025-07-02
Adhere to PyPA binary distribution format for built wheels

LuisAlfredoNu:
**Context:**
The Python packaging authority (PyPA) have specified expectations for package names that are coming into force by PyPI for shipped wheels https://packaging.python.org/en/latest/specifications/binary-distribution-format/

To quote verbatim:

```
In distribution names, any run of -_. characters (HYPHEN-MINUS, LOW LINE and FULL STOP) should be replaced 
with _ (LOW LINE), and uppercase characters should be replaced with corresponding lowercase ones. This is 
equivalent to regular [name normalization](https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization) 
followed by replacing - with _. Tools consuming wheels must be prepared to accept . (FULL STOP) and uppercase letters, 
however, as these were allowed by an earlier version of this specification.
```

Modifying the `pyproject.toml` and the script to generate it for the different backends. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-93339]


scripts/configure_pyproject_toml.py
```diff
@@ -109,10 +109,10 @@ def parse_args():
 
     plugin = "pennylane_lightning." + backend + ":Lightning" + suffix
 
-    pkg_suffix = "" if suffix == "Qubit" else "_" + suffix
+    pkg_suffix = "" if suffix == "Qubit" else "_" + suffix.lower()
```
josephleekl: The logic here is kind of weird - if suffix = `lightning_kokkos` , then `kokkos`, then `Kokkos`, then we lower it back to `kokkos`. Can this be simplified?

pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["cmake", "ninja; platform_system!='Windows'", "setuptools>=75.8.1",
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "PennyLane_Lightning"
+name = "pennylane_lightning"
 description = "PennyLane-Lightning plugin"
 readme = "README.rst"
 requires-python = ">=3.10"
```
maliasadi, line 9: Line 17 -- we could fix this for other lightning sims except LQ!

scripts/configure_pyproject_toml.py
LuisAlfredoNu: Done!

pyproject.toml
LuisAlfredoNu: Done!

LuisAlfredoNu:
> Thanks @LuisAlfredoNu! Could you also check PTM latest/latest?


[latest-latest](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/16054464003)
:heavy_check_mark: :green_circle:

josephleekl (approved):
LGTM, thanks @LuisAlfredoNu !

PR #1198, opened 2025-07-03
Add Nanobind Bindings for Lightning Qubit Backend

AmintorDusko:
## Description
This PR adds Nanobind-based Python bindings for the Lightning Qubit backend, replacing the previous Pybind11 implementation. Nanobind offers improved performance and reduced binary size compared to Pybind11.

## Key changes:

Created LQubitBindings_nb.hpp with Nanobind-specific implementations
Added support for all existing Lightning Qubit functionality in the Nanobind interface
Maintained API compatibility with the existing Pybind11 bindings
Added tests to verify the Nanobind implementation works correctly
Implemented graceful fallback (in most of the cases) to Pybind11 when Nanobind is not available
The implementation follows the same pattern used for other Lightning backends, with backend-specific functionality registered through template functions. This ensures consistent behavior across all Lightning backends while allowing for backend-specific optimizations.

Performance testing shows improved import times and reduced memory usage with the Nanobind implementation.

[sc-92715]

### DIFF
--- a/pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
+++ b/pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
@@ -87,14 +87,12 @@ createNumpyArrayFromVector(const std::vector<VectorT> &data, std::size_t rows,
  * @brief Convert complex matrices from nanobind ndarray to std::vector
  *
  * @tparam ComplexT Complex type to convert to
- * @tparam ParamT Precision type of the complex numbers
  * @param matrices Vector of ndarrays containing matrices
  * @return std::vector<std::vector<ComplexT>> Converted matrices
  */
-template <typename ComplexT, typename ParamT>
+template <typename ComplexT>
 std::vector<std::vector<ComplexT>> convertMatrices(
-    const std::vector<nb::ndarray<const std::complex<ParamT>, nb::c_contig>>
-        &matrices) {
+    const std::vector<nb::ndarray<const ComplexT, nb::c_contig>> &matrices) {
     std::vector<std::vector<ComplexT>> conv_matrices(matrices.size());
 
     for (std::size_t i = 0; i < matrices.size(); i++) {
--- a/pennylane_lightning/core/bindings/Bindings_nb.hpp
+++ b/pennylane_lightning/core/bindings/Bindings_nb.hpp
@@ -26,7 +26,6 @@
 #include <string>
 #include <tuple>
 #include <unordered_map>
-#include <variant>
 #include <vector>
 
 #include <nanobind/nanobind.h>
@@ -123,6 +122,8 @@ using namespace Pennylane::LightningTensor::TNCuda::NanoBindings;
 static_assert(false, "Backend not found.");
 #endif
 
+namespace nb = nanobind;
+
 /// @cond DEV
 namespace {
 using Pennylane::NanoBindings::Utils::createNumpyArrayFromVector;
@@ -131,11 +132,9 @@ using Pennylane::Util::CPUMemoryModel;
 } // namespace
 /// @endcond
 
-namespace nb = nanobind;
 namespace Pennylane::NanoBindings {
-
 /**
- * @brief Register applyMatrix.
+ * @brief Register applyMatrix
  */
 template <class StateVectorT>
 void registerMatrix(
@@ -154,7 +153,7 @@ void registerMatrix(
 }
 
 /**
- * @brief Register StateVector class.
+ * @brief Register StateVector class
  *
  * @tparam StateVectorT Statevector type to register
  * @tparam PyClass Nanobind's class object type
@@ -189,7 +188,7 @@ void registerGatesForStateVector(PyClass &pyclass) {
 }
 
 /**
- * @brief Register controlled gate operations for a statevector.
+ * @brief Register controlled gate operations for a statevector
  *
  * @tparam StateVectorT State vector type
  * @tparam PyClass Nanobind class type
@@ -227,12 +226,12 @@ void registerControlledGates(PyClass &pyclass) {
         });
 }
 /**
- * @brief Create an aligned array for a given type, memory model and array size.
+ * @brief Create an aligned array for a given type, memory model and array size
  *
  * @tparam VectorT Datatype of array to create
  * @param memory_model Memory model to use
  * @param size Size of the array to create
- * @return Nanobind ndarray
+ * @return nb::ndarray<VectorT, nb::numpy, nb::c_contig>
  */
 template <typename VectorT>
 auto alignedArray(CPUMemoryModel memory_model, std::size_t size, bool zeroInit)
@@ -256,35 +255,37 @@ auto alignedArray(CPUMemoryModel memory_model, std::size_t size, bool zeroInit)
 }
 
 /**
- * @brief Allocate aligned array with specified dtype.
+ * @brief Allocate aligned array with specified dtype
  *
  * @param size Size of the array to create
- * @param dtype Data type as string ("complex64", "complex128", "float32",
- * "float64")
+ * @param dtype Python dtype object to create the array with
  * @param zeroInit Whether to initialize the array with zeros
- * @return Nanobind ndarray
+ * @return nb::object A NumPy array with the specified dtype
  */
-auto allocateAlignedArray(std::size_t size, const std::string &dtype,
-                          bool zeroInit) -> nb::object {
+auto allocateAlignedArray(std::size_t size, const nb::object &dtype,
+                          bool zeroInit = false) -> nb::object {
     auto memory_model = bestCPUMemoryModel();
 
-    if (dtype == "complex64") {
-        return nb::cast(
-            alignedArray<std::complex<float>>(memory_model, size, zeroInit));
-    } else if (dtype == "complex128") {
-        return nb::cast(
-            alignedArray<std::complex<double>>(memory_model, size, zeroInit));
-    } else if (dtype == "float32") {
+    // Convert dtype to string representation
+    std::string dtype_str = nb::cast<std::string>(dtype.attr("name"));
+
+    if (dtype_str == "complex64") {
+        return nb::cast(alignedArray<std::complex<float>>(memory_model, size, zeroInit));
+    } else if (dtype_str == "complex128") {
+        return nb::cast(alignedArray<std::complex<double>>(memory_model, size, zeroInit));
+    } else if (dtype_str == "float32") {
         return nb::cast(alignedArray<float>(memory_model, size, zeroInit));
-    } else if (dtype == "float64") {
+    } else if (dtype_str == "float64") {
         return nb::cast(alignedArray<double>(memory_model, size, zeroInit));
     }
 
-    throw std::runtime_error("Unsupported dtype: " + dtype);
+    throw std::runtime_error("Unsupported dtype: " + dtype_str);
 }
 
 /**
- * @brief Register array alignment functionality.
+ * @brief Register array alignment functionality
+ *
+ * @param m Nanobind module
  */
 void registerArrayAlignmentBindings(nb::module_ &m) {
     // Add allocate_aligned_array function
@@ -294,7 +295,7 @@ void registerArrayAlignmentBindings(nb::module_ &m) {
 }
 
 /**
- * @brief Return basic information of runtime environment.
+ * @brief Return basic information of runtime environment
  */
 nb::dict getRuntimeInfo() {
     using Pennylane::Util::RuntimeInfo;
@@ -309,9 +310,9 @@ nb::dict getRuntimeInfo() {
 }
 
 /**
- * @brief Get compile information as a dictionary.
+ * @brief Get compile information as a dictionary
  *
- * @return Dictionary with compile information.
+ * @return Dictionary with compile information
  */
 nb::dict getCompileInfo() {
     using namespace Pennylane::Util;
@@ -369,9 +370,9 @@ nb::dict getCompileInfo() {
 }
 
 /**
- * @brief Register bindings for general info.
+ * @brief Register bindings for general info
  *
- * @param m Nanobind module.
+ * @param m Nanobind module
  */
 void registerInfo(nb::module_ &m) {
     // Add compile info
@@ -382,7 +383,7 @@ void registerInfo(nb::module_ &m) {
 }
 
 /**
- * @brief Register backend-agnostic observables.
+ * @brief Register backend-agnostic observables
  *
  * @tparam LightningBackendT
  * @param m Nanobind module
@@ -506,8 +507,8 @@ void registerBackendAgnosticObservables(nb::module_ &m) {
  * @tparam StateVectorT State vector type
  * @param M Measurements object
  * @param wires Vector of wire indices
- * @return nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy,
- * nb::c_contig> Array with probabilities in numpy format
+ * @return nb::ndarray<PrecisionT, nb::numpy, nb::c_contig> NumPy array with
+ * probabilities
  */
 template <class StateVectorT>
 nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy, nb::c_contig>
@@ -523,8 +524,8 @@ probsForWires(Measurements<StateVectorT> &M,
  *
  * @tparam StateVectorT State vector type
  * @param M Measurements object
- * @return nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy,
- * nb::c_contig> Array with probabilities in numpy format
+ * @return nb::ndarray<PrecisionT, nb::numpy, nb::c_contig> NumPy array with
+ * probabilities
  */
 template <class StateVectorT>
 nb::ndarray<typename StateVectorT::PrecisionT, nb::numpy, nb::c_contig>
@@ -541,8 +542,8 @@ probsForAllWires(Measurements<StateVectorT> &M) {
  * @param M Measurements object
  * @param num_wires Number of wires
  * @param num_shots Number of shots
- * @return nb::ndarray<std::size_t, nb::numpy, nb::c_contig> 2D array with
- * samples in numpy format
+ * @return nb::ndarray<std::size_t, nb::numpy, nb::c_contig> NumPy 2D array with
+ * samples
  */
 template <class StateVectorT>
 nb::ndarray<std::size_t, nb::numpy, nb::c_contig>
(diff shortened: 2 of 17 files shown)


pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -256,35 +255,37 @@ auto alignedArray(CPUMemoryModel memory_model, std::size_t size, bool zeroInit)
...
- * "float64")
+ * @param dtype Python dtype object to create the array with
  * @param zeroInit Whether to initialize the array with zeros
- * @return Nanobind ndarray
+ * @return nb::object A NumPy array with the specified dtype
  */
-auto allocateAlignedArray(std::size_t size, const std::string &dtype,
-                          bool zeroInit) -> nb::object {
+auto allocateAlignedArray(std::size_t size, const nb::object &dtype,
+                          bool zeroInit = false) -> nb::object {
```
jzaia18, line 285: What is the benefit of returning this as an `nb::object`? Is this to enable DLPack support for other libraries? If so it might be a good idea to leave a comment about this for future readability

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -44,29 +63,92 @@ using StateVectorBackends =
...
+    }
+
+    return res_map;
+}
+
+/**
+ * @brief Register controlled matrix kernel.
+ */
+template <class StateVectorT>
+void applyControlledMatrix(
```
jzaia18: Is there any reason this couldn't be unified in `Bindings_nb.hpp`?

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -147,12 +292,27 @@ void registerBackendSpecificStateVectorMethods(PyClass &pyclass) {
             }
 
             // Copy data to numpy array
-            ComplexT *data_ptr = static_cast<ComplexT *>(state.data());
+            ComplexT *data_ptr = state.data();
             std::copy(sv.getData(), sv.getData() + sv.getLength(), data_ptr);
         },
         "Copy StateVector data into a Numpy array.", nb::arg("state"));
```
jzaia18:
```suggestion
        "Copy StateVector data into a Numpy-like array.", nb::arg("state"));
```

Since this should also work with e.g. JAX arrays

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -170,7 +429,79 @@ void registerBackendSpecificMeasurements(PyClass &) {} // pyclass
...
+            return self == other_cast;
+        },
+        "Compare two observables");
+}
+
+/**
+ * @brief Register Vector Jacobian Product.
+ */
+template <class StateVectorT>
+auto registerVJP(
```
jzaia18, line 425: I think we had some discussion about whether or not this is actually unused and we should remove it. Perhaps now is the time to do so? CC: @maliasadi @tomlqc

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -179,13 +510,40 @@ void registerBackendSpecificObservables(nb::module_ &) {} // m
...
+
+    const std::string bitsize =
+        std::to_string(sizeof(std::complex<PrecisionT>) * 8);
+
+    std::string class_name;
+
+    //***********************************************************************//
+    //                        Vector Jacobian Product
+    //***********************************************************************//
+    class_name = "VectorJacobianProductC" + bitsize;
```
jzaia18, line 464: Same comment as above

tests/bindings/test_lqubit_bindings_nb.py
```diff
@@ -0,0 +1,338 @@
...
+    def test_kernel_map(self, current_nanobind_module, precision, get_statevector_class):
+        """Test getting kernel map."""
+        StateVectorClass = get_statevector_class(current_nanobind_module, precision)
+        num_qubits = 2
+        sv = StateVectorClass(num_qubits)
+        kernel_map = sv.kernel_map()
+        # Check that kernel map contains expected keys
+        assert "PauliX" in kernel_map
+        assert "Hadamard" in kernel_map
+        assert "CNOT" in kernel_map
```
jzaia18, line 138: Is it possible to loop through all expected gates somehow? e.g. by grabbing the list of supported gates from the toml file

tests/bindings/test_lqubit_bindings_nb.py
```diff
@@ -0,0 +1,338 @@
...
+        assert sparse_ham.get_wires() == wires
+
+        # Check string representation
+        assert "SparseHamiltonian" in sparse_ham.__repr__()
+
+
+class TestLQubitVJPBindings:
+    """Tests for LightningQubit-specific VectorJacobianProduct bindings."""
+
+    @pytest.mark.parametrize("precision", ["64", "128"])
```
jzaia18, line 312: Same comment as above

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
```diff
@@ -161,7 +321,106 @@ void registerBackendSpecificStateVectorMethods(PyClass &pyclass) {
...
  */
 template <class StateVectorT, class PyClass>
-void registerBackendSpecificMeasurements(PyClass &) {} // pyclass
+void registerBackendSpecificMeasurements(PyClass &pyclass) {
+    using PrecisionT = typename StateVectorT::PrecisionT;
+    using ComplexT = typename StateVectorT::ComplexT;
+
+    using np_arr_c = nb::ndarray<std::complex<PrecisionT>, nb::c_contig>;
+    using SparseIndexT = std::size_t;
+    using np_arr_sparse_ind = nb::ndarray<SparseIndexT, nb::c_contig>;
```
jzaia18: Minor nitpick, but since these don't have to actually be numpy arrays it might be better to name these something else (e.g. `ArrayT`)

jzaia18 (commented):
Nice work, just a few comments

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: Nanobind is very strict with data types; returning a nb::object offers similar functionality as before. The feature allows returning different data types without templating.

AmintorDusko: I will expand the return type explanation, though. Thanks!

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: Probably not. I assume you checked. In this case, I will move it.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: You are right. My bad. 😅

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: I will wait for more feedback on that.

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
AmintorDusko: Same comment as above

tests/bindings/test_lqubit_bindings_nb.py
AmintorDusko: It might be, but this isn't a pressing issue right now, is it?

tests/bindings/test_lqubit_bindings_nb.py
AmintorDusko: Same comment as above

AmintorDusko (commented):
Thank you for your review, @jzaia18.

jzaia18 (approved):
Looks good!

.github/CHANGELOG.md
```diff
@@ -72,6 +75,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Added Nanobind-based Python bindings for Lightning Qubit as an alternative to PyBind11.
```
maliasadi, line 78: Lightning-Qubit

pennylane_lightning/.clang-tidy
```diff
@@ -1,7 +1,7 @@
 ---
 Checks:          '-*,clang-diagnostic-*,clang-analyzer-*,-llvmlibc-*,modernize-*,-modernize-use-trailing-return-type,clang-analyzer-cplusplus*,openmp-*,performance-*,portability-*,readability-*,-modernize-avoid-c-arrays,-readability-magic-numbers,hicpp-*,-hicpp-no-array-decay,-hicpp-avoid-c-arrays,bugprone-suspicious-*,llvm-namespace-comment,cppcoreguidelines-slicing,cppcoreguidelines-special-member-functions,-readability-identifier-length'
 WarningsAsErrors: '*'
-HeaderFilterRegex: '.*'
+HeaderFilterRegex: '^(?!.*nanobind)(?!.*pybind11)(?!.*kokkos)(?!.*catalyst).*$'
```
maliasadi, line 4: nice!

pennylane_lightning/core/bindings/BindingsUtils_nb.hpp
```diff
@@ -39,17 +39,16 @@ createNumpyArrayFromVector(const std::vector<VectorT> &data,
             "Data size does not match the specified shape");
     }
 
-    // Allocate new memory and copy the data
-    VectorT *new_data = new VectorT[total_size];
-    std::memcpy(new_data, data.data(), total_size * sizeof(VectorT));
+    std::vector<VectorT> *new_data = new std::vector<VectorT>(std::move(data));
```
maliasadi: What about using `auto` instead of explicit type repetitions in these cases:
```
auto *new_data = ...
```

pennylane_lightning/core/utils/Util.hpp
```diff
@@ -665,6 +665,8 @@ template <typename DestType, typename SrcType>
 inline auto PL_reinterpret_cast(const SrcType *src_ptr) -> const DestType * {
     static_assert(sizeof(DestType) == sizeof(SrcType),
                   "Types must have the same size for reinterpretation");
+    static_assert(alignof(DestType) == alignof(SrcType),
```
maliasadi, line 668: :raised_hands:

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -589,26 +608,43 @@ void registerBackendAgnosticMeasurements(PyClass &pyclass) {
...
-                                        observables,
-                                        operations,
-                                        trainableParams};
-    adjoint_jacobian.adjointJacobian(std::span{jac}, jd, sv);
-    return createNumpyArrayFromVector<PrecisionT>(jac);
+
+    const std::string bitsize =
+        std::to_string(sizeof(std::complex<PrecisionT>) * 8);
+
+    std::string class_name = "AdjointJacobianC" + bitsize;
```
maliasadi, line 622: wouldn't be cheaper to check the `PrecisionT` type and set `bitsize_str` either `64` or `128`?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -636,13 +672,19 @@ OpsData<StateVectorT> createOpsList(
...
+    std::vector<std::vector<ComplexT>> conv_matrices(ops_matrices.size());
+
+    for (std::size_t i = 0; i < ops_matrices.size(); i++) {
+        if (ops_matrices[i].size() > 0) {
+            const auto *m_ptr =
+                PL_reinterpret_cast<const ComplexT>(ops_matrices[i].data());
+            const auto m_size = ops_matrices[i].size();
+            conv_matrices[i] = std::vector<ComplexT>(m_ptr, m_ptr + m_size);
+        }
+    }
```
maliasadi: what's the motivation behind inlining `Pennylane::NanoBindings::Utils::convertMatrices` here?

maliasadi: Could you also rewrite this with pre-defined `std::transform` from algorithm?

pennylane_lightning/core/bindings/Bindings_nb.hpp
```diff
@@ -724,12 +762,55 @@ void registerBackendAgnosticAlgorithms(nb::module_ &m) {
...
+        [](StateVectorT &sv, const std::vector<std::size_t> &state,
+           const std::vector<std::size_t> &wires, const bool async) {
+            if constexpr (requires { sv.setBasisState(state, wires, async); }) {
+                sv.setBasisState(state, wires, async);
+            } else {
+                sv.setBasisState(state, wires);
+            }
+        },
+        "Set the state vector to a computational basis state.",
+        nb::arg("state") = nullptr, nb::arg("wires") = nullptr,
```
maliasadi: Shouldn't we use `nb::none()` instead of standard c++ `nullptr?

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings.hpp
```diff
@@ -260,7 +260,7 @@ void registerBackendClassSpecificBindings(PyClass &pyclass) {
             },
             "Copy StateVector data into a Numpy array.")
         .def(
-            "UpdateData",
+            "updateData",
```
maliasadi, line 263: :+1:

pennylane_lightning/lightning_base/_serialize.py
```diff
@@ -541,7 +541,7 @@ def get_wires(operation, single_op):
                 else:
                     if hasattr(self.sv_type, name):
                         params.append(single_op_base.parameters)
-                        mats.append([])
+                        mats.append(np.array([]))
```
maliasadi, line 544: What's the problem with python lists?

pennylane_lightning/lightning_qubit/_adjoint_jacobian.py
```diff
@@ -28,6 +28,7 @@
 except ImportError as ex:
     warn(str(ex), UserWarning)
 
+
```
maliasadi: minor: extra blank line

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -112,15 +112,19 @@ def _apply_state_vector(self, state, device_wires: Wires):
         if sp.sparse.issparse(state):
             state = state.toarray().flatten()
         elif isinstance(state, self._qubit_state.__class__):
-            state_data = allocate_aligned_array(state.size, np.dtype(self.dtype), True)
+            state_data = allocate_aligned_array(state.size(), np.dtype(self.dtype), True)
```
maliasadi, line 115: is `size` an attribute or property? I thought it's an attribute of `state` :thinking:

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -50,12 +50,22 @@
...
-except ImportError as ex:
-    warn(str(ex), UserWarning)
-    LQ_CPP_BINARY_AVAILABLE = False
+except ImportError:
+    try:
+        # Fall back to pybind11 version if nanobind version is not available
+        from pennylane_lightning.lightning_qubit_ops import backend_info
+
+        warn(
+            "Using legacy pybind11 bindings. Consider upgrading to nanobind version for better performance.",
```
maliasadi: Could we raise RuntimeError instead of UserWarning so that we can capture such failures in CIs?

tests/bindings/test_bindings_nb.py
```diff
@@ -305,3 +305,44 @@ def test_compile_info(self):
         # Check for expected keys
         for key in ["cpu.arch", "compiler.name", "compiler.version", "AVX2", "AVX512F"]:
             assert key in info
+
+
+class TestAlignedArrayNB:
```
maliasadi, line 308: Nice set of unit tests!

tests/bindings/test_lqubit_bindings_nb.py
```diff
@@ -0,0 +1,338 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for LightningQubit-specific Nanobind bindings."""
+
+import numpy as np
+import pytest
+from conftest import backend
+
+# Skip all tests if not using lightning.qubit
+if backend != "qubit":
```
maliasadi: shouldn't it be "lightning.qubit"? If not, when did we change this to "qubit"?! :smile:

tests/bindings/test_lqubit_bindings_nb.py
```diff
@@ -0,0 +1,338 @@
...
+            if hasattr(module, class_name):
+                return getattr(module, class_name)
+            pytest.skip(f"Class {class_name} not available in module")
+
+        return _get_class
+
+    @pytest.mark.parametrize("precision", ["64", "128"])
+    def test_init(self, current_nanobind_module, precision, get_statevector_class):
+        """Test initialization of state vector."""
+        StateVectorClass = get_statevector_class(current_nanobind_module, precision)
```
maliasadi, line 43: What about defining a fixture for the sv class w/ C64 and C128 and use that single one instead of `(self, ..., current_nanobind_module, precision, ...)` in these unit tests?

tests/bindings/test_lqubit_bindings_nb.py
maliasadi, line 1: Thanks for adding all these unit tests :star_struck:

maliasadi (commented):
Awesome work! :100:

pennylane_lightning/core/simulators/lightning_qubit/bindings/LQubitBindings_nb.hpp
maliasadi: Let's not remove this! We'll integrate this binding with the rest of the LQ stack later in the quarter.

tests/bindings/test_lqubit_bindings_nb.py
AmintorDusko: I was working on it in the MPI AdjJac PR. I still have some places to update but this is the idea.

tests/bindings/test_lqubit_bindings_nb.py
AmintorDusko: I was prototyping, and at some point, I reduced the name because it made sense. I will bring it back, because if not, this would be one more type of naming we need to keep track of.

.github/CHANGELOG.md
AmintorDusko: We have both ways in our docs and changelog. Do you want to set Lightning-Something as the official way?

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: You are right. It makes even more sense.

pennylane_lightning/lightning_qubit/lightning_qubit.py
AmintorDusko: Of course, we can, but are we planning to keep pybind11 and nanobind coexisting for some time? I added it mostly as a sanity check; I was considering removing it before merging if we were removing pybind11. I'm not sure now.

AmintorDusko: I removed this fallback.

.github/CHANGELOG.md
maliasadi: Yeah, Lightning-Something w/ one hyphen instead of "Lightning Something" is preferable. But you're right we have both versions in CL. So, no action is required here, thanks!

pennylane_lightning/lightning_base/_serialize.py
maliasadi: chatted with Jake offline -- this will be fixed in a followup

pennylane_lightning/lightning_qubit/_state_vector.py
AmintorDusko: Are you suggesting any change?

pennylane_lightning/core/bindings/Bindings_nb.hpp
AmintorDusko: probably yes

pennylane_lightning/lightning_qubit/_state_vector.py
maliasadi: My question was for `size` whether it is an attribute or property (`.size` vs. `.size()`)

maliasadi (approved):
LGTM! 🙌

pennylane_lightning/lightning_qubit/_state_vector.py
AmintorDusko: At this point, `state` is a numpy array, and numpy arrays have both a property `size` and a method `size()`. The difference between them is subtle, and in our case, none. Just as a curiosity:

- the property `size` will always return the total number of elements, and this is all it does.
- the method `size()`, if we don't specify anything else, is equal to the property `size`, but in this case, we have the option of specifying an axis for counting the elements.

I update `size` in the LQubit state vector because we had a mix of both.

PR #1210, opened 2025-07-15
Add "device" as a supported MCM method that resolves to "tree-traversal"

astralcai:
**Context:**
It was previously determined that `mcm_method="device"` is the convention for using the device-native mcm method, one that does not require any transforms on the PL side. Now that tree-traversal is supported on the lightning devices, it can be considered a device native MCM method.

Update the Lightning devices that support tree-traversal to treat `mcm_method="device"` as tree-traversal.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-95418]

### DIFF
--- a/pennylane_lightning/lightning_gpu/lightning_gpu.py
+++ b/pennylane_lightning/lightning_gpu/lightning_gpu.py
@@ -350,7 +350,7 @@ class LightningGPU(LightningBase):
         new_device_options.update(mcmc_default)
 
         mcm_supported_methods = (
-            ("deferred", "tree-traversal", "one-shot", None)
+            ("device", "deferred", "tree-traversal", "one-shot", None)
             if not qml.capture.enabled()
             else ("deferred", "single-branch-statistics", None)
         )
@@ -359,6 +359,9 @@ class LightningGPU(LightningBase):
 
         if (mcm_method := mcm_config.mcm_method) not in mcm_supported_methods:
             raise DeviceError(f"mcm_method='{mcm_method}' is not supported with lightning.gpu.")
+
+        if mcm_config.mcm_method == "device":
+            mcm_config = replace(mcm_config, mcm_method="tree-traversal")
 
         if qml.capture.enabled():
 
@@ -374,8 +377,9 @@ class LightningGPU(LightningBase):
             elif mcm_method is None:
                 mcm_updated_values["mcm_method"] = "deferred"
 
-            updated_values["mcm_config"] = replace(mcm_config, **mcm_updated_values)
+            mcm_config = replace(mcm_config, **mcm_updated_values)
 
+        updated_values["mcm_config"] = mcm_config
         return replace(config, **updated_values, device_options=new_device_options)
 
     def preprocess(self, execution_config: ExecutionConfig = DefaultExecutionConfig):
--- a/pennylane_lightning/lightning_gpu/lightning_gpu.py
+++ b/pennylane_lightning/lightning_gpu/lightning_gpu.py
@@ -29,7 +29,7 @@ import numpy as np
 import pennylane as qml
 from numpy.random import BitGenerator, Generator, SeedSequence
 from numpy.typing import ArrayLike
-from pennylane.devices import DefaultExecutionConfig, ExecutionConfig
+from pennylane.devices import DefaultExecutionConfig, ExecutionConfig, MCMConfig
 from pennylane.devices.capabilities import OperatorProperties
 from pennylane.devices.modifiers import simulator_tracking, single_tape_support
 from pennylane.devices.preprocess import (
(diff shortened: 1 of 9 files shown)


tests/test_native_mcm.py
```diff
@@ -257,7 +258,7 @@ def test_qnode_mcm_method(self, mocker, mcm_method, shots):
             spy_one_shot.assert_called_once()
             spy_deffered.assert_not_called()
             spy_tree_traversal.assert_not_called()
-        elif mcm_method == "tree-traversal":
+        elif mcm_method in ("device", "tree-traversal"):
```
LuisAlfredoNu: Could you create a test called `test_qnode_default_mcm_method_device` that spy `tree-traversal`, please/

astralcai: Do you mean when `mcm_method` is not provided, default to tree-traversal?

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -492,5 +466,39 @@ def get_c_interface():
         return LightningBase.get_c_interface_impl("LightningGPUSimulator", "lightning_gpu")
 
 
+def _resolve_mcm_method(mcm_config: MCMConfig):
```
maliasadi, line 469: Could we move this helper method to the [base class](https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/lightning_base/lightning_base.py)? @LuisAlfredoNu this can be added in a followup PR.

maliasadi (commented):
LGTM! Just a comment about `_resolve_mcm_method` methods

tests/test_native_mcm.py
LuisAlfredoNu: I refer that when you set `mcm_method="device"`, then should call the tree-travesal method  :smile:

astralcai: Isn't that what's happening here? In this test case, if `mcm_method` is `device`, the tree traversal methods are spied.

LuisAlfredoNu: Yes, but it is better to be completely explicit on the tests and have a unique test for each default case. :smiley:

pennylane_lightning/lightning_gpu/lightning_gpu.py
LuisAlfredoNu: Good idea. It will be better to move this to the base class. @maliasadi sure

tests/test_native_mcm.py
astralcai: Sure, done.

tests/test_native_mcm.py
```diff
@@ -261,6 +262,10 @@ def test_qnode_mcm_method(self, mocker, mcm_method, shots):
             spy_tree_traversal.assert_called_once()
             spy_deffered.assert_not_called()
             spy_one_shot.assert_not_called()
+        elif mcm_method == "device":
+            spy_tree_traversal.assert_called_once()
+            spy_deffered.assert_not_called()
+            spy_one_shot.assert_not_called()
```
LuisAlfredoNu, line 285:
```suggestion


@pytest.mark.parametrize("shots", [None, 10])
def test_qnode_default_mcm_method_device(self, mocker):
        """Test the default mcm method is used for analytical simulation"""
        spy_deferred = mocker.spy(qml.defer_measurements, "_transform")
        spy_dynamic_one_shot = mocker.spy(qml.dynamic_one_shot, "_transform")
        spy_tree_traversal = mocker.patch(
            "pennylane_lightning.lightning_base.lightning_base.mcm_tree_traversal"
        )

        circuit = self.generate_mcm_circuit(
            device_kwargs={"wires": 3, "shots": shots},
            qnode_kwargs={"mcm_method": "device"},
            mcm_kwargs={},
        )

        _ = circuit(np.pi / 8)

        spy_deferred.assert_not_called()
        spy_dynamic_one_shot.assert_not_called()
        spy_tree_traversal.assert_called_once()
```

tests/test_native_mcm.py
```diff
@@ -43,6 +43,7 @@ def get_device(wires, **kwargs):
 @pytest.fixture(
     scope="function",
     params=[
+        "device",
```
astralcai:
```suggestion
```

tests/test_native_mcm.py
```diff
@@ -261,6 +262,10 @@ def test_qnode_mcm_method(self, mocker, mcm_method, shots):
             spy_tree_traversal.assert_called_once()
             spy_deffered.assert_not_called()
             spy_one_shot.assert_not_called()
+        elif mcm_method == "device":
+            spy_tree_traversal.assert_called_once()
+            spy_deffered.assert_not_called()
+            spy_one_shot.assert_not_called()
```
astralcai:
```suggestion
```

LuisAlfredoNu (approved):
Thank you @astralcai 
LGTM!

PR #1223, opened 2025-08-01
[Nanobind] Remove Pybind11 and update/clean code

AmintorDusko:
This is the final PR.
Intended for removing Pybind11 and general cleaning.


CMakeLists.txt
```diff
@@ -220,8 +211,7 @@ target_include_directories(pennylane_lightning  INTERFACE "$<INSTALL_INTERFACE:$
 #####################################################
 if(ENABLE_PYTHON)
     message(STATUS "ENABLE_PYTHON is ON.")
-    # Existing PyBind11 module
-    pybind11_add_module("${PL_BACKEND}_ops" "pennylane_lightning/core/bindings/Bindings.cpp")
+    # Only build Nanobind module
```
jzaia18:
```suggestion
    # Build Nanobind module
```

pennylane_lightning/core/bindings/Bindings.cpp
```diff
@@ -1,91 +0,0 @@
-// Copyright 2018-2024 Xanadu Quantum Technologies Inc.
```
jzaia18, line 1: Would it maybe be better to rename all of the `_nb` files to match these old filenames? For example, after removing the pybind11 `Bindings.hpp`, rename `Bindings_nb.hpp` to `Bindings.hpp`

jzaia18 (approved):
Awesome work!

pennylane_lightning/core/bindings/Bindings.cpp
AmintorDusko: This is the idea, indeed.

pennylane_lightning/core/bindings/Bindings.hpp
```diff
@@ -15,528 +15,774 @@
...
+using Pennylane::NanoBindings::Utils::createNumpyArrayFromVector;
 using Pennylane::Util::bestCPUMemoryModel;
 using Pennylane::Util::CPUMemoryModel;
-using Pennylane::Util::getMemoryModel;
+using Pennylane::Util::PL_reinterpret_cast;
 } // namespace
 /// @endcond
 
-namespace py = pybind11;
+namespace Pennylane::NanoBindings {
```
jzaia18, line 135: Should we rename this namespace to something like `Pennylane::Bindings`? (Honestly a nitpick, either way is fine)

AmintorDusko: I'm more inclined in keeping it NanoBindings

PR #1240, opened 2025-08-25
[Drafdt] Move stop condition logic to be LQubit exclusive

AmintorDusko:
Moving the new stop condition logic to be LQubit exclusive.


AmintorDusko:
I don;t think there will be more work on this front. Feel free to open and continue the effort otherwise.

PR #1254, opened 2025-09-10
QuantumDevice interface updates to lightning for dynamic allocation

dime10:
The current implementation of Catalyst's `QuantumDevice` interface in lightning has some issues, in that it doesn't follow the required semantics. Specifically, around partial qubit deallocation and their effect on other interface functions, like `GetNumQubits` and MP functions. 

In preparation for the upcoming dynamic allocation feature in PennyLane, these inconsistencies need to be fixed. We are also replacing the `ReleaseAllQubits` function with a `ReleaseQubits([ids...])`, which was a longstanding inconsistency between the Runtime C-API and the QuantumDevice interface.

Note: This PR will be merged, then the corresponding Catalyst PR will be merged. We will then have a [new separate PR](https://github.com/PennyLaneAI/pennylane-lightning/pull/1258) to undo the changes to our CI to track their main branch again.


.clang-format
```diff
@@ -0,0 +1,4 @@
+BasedOnStyle: llvm
```
paul0403: Remember to remove these before merging : )

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+                           });
+    }
+
+    auto getDeviceId(ProgramQubitID program_id) const -> DeviceQubitID {
+        RT_FAIL_IF(!isValidQubitId(program_id), "Invalid program qubit ID");
+
+        return this->qubit_id_map.at(program_id);
+    }
+
+    auto getDeviceId(const std::vector<ProgramQubitID> &program_ids) const
```
paul0403: Do we want to call the plural ones `isValidQubitIds`, `getDeviceIds`, etc.?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -65,17 +95,43 @@ auto LightningSimulator::AllocateQubits(size_t num_qubits)
...
+
+    // TODO: We don't have to check whether the released qubit is pure, but it
+    // is something we can add easily to catch program errors.
 }
 
 auto LightningSimulator::GetNumQubits() const -> size_t {
-    return this->device_sv->getNumQubits();
+    // This needs to produce the currently allocated number of qubits,
+    // not the internal state size.
+    return this->qubit_manager.getNumQubits();
```
paul0403, line 120: Thanks for fixing this!!

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+ * Additionally, the class maintains a collection of qubits that are inactive
+ * (i.e. have been freed) but whose memory is not physically released to the OS.
+ * This mechanism is an optimization to allow efficient reuse of simulator
+ * resources.
+ */
+template <typename ProgramQubitID = QubitIdType,
+          typename DeviceQubitID = size_t>
+class QubitManager {
+  private:
+    std::map<ProgramQubitID, DeviceQubitID> qubit_id_map{};
```
mlxd:
```suggestion
    std::unorderd_map<ProgramQubitID, DeviceQubitID> qubit_id_map{};
```

Tends to have better lookup scaling in practice.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+    std::set<DeviceQubitID> free_device_qubits;
+    ProgramQubitID next_program_id{0};
+
+  public:
+    QubitManager() = default;
+    ~QubitManager() = default;
```
mlxd: Can we make this `virtual`, or mark the class itself `final`?

.clang-format
dime10: I added it intentionally since the repo doesn't have one, better integration with IDEs (or just calling the tool from the command line in general).

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: I'm ambivalent, the existing one didn't do it and C++ has function overloading, but if desired I'm happy to change it

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: Whatever you want to do here, I'm no c++ wiz 😅

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -22,29 +25,56 @@
...
+            size_t upper_bits = (idx >> device_idx) << (device_idx + 1);
+            size_t full_idx = upper_bits | fixed_bit | lower_bits;
+            data[full_idx] = std::complex<double>(0., 0.);
+        }
 
-    for (auto dst = data.end() - 2; src != data.begin();
-         std::advance(src, -1), std::advance(dst, -2)) {
-        *dst = *src;
-        *src = std::complex<double>(.0, .0);
+        double norm = std::sqrt(squaredNorm(data.data(), data.size()));
```
dime10: @PennyLaneAI/performance Could you help me express this in Kokkos/GPU devices? 

I suppose `data` is actually residing on the CPU after it is fetched with `this->device_sv->getDataVector()`, is that correct?
With Kokkos the datatype is `std::vector<Kokkos::complex<double>>`, which can't be passed to `std::sqrt`. I'm guessing the issue is with `squaredNorm`, since it's supposed to [remove](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/utils/TypeTraits.hpp#L31) the complex part, but that only works with `std::complex`.

josephleekl: is what you're doing here equivalent to collapsing the specific qubit to the 0 state? in this case each of the statevectors have a collapse method ([LQ](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L881), [LGPU](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/simulators/lightning_gpu/StateVectorCudaManaged.hpp#L666), [LKokkos](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/simulators/lightning_kokkos/StateVectorKokkos.hpp#L797)).

But for [LQ](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L908) and [LK](https://github.com/PennyLaneAI/pennylane-lightning/blob/0f7e77a9d90ec39d7186331c0c2dc53b646438a8/pennylane_lightning/core/simulators/lightning_kokkos/StateVectorKokkos.hpp#L812) we also have normalize methods. We can add one for LGPU as well if that's easier.

dime10: Oh nice, yes it is just collapsing it onto 0 :)

.clang-format
mlxd: Lightning does has format files already, at the top-level

mlxd: https://github.com/PennyLaneAI/pennylane-lightning/blob/master/bin/format
Sorry, my comment was misleading, they are encapsulated into this Python generator.

mlxd: Also, if adding new CF files, I'd prefer we start to migrate to using the https://github.com/PennyLaneAI/guidance-docs/blob/master/config_files/.clang-format files instead across the board.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: [aa14ce5](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/aa14ce58f4a48bedf2ce7792dbe6f9c3f6bf93c1)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: [bcad12b](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/bcad12b90233607f898ac1e797063f2d01d99618)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: [f3450a9](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/f3450a9f392274a33f148edda320e153fd5a8813)

.clang-format
dime10: >https://github.com/PennyLaneAI/pennylane-lightning/blob/master/bin/format
Sorry, my comment was misleading, they are encapsulated into this Python generator.

Yes, the problem is they are only available when using this binary though.

>Also, if adding new CF files, I'd prefer we start to migrate to using the https://github.com/PennyLaneAI/guidance-docs/blob/master/config_files/.clang-format files instead across the board.

I just copied the existing format spec, I don't want to make any sweeping changes to the repo in this PR.

mlxd: Let's remove this for now as @paul0403 suggested and come back to it in a separate follow-on.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -65,17 +82,43 @@ auto LightningSimulator::AllocateQubits(size_t num_qubits)
...
+    }
 }
 
 void LightningSimulator::ReleaseQubit(QubitIdType q) {
+    // We do not deallocate physical memory in the statevector for this
+    // operation, instead we just mark the qubits as released.
+    // TODO: need to ensure the semantic behaviour of releasing qubits is still
+    // maintained,
+    //       in particular: measurements must not compute results for released
+    //       qubits, only for active/allocated qubits
```
dime10, line 110: I'll have a second PR update the measurement process functions to handle a partially deallocated statevector.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -65,17 +82,43 @@ auto LightningSimulator::AllocateQubits(size_t num_qubits)
...
     this->qubit_manager.Release(q);
+
+    // TODO: We don't have to check whether the released qubit is pure, but it
+    // is something we can add easily to catch program errors.
```
dime10, line 114: This is optional but a nice to have for users I think. Just something to consider for later

pennylane_lightning/core/simulators/lightning_qubit/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -81,7 +83,9 @@ TEST_CASE("test AllocateQubits", "[Driver]") {
 
     auto &&q = sim->AllocateQubits(2);
 
-    sim->ReleaseQubit(q[0]);
+    // TODO: releasing doesn't rescale the state
+    //       either remove from test or re-enable after fix
+    // sim->ReleaseQubit(q[0]);
```
dime10: This is more a question regarding the intent of the test. Does it mean to check the statevector after a qubit has been released? If so, this can be re-enabled in my second PR. Otherwise, we might want to remove it.

Note that `GetNumQubits()` now returns the number of qubits that are active and allocated, not the size of the underlying statevector. `State()` currently returns the underlying statevector including any freed qubits (to be fixed).

pennylane_lightning/core/simulators/lightning_gpu/catalyst/CMakeLists.txt
```diff
@@ -10,7 +10,10 @@ include(FetchContent)
 include("${pennylane_lightning_SOURCE_DIR}/cmake/support_catalyst.cmake")
 FindCatalyst(lightning_gpu_catalyst)
 
-target_include_directories(lightning_gpu_catalyst INTERFACE  ${CMAKE_CURRENT_SOURCE_DIR})
+target_include_directories(lightning_gpu_catalyst PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_CURRENT_SOURCE_DIR}/../../lightning_qubit/catalyst
```
dime10: Let me know if this is fine or you would like the header to be moved somewhere else. What I wanted to avoid is duplicate the same header 3 times, once for each device.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+ * This is done for safety to detect any use-after-free or other anomalies.
+ *
+ * Additionally, the class maintains a collection of qubits that are inactive
+ * (i.e. have been freed) but whose memory is not physically released to the OS.
+ * This mechanism is an optimization to allow efficient reuse of simulator
+ * resources.
+ */
+template <typename ProgramQubitID = QubitIdType,
+          typename DeviceQubitID = size_t>
+class QubitManager final {
```
josephleekl: Why is this class not based on the common QubitManager class in Catalyst? Why do we need a separate one here? I might be missing some background context

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -22,29 +25,43 @@
...
+    // that means we may not need to resize the vector.
+    size_t device_idx;
+    std::optional<size_t> candidate = this->qubit_manager.popFreeQubit();
+    if (!candidate.has_value()) {
+        assert(dsize == 1UL << num_qubits);
+        data.resize(dsize << 1UL);
+        device_idx = num_qubits;
+
+        // zero the new data and move existing amplitudes
+        auto src = data.begin();
```
josephleekl: actually this seems like a very convoluted way to do this interleaving with not much obvious benefits 😅 can we do something like this instead:
```
const auto& original_data = this->device_sv->getDataVector();
const size_t dsize = original_data.size();
...
if(...){
std::vector<std::complex<double>, AlignedAllocator<std::complex<double>>> new_data(dsize << 1UL);
for (size_t i = 0; i < data.size(); ++i) {
    new_data[2 * i] = original_data[i];
}
this->device_sv = std::make_unique<StateVectorT>(new_data);
...
```

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -65,17 +82,43 @@ auto LightningSimulator::AllocateQubits(size_t num_qubits)
     return result;
 }
 
-void LightningSimulator::ReleaseAllQubits() {
-    this->qubit_manager.ReleaseAll();
-    this->device_sv = std::make_unique<StateVectorT>(0); // reset the device
+void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
+    // fast path for single register alloc and dealloc
+    if (this->GetNumQubits() == ids.size()) {
+        std::vector<QubitIdType> allocated_ids =
```
josephleekl, line 84: this might be faster:
```
const auto& allocated_ids = this->qubit_manager.getAllQubitIds();
const std::unordered_set<QubitIdType> allocated_set(allocated_ids.begin(),allocated_ids.end());
const bool all_ids_are_allocated =
        std::all_of(ids.begin(), ids.end(), [&](QubitIdType id) {
            return allocated_set.contains(id); 
        });

    if (all_ids_are_allocated) {
        this->qubit_manager.ReleaseAll();
        this->device_sv = std::make_unique<StateVectorT>(0);
        return;
}
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -12,33 +12,51 @@
...
-    const std::size_t dsize = data.size();
-    data.resize(dsize << 1UL);
+    const size_t dsize = data.size();
+
+    // The statevector may contain previously freed qubits,
+    // that means we may not need to resize the vector.
+    size_t device_idx;
+    std::optional<size_t> candidate = this->qubit_manager.popFreeQubit();
+    if (!candidate.has_value()) {
+        assert(dsize == 1UL << num_qubits);
```
josephleekl: `RT_ASSERT`?

pennylane_lightning/core/simulators/lightning_gpu/catalyst/CMakeLists.txt
josephleekl: IMO this is OK, but I'm not sure why we're using a separate QubitManager?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: Sure thing, let me explain. The way the runtime model is setup, a QubitManager is fundamentally a component of a QuantumDevice, in fact, the runtime doesn't prescribe the existence of a QubitManager at all.

The "common" QubitManager class in Catalyst is not actually common at all, it was an implementation specifically for the lightning device plugins. In fact, it was written for the `StateVectorLQubitDynamic` that shrank the statevector upon each single-qubit release, as evidenced by the fact that the release function [shifts all larger device indices down by one](https://github.com/PennyLaneAI/catalyst/blob/1b7819f12faed9b5af6d519af7bfc141dd5a82a9/runtime/lib/backend/common/QubitManager.hpp#L55-L57), whereas the current ligthning device plugins don't shrink the statevector upon qubit release.
This is a major bug (that wen't unnoticed) because the QubitManager was treated as some universal class without care for the actual device behaviour. This PR remedies this by bringing the class close to the code it interacts with and aligning their behaviour.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: Yes please, happy to update this as I couldn't read the code myself either 😅 

Btw, is the `AlignedAllocator` necessary? I think only LQ has it, the others don't.
Similarly, some devices have `std::move` and some devices don't when shifting the amplitudes in the statevector. Let me know if this indeed needs to differ between devices or there is a preferred form I should use :)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
josephleekl: That makes sense, thanks! so will Catalyst's 'common' QubitManager be deleted?

pennylane_lightning/core/simulators/lightning_gpu/catalyst/CMakeLists.txt
josephleekl: Sorry to reopen this, I suggest we put `LightningQubitManager.hpp` in `pennylane-lightning/pennylane_lightning/core/simulators/base/catalyst/LightningQubitManager.hpp`.

It looks like there's a lot of commonalities/duplication between the 3 devices, we might consider refactoring more into a base class in the future!

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: Sure thing, happy to learn more about c++ :) 

From I what I read, a tree based can be faster than a hash-based set depending on the number of elements in it. But here you are only constructing one set instead of two. Another option I saw was just sorting both arrays and then comparing. In any case, I'll go with your suggestion!

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
dime10: Hmmm, good question. RT_ASSERT is a runtime exception I believe. Here I was thinking this is closer to an actual assertion, in the sense that I'm just checking the internal logical consistency of the code, and that it can be disabled.
Let me know if you would in fact prefer an exception (RT_ASSERT) over an assertion (assert) here.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningQubitManager.hpp
dime10: Imo it should at least be updated, but we still need QubitManagers for the devices located in the Catalyst repo. Whether they can share the same or not depends on the device behaviour which we need to look at.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
mlxd: > is the AlignedAllocator necessary? I think only LQ has it, the others don't.

Yep, LQ is the only one that has it, as it is a requirement for the AVX2/AVX512 kernels to be valid, which only LQ supports. It won't affect ARM users, but on x86 platforms it provides pretty substantial performance improvements if the better kernels can be chosen.

It *may* be possible to defer to another allocation strategy, but that'd be a project on it's own to determine if this is the case.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
josephleekl: I think they're all valid options :D hash-based indeed could be bad when there's hash-collision.

I mostly prefer this precisely because we only need to construct one extract unordered_set. Sorting both arrays will require having a mutable copy of the vector, which adds some cost. And inserting elements into the two sets in the original implementation is a bit more expensive!

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: Great, that's what I was looking for :)

josephleekl: Thanks @mlxd .

As for the std::move, I don't think that's necessary. These complex types should be trivially copyable, so there's no point in using `std::move`. You can ignore them.

However just want to note here that the implementations here will not be ideal for LKokkos or LGPU, since we're copying the data vector to the host, creating another vector on the host, then copy to the device/GPU. Ideally we want to do all of them directly on the device/GPU, but maybe we can come back to this in the future to optimize this, or we can do it now.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
josephleekl: assert does make sense here - we don't use it anywhere else in lightning probably because throwing an exception is better from Python execution. However for this case when Catalyst calls this and dies from an assert (and not handling an exception), if Catalyst is OK with this behaviour, then this is OK IMO.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: Okay good to know, that's what I thought I didn't the library types (e.g. Kokkos::) were different in this regard.

Regarding performing computations on device, I don't know much about it so I'll have to leave that to you guys for possible improvements 😅 

Thanks for all the insights though, it's great to learn more!

josephleekl: Yeah might be good to leave a TODO comment in the code to remind us to port this to the device/GPU in the future.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
dime10: Well, an assert is definitely worse from a user experience that's true. I guess the idea with asserts is that they should never trigger in production, but well... 😅 I can change it to an exception 👍

pennylane_lightning/core/simulators/lightning_gpu/catalyst/CMakeLists.txt
dime10: [82cadb9](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/82cadb979e72ef82494bda6a6c67b0c38ca78303)

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
dime10: [8565c2f](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/8565c2ff3b8cbebfd04913e30f3cd112a6873280)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: [68c6317](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/68c6317070477bb5dda3e9662a55ff3354cf62ea)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: [dbbe483](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/dbbe483033d9a15fae586185f13d75f99c8a8d26)

dime10: [3bf042f](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/3bf042f3e3f7313dff224f3be24de188f02ec0d2)
Added your suggestion, it does create an additional vector though, whereas the existing code makes the changes in-place.

.clang-format
dime10: [192846f](https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/192846f61c5d80f12661b3512ce091a0c9a35332)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: I had to revert the suggestion because it was causing a test failure, somehow it wasn't equivalent I guess :/

dime10:
@josephleekl I think all the comments have been addressed. One thing I just realized, the collapse method doesn't really work if the target subspace has 0 amplitude to begin with, or if the amplitude is really small it probably increase rounding errors quite a bit. We should probably use a better reset method here @paul0403.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
josephleekl: > I had to revert the suggestion because it was causing a test failure, somehow it wasn't equivalent I guess :/

Hmm that's weird - I'll try it out quickly

> it does create an additional vector though, whereas the existing code makes the changes in-place.

Actually the original implementation also uses a new vector:
`std::vector<std::complex<double>, AlignedAllocator<std::complex<double>>>
        data = this->device_sv->getDataVector();` is not a reference, so it also creates a new vector. We resize this to make it bigger later.

Whereas with the suggested code:
```
const auto& original_data = this->device_sv->getDataVector();
std::vector<std::complex<double>, AlignedAllocator<std::complex<double>>> new_data(dsize << 1UL);
```
we use a reference of the original SV, and directly create a new vector of the correct size. In both cases we then copy it to the `device_sv` with a new statevector object

josephleekl: I think the failure was from requiring to specify the allocator during initialization - i've updated them now.

josephleekl:
> One thing I just realized, the collapse method doesn't really work if the target subspace has 0 amplitude to begin with, or if the amplitude is really small it probably increase rounding errors quite a bit. We should probably use a better reset method here 

For this issue, we previously intentionally [removed throwing an exception](https://github.com/PennyLaneAI/pennylane-lightning/blob/abce3451acbbb7a0e0ab1a961e8b2e3dc8f26660/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L914) when the norm is close to 0 . I'm resolving this issue with @LuisAlfredoNu now
reference: https://xanaduhq.slack.com/archives/CA89H1BAN/p1748366977551839

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -19,36 +19,31 @@
...
-        std::advance(src, dsize - 1);
-        for (auto dst = data.end() - 2; src != data.begin();
-             std::advance(src, -1), std::advance(dst, -2)) {
-            *dst = std::move(*src);
-            *src = std::complex<double>(.0, .0);
-        }
+        std::vector<std::complex<double>> new_data(dsize << 1UL);
 
+        device_idx = num_qubits;
+        for (size_t i = 0; i < original_data.size(); ++i) {
```
mlxd:
```suggestion
        for (size_t i = 0; i < original_data.size(); i++) {
```

No need for pre-increment here. We default to post everywhere except for iterators that need it for correctness.

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -29,30 +29,23 @@ auto LightningKokkosSimulator::AllocateQubit() -> QubitIdType {
...
-        std::advance(src, dsize - 1);
+        std::vector<Kokkos::complex<double>> new_data(dsize << 1UL);
 
-        for (auto dst = data.end() - 2; src != data.begin();
-             std::advance(src, -1), std::advance(dst, -2)) {
-            *dst = std::move(*src);
-            *src = Kokkos::complex<double>(.0, .0);
+        device_idx = num_qubits;
+        for (size_t i = 0; i < original_data.size(); ++i) {
+            new_data[2 * i] = original_data[i];
```
mlxd:
```suggestion
        for (size_t i = 0; i < original_data.size(); i++) {
            new_data[2 * i] = original_data[i];
```

Same as above

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -30,29 +30,26 @@ auto LightningSimulator::AllocateQubit() -> QubitIdType {
...
-        // zero the new data and move existing amplitudes
-        auto src = data.begin();
-        std::advance(src, dsize - 1);
-        for (auto dst = data.end() - 2; src != data.begin();
-             std::advance(src, -1), std::advance(dst, -2)) {
-            *dst = *src;
-            *src = std::complex<double>(.0, .0);
+        device_idx = num_qubits;
+        for (size_t i = 0; i < original_data.size(); ++i) {
+            new_data[2 * i] = original_data[i];
```
mlxd: And once more

```suggestion
        for (size_t i = 0; i < original_data.size(); i++) {
            new_data[2 * i] = original_data[i];
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
josephleekl: updated https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/0a116b14e1295ac39b87d4786abd4056fd0f9f70

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/0a116b14e1295ac39b87d4786abd4056fd0f9f70

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/0a116b14e1295ac39b87d4786abd4056fd0f9f70

pennylane_lightning/core/simulators/lightning_qubit/catalyst/tests/Test_LightningDriver.cpp
josephleekl: I think we can remove this release here, and add a test after the second PR to test the updated Release/measurement behaviour

paul0403 (approved):
All good, thanks @josephleekl for the help!

Just a reminder that the issue with measurements not computing results for released qubits still needs a follow up PR, but for this one all good :+1:

.github/CHANGELOG.md
```diff
@@ -81,6 +81,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Update Catalyst's QuantumDevice interface and downstream Qubit Manager to Lightning in preparation for dynamic qubit allocation functionality.
```
maliasadi, line 84: Could you add David's name to the list of contributors?

.github/workflows/tests_gpu_cpp.yml
```diff
@@ -85,6 +85,7 @@ jobs:
         with:
           repository: PennyLaneAI/catalyst
           path: ${{ github.workspace}}/catalyst
+          ref: paul0403/rt_individual_dealloc
```
maliasadi: here and elsewhere :eyes:

cmake/support_catalyst.cmake
```diff
@@ -26,14 +26,13 @@ macro(FindCatalyst target_name)
 
     else()
         if(NOT CATALYST_GIT_TAG)
-            set(CATALYST_GIT_TAG "main" CACHE STRING "GIT_TAG value to build Catalyst")
+            set(CATALYST_GIT_TAG "paul0403/rt_individual_dealloc" CACHE STRING "GIT_TAG value to build Catalyst")
```
maliasadi: FIXME :)

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
maliasadi: don't need to update the year as we're downstreaming same code from catalyst 
```suggestion
// Copyright 2022-2025 Xanadu Quantum Technologies Inc.
```

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+#pragma once
+
+#include <algorithm>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "Exception.hpp"
+#include "Types.h"
+
+namespace Catalyst::Runtime::Simulator {
```
maliasadi, line 25: We need to update namespaces as well as double-checking the generated doxygen files in docs?

pennylane_lightning/core/catalyst/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.20)
+
+project(lightning_catalyst LANGUAGES CXX)
+
+add_library(lightning_catalyst INTERFACE)
+
+target_include_directories(lightning_catalyst INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
```
maliasadi:
```suggestion
project(lightning_catalyst_core LANGUAGES CXX)

add_library(lightning_catalyst_core INTERFACE)

target_include_directories(lightning_catalyst_core INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
```

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -0,0 +1,166 @@
...
+    QubitManager(QubitManager &&) = delete;
+    QubitManager &operator=(QubitManager &&) = delete;
+
+    auto getNumQubits() const -> size_t { return this->qubit_id_map.size(); }
+
+    auto isValidQubitId(ProgramQubitID program_id) const -> bool {
+        return this->qubit_id_map.contains(program_id);
+    }
+
+    auto isValidQubitId(const std::vector<ProgramQubitID> &program_ids) const
```
maliasadi, line 63: We actually need unit tests and a better documentation for all these public methods

.github/workflows/tests_gpu_cpp.yml
paul0403: This is temporary because the base class in catalyst cannot go in until lightning is updated to the new base class, but to update to the new base class lightning needs to see the new catalyst base class.... so this is a bit tricky. We decided to merge in lightning first, then the catalyst new base class https://github.com/PennyLaneAI/catalyst/pull/1996, then lightning goes back to catalyst main https://github.com/PennyLaneAI/pennylane-lightning/pull/1258

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
paul0403: Sorry, why do we need to update namespaces? :thinking:

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
paul0403: It's slightly different I think

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
maliasadi: This code doesn't live in Catlayst::Runtime anymore.

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
maliasadi: 2022-2025

.github/CHANGELOG.md
maliasadi: @paul0403 and also your name :smile:

.github/workflows/tests_gpu_cpp.yml
paul0403: Just merged in the catalyst PR

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
paul0403: But I think it's still used by the devices' catalyst runtime interface? Which is still in this namespace.

pennylane_lightning/core/catalyst/CMakeLists.txt
paul0403: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/2fde2a2307b431362108b5ed74b357219da6a0fd

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
paul0403: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/2fde2a2307b431362108b5ed74b357219da6a0fd

.github/CHANGELOG.md
paul0403: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/2fde2a2307b431362108b5ed74b357219da6a0fd

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
maliasadi: Ahh, we use the same namespace elsewhere

maliasadi (approved):
:raised_hands:

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
paul0403: https://github.com/PennyLaneAI/pennylane-lightning/pull/1254/commits/6b8287770affc83f090e640b208220e6d215b3be

PR #1260, opened 2025-09-18
[Catalyst] Properly reset released qubits by measuring

paul0403:
**Context:**
In #1254 , the Catalyst interfaces for lighting.qubit, kokkos and gpu were updated to ready for the new dynamic allocation feature.
One specific part was the reuse. Namely, previously released qubits are free to be reused in future allocations.

The `ReleaseQubit()` method essentially just records which qubits are in the recycle pool, and future calls to `AllocateQubit()` method should get a qubit from the pool and reset it to |0>.

However, the reset was just calling `collapse(0)`, which is usually fine, but then the qubit in the recycle pool was in |1> state, i.e. had no overlap with |0> at all, then the collapse would fail, because essentially the state vector is now having zero amplitude on both |0> and |1>, and normalization returns `nan`. 

We properly reset by performing a measurement, and inserting an X gate if |1> was measured.

**Benefits:**
Correct logic in qubit release and reuse for catalyst's dynamic allocation feature.

### DIFF
--- a/pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
+++ b/pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
 }
 
 TEST_CASE("Check dynamic qubit reuse", "[Driver]") {
-    std::unique_ptr<LSimulator> sim = std::make_unique<LSimulator>();
+    SECTION("Two total qubits") {
+        std::unique_ptr<LSimulator> sim = std::make_unique<LSimulator>();
 
-    auto qubits = sim->AllocateQubits(2);
-    sim->NamedOperation("PauliX", {}, {qubits[0]}, false);
-    sim->NamedOperation("Hadamard", {}, {qubits[1]}, false);
+        std::vector<intptr_t> Qs = sim->AllocateQubits(1);
+        CHECK(Qs[0] == 0);
+        std::vector<intptr_t> tempQs1 = sim->AllocateQubits(1);
+        CHECK(tempQs1[0] == 1);
+        sim->NamedOperation("PauliX", {}, {tempQs1[0]}, false);
+        sim->ReleaseQubits(tempQs1);
 
-    std::vector<std::complex<double>> state(1U << sim->GetNumQubits());
-    DataView<std::complex<double>, 1> view(state);
-    sim->State(view);
-    CHECK(state[0b00].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b10].real() == Approx(0.707107).epsilon(1e-5));
-    CHECK(state[0b01].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b11].real() == Approx(0.707107).epsilon(1e-5));
+        // Check that internal device state vector is still |01> after release
+        std::vector<std::complex<double>> state(4);
+        DataView<std::complex<double>, 1> view(state);
+        sim->State(view);
+        CHECK(state[0b00].real() == Approx(0.).epsilon(1e-5));
+        CHECK(state[0b10].real() == Approx(0.).epsilon(1e-5));
+        CHECK(state[0b01].real() == Approx(1.).epsilon(1e-5));
+        CHECK(state[0b11].real() == Approx(0.).epsilon(1e-5));
 
-    sim->ReleaseQubit(qubits[1]);
+        std::vector<intptr_t> tempQs2 = sim->AllocateQubits(1);
+        // Check that program ID is different for every allocation
+        CHECK(tempQs2[0] == 2);
 
-    sim->State(view);
-    CHECK(state[0b00].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b10].real() == Approx(0.707107).epsilon(1e-5));
-    CHECK(state[0b01].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b11].real() == Approx(0.707107).epsilon(1e-5));
+        // Check that the second allocation reuses bit 1, and device sv resets
+        // back to |00>
+        sim->State(view);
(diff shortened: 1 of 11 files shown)


pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
...
+        std::vector<intptr_t> Qs = sim->AllocateQubits(3); // |000>
+
+        std::vector<intptr_t> tempQs1 = sim->AllocateQubits(1); // |000> and |0>
+        sim->NamedOperation("PauliX", {}, {tempQs1[0]},
+                            false); // |000> and |1>
+        sim->NamedOperation("CNOT", {}, {tempQs1[0], Qs[1]},
+                            false);  // |010> and |1>
+        sim->ReleaseQubits(tempQs1); // |010>
+
+        std::vector<intptr_t> tempQs2 = sim->AllocateQubits(1); // |010> and |0>
```
josephleekl, line 282: Just for paranoia - can we also check the exact state after this AllocateQubits?

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
...
+        sim->NamedOperation("PauliX", {}, {tempQs2[0]},
+                            false); // |010> and |1>
+        sim->NamedOperation("CNOT", {}, {tempQs2[0], Qs[1]},
+                            false);  // |000> and |1>
+        sim->ReleaseQubits(tempQs2); // |000>
+
+        std::vector<intptr_t> tempQs3 = sim->AllocateQubits(1); // |000> and |0>
```
josephleekl, line 295: also after this AllocateQubits

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
...
+        sim->NamedOperation("PauliX", {}, {tempQs3[0]},
+                            false); // |000> and |1>
+        sim->NamedOperation("CNOT", {}, {tempQs3[0], Qs[1]},
+                            false);  // |010> and |1>
+        sim->ReleaseQubits(tempQs3); // |010>
```
josephleekl, line 306: Also the final state

josephleekl (commented):
Thanks @paul0403 , looks good! For the test, can you also check the exact state, so we can be confident that the intermediate states are as expected?

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
...
-    CHECK(state[0b11].real() == Approx(0.707107).epsilon(1e-5));
-
-    sim->ReleaseQubit(qubits[1]);
-
-    sim->State(view);
-    CHECK(state[0b00].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b10].real() == Approx(0.707107).epsilon(1e-5));
-    CHECK(state[0b01].real() == Approx(0.).epsilon(1e-5));
-    CHECK(state[0b11].real() == Approx(0.707107).epsilon(1e-5));
+    SECTION("Two total qubits") {
```
LuisAlfredoNu:
```suggestion
    SECTION("Two total qubits allocated") {
```
If I understand correctly, at the end you have 3 qubits right?

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
```diff
@@ -188,36 +188,107 @@ TEST_CASE("Check re-AllocateQubit", "[Driver]") {
...
+        std::unique_ptr<LSimulator> sim = std::make_unique<LSimulator>();
+
+        std::vector<intptr_t> Qs = sim->AllocateQubits(1);
+        CHECK(Qs[0] == 0);
+        std::vector<intptr_t> tempQs1 = sim->AllocateQubits(1);
+        CHECK(tempQs1[0] == 1);
+        sim->NamedOperation("PauliX", {}, {tempQs1[0]}, false);
```
LuisAlfredoNu, line 200: What happen if we don't apply the gate `PauliX` ? :thinking:

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
josephleekl: There are only 2 qubits, we allocate 1, allocate 1, release 1, allocate 1 (which re-uses) the released one

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
paul0403: I added a X gate because otherwise the state is just always zero, which is the default state and I feel like checking against that is not as meaningful.

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
paul0403: So technically there's 3 total allocation calls, but only 2 total qubits activated on the device, because one is reused. How about `"Two total device qubits"`?

pennylane_lightning/core/catalyst/tests/Test_LightningDriver.cpp
paul0403: https://github.com/PennyLaneAI/pennylane-lightning/pull/1260/commits/b38461cbbe27470b989a7c0e27dea8633f16c8bb

josephleekl (approved):
Awesome! Thanks @paul0403 🚀

Issue #1265, opened 2025-09-30
Forschungszentrum Jülich request: CUDA/13 support PennyLane-Lightning


AlvaroHuanay:
Dear PennyLaneAI team,

We are preparing our Stages/2026 HPC software stack, where PennyLane is widely used, and we would like to request support for CUDA 13 in the current release line (pennylane-lightning v0.42.0, [latest](https://github.com/PennyLaneAI/pennylane-lightning/tree/v0.42.0)). We followed from source installation and PYPI instructions for PennyLane_Lightning_GPU and PennyLane_Lightning_Tensor and always pulls CUDA/12 dependencies ([custatevec-cu12](https://github.com/PennyLaneAI/pennylane-lightning/blob/e65dac1acd7f85426bc3fde3fd9a1755be4d7dfd/doc/requirements.txt#L5), [cutensornet-cu12](https://github.com/PennyLaneAI/pennylane-lightning/blob/e65dac1acd7f85426bc3fde3fd9a1755be4d7dfd/doc/requirements.txt#L6C1-L6C17)...)

Could you please provide us a CUDA/13 release?

For reproducibility, here are the key dependencies in our stack:

Dependency | Version
-- | --
GLIBC | 2.28
GCC | 14.3.0
Python | 3.13.5
Numpy | 2.3.2
Scipy | 1.16.1
Pandas | 2.3.1
matplotlib | 3.10.5
matplotlib-inline | 0.1.7
Rust | 1.88.0
Pybind11 | 3.0.0
maturin | 1.9.1
Cython | 3.1.2
Cupy | 13.6.0

Any additional information you may need I will be glad to provide it.

Thank you very much in advance.

Kind regards.

mlxd:
Hi @AlvaroHuanay  thanks for reaching out.

> We are preparing our Stages/2026 HPC software stack, where PennyLane is widely used

Glad to hear. We'd always appreciate to know more about usage on your systems, if that information is available.

> Could you please provide us a CUDA/13 release?

Just for note, we tend to support only a single CUDA version in released (PyPI) Python wheels to avoid:
* Pollution of PyPI with many separate versions (e.g. `pennylane_lightning_tensor-cuXY`, `pennylane_lightning_gpu-cuXY`, etc)
* Shipping compiled for both modules, which can cause bloat in wheel size, and often only 1 is needed on a given machine
* Testing across Python versions, backends, CUDA SDKs and MPI vesrions can be quite heavy.

[...]

PR #1267, opened 2025-10-01
support compiling Lightning GPU and Tensor from source with CUDA 13

josephleekl:
### Before submitting

**Context:**
Currently Lightning GPU and Tensor do not support building with CUDA 13.

Allow users to choose building Lightning GPU and Tensor with CUDA 13. This can be enabled by:
- adding `PL_CUDA_VERSION` environment variable (or to `CMAKE_ARGS`), or
- auto detect CUDA version from NVCC, or
- default uses CUDA 12

Note: We are not pre-building wheels or performing CI tests with CUDA 13 yet - just enabling the possibility of building from source with CUDA 13.

Note: This has been tested with GCC-14.3.0

Note: The current version of Kokkos used (4.5.0) does not support compiling with CUDA 13. This support is only added  from [v4.7.1](https://github.com/kokkos/kokkos/pull/8328). Therefore we are not supporting building Lightning Kokkos with CUDA 13 yet.

**Benefits:**
Enables Lightning GPU and Tensor devices for systems with CUDA 13.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-100445]

Issue #1265

### DIFF
--- a/scripts/cuda_version.py
+++ b/scripts/cuda_version.py
@@ -26,6 +26,7 @@ import os
 import re
 import shutil
 import subprocess
+import warnings
 
 _ALLOWED_CUDA_MAJOR_VERSIONS = ("12", "13")
 _DEFAULT_CUDA_VERSION = "12"
@@ -54,6 +55,13 @@ def cuda_version():
     if version not in _ALLOWED_CUDA_MAJOR_VERSIONS:
         raise ValueError(
             f"Invalid CUDA version {version}. Allowed major versions are: {_ALLOWED_CUDA_MAJOR_VERSIONS}."
+        )
+
+    nvcc_version = _get_cuda_version_from_nvcc()
+    if version != nvcc_version:
+        warnings.warn(
+            f"Using CUDA version {version}, but found version {nvcc_version} from nvcc. Please ensure CUDA Toolkit version matches.",
+            UserWarning,
         )
 
     return version


.github/CHANGELOG.md
```diff
@@ -90,6 +90,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Support building `lightning.gpu` and `lightning.tensor` from source with CUDA 13.
+  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
```
```suggestion
  [(#1267)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1267)
```

scripts/configure_pyproject_toml.py
```diff
@@ -130,17 +133,21 @@ def parse_args():
...
+        dependencies += [f"custatevec-cu{cuda_version}"]
 
     if backend == "lightning_tensor":
-        dependencies += ["cutensornet-cu12", "nvidia-cusolver-cu12"]
+        dependencies += [f"cutensornet-cu{cuda_version}", f"nvidia-cusolver-cu{cuda_version}"]
 
     if backend in ("lightning_gpu", "lightning_tensor"):
+        package_suffix = ""
+        if cuda_version == "12":
+            package_suffix = "-cu12"
```
jzaia18: Do you need another case that handles when `cuda_version` is 13?

josephleekl: the custatevec ones have cu13 appended for CUDA 13, e.g.
https://pypi.org/project/custatevec-cu13/

but for the main packages they don't have 13 appended i.e.
https://pypi.org/project/nvidia-nvjitlink/
https://pypi.org/project/nvidia-cusparse/
https://pypi.org/project/nvidia-cublas/
https://pypi.org/project/nvidia-cuda-runtime/

(they have packages with -cu13 appended too, e.g. [nvidia-nvjitlink-cu13](https://pypi.org/project/nvidia-nvjitlink-cu13/#history) , but they're version 0 placeholders, which has not been updated)

scripts/configure_pyproject_toml.py
```diff
@@ -130,17 +133,21 @@ def parse_args():
...
 
     if backend == "lightning_gpu":
-        dependencies += ["custatevec-cu12"]
+        dependencies += [f"custatevec-cu{cuda_version}"]
 
     if backend == "lightning_tensor":
-        dependencies += ["cutensornet-cu12", "nvidia-cusolver-cu12"]
+        dependencies += [f"cutensornet-cu{cuda_version}", f"nvidia-cusolver-cu{cuda_version}"]
 
     if backend in ("lightning_gpu", "lightning_tensor"):
```
maliasadi, line 141: Could you add a dev note for the future reference?
```suggestion
    if backend in ("lightning_gpu", "lightning_tensor"):
        # We only use 12 appended for CUDA 12
        # as nvidia libraries don't use 13 appended
        # to be used for CUDA 13.
```

scripts/configure_pyproject_toml.py
maliasadi:
```suggestion
        package_suffix = "-cu12" if cuda_version == "12" else ""
```

scripts/cuda_version.py
```diff
@@ -0,0 +1,103 @@
...
+    cmake_args = os.environ.get("CMAKE_ARGS")
+    if cmake_args:
+        # Find an argument like 'PL_CUDA_VERSION=12'
+        match = re.search(r"PL_CUDA_VERSION=(\d+)", cmake_args)
+        if match:
+            return match.group(1)
+    return None
+
+
+def cuda_version():
```
maliasadi: We usually add protected methods after public methods in py modules

cmake/support_pllgpu.cmake
```diff
@@ -7,7 +7,7 @@ include_guard()
 
 # Macro to aid in finding CUDA lib
 macro(findCUDATK external_libs)
-    find_package(CUDAToolkit REQUIRED EXACT 12)
+    find_package(CUDAToolkit REQUIRED 12)
```
maliasadi, line 10: Do we need to further relax this requirement to be able to set it to 13 and 12 w.r.t. `PL_CUDA_VERSION`?

scripts/cuda_version.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Finds the CUDA version from multiple sources in a specific order of priority.
+
+    Priority:
+    1. Environment Variable: PL_CUDA_VERSION
+    2. Environment Variable: CMAKE_ARGS (containing PL_CUDA_VERSION=...)
+    3. System command: nvcc --version
```
maliasadi: Could we benefit from setting `PL_CUDA_VERSION` to `nvcc --version` to be used in support_pllgpu.cmake?

LuisAlfredoNu:
I am testing the installation and after creating the wheel for LGPU and installing it I found that the wheels try to install 
```
Downloading nvidia_cusparse-12.6.3.3-py3-none-manylinux2014_x86_64.manylinux_2_17_x86_64.whl (145.9 MB)
```

scripts/cuda_version.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns:
+        str: The determined CUDA version string (e.g., "12").
+    """
+    version = (
+        os.environ.get("PL_CUDA_VERSION")
+        or _get_cuda_version_from_cmake_args()
+        or _get_cuda_version_from_nvcc()
+        or _DEFAULT_CUDA_VERSION
+    )
```
LuisAlfredoNu: What if I set `PL_CUDA_VERSION=13` and I have `nvcc` with cu12 ? 

How we can set the default without 'nvcc' available? I mean, is it possible to install the LGPU without `nvcc` available ?

cmake/support_pllgpu.cmake
josephleekl: so with this currently `find_package(CUDAToolkit REQUIRED 12)` it accepts anything >=12, so it automatically works for 12 and 13 already. I considered making it something like `find_package(CUDAToolkit REQUIRED EXACT CUDA_VERSION)` and passing the CUDA_VERSION into CMake, but IMO that seems unnecessary, since nowhere else in the CMake do we need the CUDA version number.

scripts/cuda_version.py
josephleekl: As mentioned in another comment, the CUDA version is not used anywhere else in `support_pllgpu.cmake` except when finding the package. In the current implementation we accept both 12 and 13, so IMO we don't need to pass it in

scripts/cuda_version.py
josephleekl: > What if I set PL_CUDA_VERSION=13 and I have nvcc with cu12 ?

Then it will try and use CUDA 13. The user's option should have higher priority, and the nvcc route is just a helper tool in case users don't provide any setting and there's already a nvcc available. 

> How we can set the default without 'nvcc' available? I mean, is it possible to install the LGPU without nvcc available ?

The nvcc route is just a helper - for example the execution of `nvcc` could fail, it shouldn't fail here (it should fail when we actually try to compile), so we should still be able to continue here.

scripts/cuda_version.py
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/pull/1267/commits/d9286b917550630e238f10996d7c68af6e712ab5

scripts/configure_pyproject_toml.py
josephleekl: https://github.com/PennyLaneAI/pennylane-lightning/pull/1267/commits/d9286b917550630e238f10996d7c68af6e712ab5

scripts/cuda_version.py
LuisAlfredoNu: I agree that we should continue but at some point we should raise a warning when the user is requesting cu13 but everything will use cu12

cmake/support_pllgpu.cmake
josephleekl: So this `find_package(CUDAToolkit REQUIRED 12)` allows anything >=12, so it allows both 12 and 13. Since this version number is not used anywhere else in CMake, IMO this is simpler and more idiomatic for cmake, and doesn't introduce more complexity than it needs.

scripts/cuda_version.py
josephleekl: since `support_pllgpu.cmake` accepts both 12 and 13, it will already automatically detect which version you have, similar to nvcc --version. So I think it's not necessary to run this `cuda_version` function for cmake. So right now, if we don't specify anything (don't add any flags), both the configure step will use `nvcc` to auto detect, and `support_pllgpu.cmake` will use `find_package` to auto detect, and the versions should match. I think this is cleaner, but happy to hear any other suggestions :D

scripts/cuda_version.py
josephleekl: I've added a warning if the version supplied doesn't match the nvcc version
https://github.com/PennyLaneAI/pennylane-lightning/pull/1267/commits/afb9ace3578b1e49d93d9d8ddd0d6e87aaa79a25

scripts/cuda_version.py
maliasadi: Sounds good! :+1:

cmake/support_pllgpu.cmake
LuisAlfredoNu: What about 
```suggestion
    find_package(CUDAToolkit REQUIRED 12...13)
```
to make it more restricted with the version? 
https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature


> The [version] argument requests a version with which the package found should be compatible. There are two possible forms in which it may be specified:
>
>    A single version with the format major[.minor[.patch[.tweak]]], where each component is a numeric value.
>
>    A version range with the format versionMin...[<]versionMax where versionMin and versionMax have the same format and constraints on components being integers as the single version. By default, both end points are included. By specifying <, the upper end point will be excluded. Version ranges are only supported with CMake 3.19 or later.

josephleekl: I tested it, but it doesn't actually make a difference. I believe it's due to:
```
Note

With the exception of CPS packages, version support is currently provided 
only on a package-by-package basis. When a version range is specified but 
the package is only designed to expect a single version, the package will ignore 
the upper end point of the range and only take the single version at the lower 
end of the range into account. Non-CPS packages that do support version 
ranges do so in a manner that is determined by the individual package.
```

it ignores the upper bound, e.g. if you specify 11...12, it still passes with CUDA 13.

LuisAlfredoNu (approved):
Thanks @josephleekl 
LGTM

PR #1290, opened 2025-10-23
[WIP] The preprocess transform program use the correct gate set

astralcai:
**Context:**
Update the preprocess transform programs to use the `gate_set` generated from the `capabilities` object.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
Requires https://github.com/PennyLaneAI/pennylane/pull/8532
[sc-102069]

### DIFF
--- a/pennylane_lightning/lightning_gpu/lightning_gpu.py
+++ b/pennylane_lightning/lightning_gpu/lightning_gpu.py
@@ -174,7 +174,7 @@ def _add_adjoint_transforms(program: TransformProgram, device_wires=None) -> Non
         name=name,
         skip_initial_state_prep=False,
         device_wires=device_wires,
-        target_gates=LightningGPU.capabilities.gate_set(differntiable=True),
+        target_gates=LightningGPU.capabilities.gate_set(differentiable=True),
     )
     program.add_transform(validate_observables, accepted_observables, name=name)
     program.add_transform(


JerryChen97:
Is this `differentiable=True` enough for fixing this bug? @astralcai also, lightning tensor has its own hardcoded operation list, should we also fix it

astralcai:
> Is this `differentiable=True` enough for fixing this bug? @astralcai also, lightning tensor has its own hardcoded operation list, should we also fix it

`differentiable=True` basically skips all operators in the TOML file that doesn't have "differentiable" on it. Does `lightning.tensor` even supports adjoint differentiation? Note that `differentiable=True` is only added to the decompose in the adjoint transform. The regular `decompose` does not have that.

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -174,7 +174,7 @@ def _add_adjoint_transforms(program: TransformProgram, device_wires=None) -> Non
         name=name,
         skip_initial_state_prep=False,
         device_wires=device_wires,
-        target_gates=LightningGPU.capabilities.operations.keys(),
+        target_gates=LightningGPU.capabilities.gate_set(differntiable=True),
```
astralcai: For the adjoint transform, only differentiable gates are in the target gate set.

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -402,7 +406,7 @@ def preprocess_transforms(
             skip_initial_state_prep=True,
             name=self.name,
             device_wires=self.wires,
-            target_gates=self.capabilities.operations.keys(),
+            target_gates=self.capabilities.gate_set(),
```
astralcai, line 409: Normally you don't need `differentiable=True` @JerryChen97

JerryChen97: Oh thanks a lot for clarification!

pennylane_lightning/lightning_gpu/lightning_gpu.py
jzaia18:
```suggestion
        target_gates=LightningGPU.capabilities.gate_set(differentiable=True),
```
Assuming this was a typo?

jzaia18 (approved):
LGTM

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
```diff
@@ -165,7 +165,7 @@ def _add_adjoint_transforms(program: TransformProgram, device_wires=None) -> Non
         name=name,
         skip_initial_state_prep=False,
         device_wires=device_wires,
-        target_gates=LightningKokkos.capabilities.operations.keys(),
+        target_gates=LightningKokkos.capabilities.gate_set(differentiable=True),
```
comp-phys-marc, line 168: Why do we set `differentiable=True` here but not in other places?

pennylane_lightning/lightning_gpu/lightning_gpu.py
JerryChen97: @comp-phys-marc Here's the reason we only put `differentiable=True` for adjoint transforms

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
JerryChen97: This adjoint method is only for differentiable ops

pennylane_lightning/lightning_gpu/lightning_gpu.py
comp-phys-marc: Ok, I got it. Thanks!

pennylane_lightning/lightning_kokkos/lightning_kokkos.py
comp-phys-marc: ok!

PR #1295, opened 2025-11-17
fix: latest/latest after PennyLane updated MCM validation

andrijapau:
https://github.com/PennyLaneAI/pennylane/pull/8596 changed the handling of mcm method validation. 

Passes L/L [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/19513175226) and S/S [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/19514554399).

[sc-86555]

### DIFF
--- a/tests/lightning_base/test_device.py
+++ b/tests/lightning_base/test_device.py
@@ -741,18 +741,6 @@ class TestExecution:
 
     @pytest.mark.skipif(
         device_name == "lightning.tensor",
-        reason="lightning.tensor device doesn't have support for program capture.",
-    )
-    def test_preprocess_invalid_mcm_method_error(self, enable_disable_plxpr):
-        """Test that an error is raised if mcm_method is invalid."""
-        device = LightningDevice(wires=1)
-        config = ExecutionConfig(mcm_config=MCMConfig(mcm_method="foo"))
-
-        with pytest.raises(DeviceError, match="mcm_method='foo' is not supported"):
-            _ = device.setup_execution_config(config)
-
-    @pytest.mark.skipif(
-        device_name == "lightning.tensor",
         reason="lightning.tensor device does not support mcms",
     )
     def test_decompose_conditionals(self):


tests/test_native_mcm.py
```diff
@@ -87,7 +87,7 @@ def test_unsupported_method(self):
             obs=qml.PauliZ(0),
         )
         with pytest.raises(
-            DeviceError, match=f"mcm_method='{method}' is not supported with {device_name}"
```
josephleekl, line 90: Thanks @andrijapau - please also add an entry to the changelog!

tests/test_native_mcm.py
JerryChen97, line 1: Seems Pennylane's new validation at post init works well now. Is the original expected error msg from Lightning? If so should we care about its coverage?

tests/test_native_mcm.py
```diff
@@ -87,7 +87,7 @@ def test_unsupported_method(self):
...
+            DeviceError, match=f"'{method}' is not a valid mcm_method."
```
andrijapau:
```suggestion
            ValueError, match=f"'{method}' is not a valid mcm_method."
```

tests/test_native_mcm.py
andrijapau: Done!

LuisAlfredoNu: if we want to remove this test, we also need to modify the following function:
https://github.com/PennyLaneAI/pennylane-lightning/blob/ebfcae49d9d5637553424f440157577c3911126c/pennylane_lightning/lightning_base/lightning_base.py#L754

tests/test_native_mcm.py
andrijapau: Yeah, the old message is from the device's resolution of the mcm method. But now the executionconfig takes care of that.

tests/lightning_base/test_device.py
```diff
@@ -739,18 +739,6 @@ def test_sbs_and_postselect_warning(self, enable_disable_plxpr, postselect_mode)
...
             _ = device.setup_execution_config(config)
 
-    @pytest.mark.skipif(
-        device_name == "lightning.tensor",
-        reason="lightning.tensor device doesn't have support for program capture.",
-    )
-    def test_preprocess_invalid_mcm_method_error(self, enable_disable_plxpr):
-        """Test that an error is raised if mcm_method is invalid."""
-        device = LightningDevice(wires=1)
-        config = ExecutionConfig(mcm_config=MCMConfig(mcm_method="foo"))
```
andrijapau, line 749: This would fail with latest PL - not a supported MCM method

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -757,7 +757,7 @@ def resolve_mcm_method(mcm_config: MCMConfig, tape: QuantumScript | None, device
     mcm_supported_methods = (
         ("device", "deferred", "tree-traversal", "one-shot", None)
         if not qml.capture.enabled()
-        else ("deferred", "single-branch-statistics", None)
+        else ("device", "deferred", "single-branch-statistics", None)
```
andrijapau, line 760: "device" should be allowed if capture is enabled

tests/test_native_mcm.py
andrijapau: Removing as PL handles this validation - not lightning

JerryChen97:
project coverage of lightning PR's is ignoreable

andrijapau:
> project coverage of lightning PR's is ignoreable

yep just was waiting to see if L/L and S/S pass - they do!

PR #1312, opened 2025-12-19
Cache intermediate states in adjoint/VJP to eliminate redundant forward pass in backward

maliasadi:
**Context:**
Cache intermediate states in adjoint/VJP to eliminate redundant forward pass in backward.

Currently, in the backward pass (adjoint-jacobian, vjp), there are redundant calls to the forward pass to get the final state. This new cache approach will help to avoid redundant computations by storing the final state after the forward pass and reusing it in the backward pass. This works with single or multiple backward calls after a single forward pass. The current cache is per-device, so if multiple devices are used, each device will have its own cache and will be freed when the device is deallocated. This boosts the performance of backward execution on CPUs significantly by avoiding extra forward executions of the same circuit. For GPU devices, the memory walls and high compute demands make thie caching ineffective at scale, so it is disabled by default for GPU devices.

**Benefits:**
- Checked the backward/fordward ratio of referenced example (PR 1311) for LQ (24, 4, 2), this PR reduces the wall-time ratio from `8.1` -> `5.6`.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-106746]

### DIFF
--- a/pennylane_lightning/lightning_qubit/lightning_qubit.py
+++ b/pennylane_lightning/lightning_qubit/lightning_qubit.py
@@ -429,9 +429,8 @@ class LightningQubit(LightningBase):
             execution_config = ExecutionConfig()
 
         # Cache intermediate states only if using adjoint differentiation with device VJP
-        self._intermediate_states = (
-            {} if execution_config.use_device_jacobian_product else None
-        )  # pylint: disable=attribute-defined-outside-init
+        # # pylint: disable=attribute-defined-outside-init
+        self._intermediate_states = {} if execution_config.use_device_jacobian_product else None
 
         mcmc = {
             "mcmc": self._mcmc,


pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -477,6 +564,82 @@ def supports_derivatives(
...
+            [circuit], _ = qml.map_wires(circuit, wire_map)
+
+        final_state = None
+        if self._intermediate_states is not None:
+            final_state = self._intermediate_states.get(circuit.hash, None)
+            print("Using cached state for jacobian:", final_state is not None)
+        else:
+            print("No cached state for jacobian")
+            state.reset_state()
+            final_state = state.get_final_state(circuit)
```
jzaia18:
```suggestion
        if self._intermediate_states is not None and (final_state := self._intermediate_states.get(circuit.hash, None) is not None):
            print("Using cached state for jacobian:", final_state is not None)
        else:
            print("No cached state for jacobian")
            state.reset_state()
            final_state = state.get_final_state(circuit)
```

What if there are intermediate states but not for the circuit we care about? Something like this seems a bit safer

jzaia18: (Same comment below)

maliasadi: It’s still WIP and not the final version, hence all the print debugging 😄

maliasadi: > What if there are intermediate states but not for the circuit we care about? Something

For circuits without forward execution, we fallback to the prev implementation w/o necessarily copying the intermediate states. I also updated the patch to ensure that we only cache circuits when there are follow-up calls to either vjp, grad or adjoint-jacobian.

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -75,6 +75,9 @@ class LightningBase(Device):
             computing the jacobian.
     """
 
+    # Dictionary to store intermediate states for backward methods
+    _intermediate_states: dict | None = None
```
jzaia18, line 81: Is there any guarantee of a maximum size this can become? If not, should we add some mechanism to prevent this from "leaking" memory? (not a true leak but to keep us from holding onto large chunks when unneeded)

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -333,12 +347,20 @@ def vjp(  # pylint: disable=too-many-arguments, too-many-positional-arguments
...
+                circuit, cotangents
+            )
+
         state.reset_state()
-        final_state = state.get_final_state(circuit)
         # pylint: disable=not-callable
-        return self.LightningAdjointJacobian(final_state, batch_obs=batch_obs).calculate_vjp(
-            circuit, cotangents
-        )
+        return self.LightningAdjointJacobian(
```
jzaia18: We should probably cache the final state that we use as input here in `_intermediate_states`

pennylane_lightning/lightning_qubit/lightning_qubit.py
```diff
@@ -429,6 +429,8 @@ def execute(
         if execution_config is None:
             execution_config = ExecutionConfig()
 
+        super().execute(circuits, execution_config)
```
jzaia18: What's the reason for this addition?

pennylane_lightning/lightning_base/lightning_base.py
maliasadi: I see your point, but given the CPU's compute bottleneck and the fact that our LQ simulator is already memory-bound at the same ratio, additional checks would be redundant.

pennylane_lightning/lightning_base/lightning_base.py
maliasadi: Good point! Fixed [0910652](https://github.com/PennyLaneAI/pennylane-lightning/pull/1312/commits/0910652b5039ec31071a12fdac838e9c5367b8e0)

pennylane_lightning/lightning_qubit/lightning_qubit.py
maliasadi: I've added a check in the base class to only enable caching only in the backward pass https://github.com/PennyLaneAI/pennylane-lightning/pull/1312/files/cd99e570551ed52961eef3d6c3eb49e3f20955ad#diff-d17764476a1fbb30d4327f8f0aa15b79af8c58f03ce5d5c83323945961474e4dR168

maliasadi: This will let us to explore this caching for other devices now and then

pennylane_lightning/lightning_base/lightning_base.py
jzaia18: Would something the following code not cause an OOM issue?

```
dev = qml.device("lightning.qubit", wires=24, device_vjp=True)

for _ in range(100):
    qnode = # create a new qnode
    qml.grad(qnode)() # <--- cache the qnode, leaking a statevector size of memory each iteration
```

jzaia18: Oh nevermind, it seems like each execute resets this cache so we shouldn't have an issue

.github/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 
 <h3>Improvements 🛠</h3>
 
+- Cache intermediate states in the adjoint-jacobian pipeline to eliminate redundant forward pass in backward in `lightning.qubit`.
```
jzaia18:
```suggestion
- Cache intermediate states in the adjoint-jacobian pipeline to eliminate redundant forward pass computed during backward pass in `lightning.qubit`.
```

pennylane_lightning/core/simulators/lightning_gpu/algorithms/AdjointJacobianGPU.hpp
```diff
@@ -246,7 +246,8 @@ class AdjointJacobian final
             return;
         }
 
-        std::chrono::duration<double, std::milli> zero_time = std::chrono::duration<double, std::milli>::zero();
+        std::chrono::duration<double, std::milli> zero_time =
```
tomlqc: Only need to remove all the chrono stuff

pennylane_lightning/lightning_qubit/_state_vector.py
```diff
@@ -91,6 +91,18 @@ def _state_dtype(self):
         """
         return StateVectorC128 if self.dtype == np.complex128 else StateVectorC64
 
+    def copy_state_class(self):
```
tomlqc: why is this called `copy_state_class`? why `class`?

maliasadi: I mingled the name with `_class` as this method copies the state-vector class instance and not  just the underlying state-vector. I can update the name to `copy_sv` or something else if `copy_state_class` is confusing?

PR #1315, opened 2025-12-22
Fix CIs: numpy 2.4.0 and transform attr updates

maliasadi:
**Context:**
- Address transform attr name change https://github.com/PennyLaneAI/pennylane/pull/8756
- Pin numpy <2.4 on CIs as `pyscf` is not compatible with numpy 2.4 yet https://github.com/PennyLaneAI/pennylane/pull/8813/files

**Benefits:**
[latest/latest](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/20445147042) :green_circle: 

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/tests_gpu_python.yml
+++ b/.github/workflows/tests_gpu_python.yml
@@ -152,7 +152,7 @@ jobs:
         run: |
           python -m pip install -r requirements-tests.txt
           # pyscf is not compatible with numpy 2.4 yet
-          python -m pip install cmake openfermionpyscf numpy<2.4
+          python -m pip install cmake openfermionpyscf "numpy<2.4"
 
       - name: Install required lightning_gpu only packages
         if: matrix.pl_backend == 'lightning_gpu'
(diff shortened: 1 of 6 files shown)


tests/lightning_base/test_device.py
```diff
@@ -725,7 +725,7 @@ def test_preprocess_incorrect_device_config(self):
         device_name == "lightning.tensor",
         reason="lightning.tensor device doesn't have support for program capture.",
     )
-    @pytest.mark.parametrize("postselect_mode", ["hw-like", "fill-shots"])
```
tomlqc, line 728: why removing `fill-shots`?

maliasadi: I don't know how this was checked and not failed before, but we don't have support for this case; here's the [SC ticket](https://app.shortcut.com/xanaduai/story/82462/lightning-devices-support-postselection-with-defer-measurements) and we also test the error message [here](https://github.com/PennyLaneAI/pennylane-lightning/blob/71069eb69701ca2622892501aaccadbd04a3b291/tests/test_native_mcm.py#L185).

tomlqc: ok, I see `"Using postselect_mode='fill-shots' is not supported."`

PR #1321, opened 2026-01-06
Fix dynamic wires with samples

rniczh:
**Context:**

```python
from catalyst import *
import pennylane as qml

qml.capture.enable()

@qjit
@qml.qnode(qml.device("lightning.qubit", wires=3, shots=10))
def circuit():
    with qml.allocate(2) as qs:
        qml.X(qs[1])

    return qml.sample(wires=[0, 1])

print(circuit())
```

It get unexpected result due to the aux wires involved.

```
[[0 0]
 [0 1]
 [0 0]
 [1 0]
 [0 0]
 [0 0]
 [0 1]
 [0 0]
 [1 0]
 [0 0]]
```

1. State Vector Reduction
~Only take the device wires for generating samples~
Compact state vector before measurement.

For state vector normalization:

$\|\psi\| = \sqrt{\sum_i |\psi_i|^2}$
$normalized_i = \psi_i / \|\psi\|$

2. Raise an error if in mixed after qubits released
When dynamically releasing qubits in a quantum simulator, we need to ensure that the released qubits are disentangled from the remaining active qubits. If the qubits are entangled, releasing them would result in an incorrect reduced state vector.

For a state to be safely released, it must be separable:

```
|phi> = |released> x |active>
```

When we perform a partial trace over the active qubits, we get the reduced density matrix of the released qubits. Then we can get the Purity from the `ρ_released`. If purity equal is smaller than 1, this indicates it's in the mixed state.
```
ρ_released = Tr_active(|phi><phi|) = |released><released|
purity = Tr(ρ_released^2)
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

https://github.com/PennyLaneAI/catalyst/issues/2339
[[sc-107206]]
[[sc-98156]]


pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -478,10 +482,8 @@ void LightningSimulator::PartialCounts(DataView<double, 1> &eigvals,
     RT_FAIL_IF((eigvals.size() != numElements || counts.size() != numElements),
                "Invalid size for the pre-allocated partial-counts");
 
-    // get device wires
-    auto &&dev_wires = getDeviceWires(wires);
-
-    auto li_samples = this->GenerateSamples(device_shots);
+    auto dev_wires = getDeviceWires(wires);
+    auto li_samples = this->GenerateSamples(device_shots, dev_wires);
```
dime10: Huh, so was the existing strategy to generate samples on all wires first and then filter the produced samples to the provided wires (since GenerateSamples didn't know about the wires)?

If so, do you know why the filtering doesn't work once dynamic allocations are present?

pennylane_lightning/core/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp
```diff
@@ -552,7 +552,17 @@ class Measurements final
...
+        std::iota(wires.begin(), wires.end(), 0);
+        return generate_samples_metropolis(wires, kernelname, num_burnin,
+                                           num_samples);
+    }
+
+    std::vector<std::size_t> generate_samples_metropolis(
+        const std::vector<std::size_t> &wires, const std::string &kernelname,
+        std::size_t num_burnin, std::size_t num_samples) {
         std::uniform_real_distribution<PrecisionT> distrib(0.0, 1.0);
+        std::size_t num_qubits = wires.size();
```
dime10: Do we know that this is sufficient to identify which wires should be selected in a more complex dynamic (de)allocation scheme, or does it rely on the assumption that the selected wires are always the bottom n wires?

dime10 (commented):
Thank you @rniczh !

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: For the case with aux quibts, `samples` has only 20 numbers, but now it writes not only 20 times and out-of-bound. That's the reason why it caused the `samples` looks weird.
```
*(samplesIter++) =
                static_cast<double>(li_samples[shot * numQubits + wire_idx]);
```

log:
```
li_samples: 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 1
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 0 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 1 li_samples[shot * numQubits + wire_idx]: 0
samplesIter: 0
wire_idx: 2 li_samples[shot * numQubits + wire_idx]: 1
samplesIter: 1
samples: 0 0 0 1 0 0 0 0 1 1 0 0 0 0 1 0 0 0 0 1 
[[0 0]
 [0 1]
 [0 0]
 [0 0]
 [1 1]
 [0 0]
 [0 0]
 [1 0]
 [0 0]
 [0 1]]
 ```

Comment (dime10) [reply pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp]: Thanks, that makes sense!

File: pennylane_lightning/core/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp (hunk 2 above)
Comment (dime10) [reply pennylane_lightning/core/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp]: Mentioned this offline but I think we'd be better off providing a universal solution that we never got around to implementing (see https://github.com/PennyLaneAI/pennylane-lightning/pull/1254#discussion_r2344381991).

Comment (rniczh) [reply pennylane_lightning/core/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp]: I updated the PR, basically removed the original solution that only pass the device_wires to `GenerateSample`. Now, it will compact the state vector before the measurement process to ensure the erased qubits are erased from state vector as well.
~https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/ca0ff43c6f187998111af61d2ff223f48434264b~ Edit: https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/77fbbe908646de00abbd9031005543fea77ffbf5

Comment (dime10) [reply pennylane_lightning/core/simulators/lightning_qubit/measurements/MeasurementsLQubit.hpp]: Wow that was fast! 🤩 
@maliasadi are you able to look this over?

**Review by maliasadi — COMMENTED** [review]
Thanks @rniczh! A few things:

- If we need this to go out with the release, you will need to rebase it against #1317 and sync ASAP with @jzaia18 
- Update Lighnting-Kokkos and Lightning-GPU catalyst impl similar to Lightning-Qubit
- Please add a changelog entry for this fix

File: pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp [hunk 3]
```diff
@@ -417,11 +498,11 @@ void LightningSimulator::PartialSample(DataView<double, 2> &samples,
...
                "Invalid size for the pre-allocated partial-samples");
 
-    // get device wires
-    auto &&dev_wires = getDeviceWires(wires);
-
     auto li_samples = this->GenerateSamples(device_shots);
 
+    // Get device wires
+    auto &&dev_wires = getDeviceWires(wires);
+
```
jzaia18, line 684: Any reason for switching the order of these steps?

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -163,5 +163,28 @@ class QubitManager final {
...
+        std::unordered_set<DeviceQubitID> new_free_qubits;
+        for (auto old_device_id : this->free_device_qubits) {
+            auto it = old_to_new_mapping.find(old_device_id);
+            if (it != old_to_new_mapping.end()) {
+                new_free_qubits.insert(it->second);
+            } else {
+                new_free_qubits.insert(old_device_id);
+            }
+        }
+        this->free_device_qubits = std::move(new_free_qubits);
```
jzaia18: How often do we expect this section to be run? Will it be expensive to create and free the `free_device_qubits` set each time this is run as we currently do?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -117,10 +136,88 @@ void LightningSimulator::ReleaseQubit(QubitIdType q) {
...
+            size_t old_bit_pos = old_num_qubits - 1 - old_wire;
+            size_t new_bit_pos = num_qubits_after - 1 - i;
+
+            if ((old_idx >> old_bit_pos) & 1) {
+                new_idx |= (1UL << new_bit_pos);
+            }
+        }
+
+        new_data[new_idx] += old_data[old_idx];
+    }
```
jzaia18, line 227: This outer loop goes over the entire statevector, we probably want to use parallelism to prevent this from being a very expensive operation. Unsure if it's worth doing here or fixing in a follow-up PR. @maliasadi

jzaia18 (commented):
Nice work Hongsheng! Just a few questions and comments. ~~Is there any reason we're implementing this for `lightning.qubit` but not the other devices? Does this issue exist for the others as well?~~ nvm, just saw Ali already addressed this

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: That's because the `getMeasurements()` is within the `GenerateSamples`. The `getMeasurements()` will compact the state vector to get the measurement object, that means if the `getDeviceWires` happens before `GenerateSamples` may not get the correct device wires.

```
std::vector<size_t> LightningSimulator::GenerateSamples(size_t shots) {
    auto m = getMeasurements();
    // ... rest of code ...
```

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
rniczh: Few, not even happened in the test so far. That's what I currently need to add in the next revision.
Edit: I added the test https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/54a87e135b90caae52ccdca83131806407ee4be5

So it's only happened if you break the continuity of the device id. For example, if you release the q1 in |q0 q1 q2>, it should map the device id 2 to 1 then before sampling

rniczh: And it's only `O(n)` based on the size of device_id

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
maliasadi: Good point! Let's fix this in the followup PR. There are some parallel discussions to introduce OMP in the catalyst runtime methods/classes.

.github/CHANGELOG.md
```diff
@@ -34,11 +34,14 @@
 
 <h3>Bug fixes 🐛</h3>
 
+- Fix sampling with dynamically allocated and released qubits in Catalyst. The state vector is now compacted before measurements to correctly handle released qubits, ensuring `qml.sample()` and other measurement operations return correct results when using `qml.allocate()` and `qml.release()`.
```
maliasadi: @AntonNI8 will you happy with this entrypoint? :) 
```suggestion
- Fixed sampling with dynamically allocated and released qubits in Catalyst. The state vector is now compacted before measurements to correctly handle released qubits, ensuring `qml.sample()` and other measurement operations return correct results when using `qml.allocate()` and `qml.release()`.
```

.github/CHANGELOG.md
```diff
@@ -34,11 +34,14 @@
...
+  [(#1321)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1321)
+
 - Corrected an issue in tests where a PennyLane operator was used within a QNode to compute a
   matrix, which would lead to wrongful queuing as of PennyLane
   pull request [#8131](https://github.com/PennyLaneAI/pennylane/pull/8131).
   [(#1292)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1292)
-
+｀｀
```
maliasadi:
```suggestion
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -18,6 +18,22 @@
 
 namespace Catalyst::Runtime::Simulator {
 
+namespace {
+inline void normalizeStateVector(std::vector<std::complex<double>> &data) {
```
maliasadi: Where are the unit tests for this method? In Catalyst? :thinking:

maliasadi: I'd add this function to a helper `Utils.hpp` file, maybe [here](https://github.com/PennyLaneAI/pennylane-lightning/tree/master/pennylane_lightning/core/catalyst), so that we could use it for other devices.

We can do this in the follow-up PR though.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -101,19 +122,95 @@ void LightningGPUSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
 }
 
+auto LightningGPUSimulator::getMeasurements()
+    -> Pennylane::LightningGPU::Measures::Measurements<StateVectorT> {
+    CompactStateVector();
+    return Pennylane::LightningGPU::Measures::Measurements<StateVectorT>{
+        *(this->device_sv)};
+}
+
+void LightningGPUSimulator::CompactStateVector() {
```
maliasadi: Where are these tested?  I couldn't find any tests for these methods for all devices.

pyproject.toml
```diff
@@ -8,22 +8,22 @@ description = "PennyLane-Lightning plugin"
...
-    "Operating System :: POSIX :: Linux", 
-    "Programming Language :: Python", 
-    "Programming Language :: Python :: 3", 
-    "Programming Language :: Python :: 3 :: Only", 
-    "Programming Language :: Python :: 3.11", 
-    "Programming Language :: Python :: 3.12", 
-    "Programming Language :: Python :: 3.13", 
-    "Programming Language :: Python :: 3.14", 
+classifiers = [
+    "Development Status :: 4 - Beta",
```
maliasadi: :open_mouth:

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -163,5 +163,23 @@ class QubitManager final {
...
+            }
+        }
+
+        // Update free device qubits as well
+        std::unordered_set<DeviceQubitID> new_free_qubits;
+        for (auto old_device_id : this->free_device_qubits) {
+            new_free_qubits.insert(old_device_id);
+        }
+        this->free_device_qubits = std::move(new_free_qubits);
+    }
```
jzaia18: What is this step doing? It just seems to overwrite free_device_qubits with a copy of itself

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -82,9 +98,14 @@ auto LightningGPUSimulator::AllocateQubits(std::size_t num_qubits)
...
-    // We do not deallocate physical memory in the statevector for this
-    // operation, instead we just mark the qubits as released.
+    RT_FAIL_IF(!this->qubit_manager.isValidQubitId(q),
+               "Invalid qubit to release");
+
+    // Mark the qubit as released in the qubit manager
     this->qubit_manager.Release(q);
+
+    // Mark that compaction is needed
+    this->needs_compaction = true;
```
maliasadi: need to update the header file too.

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
```diff
@@ -163,5 +163,23 @@ class QubitManager final {
...
     }
+
+    void RemapDeviceIds(const std::unordered_map<DeviceQubitID, DeviceQubitID>
+                            &old_to_new_mapping) {
+        // Update each program_id's device_id according to the mapping
+        for (auto &[program_id, device_id] : this->qubit_id_map) {
+            auto it = old_to_new_mapping.find(device_id);
+            if (it != old_to_new_mapping.end()) {
+                device_id = it->second;
+            }
```
maliasadi:
```suggestion
            if (auto it = old_to_new_mapping.find(device_id); it != old_to_new_mapping.end()) {
                    device_id = it->second;
            }
```

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.hpp
```diff
@@ -98,6 +102,13 @@ class LightningSimulator final : public Catalyst::Runtime::QuantumDevice {
     auto GenerateSamplesMetropolis(size_t shots) -> std::vector<size_t>;
     auto GenerateSamples(size_t shots) -> std::vector<size_t>;
 
+    // Compact state vector by removing released qubits
+    void CompactStateVector();
```
maliasadi: minor: this is the style we've been following for a while:

```suggestion
    void compactStateVector();
```

maliasadi: Should we call this `reducedStateVector` instead of `compact`?

jzaia18:
Could you also bump the rc version to 3 within `pennylane_lightning/core/_version.py`?

pyproject.toml
rniczh: oh i formatted it unconsciously. let me revert it

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.hpp
rniczh: Sure! I updated the name here: https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/a5e2132841ce5f501f82425e68627d1b2a5193f7

pennylane_lightning/core/catalyst/LightningQubitManager.hpp
rniczh: Thanks for pointing out! It's something I forgot to clean when some lines of code are removed from here. I updated in this commit https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/81c4859d481e71d374df24cbfb008af4c08e08d9

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: What do you mean for updating the header file? This bool value is initialized in this class:

```cpp
class LightningGPUSimulator final : public Catalyst::Runtime::QuantumDevice {
     ...

    // Flag to indicate if state vector needs compaction
    bool needs_compaction{false};
    
    ...
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: We didn't have the tests solely for this helper function, I added tests for this:
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/60511a83b684adfaec7d18b231b936157b4cf3a6

rniczh: Done! https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/60511a83b684adfaec7d18b231b936157b4cf3a6

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: The function is triggered by `getMeasurements()`, ensuring that any test executing the measurement process also updates the state vector. Currently, only one test case is added (the one that reported from the issue https://github.com/PennyLaneAI/catalyst/issues/2339). The specific test can be found here:

https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/changes#diff-6f0e99f87b0a9753689936392636b52eb19b81aac144b24d12de297f3c617447R335

pyproject.toml
rniczh: Done! https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/0ec104fdcf5f54c1e5d10cefc78e38fd861bab30

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -148,12 +131,12 @@ void LightningSimulator::ReleaseQubit(QubitIdType q) {
...
 auto LightningSimulator::getMeasurements()
     -> Pennylane::LightningQubit::Measures::Measurements<StateVectorT> {
-    CompactStateVector();
+    reducedStateVector();
     return Pennylane::LightningQubit::Measures::Measurements<StateVectorT>{
         *(this->device_sv)};
 }
 
-void LightningSimulator::CompactStateVector() {
+void LightningSimulator::reducedStateVector() {
```
jzaia18: Not to be too nitpicky, but since this modifies the device and doesn't return a new statevector, would a better name be either `reduceStateVector` or `toReducedStateVector`? Similarly should we also rename the `needs_compaction` variable to `needs_reduction`?

CC @maliasadi

jzaia18 (commented):
Looking good, just 1 last comment

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
jzaia18: (Also we should update Lightning-Kokkos and Lightning-GPU to match once we do settle on the ideal names)

rniczh: I think `reduceStateVector` is good? How do you think @maliasadi

rniczh: I updated the function to `reduceStateVector` for all device https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/6c8cdb2e7c7ccc1d38513a0ba7008b0c36701359

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
maliasadi: ah, didn't see it. looks good!

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
maliasadi: Will this be added to Catalyst for the release? 

per our offline discussions, what if these aux wires are entangled, we should test and add some handling mechanism for these cases in both Lightning & Catalyst?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -117,10 +119,88 @@ void LightningSimulator::ReleaseQubit(QubitIdType q) {
     // maintained,
     //       in particular: measurements must not compute results for released
     //       qubits, only for active/allocated qubits
```
dime10: We can remove the TODO now :)

pennylane_lightning/core/utils/Util.hpp
```diff
@@ -690,4 +690,29 @@ inline auto PL_reinterpret_cast(SrcType *src_ptr) -> DestType * {
...
+ * @brief Normalize a complex state vector to unit norm.
+ *
+ * This function normalizes a state vector in-place
+ *
+ * @tparam ComplexT Complex number type (e.g., std::complex<double>).
+ * @tparam Alloc Allocator type.
+ * @param data State vector data to normalize.
+ */
+template <class ComplexT, class Alloc = std::allocator<ComplexT>>
+inline void normalizeStateVector(std::vector<ComplexT, Alloc> &data) {
```
dime10: Just wanted to mention I believe most devices already have a normalize method, since it's needed for the `measure` implementation:
https://github.com/PennyLaneAI/pennylane-lightning/blob/4739a0f6f42a39ae8c32da08ce00be942a604144/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L908

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -117,10 +119,88 @@ void LightningSimulator::ReleaseQubit(QubitIdType q) {
...
+    RT_FAIL_IF(!this->qubit_manager.isValidQubitId(q),
+               "Invalid qubit to release");
+
+    // Mark the qubit as released in the qubit manager
     this->qubit_manager.Release(q);
 
-    // TODO: We don't have to check whether the released qubit is pure, but it
-    // is something we can add easily to catch program errors.
```
dime10, line 122: We should get consensus on the desired behaviour. I don't think this needs to be a big thing, just picking one of the options and checking with product what they think. As far as I know the options are:
- (1) it is enforced that qubits must be pure (disentangled) before release
  - implementation: (simulator-only) raise an error if the user releases an entangled qubit
  - benefit: nice to catch (likely) programming errors for the user, no undefined or non-deterministic behaviour
  - issues: 
    - if an entire subset of qubits is released at once, it should be okay if they are entangled with each other, but here we do the release one-by-one so that complicates the detection
    - takes a bit of work each qubit release
- (2) it is well-defined what happens when qubits are not pure:
  - implementation: reset affected qubits projecting the remainder of state onto of the two subspaces
    - this could be random (like on hw) or a forced projection onto 0 for example (simulator-only)
    - since we want the behaviour to be well defined and observable, I think this should happen immediately upon release (e.g. a snapshot would reflect this right away, but I'm not sure that even with option 3 the partial statevector would ever be observable?)
  - benefit: can be deterministic still, is observable
  - issues: takes a bit of work each qubit release
- (3) it is undefined behaviour what happens when releasing non-pure qubits:
  - implementation: undefined means we can do whatever is convenient, but at the very least we have to project (and compact) the state before measurement processes and before handing out fresh qubits that reuse previously deallocated space in the statevector
  - benefit: could be faster to delay than doing something immediately like in the other options
  - issues: least clean?

Although, even if the projection is done lazily we could still raise an error at that point, it might just be less precise than if we do it immediately. So maybe there is only two choices (error or silent projection), and the question of whether to do it eagerly or lazily is orthogonal to that choice. Another variant could check for an error eagerly, but compact lazily. Or another one could be have a safe release and an unsafe release, which can be toggled by the user.

dime10: My gut decision would probably be option 1 for safety (and compact later, and/or safety toggle).

dime10 (commented):
Nice work! Let's get this across the finish line to make sure the dynamic allocation feature actually works :)

Only one question to resolve really:

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
lazypanda10117, line 578: We probably want to call `reduceStateVector()` here as well.

rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c99af2731af9171a15dec107c4a43ea33f24dcc6

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Thank you for such a detailed explanation! In the end, I tried the lazy approach with calculating the purity to check if the final reduced state vector is in a mixed state, and report an error if it is.

So if purity < 1 then it will report:
```
Cannot release qubits: released qubits are entangled with remaining qubits (purity = ...)
```

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c99af2731af9171a15dec107c4a43ea33f24dcc6

pennylane_lightning/core/utils/Util.hpp
rniczh: Done. https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c99af2731af9171a15dec107c4a43ea33f24dcc6

I changed to use the normalize() directly

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -59,6 +60,26 @@ auto LightningSimulator::AllocateQubit() -> QubitIdType {
         // The collapse is performed by a measurement followed by an X gate if
         // measured 1.
         new_program_idx = this->qubit_manager.Allocate(device_idx);
+
+        // If state vector size doesn't match active qubits, expand it
```
dime10: What is this block doing here, doesn't the branch above already handle expanding the statevector when there is no free qubits?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +121,189 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+        active_device_ids.insert(this->qubit_manager.getDeviceId(qid));
+    }
+
+    // Collect released qubit device IDs
+    std::vector<size_t> released_device_ids;
+    for (size_t i = 0; i < num_qubits; i++) {
+        if (!active_device_ids.contains(i)) {
+            released_device_ids.push_back(i);
+        }
+    }
```
dime10: This should work but doesn't the qubit manager already track the freed IDs? (which is how the Allocate function reuses existing space)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +121,189 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+
+    // Check if purity is close to 1
+    constexpr double epsilon = 1e-6;
+    if (std::abs(purity - 1.0) > epsilon) {
+        const std::string error_msg =
+            "Cannot release qubits: released qubits are "
+            "entangled with remaining qubits (purity = " +
+            std::to_string(purity) + ")";
+        RT_FAIL(error_msg.c_str());
+    }
```
dime10: Nice work on the purity check! Just for reference I pulled up this prior implementation we had for lightning: https://github.com/PennyLaneAI/catalyst/blob/d80915551ba3672cbb73fe186513d9897b04b31c/runtime/lib/backend/lightning/lightning_dynamic/StateVectorLQubitDynamic.hpp#L246

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
dime10: Nice work! 

One thing I just realized is that for qubit reuse we automatically reset the qubit we are reusing. What this means is that with the lazy purity check, an entangled qubit that is being released could either raise an error (if we are compacting it before a measurement process), or project the state randomly without error (if we are reusing it), which is inconsistent.
If we want to do away with the inconsistency we would either
- get rid of the purity check and always project
- always do the purity check eagerly
- add the purity check before the reuse

@rniczh What are your thoughts here?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: The issue is there are free qubits for reusing, however the state vector has been reduced.

So the steps are:
1. When a qubit is released, `reduceStateVector()` is called lazily (e.g. during `State()` or `getMeasurements()`). This shrinks the state vector if needed.
2. Run `State()` -> It cause the state to be reduced
3. There are free qubits -> go to this branch to reuse the space
3. **Here** Allocate updates the qubit_manager mapping (program_id -> device_id) and doens't modify the state vector size.
```python
    auto Allocate(DeviceQubitID device_id) -> ProgramQubitID {
        this->qubit_id_map[this->next_program_id] = device_id;
        return this->next_program_id++;
    }
```
3. The state vector already be shrinked, need to be expanded to the new size for reusing

dime10: Then we need to clear the free qubits from the manager when compacting, much simpler :)

rniczh: Thanks! that make sense, I introduced a `ClearFreeQubits` function to do the `this->free_device_qubits.clear();`. So it will call this function before the end of the `reduceStateVector()`
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/1f02268cc34bde527165ede2f02b177a53f257eb

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Yes, I added a `getFreeDeviceQubits()` function for doing this job
```
auto getFreeDeviceQubits() const
        -> const std::unordered_set<DeviceQubitID> & {
        return this->free_device_qubits;
    }
```
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/1f02268cc34bde527165ede2f02b177a53f257eb

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Oh, should reply to this thread. 😅 https://github.com/PennyLaneAI/pennylane-lightning/pull/1321#issuecomment-3813536011

dime10: Yup looks good, thanks :D

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -49,10 +52,17 @@ auto LightningGPUSimulator::AllocateQubit() -> QubitIdType {
     } else {
         device_idx = candidate.value();
 
+        // Check if the qubit to be reused is disentangled before reusing
+        if (!checkSingleQubitDisentangled(device_idx)) {
+            RT_FAIL("Cannot reuse qubit: qubit is entangled with remaining "
+                    "qubits. Release qubits must be disentangled.");
+        }
```
jzaia18, line 59: This failure seems like it should happen at deallocation time rather than at the moment the user tries to allocate a new qubit. Is there any technical reason the failure has to happen here?

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -101,19 +116,234 @@ void LightningGPUSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+    return std::abs(purity.real() - 1.0) < epsilon &&
+           std::abs(purity.imag()) < epsilon;
+}
+
+void LightningGPUSimulator::checkReleasedQubitsDisentangled() {
+    const size_t num_qubits = this->device_sv->getNumQubits();
+    const size_t num_active = this->qubit_manager.getNumQubits();
+
+    // Skip check if no qubits to release or all qubits are active
+    if (num_qubits == num_active || num_active == 0) {
```
jzaia18:
```suggestion
    // Skip check if no qubits to release or no qubits are active
    if (num_qubits == num_active || num_active == 0) {
```
I think the condition is correct, but the comment seems off

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -101,19 +116,234 @@ void LightningGPUSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+            bool is_released = free_device_qubits.contains(i);
+            if (is_released) {
+                const size_t bit_pos = num_qubits - 1 - i;
+                const size_t new_bit_pos = num_released - 1 - released_bit;
+                if ((released_idx >> new_bit_pos) & 1) {
+                    idx |= (1UL << bit_pos);
+                }
+                released_bit++;
+            } else {
+                const size_t bit_pos = num_qubits - 1 - i;
```
jzaia18:
```suggestion
            bool is_released = free_device_qubits.contains(i);
            const size_t bit_pos = num_qubits - 1 - i;
            if (is_released) {
                const size_t new_bit_pos = num_released - 1 - released_bit;
                if ((released_idx >> new_bit_pos) & 1) {
                    idx |= (1UL << bit_pos);
                }
                released_bit++;
            } else {
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -101,19 +116,234 @@ void LightningGPUSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+    // Check if released qubits are disentangled from active qubits using purity
+    // If separable, then |phi> = |released> x |active>
+    //     then ρ_released = Tr_active(|phi><phi|) = |released><released|
+    //     and Tr(ρ_released^2) = 1 (pure state)
+    const auto &state_data = this->device_sv->getDataVector();
+    const size_t num_released = free_device_qubits.size();
+    const size_t released_space_size = 1UL << num_released;
+    const size_t active_space_size = 1UL << num_active;
+
+    // function for mapping indices
```
jzaia18: It might be helpful to expand on this comment a bit, what is this mapping needed for?

requirements-dev.txt
```diff
@@ -5,7 +5,7 @@ cmake
 scipy-openblas32>=0.3.26
 custatevec-cu12; sys_platform == "linux"
 cutensornet-cu12; sys_platform == "linux"
-git+https://github.com/PennyLaneAI/pennylane.git@master
+git+https://github.com/PennyLaneAI/pennylane.git@v0.44.0-rc0
```
jzaia18: Probably safe to set this back to normal (might be handled by merging master into this branch). Same for `requirements-tests.txt`

pennylane_lightning/core/utils/Util.hpp
```diff
@@ -690,4 +690,42 @@ inline auto PL_reinterpret_cast(SrcType *src_ptr) -> DestType * {
...
 
+/**
+ * @brief Compute purity from released parts
+ *
+ * Tr(ρ^2) = Σ_i,j |Sum_a(released_a[i] x conj(released_a[j]))|^2
+ *
+ * @tparam ComplexT Complex number type
+ * @param released_parts Vector of released parts
+ *                       {active -> [released] for all active \in active states}
+ * @return Purity value
```
jzaia18: Is the term "purity" well defined? It might be helpful to have a short summary of what this actually represents

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+
+    for (auto qid : all_qubits) {
+        size_t device_wire = this->qubit_manager.getDeviceId(qid);
+        wire_id_pairs.push_back({device_wire, qid});
+    }
+
+    // Sort by device wire index
+    std::sort(wire_id_pairs.begin(), wire_id_pairs.end());
+
+    // Extract reduced state vector
```
jzaia18: This comment doesn't seem accurate to the following line

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+    auto &old_data = this->device_sv->getDataVector();
+    size_t num_qubits_after = wire_id_pairs.size();
+    size_t new_size = 1UL << num_qubits_after;
+
+    std::vector<std::complex<double>> new_data(new_size);
+
+    size_t old_num_qubits = this->device_sv->getNumQubits();
+
+    // Find a reference old_idx with non-zero amplitude
```
jzaia18: Why is a non-zero amplitude index needed? It might be helpful to add a bit more context

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+
+            if ((new_idx >> new_bit_pos) & 1) {
+                old_idx |= (1UL << old_bit_pos);
+            } else {
+                old_idx &= ~(1UL << old_bit_pos);
+            }
+        }
+
+        new_data[new_idx] = old_data[old_idx];
+    }
```
jzaia18, line 227: Some additional comments explaining what this does would be helpful. e.g. "Translate old amplitude indices within the statevector to new indices by following change in qubit ordering"

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+            }
+            rho[2 * i + j] = sum;
+        }
+    }
+
+    std::complex<double> purity =
+        (rho[0] * rho[0]) + (std::complex<double>{2.0, 0.0} * rho[1] * rho[2]) +
+        (rho[3] * rho[3]);
+
+    constexpr double epsilon = 1e-6;
```
jzaia18: Would there be value in making this a default argument for the function?

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+    return std::abs(purity.real() - 1.0) < epsilon &&
+           std::abs(purity.imag()) < epsilon;
```
jzaia18: `purity.imag()` should always be (effectively) 0, shouldn't it? It might be better to have a `PL_ABORT_IF(std::abs(purity.imag()) < epsilon)` rather than including this in the return value

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -100,26 +110,241 @@ void LightningSimulator::ReleaseQubits(const std::vector<QubitIdType> &ids) {
...
+}
+
+void LightningSimulator::checkReleasedQubitsDisentangled() {
+    const size_t num_qubits = this->device_sv->getNumQubits();
+    const size_t num_active = this->qubit_manager.getNumQubits();
+
+    // Skip check if no qubits to release or all qubits are active
```
jzaia18:
```suggestion
    // Skip check if no qubits to release or no qubits are active
```

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -104,19 +119,235 @@ void LightningKokkosSimulator::ReleaseQubits(
...
 }
 
+auto LightningKokkosSimulator::getMeasurements()
+    -> Pennylane::LightningKokkos::Measures::Measurements<StateVectorT> {
+    reduceStateVector();
+    return Pennylane::LightningKokkos::Measures::Measurements<StateVectorT>{
+        *(this->device_sv)};
+}
+
+void LightningKokkosSimulator::reduceStateVector() {
```
jzaia18, line 143: This is extremely similar to the one within LightningQubitSimulator (just a couple type differences). Is there some way we could put this within the base class instead of having multiple impls? CC @maliasadi

jzaia18: (same for the other 2 methods below)

jzaia18 (commented):
Awesome work! Just a few comments. There are a few changes here that if you do in 1 device you should make sure to replicate for the others as well

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: Since we do the lazy purity check, the purity check won't be triggered eagerly. It happens only when you are going to use it, like `reuse` or do the measurment. Details can be found in these comments:
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321#discussion_r2733944234
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321#issuecomment-3813536011

pennylane_lightning/core/utils/Util.hpp
rniczh: Yes, purity is just a term in quantum theory https://en.wikipedia.org/wiki/Purity_(quantum_mechanics)
Let me write a short summary here.

jzaia18: In that case I would just link that wiki page in the comment

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c77ab80703d4c28bee874c48abefca4c865eca28

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c77ab80703d4c28bee874c48abefca4c865eca28

pennylane_lightning/core/utils/Util.hpp
rniczh: I wrote a short summary of the purity along with this commit
https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c77ab80703d4c28bee874c48abefca4c865eca28

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Yes, it should be 0 https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/7aa675e9ec90896479c1b1a7a0a8dc8ab13cd630

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Sure, just added it as a param https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/13a909ead5c8c611fd7847cd7fe9c5d2fbb7292e

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: I updated the comment: https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/c77ab80703d4c28bee874c48abefca4c865eca28

```
// Create a new state vector with reduced size
```

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/f5838a7dff9a6de5ff0092e652fa3d4080008814

requirements-dev.txt
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/9d93e936bc2403ce7834938a50f4df37179b99f8

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
rniczh: The implementations differ in:
- Complex type (std::complex vs Kokkos::complex)
- Data access patterns (getDataVector() returns reference for CPU, copy for Kokkos/GPU)
- Absolute value function (std::abs vs Kokkos::abs)
How do you think?

pennylane_lightning/core/utils/Util.hpp
rniczh: Added a link to wiki as well https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/b9095c4b798551551bb4474ffc192cb3f76a6140

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
dime10: I laid out some of the options we have around this above in [this thread](https://github.com/PennyLaneAI/pennylane-lightning/pull/1321#discussion_r2706093803), and Hongsheng decided to go with the lazy approach around error checking. That being said, if there are better arguments for a different approach we are happy to discuss it :)

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
jzaia18: I think it might still be worth having a sanity-check assert to make sure the imaginary part is 0 (raising an error if it isn't)

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
jzaia18: Okay I think that works. Leaving this thread open in case @maliasadi wants to weigh in

jzaia18 (approved):
Awesome work, no more comments on my end. Just make sure you get 1 more approval from perf team before merging CC @maliasadi

maliasadi (approved):
Thanks @rniczh! 👍 

My remaining concerns can be addressed in follow-up PRs:


1. `reduceStateVector` currently loops over the entire state-vector and it's not parallelized. We can try to update the logic with OMP enabled.
2.  Same logic is implemented in `LightningSimulator.cpp`, `LightningGPUSimulator.cpp`, `LightningKokkosSimulator.cpp`! We should get back to the device implementations and remove code duplication across lightning QuantumDevice impl.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
maliasadi: Looks good, as long as this raises clear runtime errors when normalization/reduction fails. Lightning isn't a mixed-state simulator!

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
rniczh: Done https://github.com/PennyLaneAI/pennylane-lightning/pull/1321/commits/065f86693ec5cc90f0b6ab056ba5e657f200abb0

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
rniczh: > what if these aux wires are entangled

So now, the purity check has been added to lightning to ensure there is no mixed state happen in lightning. And once the PR get merged, we would also add the integration tests in cataylst just like what we did here as well.

maliasadi: Sounds good! thanks for the clarification

PR #1327, opened 2026-01-13
Add PauliMeasure simulation support to Catalyst

maliasadi:
**Context:**
Enable execution of QJIT-compiled programs with Pauli Measurement on Lightning state-vector simulators (LQ, LK, LGPU).

Depends on https://github.com/PennyLaneAI/catalyst/pull/2348

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-107353]


pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -491,6 +492,77 @@ auto LightningGPUSimulator::Measure(QubitIdType wire,
...
+    }
+
+    // Measure in the computational basis
+    bool mres = true;
+    for (const auto &wire : wires) {
+        Result res = this->Measure(wire);
+        mres = mres && (*res);
+    }
+
+    for (std::size_t i = 1; i < pauli_word.size(); i++) {
```
lazypanda10117, line 547: If I am not mistaken, I think this needs to be a CNOT ladder of the form `wires[i], wires[i-1]`, similar to here https://github.com/PennyLaneAI/catalyst/pull/2057/files#diff-a521767211f26c8b4288939db6f2fa69545ac53f0e585921773d8322171ccde1R113

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -491,6 +492,77 @@ auto LightningGPUSimulator::Measure(QubitIdType wire,
...
+            break;
+        case 'Z':
+            // No basis change needed for Z
+            break;
+        default:
+            RT_FAIL("Invalid character in Pauli word");
+        }
+    }
+
+    for (std::size_t i = 1; i < pauli_word.size(); i++) {
```
lazypanda10117, line 539: Same comment about the CNOT ladder and everywhere else that uses this structure.

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
```diff
@@ -519,6 +520,77 @@ auto LightningKokkosSimulator::Measure(QubitIdType wire,
                 : const_cast<Result>(&GLOBAL_RESULT_FALSE_CONST);
 }
 
+auto LightningKokkosSimulator::PauliMeasure(
```
lazypanda10117, line 537: Not sure if this is possible/recommended in Lightning, but I am curious if there is a way to abstract this logic out so it can be used by all devices rather than duplicating them.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
maliasadi: @lazypanda10117 @sengthai  here's the referenced python impl 

``` python
def pauli_measurement(pauli, wires):

    for i, wire in enumerate(wires):
        if pauli[i] == "X":
            qml.Hadamard(wire)
        elif pauli[i] == "Y":
            qml.adjoint(qml.S(wire))
            qml.Hadamard(wire)

    for wire in wires[1:]:
        qml.CNOT(wires = [wire, wires[0]])

    meas = qml.measure(wires[0])
    # print(meas)

    for wire in wires[1:]:
        qml.CNOT(wires = [wire,wires[0]])

    for i, wire in enumerate(wires):
        if pauli[i] == "X":
            qml.Hadamard(wire)
        elif pauli[i] == "Y":
            qml.Hadamard(wire)
            qml.S(wire)

    return meas
```

Feel free to update the impl as fits the end-to-end pipeline.

pennylane_lightning/core/simulators/lightning_kokkos/catalyst/LightningKokkosSimulator.cpp
maliasadi: That was the original plan! However, we cannot implement this in the base meas class because this feature is not required without QJIT. Additionally, the meas class in Lightning assumes the underlying state vector remains constant throughout all meas processes. Adding this to the Lightning core API would require legacy changes that are currently low priority. Also, most of these Lightning `QuantumDevice` impl can be abstracted. This is a code cleanup task for a follow-up PR, rather than a fix for the current one

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
lazypanda10117: Per discussion with Sengthai, it seems like this one is the correct implementation, so we will have to revisit the `ppr-to-mbqc` pass and fix that later.

sengthai: Based on this paper (https://arxiv.org/pdf/2305.04807), implementing with the CNOT ladder is correct. However, it is good to unify them across all implementation.

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
```diff
@@ -491,6 +492,77 @@ auto LightningGPUSimulator::Measure(QubitIdType wire,
...
+        this->NamedOperation("CNOT", {}, {wires[i], wires[0]});
+    }
+
+    // Measure in the computational basis
+    bool mres = true;
+    for (const auto &wire : wires) {
+        Result res = this->Measure(wire);
+        mres = mres && (*res);
+    }
```
sengthai: Based on [ref](https://github.com/PennyLaneAI/pennylane-lightning/pull/1327#discussion_r2691787745) here, it seems measure all qubits; however, it should only measure `wire[0]`.

pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp
```diff
@@ -524,6 +524,90 @@ auto LightningSimulator::Measure(QubitIdType wire,
...
+    for (std::size_t i = 1; i < pauli_word.size(); i++) {
+        this->NamedOperation("CNOT", {}, {wires[i], wires[0]});
+    }
+
+    // Measure in the computational basis
+    bool mres = true;
+    for (const auto &wire : wires) {
+        Result res = this->Measure(wire);
+        mres = mres && (*res);
+    }
```
sengthai: Same with this https://github.com/PennyLaneAI/pennylane-lightning/pull/1327/changes#r2695111772

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
lazypanda10117: should we fix that here or in the mbqc pass then?

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
lazypanda10117: Thanks for noticing this. Fixed https://github.com/PennyLaneAI/pennylane-lightning/pull/1327/commits/0e7f85c1d516a1b0437c9e4ded12b8de0c41dd64

pennylane_lightning/core/simulators/lightning_gpu/catalyst/LightningGPUSimulator.cpp
sengthai: Since we have been using this implementation, we should fix it in mbqc then.

PR #1339, opened 2026-01-29
remove pennylane from docs group

runora95:
**Context:**
Push to TestPyPI [failed on master](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/21483986213/job/61888856259) for version 0.45.0-dev12

Removing `pennylane @master ` from docs dependency group.

**Benefits:**
Upload to TestPyPi will work

**Possible Drawbacks:**
None
**Related GitHub Issues:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,18 +82,13 @@ base = [
     "isort~=7.0.0",
     "pylint~=4.0.0"
 ]
-
-[project.optional-dependencies]
-kokkos = [ "pennylane-lightning-kokkos",]
-amdgpu = [ "pennylane-lightning-amdgpu",]
-gpu = [ "pennylane-lightning-gpu",]
-tensor = [ "pennylane-lightning-tensor",]
 docs = [
     "build",
     "wheel",
     "tomlkit",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -94,7 +94,6 @@ docs = [
     "tomlkit",
     "custatevec-cu12; sys_platform == 'linux'",
     "cutensornet-cu12; sys_platform == 'linux'",
-    "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git@master",
```
maliasadi, line 97: But we need to still install it for building the 'latest' version of the docs? 🤔

runora95: true, the restriction seems to be with just `[project.optional-dependencies]` because of the metadata it attaches. There might be a way to get rtd to use a regular dependency group

maliasadi (approved):
Thanks @runora95! Let's see if this will fix the issue with pushing the wheels to TestPyPI 🤞

josephleekl (approved):
Thanks @runora95 ! Let's hope this fixes the issue

PR #1292, opened 2025-11-03
Remove operator instantiation from test to obtain matrix

dwierichs:
**Context:**
A test uses an operator instantiation within a QNode to compute a sparse matrix. This should not be done within a QNode
and will surface as a problem once https://github.com/PennyLaneAI/pennylane/pull/8131 is merged.

**Description of the Change:**
Compute the matrix outside of the QNode instead.

**Benefits:**
Code quality, and compliance with new queuing behaviour that will be created by the mentioned PR. 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

PR #1340, opened 2026-02-04
fix stable and latest CI

josephleekl:
**Context:**
There are several issues causing our CIs to fail:
- Since we removed our requirements.txt files and have all dependencies specified in pyproject.toml, our stable CI tests have failed, as it cannot install the test dependencies with the same command. 
- Since pytest 9.0 is introduced, one of test fixtures is deprecated and caused an error
- This uncovered one [issue](https://xanaduhq.slack.com/archives/C0271MF2M50/p1770241318650249) with PennyLane's queueing of matrix decomposition in a test
- This uncovered another [issue](https://xanaduhq.slack.com/archives/C0271MF2M50/p1770304537657249) with PennyLane's handling of wire_order in Composite Op

1. We introduce a temporary patch (which needs to be removed in the next release cycle) to install test dependencies with `pip install -r requirements-tests.txt` (the old way) for stable CI runs.

2. We updated the fixture to be compliant with pytest 9.0.For stable test, this is not reflected, so we pull the older version of pytest, which will also be removed in the next release cycle.

3. We apply the same fix [applied before ](https://github.com/PennyLaneAI/pennylane-lightning/pull/1292) to an offending test in the mpitests

4. We update the argument type of `wire_order` in a test

**Benefits:**
[stable/stable](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/21721347342) and [latest/latest](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/21721343991) CIs both pass again.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-110620]

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -33,7 +33,7 @@
 <h3>Internal changes ⚙️</h3>
 
 - Temporarily updated CI for stable versions to install from `requirements-tests.txt`.
-  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
+  [(#1340)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1340)
 
 - Cleaned up vector-matrix methods used by Lightning devices at `LinearAlgebra.hpp`.
   [(#1314)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1314)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -32,7 +32,7 @@
 
 <h3>Internal changes ⚙️</h3>
 
-- Temporarily updated CI for stable versions to install from `requirements-tests.txt`.
+- Temporarily updated CI for stable versions to install from `requirements-tests.txt`, and fixed tests to work with pytest 9.0 and updated `queue_category` in PennyLane.
   [(#1340)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1340)
 
 - Cleaned up vector-matrix methods used by Lightning devices at `LinearAlgebra.hpp`.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -32,6 +32,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Temporarily updated CI for stable versions to install from `requirements-tests.txt`.
+  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
+
 - Cleaned up vector-matrix methods used by Lightning devices at `LinearAlgebra.hpp`.
   [(#1314)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1314)
 
--- a/.github/workflows/tests_gpu_python.yml
+++ b/.github/workflows/tests_gpu_python.yml
@@ -153,7 +153,11 @@ jobs:
 
       - name: Install required packages
         run: |
-          python -m pip install --group tests
+          if [[ "${{ inputs.lightning-version }}" == "stable" ]]; then # FIXME: remove after v0.45 release
+            python -m pip install -r requirements-tests.txt
+          else
+            python -m pip install --group tests
+          fi
           # pyscf is not compatible with numpy 2.4 yet
           python -m pip install cmake openfermionpyscf "numpy<2.4"
 
--- a/.github/workflows/tests_lgpumpi_python.yml
+++ b/.github/workflows/tests_lgpumpi_python.yml
@@ -134,7 +134,11 @@ jobs:
           PL_BACKEND: lightning_qubit
         run: |
           source /etc/profile.d/modules.sh && module use /opt/modules/ && module load ${{ matrix.mpilib }}
-          python -m pip install --group tests
+          if [[ "${{ inputs.lightning-version }}" == "stable" ]]; then # FIXME: remove after v0.45 release
+            python -m pip install -r requirements-tests.txt
+          else
+            python -m pip install --group tests
+          fi
           python -m pip install custatevec-cu12 # required for CUQUANTUM_SDK path
           # pyscf is not compatible with numpy 2.4 yet
           python -m pip install mpi4py openfermionpyscf "numpy<2.4"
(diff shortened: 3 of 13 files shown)


.github/CHANGELOG.md
```diff
@@ -32,6 +32,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Temporarily updated CI for stable versions to install from `requirements-tests.txt`.
+  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
```
```suggestion
  [(#1340)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1340)
```

mpitests/test_measurements_sparse.py
```diff
@@ -119,9 +119,9 @@ class TestSparseExpvalQChem:
     singles, doubles = qchem.excitations(active_electrons, qubits)
     excitations = singles + doubles
 
-    @pytest.fixture(
-        params=[np.complex64, np.complex128] if device_name != "lightning.gpu" else [np.complex128]
-    )
+    _dtypes = [np.complex64, np.complex128] if device_name != "lightning.gpu" else [np.complex128]
```
josephleekl: this fixes a different issue, where this is deprecated in pytest 9.0, and causes [this failure](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/21653479788/job/62513335666) in latest/latest . However, for stable tests, this won't be reflected, so we have to use the previous version of pytest (8.4) before this is deprecated.

.github/workflows/tests_lgpumpi_python.yml
```diff
@@ -134,7 +134,12 @@ jobs:
           PL_BACKEND: lightning_qubit
         run: |
           source /etc/profile.d/modules.sh && module use /opt/modules/ && module load ${{ matrix.mpilib }}
-          python -m pip install --group tests
+          if [[ "${{ inputs.lightning-version }}" == "stable" ]]; then # FIXME: remove after v0.45 release
+            python -m pip install -r requirements-tests.txt
+            python -m pip install pytest==8.4.2
```
josephleekl, line 139: Since pytest has been updated to 9.0, it broke one of the MPI tests. The test has been updated, but for testing with stable version, it doesn't use the new/fixed test. So we have to revert to the older pytest version that works.

mpitests/test_measurements_sparse.py
```diff
@@ -119,9 +119,7 @@ class TestSparseExpvalQChem:
     singles, doubles = qchem.excitations(active_electrons, qubits)
     excitations = singles + doubles
 
-    @pytest.fixture(
-        params=[np.complex64, np.complex128] if device_name != "lightning.gpu" else [np.complex128]
-    )
+    @pytest.mark.parametrize("dtype", [np.complex128])
```
josephleekl: this fixes a different issue, where this is deprecated in pytest 9.0, and causes [this failure](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/21653479788/job/62513335666) in latest/latest . However, for stable tests, this won't be reflected, so we have to use the previous version of pytest (8.4) before this is deprecated.

mpitests/test_measurements_sparse.py
```diff
@@ -76,27 +76,23 @@ def dev(self, request):
     def test_sparse_Pauli_words(self, cases, tol, dev):
         """Test expval of some simple sparse Hamiltonian"""
 
+        # Compute the sparse matrix of the input operator
```
josephleekl, line 79: this fix is the same as [this one](https://github.com/PennyLaneAI/pennylane-lightning/pull/1292) applied before in the non-mpi version of the same test

mpitests/test_measurements_sparse.py
```diff
@@ -119,28 +115,26 @@ class TestSparseExpvalQChem:
...
-    )
+    @pytest.mark.parametrize("dtype", [np.complex128])
     @pytest.mark.parametrize(
         "qubits, wires, H, hf_state, excitations",
         [
             [qubits, range(qubits), H, hf_state, excitations],
             [
                 qubits,
-                np.random.permutation(np.arange(qubits)),
+                list(np.random.permutation(np.arange(qubits))),
```
josephleekl: this is yet [another issue](https://xanaduhq.slack.com/archives/C0271MF2M50/p1770304537657249) raised with core

.github/workflows/tests_gpu_python.yml
```diff
@@ -153,7 +153,11 @@ jobs:
 
       - name: Install required packages
         run: |
-          python -m pip install --group tests
+          if [[ "${{ inputs.lightning-version }}" == "stable" ]]; then # FIXME: remove after v0.45 release
```
JerryChen97, line 156: Dowe have a story ticket for this?

JerryChen97 (approved):
Thanks for fixing!

tests/test_measurements_sparse.py
```diff
@@ -89,26 +89,26 @@ class TestSparseExpvalQChem:
...
-                np.random.permutation(np.arange(qubits)),
+                list(np.random.permutation(np.arange(qubits))),
                 H,
                 hf_state,
                 excitations,
             ],
         ],
     )
-    def test_sparse_Pauli_words(self, qubits, wires, H, hf_state, excitations, tol, request):
+    def test_sparse_Pauli_words(self, qubits, wires, H, hf_state, excitations, tol, dtype):
```
maliasadi: Fix this warning:
```
Too many positional arguments (8/5) (too-many-positional-arguments)
```

.github/workflows/tests_gpu_python.yml
josephleekl: good reminder: [here we go](https://app.shortcut.com/xanaduai/story/110773/remove-fixme-in-ci-script-for-installing-deps-for-stable-tests)

maliasadi (approved):
Thanks @josephleekl! 

Don't forget to update the changelog :)

mpitests/test_measurements_sparse.py
```diff
@@ -119,28 +115,26 @@ class TestSparseExpvalQChem:
...
                 H,
                 hf_state,
                 excitations,
             ],
         ],
     )
-    def test_sparse_Pauli_words(self, qubits, wires, H, hf_state, excitations, tol, request):
+    def test_sparse_Pauli_words(self, qubits, wires, H, hf_state, excitations, tol, dtype):
```
jzaia18:
```suggestion
    def test_sparse_pauli_words(self, qubits, wires, H, hf_state, excitations, tol, dtype):
```

silly nitpick unrelated to the PR

jzaia18 (approved):
Thanks for the fix, just don't forget to create a story for after the release

mpitests/test_measurements_sparse.py
LuisAlfredoNu:
```suggestion
    @pytest.mark.parametrize("dtype", [np.complex64, np.complex128])
```
Since the only device that uses this test is LGPU. L30

LuisAlfredoNu: ~Could you also fix L37 please?~
nevermind it is necessary that fixture

PR #1341, opened 2026-02-05
Clean up preprocess and pass correct gate set

astralcai:
**Context:**
- Update the preprocess transforms to pass the correct gate set to `decompose`.
- Clean up duplicate code in the preprocess transforms.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-110772]

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -50,12 +50,16 @@
 - Updated `pyproject.toml` with project dependencies to replace the requirements files. Updated workflows to use installations from `pyproject.toml`.
   [(1334)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1334)
 
+- Cleaned up the preprocess transforms of the lightning devices, updating the calls to `decompose` with the correct gate set.
+  [(#1341)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1341)
+
 <h3>Contributors ✍️</h3>
 
 This release contains contributions from (in alphabetical order):
 
 Runor Agbaire,
 Ali Asadi,
+Astral Cai,
 Ashish Kanwar Singh,
 Jeffrey Kam,
 Joseph Lee,
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -32,6 +32,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Cleaned up the preprocess transforms of the lightning devices, updating the calls to `decompose` with the correct gate set.
+  [(#1341)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1341)
+
 - Remove MPICH checks from CI pipelines for Lightning devices with MPI distributed support.
   [(#1342)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1342)
 
@@ -55,9 +58,6 @@
 
 - Updated `pyproject.toml` with project dependencies to replace the requirements files. Updated workflows to use installations from `pyproject.toml`.
   [(1334)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1334)
-
-- Cleaned up the preprocess transforms of the lightning devices, updating the calls to `decompose` with the correct gate set.
-  [(#1341)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1341)
 
 <h3>Contributors ✍️</h3>
 
--- a/pennylane_lightning/lightning_base/lightning_base.py
+++ b/pennylane_lightning/lightning_base/lightning_base.py
@@ -33,8 +33,19 @@ from numpy.random import BitGenerator, Generator, SeedSequence
 from numpy.typing import ArrayLike
 from pennylane.devices import Device, ExecutionConfig, MCMConfig
 from pennylane.devices.modifiers import simulator_tracking, single_tape_support
-from pennylane.exceptions import DeviceError
-from pennylane.measurements import MidMeasureMP, Shots, ShotsLike
+from pennylane.devices.preprocess import (
+    decompose,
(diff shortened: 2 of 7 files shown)


.github/CHANGELOG.md
```diff
@@ -53,12 +53,16 @@
 - Updated `pyproject.toml` with project dependencies to replace the requirements files. Updated workflows to use installations from `pyproject.toml`.
   [(1334)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1334)
 
+- Cleaned up the preprocess transforms of the lightning devices, updating the calls to `decompose` with the correct gate set.
```
LuisAlfredoNu: Could you move it on top of the section, please?

pennylane_lightning/lightning_base/lightning_base.py
```diff
@@ -819,3 +831,62 @@ def resolve_mcm_method(mcm_config: MCMConfig, tape: QuantumScript | None, device
...
+            name=name,
+        )
+        + validate_observables(stopping_condition=_adjoint_observables, name=name)
+        + validate_measurements(analytic_measurements=_adjoint_measurement, name=name)
+        + validate_adjoint_trainable_params
+    )
+
+
+def supports_adjoint(device, circuit) -> bool:
+    """Returns True iff the device supports adjoint differentiation of the given circuit."""
```
LuisAlfredoNu:
```suggestion
    """Returns True if the device supports adjoint differentiation of the given circuit."""
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -90,96 +89,13 @@
 }
 
 
-def stopping_condition(op: Operator, allow_mcms: bool = True) -> bool:
-    """A function that determines whether or not an operation is supported by ``lightning.gpu``."""
-    if isinstance(op, MidMeasureMP):
-        # Conditional and MidMeasureMP should not be decomposed
-        return allow_mcms
+def make_stopping_condition(gate_set: GateSet) -> Callable[[Operator], bool]:
```
maliasadi: While you're cleaning up, could we also fix the inconsistency of the stooping condition method in Lightning-GPU with other lightning devices?

pennylane_lightning/lightning_gpu/lightning_gpu.py
```diff
@@ -419,7 +342,8 @@ def preprocess_transforms(
         pipeline.add_transform(qml.transforms.broadcast_expand)
 
         if exec_config.gradient_method == "adjoint":
-            _add_adjoint_transforms(pipeline, device_wires=self.wires)
+            adjoint_transforms(self, allow_mcms)
```
maliasadi: So does this mean that even if Lightning supports a gate, this op will still be decomposed as long as it isn’t differentiable? 🤔

maliasadi: shouldn't be updated as follows:

```suggestion
            pipeline += adjoint_transforms(self, allow_mcms)
```

pennylane_lightning/lightning_gpu/lightning_gpu.py
astralcai: The Lightning-GPU has the simplest stopping condition (checks whether the op is in the gate set), while the other devices all have additional logic in their stopping conditions. I'm not sure I have enough context to unify them.

pennylane_lightning/lightning_gpu/lightning_gpu.py
astralcai: Ooooops

astralcai: My understanding is that if a gate is not labeled "differentiable" in the TOML file, it cannot be treated as supported if we're using adjoint gradient. Is that understanding inaccurate?

maliasadi: This should be still supported in the forward mode

pennylane_lightning/lightning_gpu/lightning_gpu.py
maliasadi: my concern was mostly with the name of the stopping condition function :)


Could we remove the stopping_condition func for this device and rely solely on the supported ops from the device capabilities?

astralcai: Unfortunately not yet. `preprocess.decompose` still requires a `stopping_condition` even when a `gate_set` is provided, as opposed to `transforms.decompose` which is able to infer the stopping condition from the gate set. I probably should've done the same for `preprocess.decompose` long ago...

pennylane_lightning/lightning_gpu/lightning_gpu.py
astralcai: How would you know in `preprocess` if we're in forward mode or doing the gradient?

pennylane_lightning/lightning_gpu/lightning_gpu.py
astralcai: It turns out that what I did didn't work, so I made it just look like the rest of them.

LuisAlfredoNu (approved):
Thank you @astralcai 
LGTM

pennylane_lightning/lightning_gpu/lightning_gpu.py
maliasadi: I don't know! It's also my question :smile:

maliasadi (approved):
LGTM! My only remaining concern https://github.com/PennyLaneAI/pennylane-lightning/pull/1341#discussion_r2785734976 is unrelated to this PR and requires further discussion.

PR #1348, opened 2026-02-24
Avoid decomposing unsupported Ops into QubitUnitary when not specified in the device TOML file

maliasadi:
**Context:**
Avoid decomposing unsupported gates and templates into `QubitUnitary` when not specified in the device TOML file of Lightning simulators from Catalyst. re: https://github.com/PennyLaneAI/catalyst/pull/2426 

- Skip the initialization of `_to_matrix_ops` when `QubitUnitary` is not in the list of supported ops in Lightning 

**Benefits:**
- Enable decomposing to a custom target gateset without `QubitUnitary` when using Lightning simulators with `qml.qjit`.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-111263]

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -245,3 +245,14 @@ def test_supported_macos_platform_qubit():
 
     assert dev_name == "LightningSimulator"
     assert "liblightning_qubit_catalyst.dylib" in shared_lib_name
+
+
+@pytest.mark.skipif(
+    (device_name == "lightning.tensor"),
+    reason="Lightning-Tensor is not integrated with Catalyst and doesn't support to_matrix_ops.",
+)
+def test_device_to_matrix_ops():
+    """Test that the device's to_matrix_ops capability is correctly set based on the config file."""
+    dev = qml.device(device_name)
+    to_mat_ops = dev._to_matrix_ops
+    assert to_mat_ops is not None and isinstance(to_mat_ops, dict)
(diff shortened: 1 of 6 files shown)


josephleekl (approved):
Looks good, thanks @maliasadi 

I guess we can only test that this works in Catalyst?

tests/test_device.py
```diff
@@ -245,3 +245,14 @@ def test_supported_macos_platform_qubit():
...
+
+@pytest.mark.skipif(
+    (device_name == "lightning.tensor"),
+    reason="Lightning-Tensor is not integrated with Catalyst and doesn't support to_matrix_ops.",
+)
+def test_device_to_matrix_ops():
+    """Test that the device's to_matrix_ops capability is correctly set based on the config file."""
+    dev = qml.device(device_name)
+    to_mat_ops = dev._to_matrix_ops
+    assert to_mat_ops is not None and isinstance(to_mat_ops, dict)
```
josephleekl, line 258: so this depends on the current configuration of the TOML files, right? can we test a case where the toml says QubitUnitary is not supported?

maliasadi: @josephleekl yeah, I don't know how to check `None` without changing the lightning device toml file.

jzaia18: +1, I think we should have a test to make sure this works

jzaia18 (commented):
Happy to approve once we have a unit test

jzaia18 (approved):

PR #1354, opened 2026-03-16
Fix LGPU synchronise before ComputeExpectation

josephleekl:
**Context:**
Upgrading to CUDA 12.9 uncovered an error (e.g. [here](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/23126420272/job/67170365591?pr=1353)) when performing some expectation value computation for LGPU with MPI, where the result read by the host if 0.

The issue comes from the fact that `custatevecComputeExpectation` is potentially asynchronous, as noted in their [documentation](https://docs.nvidia.com/cuda/cuquantum/25.03.0/custatevec/api/functions.html#custateveccomputeexpectation):

> This function might be asynchronous with respect to host depending on the arguments. Please use cudaStreamSynchronize (for synchronization) or cudaStreamWaitEvent (for establishing the stream order) with the stream set to the handle of the current device before using the results stored in expectationValue.

By applying a cuda sync between the ComputeExpectation call and the host read call, we prevent the error observed where we read a value of 0.

**Benefits:**
Fix bug and we can upgrade to CUDA 12.9

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-112893]

### DIFF
--- a/.github/workflows/tests_lgpumpi_cpp.yml
+++ b/.github/workflows/tests_lgpumpi_cpp.yml
@@ -47,7 +47,7 @@ jobs:
       matrix:
         mpilib: ["openmpi"]
         cuda_version_maj: ["12"]
-        cuda_version_min: ["4"]
+        cuda_version_min: ["9"]
     timeout-minutes: 60
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests_lgpumpi_cpp.yml
```diff
@@ -138,6 +138,9 @@ jobs:
           mpirun --version
           which -a mpicxx
           mpicxx --version
+          ompi_info --parsable --all | grep -i cuda
+          ompi_info --parsable --all | grep mpi_built_with_cuda_support
+          ldd $(which mpirun) | grep -i cuda
```
josephleekl:
```suggestion
```

maliasadi (approved):
Thanks @josephleekl!

PR #1365, opened 2026-04-21
Fix MPI test to use PennyLane Numpy

josephleekl:
**Context:**
Some MPI Python tests involving differentiation did not use PennyLane Numpy, therefore the circuits did not execute. 

Update to use PennyLane Numpy for relevant tests to make sure circuits are executed and the tests are run properly. The tests were identified using `-W error` in pytest which outputs `UserWarning: Attempted to differentiate a function with no trainable parameters` as error.

**Benefits:**
Proper MPI tests.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-117519]
[sc-117483]

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -35,6 +35,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Fixed Python MPI tests involving differentiation to use PennyLane Numpy.
+  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
+
 - Upgraded Kokkos from v4.5.0 to v5.1.0.
   [(#1308)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1308)
 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -36,7 +36,7 @@
 <h3>Internal changes ⚙️</h3>
 
 - Fixed Python MPI tests involving differentiation to use PennyLane Numpy.
-  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
+  [(#1365)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1365)
 
 - Upgraded Kokkos from v4.5.0 to v5.1.0.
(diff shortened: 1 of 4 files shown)


.github/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+- Fixed Python MPI tests involving differentiation to use PennyLane Numpy.
+  [(#XXXX)](https://github.com/PennyLaneAI/pennylane-lightning/pull/XXXX)
```
```suggestion
  [(#1365)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1365)
```

mpitests/test_adjoint_jacobian.py
```diff
@@ -523,7 +523,7 @@ def circ(x):
                 qml.RX(x, wires=0)
                 return qml.expval(qml.PauliZ(0))
 
-            qml.grad(circ)(0.1)
+            qml.grad(circ)(np.array(0.1))
```
josephleekl, line 526: for this file we `from pennylane import numpy as np`, so np is the correct one.

mpitests/test_apply.py
```diff
@@ -857,7 +858,7 @@ def circuit_ansatz(params, wires):
     qml.IsingXX(params[13], wires=[wires[1], wires[0]])
     qml.IsingYY(params[14], wires=[wires[3], wires[2]])
     qml.IsingZZ(params[15], wires=[wires[2], wires[1]])
-    qml.PSAWP(params[16], wires=[wires[3], wires[0]])
+    qml.PSWAP(params[16], wires=[wires[3], wires[0]])
```
maliasadi, line 861: 😄

maliasadi (approved):

mpitests/test_adjoint_jacobian.py
jzaia18: Would it be worth changing all the `np`s to `pnp`s for consistency while we're at it?

jzaia18 (approved):
LGTM, thanks for the fix!

mpitests/test_adjoint_jacobian.py
josephleekl: i think i'll leave it as it is for consistency with the non-mpi version of the test

PR #1366, opened 2026-04-22
Update lightning-core with qml->qp changes

jzaia18:
**Context:** 
We would like to update a number of PennyLane modules to import pennylane as `qp` and to use `qp` instead of `qml` throughout. This is a part of a wider rebrand of the package.

**Description of the Change:**
Updates names, imports and references to `qml`.

**Benefits:**
Updated to the new branding.

**Possible Drawbacks:**
A lot of changes.

**Related GitHub Issues:**
[sc-117172]
Builds on #1363

PR #1363, opened 2026-04-21
Update tests with qml->qp changes

jzaia18:
**Context:** 
We would like to update a number of PennyLane modules to import pennylane as `qp` and to use `qp` instead of `qml` throughout. This is a part of a wider rebrand of the package.

**Description of the Change:**
Updates names, imports and references to `qml`.

**Benefits:**
Updated to the new branding.

**Possible Drawbacks:**
A lot of changes.

**Related GitHub Issues:**
[sc-117172]
Builds on #1362

PR #1362, opened 2026-04-21
Update docs and .github folders with qml->qp changes

jzaia18:
**Context:** 
We would like to update a number of PennyLane modules to import pennylane as `qp` and to use `qp` instead of `qml` throughout. This is a part of a wider rebrand of the package.

**Description of the Change:**
Updates names, imports and references to `qml`.

**Benefits:**
Updated to the new branding.

**Possible Drawbacks:**
A lot of changes.

**Related GitHub Issues:**
[sc-117172]

PR #1368, opened 2026-04-24
Add .git-blame-ignore-revs

jzaia18:
Context:
#1362, #1363, and #1366 updated a number of PennyLane modules to import pennylane as qp and to use qp instead of qml throughout. This mangles the git blame and makes it seem like nearly every line containing `qp.` was last modified by one of these PRs.

Description of the Change:
Adds the commits to `main` from #1362, #1363, and #1366 to a new `.git-blame-ignore-revs` file

Benefits:
Cleaner version history when viewing the "git blame" of files

Possible Drawbacks:

**Related GitHub Issues:**
[sc-117172]

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,11 @@
+# Add commits here that should be ignored in the output of git blame. Include a comment
+# about why the commit is ignored, with a link to the corresponding PR if applicable.
+
+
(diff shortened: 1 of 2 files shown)


maliasadi (approved):

josephleekl (approved):
Thanks @jzaia18 ! Please update the changelog as well :D

PR #1378, opened 2026-05-06
Revert "Adding nightly RC builds (#1344)"

jzaia18:
**Context:**
RC wheels are failing to upload to TestPyPI because reusable workflows cannot be used with trusted publishing

Reverts #1344 to unblock RC wheels from being uploaded

**Benefits:**
Unblocks RC wheel uploads in the short term

**Possible Drawbacks:**
We will want to change this again for next release

**Related GitHub Issues:**


runora95:
If we're reverting the old PR, can we add this in to one of the workflows so it still triggers catalyst:

```
  trigger-catalyst-rc:
    needs: [upload-pypi]
    if: ${{ always() && !cancelled() }}
    name: Trigger Catalyst RC workflow
    runs-on: ubuntu-24.04
    
    steps:
      - name: Checkout Lightning repo
        uses: actions/checkout@v4
        with:
          ref: main

      - name: Trigger Catalyst RC workflow
        env:
          GITHUB_TOKEN: "${{ secrets.PENNYLANE_TRIGGER_RC_WORKFLOW }}"
        run: |
          git config user.name "GitHub Actions Bot"
          git config user.email "<>"         
          gh workflow run build-nightly-release-candidate.yaml --repo PennyLaneAI/catalyst
 ```

jzaia18:
> If we're reverting the old PR, can we add this in to one of the workflows so it still triggers catalyst:

@runora95 any preference on which workflow we should add this to? If not then probably `wheel_noarch` would be the most natural? Or would it be best to give it its own workflow entirely?

jzaia18:
Closing in favor of #1379

PR #1379, opened 2026-05-06
Upload to testpypi from 1 source

runora95:
**Context:**
PyPI doesn't allow trusted publishing from reusable workflows. The current `upload-nightly-rc-release.yml` workflow is configured to upload the wheel from each reusable workflow. The wheels can be downloaded then uploaded from the calling workflow.

Add download of built wheels and upload them to testpypi from `upload-nightly-rc-release.yml` 

**Benefits:**
Trusted publishing works.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane-lightning/pull/1344

### DIFF
--- a/.github/workflows/wheel_linux_aarch64.yml
+++ b/.github/workflows/wheel_linux_aarch64.yml
@@ -23,6 +23,10 @@ on:
       branch:
         required: false
         type: string
+      skip_upload:
+        required: false
+        type: boolean
+        default: false
 
 concurrency:
   group: wheel_linux_aarch64-${{ github.ref }}
@@ -33,7 +37,6 @@ env:
 
 permissions:
   contents: read
-  id-token: write
 
 jobs:
   set_wheel_build_matrix:
@@ -223,6 +226,7 @@ jobs:
 
   upload-pypi:
     needs: [set_wheel_build_matrix, linux-wheels-aarch64]
+    if: ${{ !inputs.skip_upload }}
     strategy:
       matrix:
         arch: [aarch64]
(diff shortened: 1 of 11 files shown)


.github/workflows/wheel_linux_aarch64.yml
```diff
@@ -23,6 +23,10 @@ on:
       branch:
         required: false
         type: string
+      skip_upload:
+        required: false
+        type: boolean
+        default: false
```
jzaia18, line 29: Don't we always have to have this as false when using `workflow_call`? Maybe instead of an input we could change the conditions later down to be "if not triggered by workflow_call"?

runora95: true, i'll update it

runora95: actually, let me double check. The event_name that triggers it usually cascades into the other workflow. It might end up reading as `scheduled` instead of `workflow_call`

runora95: I'll leave it as is,`github.event_name` will show as `scheduled` in the wheel*.yml workflows. E.g. [this job](https://github.com/runora95/pennylane/actions/runs/25463399665/job/74710974602#step:2:7) was triggered manually and the reusable workflow shows github.event_name as `workflow_dispatch` .

jzaia18: Sounds good

jzaia18 (approved):
Looks good, excited to test this out

maliasadi (approved):

PR #1403, opened 2026-07-17
657

### LINKED ISSUE #657: The `IndexError` key-not-found issue from `updateCache` in LQ
Using the C++ modules of LQ in Catalyst along with the Python interface may raise `IndexError: unordered_map::at: key not found` from `updateCache` in `./gates/KernelMap.hpp` when multiple versions of LQ for accessing the Python and C++ modules are loaded.

The issue is related to the fact that Catalyst Runtime builds the C++ modules of LQ from source and Catalyst Frontend imports the Python interface shipped with the PyPI package. With the current setup, the LQ dynamic dispatcher will be triggered twice in the Catalyst pipeline,

1. with the PL device initialization: `qml.device('lighnting.qubit', wires=n)`
2. when the compiled program is linked against the LQ shared library

The cached kernels in the dynamic dispatcher is not well-scoped and thread-local and as such using multiple versions of LQ in one process (or from both Python and C++ in Catalyst) will lead to some error messages about missing kernels or operations.

JJRBEST:
**Context:**
The IndexError key-not-found issue from updateCache in LQ
At CMakelist.txt:
Change "add_library(lightning_qubit STATIC ${LQUBIT_FILES})" to "add_library(lightning_qubit SHARED ${LQUBIT_FILES})."
**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
Closes #657

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -28,6 +28,9 @@
 - Fixed `qp.PauliRot` execution bug when handling identity Pauli word within the compilation pipeline.
   [(#1390)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1390)
 
+- Fixed `CMakelist.txt` Change add_library(lightning_qubit STATIC ${LQUBIT_FILES}) to add_library(lightning_qubit SHARED ${LQUBIT_FILES}).
+  [(#1403)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1403)
+
 <h3>Internal changes ⚙️</h3>
 
 - Upgrade Mac runner to macos-15.
@@ -63,6 +66,7 @@
 This release contains contributions from (in alphabetical order):
 
 Runor Agbaire,
+JiaRung Jian,
 Jeffrey Kam,
 Joseph Lee,
 Jake Zaia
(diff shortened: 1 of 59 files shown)


maliasadi:
@JJRBEST How's this patch fixing the issue?

JJRBEST:
> @JJRBEST How's this patch fixing the issue?

I just changed the file of CMakelist.txt, the line of  "add_library(lightning_qubit SHARED ${LQUBIT_FILES})" instead of "add_library(lightning_qubit STATIC ${LQUBIT_FILES})", because I need to let them share the same memory, SHARED instead of STATIC in file of CMakelist.txt

maliasadi:
> > @JJRBEST How's this patch fixing the issue?
> 
> I just changed the file of CMakelist.txt, the line of "add_library(lightning_qubit SHARED ${LQUBIT_FILES})" instead of "add_library(lightning_qubit STATIC ${LQUBIT_FILES})", because I need to let them share the same memory, SHARED instead of STATIC in file of CMakelist.txt

Thanks @JJRBEST! This tells us what the change is, but not yet why it fixes #657, which is the part you need to show for this PR. 

Do you have a minimal case that triggers the `IndexError` and does it pass after this change? Sharing the before/after would answer the core question directly. and as a follow-up: what makes both paths resolve to the same cache instance rather than each keeping its own?

PR #1405, opened 2026-07-17
[fix] Use `base_operation.parameters` for controlled Operators

JerryChen97:
**Context:**

The scheduled "Compat Check w/PL — latest/latest" CI run (e.g. [run 29545522067](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/29545522067)) fails on Lightning-Kokkos and Lightning-GPU with:

```
TypeError: S(): incompatible function arguments.
Invoked with: StateVectorC64, list, list, list, bool, list
```

All 60 failures are in `test_controlled_qubit_gates` / `test_adjoint_controlled_qubit_gates` on controlled S/T/SX gates; other job failures are fail-fast cancellations.

Root cause: PennyLane `master` migrated `S`/`T`/`SX` to the new operator base class (PennyLaneAI/pennylane#9859). `qml.ctrl` on a migrated gate now returns a `ControlledOp2` whose `data`/`parameters` include the control values (e.g. `[[True, False]]`). The controlled-gate dispatch in Lightning-Kokkos, Lightning-GPU, and Lightning-Tensor read `operation.parameters` from the `Controlled` wrapper and passed it as the `params` argument of the C++ binding, which expects `Sequence[float]` — so nanobind rejects the call. Unmigrated gates (`PauliX`, `Hadamard`, …) still return `[]` from the wrapper (and `ctrl(PauliX)` lowers to `MultiControlledX`), which is why only S/T/SX failed.

Lightning-Qubit was already reading `base_operation.parameters` (since #1366) and passed.

Read gate parameters from the unwrapped base operation instead of the `Controlled` wrapper in `_apply_lightning_controlled`, mirroring Lightning-Qubit:

- `lightning_kokkos/_state_vector.py`: `param = base_operation.parameters`
- `lightning_gpu/_state_vector.py`: `param = base_operation.parameters`
- `lightning_tensor/_tensornet.py`: `param = operation.base.parameters`

`base_operation.parameters` is identical to the previous behavior on released PennyLane (both empty for non-parametric gates, actual gate params for parametric ones) and correct on PennyLane `master`, so this is backward and forward compatible.

**Benefits:**

- Fixes the latest/latest and likely some potential LLL failures.
- Removes a latent inconsistency between Lightning-Qubit and the other backends.

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/pull/9818

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -68,6 +68,7 @@
 This release contains contributions from (in alphabetical order):
 
 Runor Agbaire,
+Yushao Chen,
 Jeffrey Kam,
 Joseph Lee,
 Jake Zaia
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -25,6 +25,11 @@
 
 <h3>Bug fixes 🐛</h3>
 
+- Fixed controlled-gate dispatch in Lightning-Kokkos, Lightning-GPU and Lightning-Tensor to read gate
+  parameters from the base operation instead of the `Controlled` wrapper, restoring compatibility with
+  PennyLane where `Controlled.parameters` now includes the control values.
+  [(#1405)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1405)
+
 - Fixed `qp.PauliRot` execution bug when handling identity Pauli word within the compilation pipeline.
   [(#1390)](https://github.com/PennyLaneAI/pennylane-lightning/pull/1390)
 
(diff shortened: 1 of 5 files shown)


multiphaseCFD (approved):
Thanks @JerryChen97 . Don't forget to add yourself to the contributor.

josephleekl (approved):
thanks @JerryChen97 . Please make sure this passes before merging https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/29611502785

PR #1411, opened 2026-07-28
fix: remove superfluous try/catch in MPI teardown

kszenes:
Nothing in the line `MPIRuntime::getInstance().finalizeIfOwned()` should be able to throw an exception, therefore making the try/catch block superfluous.

[sc126075]


josephleekl (approved):
Thanks @kszenes !

multiphaseCFD (approved):
LGTM! Thanks @kszenes

PR #1425, opened 2026-09-16
Use installed nanobind for Python builds

### LINKED ISSUE #1423: Use installed nanobind instead of fetching from Git
Replace FetchContent of nanobind from GitHub with a find_package call
that discovers a pre-installed nanobind via "python -m nanobind
--cmake_dir".

Our build system (Fromager) compiles wheels inside network-isolated
containers where each package's build step runs in an unshared network
namespace (via unshare(2)). Any FetchContent_Declare that attempts to
clone from a remote Git repository fails at CMake configure time
because no external network is available.

Instead, nanobind is declared as a PEP 517 build-requires entry
(pinned in the package's pyproject.toml) and installed into the build
virtualenv before compilation begins. The execute_process call invokes
"python -m nanobind --cmake_dir" to locate the installed nanobind's
CMake config, and find_package resolves it locally.

This is the standard approach used by other projects that build
nanobind-based extensions in offline or hermetic environments (e.g.
MLIR Python bindings in LLVM).

danielgaskins:
**Context:**

Python-enabled builds currently use FetchContent to clone nanobind during CMake configuration. This fails for network-isolated wheel builds even when nanobind is available in the PEP 517 build environment.

- Locate nanobind by running `${Python_EXECUTABLE} -m nanobind --cmake_dir`, then load it with `find_package`.
- Pin `nanobind==2.11.0` in both the checked-in and generated build requirements.
- Remove the unused `ENABLE_NANOBIND` conditional.

**Benefits:**

Wheel builds can use the installed nanobind package without cloning its GitHub repository during CMake configuration.

**Possible Drawbacks:**

The build requirement pin will need to be updated when Lightning adopts another nanobind release.

**Validation:**

- Generated a Lightning-Qubit `pyproject.toml` and confirmed that it contains exactly one `nanobind==2.11.0` requirement.
- Configured Lightning-Qubit in an isolated virtual environment with `FETCHCONTENT_FULLY_DISCONNECTED=ON`.
- Confirmed that `nanobind_DIR` resolved to the package installed in that environment.
- Ran `git diff --check`.

**Related GitHub Issues:**

Closes #1423

**AI Assistance:**

Full disclosure, I used Codex to inspect the build paths, implement the CMake and packaging changes, and run validation. I reviewed the diff and validation results before submitting this pull request.


josephleekl:
Hi @danielgaskins ,
Thank you so much for taking the time to put this PR together!
We have reviewed this with the team. While we really appreciate the effort you’ve put into this, we have decided not to merge it into main. Currently, this implementation falls a bit outside our core roadmap and priorities for pennylane-lightning. Additionally, as the PR currently stands, it may break some of our existing CI workflows, and we want to be cautious about adding new maintenance overhead for features that aren’t tightly aligned with our immediate goals.
Because of this, we are going to go ahead and close this PR. We truly appreciate your interest in contributing to PennyLane and hope to see you around in future discussions or issues!

PR #1426, opened 2026-09-18
Fix version update action

josephleekl:
Semver updated to 3.1 recently which broke the dev version update action.

[sc-131113]


pennylane_lightning/core/_version.py
```diff
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.46.0-dev27"
+__version__ = "0.46.0-rc.1"
```
josephleekl:
```suggestion
__version__ = "0.46.0-dev28"
```

jzaia18 (approved):

maliasadi (approved):

